PR #1, opened 2020-03-19
Add documentation for cli and http api

oed:


packages/ceramic-cli/README.md
```diff
@@ -1,40 +1,98 @@
...
+}
 ```
-/api/v0/create - POST
 
-/api/v0/show/<document-id> - GET
+### `/api/v0/show/<document-id>` - `GET`
+Returns the content of the given document.
+
+##### Optional query parameters:
+* `?version=<number>` - the version of the document to show
```
michaelsena: @oed how will we implement versions? Based on numbers as you have here, or would it be better to request the CID of the given anchor record? Or does the protocol automatically add numbers to the records so they can be queried by either?

oed: So this needs to be documented in the spec. The number refers to the n:th sequential anchor record. So essentially:
* 0: the genesis record
* 1: the first anchor record
* 2: the second anchor record
* n: the n:th anchor record

PR #6, opened 2020-03-31
Feat/doctypes

oed:
Already pushed the first commit to master 🙈 
This adds support for the signed doctypes 3id, and tile in the cli.


packages/ceramic-cli/src/ceramic-client.ts
```diff
@@ -65,7 +66,7 @@ class Document extends EventEmitter {
   }
 
   async change (newContent: any, apiUrl: string): Promise<boolean> {
-    const { docId, state } = await fetchJson(apiUrl + '/change' + this.id, { content: newContent })
+    const { state } = await fetchJson(apiUrl + '/change' + this.id, { content: newContent })
```
simonovic86: Let's normalize the URL concatenation in order to avoid multiple slashes if the `apiUrl` already ends with `/`. Maybe a separate utility function.

packages/ceramic-cli/src/ceramic-daemon.ts
```diff
@@ -16,53 +19,73 @@ class CeramicDaemon {
       next()
     })
     this.app.post(toApiPath('/create'), this.createDoc.bind(this))
-    this.app.get(toApiPath('/show/ceramic/:doctype/:cid'), this.show.bind(this))
-    this.app.get(toApiPath('/state/ceramic/:doctype/:cid'), this.state.bind(this))
-    this.app.post(toApiPath('/change/ceramic/:doctype/:cid'), this.change.bind(this))
+    this.app.get(toApiPath('/show/ceramic/:cid'), this.show.bind(this))
+    this.app.get(toApiPath('/state/ceramic/:cid'), this.state.bind(this))
+    this.app.post(toApiPath('/change/ceramic/:cid'), this.change.bind(this))
     const server = this.app.listen(7007, () => {
```
simonovic86: We can put a daemon `port` in the configuration file

packages/ceramic-cli/src/index.ts
```diff
@@ -18,9 +19,13 @@ program
   .action(async (doctype, content, { onlyGenesis }) => {
     content = JSON.parse(content)
     const ceramic = new CeramicClient()
-    const doc = await ceramic.createDocument(content, doctype, { onlyGenesis })
-    console.log(doc.id)
-    console.log(JSON.stringify(doc.content, null, 2))
+    try {
+      const doc = await ceramic.createDocument(content, doctype, { onlyGenesis })
+      console.log(doc.id)
```
simonovic86: We can introduce a specific logger (e.g. `winston`) instead of directly using `console.log`

packages/ceramic-cli/src/index.ts
```diff
@@ -18,9 +19,13 @@ program
...
+      console.log(JSON.stringify(doc.content, null, 2))
+    } catch (e) {
+      console.error(e)
```
simonovic86: Maybe `process.exit()` with a specific code?

packages/ceramic-cli/src/index.ts
```diff
@@ -30,8 +35,12 @@ program
   .description('Show the content of a document')
   .action(async (docId) => {
     const ceramic = new CeramicClient()
-    const doc = await ceramic.loadDocument(docId)
-    console.log(JSON.stringify(doc.content, null, 2))
+    try {
+      const doc = await ceramic.loadDocument(docId)
+      console.log(JSON.stringify(doc.content, null, 2))
+    } catch (e) {
+      console.error(e)
```
simonovic86: Maybe `process.exit()` with a specific code?

packages/ceramic-cli/src/index.ts
```diff
@@ -40,8 +49,12 @@ program
   .description('Show the state of a document')
   .action(async (docId) => {
     const ceramic = new CeramicClient()
-    const doc = await ceramic.loadDocument(docId)
-    console.log(JSON.stringify(doc.state, null, 2))
+    try {
+      const doc = await ceramic.loadDocument(docId)
+      console.log(JSON.stringify(doc.state, null, 2))
+    } catch (e) {
+      console.error(e)
```
simonovic86: Maybe `process.exit()` with a specific code?

packages/ceramic-cli/src/index.ts
```diff
@@ -50,12 +63,16 @@ program
...
+    try {
+      const doc = await ceramic.loadDocument(docId)
       console.log(JSON.stringify(doc.content, null, 2))
-    })
+      doc.on('change', () => {
+        console.log('--- document changed ---')
+        console.log(JSON.stringify(doc.content, null, 2))
+      })
+    } catch (e) {
+      console.error(e)
```
simonovic86: Maybe `process.exit()` with a specific code?

packages/ceramic-cli/src/index.ts
```diff
@@ -65,8 +82,12 @@ program
   .action(async (docId, content) => {
     content = JSON.parse(content)
     const ceramic = new CeramicClient()
-    const doc = await ceramic._updateDocument(docId, content)
-    console.log(JSON.stringify(doc.id, null, 2))
+    try {
+      const doc = await ceramic._updateDocument(docId, content)
+      console.log(JSON.stringify(doc.content, null, 2))
+    } catch (e) {
+      console.error(e)
```
simonovic86: Maybe `process.exit()` with a specific code?

packages/ceramic-core/src/document.ts
```diff
@@ -27,6 +27,20 @@ export interface InitOpts {
 
 const deepCopy = (obj: any): any => JSON.parse(JSON.stringify(obj))
```
simonovic86: Really useful function for cloning. Maybe we can put it in the common utility class?

packages/ceramic-cli/src/ceramic-daemon.ts
```diff
@@ -16,53 +19,73 @@ class CeramicDaemon {
...
   async createDoc (req: Request, res: Response, next: NextFunction) {
     const { doctype, genesis, onlyGenesis } = req.body
     if (!doctype || !genesis) {} // TODO - reject somehow
-    const doc = await this.ceramic.createDocument(genesis, doctype, { onlyGenesis })
-    res.json({ docId: doc.id, state: doc.state })
+    try {
+      const doc = await this.ceramic.createDocument(genesis, doctype, { onlyGenesis })
+      res.json({ docId: doc.id, state: doc.state })
+    } catch (e) {
+      res.json({ error: e.message })
```
simonovic86: Do we need to handle `Response` error status codes?

packages/ceramic-cli/src/index.ts
oed: Definitely, created an issue here: https://github.com/ceramicnetwork/js-ceramic/issues/8

packages/ceramic-cli/src/ceramic-daemon.ts
oed: Yep, that should be configurable once we have a config file!

packages/ceramic-cli/src/index.ts
oed: Yeah, that makes sense!

oed: https://github.com/ceramicnetwork/js-ceramic/issues/9

packages/ceramic-core/src/document.ts
oed: Yeah I think there are more of these useful functions that we can pull out to a utils module.

packages/ceramic-cli/src/ceramic-daemon.ts
oed: Yeah, right now this isn't really using best practices for express or rest apis in general. We should make sure to do this properly!

PR #11, opened 2020-04-03
add account link doctype

msterle:


packages/ceramic-cli/package.json
```diff
@@ -34,7 +34,7 @@
     "@ceramicnetwork/ceramic-core": "^0.0.1",
     "commander": "^4.1.1",
     "express": "^4.17.1",
-    "identity-wallet": "file:../../../../3box/identity-wallet",
+    "identity-wallet": "^1.1.3",
```
oed, line 37: heh, sorry about that :)

packages/ceramic-core/src/doctypes/accountLinkHandler.ts
```diff
@@ -0,0 +1,85 @@
...
+      content: null,
+      nextContent: null,
+      signature: SignatureStatus.GENESIS,
+      anchored: 0,
+      log: []
+    }
+  }
+
+  async applySigned (record: any, state: DocState): Promise<DocState> {
+    const nextContent = jsonpatch.applyPatch(state.content, record.content).newDocument
```
oed, line 58: Next content should just be set to whatever DID is in the record right?

packages/ceramic-core/src/doctypes/accountLinkHandler.ts
```diff
@@ -0,0 +1,85 @@
...
+    } else if (record.proof) {
+      newState = await this.applyAnchor(record, state)
+    } else if (record.content || record.owners) {
+      newState = await this.applySigned(record, state)
+    }
+    newState.log.push(cid)
+    return newState
+  }
+
+  async makeRecord (state: DocState, newContent: any): Promise<any> {
```
oed, line 30: So newContent here is the proof right? I would just put this proof directly as the content of this record.

packages/ceramic-core/src/doctypes/accountLinkHandler.ts
```diff
@@ -0,0 +1,83 @@
...
+  async applySigned (record: any, state: DocState): Promise<DocState> {
+    const proof = record.content
+    if (!await validateLink(proof)) {
+      throw new Error('Invalid proof for signed record')
+    }
+    return {
+      ...state,
+      signature: SignatureStatus.SIGNED,
+      anchored: 0,
+      nextContent: proof
```
oed, line 64: I suppose that here we should extract the DID from `proof.message` and set nextContent to be `{ did }`.

packages/ceramic-core/src/doctypes/accountLinkHandler.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  async makeGenesis (content: any, owners: [string]): Promise<any> {
+    if (content) throw new Error('Account link genesis cannot have content')
+    if (!owners) throw new Error('Owner must be specified')
+    return {
+      doctype: this.doctype,
+      owners,
+    }
+  }
```
oed, line 28: We should make sure that the owner address is a CAIP-10 address: https://github.com/ceramicnetwork/specs/blob/master/doctypes/account-link.md#examples

oed: We can simply just assume ethereum mainnet for now (`<address>@eip155:1`)

PR #13, opened 2020-04-04
Feat/anchor module

oed:
This PR does a few things:
* Update typescript version and add type only imports
* Add anchor module which works async
* Refactor abstract doctypeHandler interface to accommodate for the above

### DIFF
--- a/packages/ceramic-core/src/document.ts
+++ b/packages/ceramic-core/src/document.ts
@@ -1,5 +1,5 @@
-import Dispatcher from './dispatcher' // import only type ts 3.8
-import DoctypeHandler from './doctypes/doctypeHandler'
+import type Dispatcher from './dispatcher'
+import type DoctypeHandler from './doctypes/doctypeHandler'
 import { EventEmitter } from 'events'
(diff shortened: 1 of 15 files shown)


packages/ceramic-core/src/document.ts
```diff
@@ -179,38 +202,48 @@ class Document extends EventEmitter {
...
-      blockNumber: Date.now(),
-      txHash: 'eth-cid',
-      root: 'cid'
-    }
-    const record = { proof, path: 'ipld path for witness', next: { '/': this.head } }
-    const cid = (await this.dispatcher.storeRecord(record)).toString()
-    this._state = await this._doctype.applyRecord(record, cid, this._state)
-    return true
+  async anchor (): Promise<void> {
+    this._anchorService.on(this.id, async (cid): Promise<void> => {
```
msterle, line 242: Where is this listener cleaned up?

oed: It's removed on line 28 of `packages/ceramic-core/src/anchor-service.ts`.

PR #18, opened 2020-04-06
Ref/use prev

oed:
Two things in this PR.
* Use prev instead of next (change to spec is required)
* Use the CID type instead of strings to represent CIDs


packages/ceramic-core/src/__tests__/document.test.ts
```diff
@@ -120,7 +133,7 @@ describe('Document', () => {
       const doc2 = await Document.load(docId, doctypeHandlers, anchorService, dispatcher, { skipWait: true })
       await doc2._handleHead(headPreUpdate)
       // add short wait to get different anchor time
-      // sometime the tests are damn fast
+      // sometime the tests are very fast
```
msterle, line 136: 🤪

PR #21, opened 2020-04-08
Fix/cli improvements

oed:
The cli was a little broken after the last refactor of ceramic-core. This fixes that.

### DIFF
--- a/packages/ceramic-cli/src/index.ts
+++ b/packages/ceramic-cli/src/index.ts
@@ -1,7 +1,14 @@
 import CeramicDaemon from './ceramic-daemon'
 import CeramicClient from '@ceramicnetwork/ceramic-http-client'
 import program from 'commander'
+import { serializeState } from './utils'
 
+
+const PREFIX_REGEX = /^ceramic:\/\/|^\/ceramic\//
+function validateDocId (docId: string): string {
+  const match = docId.match(PREFIX_REGEX)
+  return match ? match[0] : null
+}
 
 program
   .command('daemon')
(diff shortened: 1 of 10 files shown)


packages/ceramic-cli/src/index.ts
```diff
@@ -35,6 +42,10 @@ program
   .command('show <docId> [<anchor>]')
   .description('Show the content of a document')
   .action(async (docId) => {
+    if (!validateDocId(docId)) {
+      console.error(`Invalid docId: ${docId}`)
```
simonovic86, line 46: maybe exit with `process.exit(1)`

packages/ceramic-cli/src/index.ts
```diff
@@ -1,8 +1,15 @@
 import CeramicDaemon from './ceramic-daemon'
 import CeramicClient from '@ceramicnetwork/ceramic-http-client'
 import program from 'commander'
+import { serializeState } from './utils'
 
 
+const PREFIX_REGEX = /^ceramic:\/\/|^\/ceramic\//
+function validateDocId (docId: string): string {
```
msterle, line 8: It shouldn't matter here because it's not invoked frequently, but regexes are very inefficient (and hard to read), so often a more verbose method is more efficient.

packages/ceramic-cli/src/index.ts
oed: yeah, good point. We can update this as part of #9

packages/ceramic-cli/src/index.ts
oed: Good point. We could update this to two `.startsWtih` if we need more clarity later.

PR #27, opened 2020-04-13
Chore/add missing tests

oed:
This adds tests for the tileHandler and interop tests between core and http client.


packages/ceramic-http-client/package.json
```diff
@@ -27,7 +27,9 @@
     "clean": "rm -rf ./lib"
   },
   "dependencies": {
+    "@types/lodash.clonedeep": "^4.5.6",
```
simonovic86, line 30: we should move it to `devDependencies`

PR #34, opened 2020-04-14
add cli dockerfile, add ipfs-api daemon option

msterle:
Added a dockerfile for the ceramic daemon, and implemented the the CLI option to set the IPFS API URL


oed:
Try to use semantic commit messages btw (feat, chore, fix, etc)

PR #37, opened 2020-04-15
fix: on createDocument, return copy in _docmap

msterle:
Ensure that even if a document already existed, we always refer to the same one.

Had added cleanup code, but removed because the dispatcher assumes only one document per docId, so cleaning up one copy remove all copies from the dispatcher's list.

### DIFF
--- a/packages/3id-did-resolver/package.json
+++ b/packages/3id-did-resolver/package.json
@@ -35,7 +35,6 @@
     "@typescript-eslint/eslint-plugin": "^2.19.0",
     "@typescript-eslint/parser": "^2.19.0",
     "babel-jest": "^25.1.0",
-    "did-resolver": "^1.1.0",
     "eslint": "^6.8.0",
     "eslint-plugin-jest": "^23.8.2",
     "jest": "^25.1.0",
@@ -45,5 +44,8 @@
     "transformIgnorePatterns": [
       "/node_modules(?!/did-jwt)/"
     ]
+  },
+  "dependencies": {
+    "did-resolver": "^1.1.0"
   }
 }
--- a/packages/3id-did-resolver/package.json
+++ b/packages/3id-did-resolver/package.json
@@ -38,14 +38,12 @@
     "eslint": "^6.8.0",
     "eslint-plugin-jest": "^23.8.2",
     "jest": "^25.1.0",
-    "tmp-promise": "^2.0.2"
+    "tmp-promise": "^2.0.2",
+    "did-resolver": "^1.1.0"
   },
   "jest": {
     "transformIgnorePatterns": [
       "/node_modules(?!/did-jwt)/"
     ]
-  },
-  "dependencies": {
-    "did-resolver": "^1.1.0"
   }
 }
--- a/packages/ceramic-cli/package.json
+++ b/packages/ceramic-cli/package.json
@@ -39,12 +39,12 @@
     "ipfs-http-client": "^42.0.0"
   },
   "devDependencies": {
-    "@babel/core": "^7.8.4",
+    "@babel/core": "^7.9.0",
     "@babel/preset-env": "^7.8.4",
-    "@babel/preset-typescript": "^7.8.3",
+    "@babel/preset-typescript": "^7.9.0",
     "@types/events": "^3.0.0",
     "@types/express": "^4.17.2",
-    "@types/node": "^13.7.4",
+    "@types/node": "^13.9.5",
     "@types/node-fetch": "^2.5.5",
     "@typescript-eslint/eslint-plugin": "^2.19.0",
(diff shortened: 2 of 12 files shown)


oed:
Lmk if you need any pointers to fix the tests.

packages/ceramic-cli/package.json
```diff
@@ -39,12 +39,12 @@
...
   "devDependencies": {
-    "@babel/core": "^7.8.4",
+    "@babel/core": "^7.9.0",
     "@babel/preset-env": "^7.8.4",
-    "@babel/preset-typescript": "^7.8.3",
+    "@babel/preset-typescript": "^7.9.0",
     "@types/events": "^3.0.0",
     "@types/express": "^4.17.2",
-    "@types/node": "^13.7.4",
+    "@types/node": "^13.9.5",
```
oed, line 47: ty!

packages/ceramic-cli/src/__tests__/ceramic-daemon.test.ts
```diff
@@ -64,17 +64,10 @@ describe('Ceramic interop: core <> http-client', () => {
...
-    expect(doc1.state.log.length).toEqual(2)
-    expect(doc1.state.anchored).toBeGreaterThan(0)
-    const doc2 = await client.createDocument({ test: 123 }, DOCTYPE_TILE)
-    await anchorUpdate(doc2)
-    expect(doc2.state.log.length).toEqual(2)
-    expect(doc2.state.anchored).toBeGreaterThan(0)
+    const doc = await core.createDocument({ test: 123 }, DOCTYPE_TILE)
+    await anchorUpdate(doc)
+    expect(doc.state.log.length).toEqual(2)
+    expect(doc.state.anchored).toBeGreaterThan(0)
```
oed, line 70: Actually can you just add a test for the `client` that creates a different document?

packages/ceramic-http-client/package.json
```diff
@@ -27,19 +27,19 @@
     "clean": "rm -rf ./lib"
   },
   "dependencies": {
+    "@types/node-fetch": "^2.5.5",
```
oed, line 30: This should be in dev-deps.

packages/3id-did-resolver/package.json
```diff
@@ -45,5 +44,8 @@
     "transformIgnorePatterns": [
       "/node_modules(?!/did-jwt)/"
     ]
+  },
+  "dependencies": {
+    "did-resolver": "^1.1.0"
```
oed, line 49: Why was this moved to regular deps?

packages/ceramic-http-client/package.json
msterle: We use `node-fetch` in production code for this package, so the types are needed as well

packages/3id-did-resolver/package.json
msterle: https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/3id-did-resolver/src/index.ts#L1

packages/ceramic-http-client/package.json
oed: hm, not sure I follow. The types should only be needed when compiling?

packages/3id-did-resolver/package.json
oed: We only use the types so we shouldn't need to have it as a dep. What command are you using the `--production` flag on? And when do we need to do that? Usually we build before we publish to npm.

oed: Still confused by this.

msterle: @oed  Without this, when the package is imported into `3box-js` and the example server is run, I get an error in the browser
```
Uncaught (in promise) TypeError: did_resolver_1.Resolver is not a constructor
    at new TileHandler (tileHandler.js:29)
    at new Ceramic (ceramic.js:39)
    at Function.eval (ceramic.js:59)
    at Generator.next (<anonymous>)
    at eval (ceramic.js:8)
    at new Promise (<anonymous>)
    at __awaiter (ceramic.js:4)
    at Function.create (ceramic.js:57)
    at Box._callee$ (3box.js:155)
    at tryCatch (runtime.js:45)
```

This is possibly because of the different versions of `did-resolver` used by 3box-js and ceramic. It might be only because I am volume mounting the package instead of using a published version, but since this was moved into dev dependencies since the last published version, it looked like a mistake. Can you please publish the latest version so we can test it?

msterle: Nevermind, this error was being caused by the combination of lerna hoisting and docker volume mounting - just not going to hoist in local development.

oed: ok, I guess we can move it to here for now then. Let's merge this then make a new release!

PR #64, opened 2020-04-21
bug(core): emit processing state

simonovic86:
- emit `PROCESSING` event for anchor status


simonovic86:
> 👍 I would use fix() instead of bug() next time :)

got it :)

PR #67, opened 2020-04-21
feat(core): add id property to signed records

oed:
This adds an id property to all signed records. The `id` property contains the CID of the genesis record.

There should also be an update to the specs repo to reflect this. 

We should also consider adding an id property to the anchor records.

This change should enable some simplification of the anchor service. Namely that we don't need to pass a docId in the request.


packages/ceramic-core/src/document.ts
```diff
@@ -236,7 +236,7 @@ class Document extends EventEmitter {
   async _verifyAnchorRecord (record: AnchorRecord): Promise<AnchorProof> {
     const proofRecord = await this.dispatcher.retrieveRecord(record.proof)
 
-    let prevRootPathRecord;
```
simonovic86, line 239: I'll setup the `WebStorm` config to remove `;` as well :)

PR #68, opened 2020-04-21
fix(core): add prev verification for anchor record

simonovic86:
- add `prev` verification for anchor record


simonovic86:
> Should be good to merge. Would you mind just squashing the commits as you do that? :)

I'll squash them definitely. These commits represent the shameful history of trying to get it work :)

PR #86, opened 2020-05-07
Feat/43 spec account model

msterle:


packages/ceramic-account/src/did.ts
```diff
@@ -0,0 +1,33 @@
+import type { CeramicApi, CeramicDocument } from './types'
+
+class DIDDocument {
+  constructor (public ceramicDoc: CeramicDocument, private _ceramic: CeramicApi) { }
+
+  get did (): string {
+    return 'did:3:' + this.ceramicDoc.state.log[0]
+  }
+
+  async linkAccount () { }
```
oed, line 10: Not sure why this is in here.

packages/ceramic-account/src/did.ts
```diff
@@ -0,0 +1,33 @@
+import type { CeramicApi, CeramicDocument } from './types'
+
+class DIDDocument {
+  constructor (public ceramicDoc: CeramicDocument, private _ceramic: CeramicApi) { }
```
oed, line 4: This class looks quite nice. It might become the main interface for 3ID doctypes in the future!

packages/ceramic-account/src/types.ts
```diff
@@ -0,0 +1,9 @@
+import type Ceramic from '@ceramicnetwork/ceramic-core'
+import type CeramicClient from '@ceramicnetwork/ceramic-http-client'
+import type CoreDocument from '@ceramicnetwork/ceramic-core/lib/document'
+import type ClientDocument from '@ceramicnetwork/ceramic-http-client/lib/document'
+
+export type CeramicApi = Ceramic | CeramicClient;
+
+// Because Ceramic and CeramicClient use different Document definition:
+export type CeramicDocument = CoreDocument | ClientDocument;
```
oed, line 9: 😹

packages/test/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "@ceramicnetwork/test",
+  "version": "0.1.0",
+  "description": "",
```
oed, line 4: Why is this a separate package? Couldn't it just be tests in the account package?

msterle: This is a dummy package that won't end up in the final PR, but I'm using it to act as a reference implementation of a package that would use ceramic-account.

packages/ceramic-account/src/did.ts
msterle: This is something I'm still working on, but I'm thinking that a class should exclusively own the ceramic documents it represents, and be the only entity writing to them - this might even become enforced by setting them private and exposing the required functionality for fetching info from the document itself. (This `linkAccount` function would be to write the link to the `account` tile in the `did` document, instead of it being done by the `ThreeIDAccount` class)

oed: Seems like this should go in the `Account Links Tile` though?

PR #88, opened 2020-05-08
feat: allow legacy did docs to be used as genesis records

msterle:


packages/ceramic-core/src/ceramic.ts
```diff
@@ -41,6 +41,7 @@ class Ceramic {
       '3id': new ThreeIdHandler(),
       'tile': new TileHandler(this),
       'account-link': new AccountLinkHandler(),
+      undefined: new ThreeIdHandler()
```
msterle, line 44: This is temporary, would like to replace this object mapping with a function or something a bit smarter to allow for default values

packages/ceramic-cli/src/ceramic-daemon.ts
```diff
@@ -35,6 +46,10 @@ class CeramicDaemon {
     app.get(toApiPath('/show/ceramic/:cid'), this.show.bind(this))
     app.get(toApiPath('/state/ceramic/:cid'), this.state.bind(this))
     app.post(toApiPath('/change/ceramic/:cid'), this.change.bind(this))
+    if (DEBUG) {
+      app.use(logErrors)
+    }
+    app.use(sendErrorResponse)
```
oed, line 52: Nice!

packages/ceramic-cli/src/ceramic-daemon.ts
```diff
@@ -21,6 +22,16 @@ interface CreateOpts {
   anchorServiceUrl?: string;
 }
 
+function logErrors (err: Error, req: Request, res: Response, next: NextFunction) {
+  console.error(err)
+  next(err)
+}
```
oed, line 28: Should we really modify the behavior of the daemon if we log the errors? Why not both log the error and send the response?

packages/ceramic-core/src/doctypes/threeIdHandler.ts
```diff
@@ -20,16 +20,36 @@ class ThreeIdHandler extends DoctypeHandler {
...
+      return {
+        doctype: DOCTYPE,
+        owners: record.owners,
+        content: record.content,
+        nextContent: null,
+        signature: SignatureStatus.GENESIS,
+        anchorStatus: AnchorStatus.NOT_REQUESTED,
+        log: [cid]
+      }
+    } else if (record['@context'] === "https://w3id.org/did/v1") {
```
oed, line 33: 👍

packages/ceramic-cli/src/ceramic-daemon.ts
msterle: This does both (if `DEBUG` is set) - when we call `next(err)` at the end, it gets passed to the next handler, in the order they are registered.

PR #95, opened 2020-05-12
feat(core): add local pinning

simonovic86:
- add local pinning functionality
- hide `anchorScheduledFor` if the anchor has failed/completed


packages/ceramic-core/src/pin/mock/mock-pinning-service.ts
```diff
@@ -0,0 +1,53 @@
+import { DocState } from "../../document"
+import PinningService from "../pinning-service"
+
+export default class MockPinningService implements PinningService {
```
oed, line 4: Is this mock pinning service really needed? Seems like it could be used for testing, but other than that it would only be bloating the codebase.

packages/ceramic-core/src/pin/level/level-pinning-service.ts
```diff
@@ -0,0 +1,134 @@
+import levelup from "levelup";
+import leveldown from "leveldown";
```
oed, line 3: Does leveldown work in browsers? We need to be careful with the dbs that we use so that we get full compatibility.

packages/ceramic-core/src/pin/level/level-pinning-service.ts
```diff
@@ -0,0 +1,134 @@
...
+        }
+    }
+
+    /**
+     * Is document pinned locally?
+     * @param docId - Document ID
+     */
+    async isDocPinned(docId: string): Promise<boolean> {
+        const state = await this.loadState(docId);
+        return state != null;
```
oed, line 116:
```suggestion
        return Boolean(state)
```

packages/ceramic-http-client/src/document.ts
```diff
@@ -1,21 +1,7 @@
...
-      method: 'post',
-      body: JSON.stringify(payload),
-      headers: { 'Content-Type': 'application/json' }
-    }
-  }
-  const res = await (await fetch(url, opts)).json()
-  if (res.error) throw new Error(res.error)
-  return res
-}
+import { fetchJson } from "./utils"
```
oed, line 4: 👍

packages/ceramic-core/src/ceramic.ts
```diff
@@ -88,6 +93,20 @@ class Ceramic {
...
+  }
+
+  async unpinDocument (docId: string): Promise<void> {
+    const document = await this.loadDocument(docId)
+    await this.dispatcher.pinningService.rm(document)
+  }
+
+  async listPinned (docId?: string): Promise<any> {
+    return this.dispatcher.pinningService.ls(docId)
+  }
```
oed, line 108: Looks like api now is:
* `ceramic.pinDocment`
* `ceramic.unpinDocment`
* `ceramic.listPinned`

We want to mirror the [ipfs api](https://github.com/ipfs/js-ipfs/blob/master/docs/core-api/PIN.md), so it should look like this:

* `ceramic.pin.add(docId)`
* `ceramic.pin.rm(docId)`
* `ceramic.pin.ls(docId?)`

packages/ceramic-core/src/ceramic.ts
```diff
@@ -66,7 +67,11 @@ class Ceramic {
   }
 
   static async create(ipfs: Ipfs.Ipfs, config: CeramicConfig = {}): Promise<Ceramic> {
-    const dispatcher = new Dispatcher(ipfs)
+    const pinningServiceFactory = new PinningServiceFactory(ipfs, config)
+    const pinningService = pinningServiceFactory.get()
+
+    const dispatcher = new Dispatcher(ipfs, pinningService)
```
oed, line 73: Why do you pass the pinningService into the dispatcher? Any reason to bundle it like this?

packages/ceramic-http-client/src/ceramic-http-client.ts
```diff
@@ -36,6 +37,24 @@ class CeramicClient {
...
+  }
+
+  async _listPinned (docId?: string): Promise<any> {
+    let url = this._apiUrl + '/pin/ls'
+    if (docId !== undefined) {
+      url += docId
+    }
+    return await fetchJson(url)
+  }
+
```
oed, line 57: Not sure why the methods are different here from what they look like in  `ceramic-core`.

In general `ceramic-core` and `ceramic-http-client` should have the same api (see comments above about the pinning api).

packages/ceramic-http-client/src/document.ts
```diff
@@ -109,6 +103,14 @@ class Document extends EventEmitter {
...
   }
 
+  async pinDocument (): Promise<any> {
+    return await fetchJson(this._apiUrl + '/pin/add' + this.id)
+  }
+
+  async unpinDocument (): Promise<any> {
+    return await fetchJson(this._apiUrl + '/pin/rm' + this.id)
+  }
+
```
oed, line 113: Why are these here? We don't have any similar methods in the document class of ceramic-core.

packages/ceramic-cli/src/index.ts
```diff
@@ -104,24 +111,57 @@ program
   })
 
 program
-  .command('change <docId> [content]')
-  .option('--owners <owners>', 'Change owner of this document (only 3ID)')
-  .description('Update the content of a document')
-  .action(async (docId, content, { owners }) => {
+  .command('pin_add <docId>')
+  .description('Pin the document locally')
```
oed, line 115: Looking at the api here we have:
* `ceramic pin_add <docId>`
* `ceramic pin_rm <docId>`
* `ceramic pin_list [<docId>]`

Can we change it to this? (might be a bit more difficult, but will be much nicer :) 
* `ceramic pin add <docId>`
* `ceramic pin rm <docId>`
* `ceramic pin ls [<docId>]`

packages/ceramic-core/src/pin/mock/mock-pinning-service.ts
oed: In the tests we can probably just automock it using jest.

packages/ceramic-core/src/pin/pinning-service-factory.ts
```diff
@@ -0,0 +1,34 @@
...
+import PinningService from "./pinning-service"
+import MockPinningService from "./mock/mock-pinning-service"
+import LevelPinningService from "./level/level-pinning-service"
+
+/**
+ * PinningService factory class
+ */
+export default class PinningServiceFactory {
+    private readonly mockPinningService: MockPinningService
+    private readonly levelPinningService: LevelPinningService
```
oed, line 14: General thought about the factory pattern (used here and for the anchor service). It seems like it might add more code bloat than what we actually get from it? In general I'm quite concerned with the size of our libraries and would like to keep the code base as minimal as possible while still being very readable. In the example here it seems like it would be quite straightforward to just construct the "pinningService" when we need it.

Related note: I think the name "pinning service" might be confusing since this is an internal interface in the ceramic-core codebase. In the future we will have actual pinning services that pin data as a service in the network. This might cause unecessary confusing. We should consider renaming this to something like "StateStore" or something similar that points to this being a local persistance layer.

packages/ceramic-cli/src/ceramic-daemon.ts
```diff
@@ -46,6 +47,10 @@ class CeramicDaemon {
     app.get(toApiPath('/show/ceramic/:cid'), this.show.bind(this))
     app.get(toApiPath('/state/ceramic/:cid'), this.state.bind(this))
     app.post(toApiPath('/change/ceramic/:cid'), this.change.bind(this))
+    app.get(toApiPath('/pin/add/ceramic/:cid'), this.pinDocument.bind(this))
+    app.get(toApiPath('/pin/rm/ceramic/:cid'), this.unpinDocument.bind(this))
+    app.get(toApiPath('/pin/ls/ceramic/:cid'), this.listPinned.bind(this))
+    app.get(toApiPath('/pin/ls'), this.listPinned.bind(this))
```
oed, line 53: Looks great 👍

oed:
Also note that `ipfs.pin.ls` returns an async iterator: https://github.com/ipfs/js-ipfs/blob/master/docs/core-api/PIN.md

Does it make sense for us to do the same?

packages/ceramic-cli/src/index.ts
```diff
@@ -103,25 +110,61 @@ program
...
-    }
-    ceramic.close()
-  })
+const pin = program.command('pin')
+pin.description('Ceramic local pinning API')
+
+pin
+    .command('add <docId>')
+    .description('Pin document')
+    .action(async (docId) => {
```
oed, line 119: Cool, I didn't know you could do it in this way :D

packages/ceramic-core/src/store/level-state-store.ts
```diff
@@ -0,0 +1,171 @@
...
+                    recursive: false,
+                })
+            })
+            await Promise.all(pinPromises);
+        }
+    }
+
+    /**
+     * Serializes document state
+     * TODO move to common utility
```
oed, line 46: 👍 Let's do this as part of the refactor!

packages/ceramic-core/src/store/level-state-store.ts
```diff
@@ -0,0 +1,171 @@
...
+        });
+        await Promise.all(pinPromises);
+        return this.store.del(docId)
+    }
+
+    /**
+     * List pinned document
+     * @param docId - Document ID
+     */
+    async ls(docId?: string): Promise<string[]> {
```
oed, line 139: I noticed that https://github.com/ipfs/js-ipfs/blob/master/docs/core-api/PIN.md#pinls returns an async iterable instead of an array. Does it maybe make sense for us to do the same?

packages/ceramic-http-client/src/ceramic-http-client.ts
```diff
@@ -19,6 +33,24 @@ class CeramicClient {
...
+      ls: async (docId?: string): Promise<any> => {
+        let url = this._apiUrl + '/pin/ls'
+        if (docId !== undefined) {
+          url += docId
+        }
+        return await fetchJson(url)
+      }
+    }
+  }
+
```
oed, line 53: Nice 👍

packages/ceramic-core/src/store/level-state-store.ts
```diff
@@ -0,0 +1,171 @@
...
+import Level from 'level-ts';
+
+import type Ipfs from 'ipfs'
+import Document, { AnchorStatus, DocState } from "../document"
+import StateStore from "./state-store"
+
+/**
+ * Level backed State Store
+ */
+export default class LevelStateStore implements StateStore {
```
oed, line 11: We should probably have unit tests for this class?

packages/ceramic-cli/src/index.ts
simonovic86: yep :D

packages/ceramic-http-client/src/document.ts
simonovic86: Got it. Removed it :)

packages/ceramic-core/src/store/level-state-store.ts
simonovic86: yep! 👍

packages/ceramic-core/src/store/level-state-store.ts
simonovic86: of course! 👍

packages/ceramic-core/src/store/level-state-store.ts
simonovic86: makes sense

packages/ceramic-core/src/pin/level/level-pinning-service.ts
```diff
@@ -0,0 +1,134 @@
...
+
+        if (pinOnIpfs) {
+            const { log } = state;
+            const pinPromises = log.map(async (cid) => {
+                return this.ipfs.pin.add(cid.toString(), {
+                    recursive: false,
+                })
+            })
+            await Promise.all(pinPromises);
+        }
```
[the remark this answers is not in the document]
simonovic86: Added pins 👍

packages/ceramic-core/src/store/level-state-store.ts
```diff
@@ -0,0 +1,217 @@
...
+
+                    let currentPath = ""
+                    for (const subPath of subPaths) {
+                        if (subPath.length === 0) {
+                            continue
+                        }
+                        currentPath += "/" + subPath
+
+                        // fetch and pin
+                        await this.ipfs.dag.get(record.proof, currentPath)
```
oed, line 60: Do we really need to call `dag.get` here? Doesn't `pin.add` do this automatically?

packages/ceramic-http-client/src/document.ts
oed, line 4: Can we try not to mix the use of `'` and `"` :)

simonovic86: yep 👍

packages/ceramic-core/src/store/level-state-store.ts
simonovic86: I assumed that we need to fetch it. If `pin.add` does that automatically, even better :)

simonovic86:
> Looks good. Anything else missing before we can merge this?

I'll just change the API for `ceramic.pin.ls` to return async iterator

packages/ceramic-core/src/store/level-state-store.ts
simonovic86: done

PR #151, opened 2020-07-15
feat(cli): remove hardcoded seed

simonovic86:
- remove hardcoded identity wallet seed from Ceramic CLI
- generate seed if there is no and save it to `~/.ceramic/config.json`

### DIFF
--- a/packages/ceramic-cli/package.json
+++ b/packages/ceramic-cli/package.json
@@ -38,7 +38,8 @@
     "commander": "^5.0.0",
     "express": "^4.17.1",
     "identity-wallet": "1.2.0-ceramic.1",
-    "ipfs-http-client": "^44.0.0"
+    "ipfs-http-client": "^44.0.0",
+    "tweetnacl": "^1.0.3"
   },
   "devDependencies": {
     "@babel/core": "^7.9.0",
--- a/packages/ceramic-cli/package.json
+++ b/packages/ceramic-cli/package.json
@@ -38,8 +38,7 @@
     "commander": "^5.0.0",
     "express": "^4.17.1",
     "identity-wallet": "1.2.0-ceramic.1",
-    "ipfs-http-client": "^44.0.0",
-    "tweetnacl": "^1.0.3"
+    "ipfs-http-client": "^44.0.0"
   },
   "devDependencies": {
     "@babel/core": "^7.9.0",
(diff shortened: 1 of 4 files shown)


packages/ceramic-cli/src/ceramic-daemon.ts
```diff
@@ -87,11 +100,41 @@ class CeramicDaemon {
...
+
+    await fs.mkdir(DEFAULT_CLI_CONFIG_PATH, { recursive: true })
+
+    console.log('Generating identity wallet seed...')
+    const config: CliConfig = {
+      seed: '0x' + Buffer.from(nacl.randomBytes(32)).toString('hex') // create new seed
+    }
+
+    await fs.writeFile(fullCliConfigPath, JSON.stringify(config, null, 2))
+    console.log('Identity wallet seed generated')
```
oed, line 134: One console log is probably enough here? Let's keep the second one?

packages/ceramic-cli/package.json
```diff
@@ -38,7 +38,8 @@
     "commander": "^5.0.0",
     "express": "^4.17.1",
     "identity-wallet": "1.2.0-ceramic.1",
-    "ipfs-http-client": "^44.0.0"
+    "ipfs-http-client": "^44.0.0",
+    "tweetnacl": "^1.0.3"
```
oed, line 42: Doesn't nodejs have a built in way to generate randomness? Seems exessive to import tweetnacl just for that.

packages/ceramic-cli/src/ceramic-daemon.ts
simonovic86: makes sense! 👍

packages/ceramic-cli/package.json
simonovic86: yep, but I thought there's a reason why the identity wallet uses `tweetnacl` for randomness. I'll change it! 👍

oed: I think IDW uses it because it can be run in browsers as well. IDW also uses tweetnacl for other things :)

simonovic86: makes sense 👍

packages/ceramic-cli/src/ceramic-daemon.ts
```diff
@@ -87,11 +100,41 @@ class CeramicDaemon {
...
+    if (exists) {
+      const configJson = await fs.readFile(fullCliConfigPath, { encoding: 'utf8' })
+      return JSON.parse(configJson)
+    }
+
+    await fs.mkdir(DEFAULT_CLI_CONFIG_PATH, { recursive: true })
+
+    console.log('Generating identity wallet seed...')
+    const config: CliConfig = {
+      seed: '0x' + Buffer.from(crypto.randomBytes(32)).toString('hex') // create new seed
```
oed, line 130: Nice :+1:

PR #185, opened 2020-07-26
node-fetch -> cross-fetch

elmariachi111:
so we can call the anchoring service from a browser

executed the tests & ran the cli / daemon + an anchoring node. Seems to work like that ;) node-fetch unfortunately doesn't support isomorphic browser & node fetching, cross-fetch falls back to a whatwg brower fetch when it detects a windows object (think it's also officially used by js-ipfs)

### DIFF
--- a/packages/ceramic-core/package-lock.json
+++ b/packages/ceramic-core/package-lock.json
@@ -1198,50 +1198,53 @@
 			"dev": true
 		},
 		"@ceramicnetwork/3id-did-resolver": {
-			"version": "0.1.5",
-			"resolved": "https://registry.npmjs.org/@ceramicnetwork/3id-did-resolver/-/3id-did-resolver-0.1.5.tgz",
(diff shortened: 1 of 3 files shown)


simonovic86:
nice 🚀 

good point! I'll merge this PR and I've already opened an [issue](https://github.com/ceramicnetwork/js-ceramic/issues/186)

thanks!

michaelsena:
Good point and thanks for the contribution @elmariachi111. 😄 On the surface looks good. I think this will increase the bundle size a bit. @simonovic86 thoughts?

simonovic86:
> Good point and thanks for the contribution @elmariachi111. 😄 On the surface looks good. I think this will increase the bundle size a bit. @simonovic86 thoughts?

makes sense! I've merged the PR and opened a [story](https://github.com/ceramicnetwork/js-ceramic/issues/186). This definitely solves the compatibility issue. As for the bundle size, I'll make a diff and if there's an issue we'll switch to some of the alternatives.

michaelsena:
Seems like the alternatives are around the same size: https://www.npmtrends.com/cross-fetch-vs-isomorphic-fetch-vs-node-fetch-vs-whatwg-fetch

PR #188, opened 2020-07-30
Doctype schema(s)

simonovic86:
- introduce schema support in ceramic-core
- implement script for schema<->types generation
- implement schema unit tests
- refactor `Doctype`(s) to match CIPs (metadata, etc)
- replace `node-fetch` with `cross-fetch`
- fix lint


packages/ceramic-doctype-three-id/src/three-id-doctype.ts
```diff
@@ -50,17 +55,15 @@ export class ThreeIdDoctype extends Doctype {
...
+    static async makeGenesis(params: Record<string, any>): Promise<Record<string, any>> {
+        const { content, metadata } = params
 
-        if (!owners) {
+        if (!metadata?.owners) {
             throw new Error('The owner of the 3ID needs to be specified')
         }
         return {
-            doctype: DOCTYPE, owners, content
+            doctype: DOCTYPE, metadata, content
```
oed, line 66: To avoid confusion, the metadata should be put into the `header` of records: https://github.com/ceramicnetwork/specs#document-records

packages/ceramic-doctype-three-id/src/three-id-doctype.ts
```diff
@@ -82,7 +85,7 @@ export class ThreeIdDoctype extends Doctype {
         }
 
         const patch = jsonpatch.compare(doctype.state.content, newContent)
-        const record: any = { owners: newOwners, content: patch, prev: doctype.head, id: doctype.state.log[0] }
+        const record: any = { metadata: newMetadata, content: patch, prev: doctype.head, id: doctype.state.log[0] }
```
oed, line 88: Here we should not neccessarily include a whole new metadata object. Instead we should include in the `header` only the properties of the metadata that we want to update.

packages/ceramic-doctype-tile/src/tile-doctype.ts
```diff
@@ -60,24 +59,24 @@ export class TileDoctype extends Doctype {
...
 
-        const record = { doctype: DOCTYPE, owners, content, unique }
+        const metadata = params.metadata? params.metadata : {}
+        const { owners } = metadata
+        if (!owners) {
+            metadata.owners = [context.user.DID]
+        }
+
+        const { content } = params
+        const record = { doctype: DOCTYPE, data: content, metadata, unique }
```
oed, line 79: Same thing here, lets use the `header` property in the *records*, and use the `metadata` property in *DocState*.

packages/ceramic-core/src/document.ts
```diff
@@ -423,8 +453,31 @@ class Document extends EventEmitter {
...
+      const schemaDoc = await ceramicApi.loadDocument(schemaDocId)
+      return schemaDoc.content
+    }
+    return null
+  }
+
   get content (): any {
-    return this._doctype.state.nextContent || this._doctype.state.content
+    const { next, content } = this._doctype.state
+    return next? next.content || content : content
```
PaulLeCam, line 480: Not an issue but could be simplified as `next?.content ?? content`

packages/ceramic-core/src/document.ts
```diff
@@ -440,7 +493,7 @@ class Document extends EventEmitter {
   }
 
   get owners (): string[] {
```
PaulLeCam, line 495: Looks like this should be `string[] | undefined` if the state or metadata object is undefined.

packages/ceramic-core/src/document.ts
oed: Do you mean `next?.content || content`? Or what does the `??` operator do?

packages/ceramic-core/src/document.ts
oed: Metadata should never be undefined, neither should `owners`.

packages/ceramic-common/schemas/basic-profile.json
```diff
@@ -0,0 +1,80 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "BasicProfile",
+  "type": "object",
+  "properties": {
+    "name": {
```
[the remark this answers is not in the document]
simonovic86: I put them here on purpose because these schemas are the ones described in our CIPs and will be used in the future (Ceramic Explorer, etc.)

packages/ceramic-common/src/doctype.ts
```diff
@@ -43,181 +44,50 @@ export interface AnchorProof {
...
 }
 
+/**
+ * Document metadata
+ */
+export interface DocMetadata {
+    owners?: Array<string>;
+    schema?: string;
+    tags?: Array<string>;
+}
```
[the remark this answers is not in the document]
simonovic86: Yep, I wanted to add something like that :) I'll change that 👍

packages/ceramic-common/src/utils/doctype-utils.ts
```diff
@@ -0,0 +1,173 @@
+import ajv from "ajv"
```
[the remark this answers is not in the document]
simonovic86: I'll take a look at it, tnx!

packages/ceramic-core/src/__tests__/ceramic-api.test.ts
```diff
@@ -0,0 +1,172 @@
...
+
+    const schemaDoc = await ceramic.createDocument<TileDoctype>(DOCTYPE_TILE, { content: stringMapSchema, metadata: { owners: [owner] }})
+
+    const tileDocParams: TileParams = {
+      metadata: {
+        schema: schemaDoc.id
+      },
+      content: { a: 1 },
+      owners: [owner]
+    }
```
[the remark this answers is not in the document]
simonovic86: yep, nice catch!

packages/ceramic-doctype-three-id/src/three-id-doctype.ts
simonovic86: I'll refactor it, tnx!

packages/ceramic-doctype-three-id/src/three-id-doctype.ts
simonovic86: 👍

packages/ceramic-doctype-tile/src/tile-doctype.ts
simonovic86: 👍

packages/ceramic-core/src/document.ts
simonovic86: @oed 👍

packages/ceramic-core/src/document.ts
PaulLeCam: It's similar to `||` but only if the left-hand side is `null` or `undefined`.
I find it safer than `||` because it treats falsy values such as `false`, `0` and `''` as expected values on the left-hand side.

packages/ceramic-core/src/document.ts
PaulLeCam: There's a mismatch between the type annotations and the code then, the `?` shouldn't be necessary in `this._doctype.state?.metadata?.owners` if these objects can never be undefined.

packages/ceramic-common/schemas/basic-profile.json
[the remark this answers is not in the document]
oed: Right, in the future these will just be stored in Ceramic :)

packages/ceramic-core/src/document.ts
oed: Cool!

packages/ceramic-core/src/document.ts
oed: Hm, afaik the `state` should never be undefined either? I might be missing something in the implementation details though?

packages/account-template/src/index.ts
```diff
@@ -15,7 +15,7 @@ class Keychain {
       "auth-data": [],
       "legacy-data": []
     }
-    const ceramicDoc = await ceramic.createDocument<TileDoctype>('tile', { content: genesisContent, owners: [owner] }, { isUnique: true })
+    const ceramicDoc = await ceramic.createDocument<TileDoctype>('tile', { content: genesisContent, metadata: { owners: [owner] } }, { isUnique: true })
```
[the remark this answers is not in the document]
simonovic86: we should update [Doctypes](https://github.com/ceramicnetwork/CIP/issues/37) then. I'll update it 👍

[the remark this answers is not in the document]
oed: This is why it would be nice with the generic property :)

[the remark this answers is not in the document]
simonovic86: yep, added the generic one as well :)

packages/ceramic-common/schemas/basic-profile.json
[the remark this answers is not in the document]
simonovic86: I'll put them in tests then 👍

packages/ceramic-core/src/document.ts
simonovic86: nope, `state` should be always defined. Fixed it 👍

packages/ceramic-common/src/doctype.ts
```diff
@@ -232,23 +102,27 @@ export abstract class Doctype extends EventEmitter {
...
     get content(): any {
-        return cloneDeep(this.state.content)
+        return cloneDeep(this._state.content)
+    }
+
+    get metadata(): DocMetadata {
+        return this._state?.metadata? cloneDeep(this._state.metadata) : null
     }
 
     get owners(): Array<string> {
```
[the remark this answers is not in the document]
simonovic86: `metadata` should always be defined 👍

packages/ceramic-common/src/doctype.ts
```diff
@@ -232,23 +102,27 @@ export abstract class Doctype extends EventEmitter {
...
     get doctype(): string {
         return this._state.doctype
     }
 
     get content(): any {
-        return cloneDeep(this.state.content)
+        return cloneDeep(this._state.content)
+    }
+
+    get metadata(): DocMetadata {
```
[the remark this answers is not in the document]
simonovic86: `metadata` should always be defined 👍

simonovic86:
> Few changes needed in the metadata refactor.
> 
> For discussion:
> I also think that we should do schema validation when `change` is called in the doctype. This means that when you update a document you need to pass valid data. I think this is the place where user error is most likely. If we don't validate here we may run into trouble? Either that, or we have to enforce that schema validation is always done. Otherwise schemas become quite ineffective?

yeah, let's discuss more about enforcing schemas before calling the Ceramic API. I have some ideas.

packages/ceramic-core/src/document.ts
simonovic86: nice!

packages/ceramic-common/src/utils/doctype-utils.ts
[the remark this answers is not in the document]
simonovic86: Makes sense 👍  I'll take a look more deeply into the alternatives and I'll create a separate PR if there's smaller/better one.

packages/ceramic-http-client/src/ceramic-http-client.ts
```diff
@@ -38,10 +37,10 @@ class CeramicClient implements CeramicApi {
         const result = await fetchJson(url)
         const { pinnedDocIds } = result
         return {
-          [Symbol.asyncIterator]() {
+          [Symbol.asyncIterator](): any {
```
PaulLeCam, line 40: `AsyncIterable<string>` no?

packages/ceramic-http-client/src/ceramic-http-client.ts
```diff
@@ -38,10 +37,10 @@ class CeramicClient implements CeramicApi {
...
             let index = 0
             return {
-              next() {
+              next(): any {
```
PaulLeCam, line 43: Should be `Promise<IteratorResult<string>>` I think

simonovic86: 👍

packages/ceramic-http-client/src/ceramic-http-client.ts
simonovic86: it should actually be `AsyncIterator<string, any, undefined>` if I'm not mistaken :)

PR #195, opened 2020-08-04
CLI schema support, doctype schema fixes

simonovic86:
- add CLI schema support
- add http-client schema support
- fix doctype schema handling
- add more tests

### DIFF
--- a/packages/ceramic-cli/src/ceramic-cli-utils.ts
+++ b/packages/ceramic-cli/src/ceramic-cli-utils.ts
@@ -151,6 +151,7 @@ export default class CeramicCliUtils {
             }
         } catch (e) {
             console.error(e)
+            return
         }
 
         return CeramicCliUtils.createDoc('tile', content, owners, onlyGenesis, isUnique)
@@ -176,6 +177,7 @@ export default class CeramicCliUtils {
             }
         } catch (e) {
             console.error(e)
+            return
         }
 
         return CeramicCliUtils.change(schemaDocId, content, owners, null)
--- a/packages/ceramic-cli/src/ceramic-cli-utils.ts
+++ b/packages/ceramic-cli/src/ceramic-cli-utils.ts
@@ -1,5 +1,5 @@
 import CeramicClient from "@ceramicnetwork/ceramic-http-client"
-import { CeramicApi, DoctypeUtils } from "@ceramicnetwork/ceramic-common/lib"
+import { CeramicApi, DoctypeUtils } from "@ceramicnetwork/ceramic-common"
 
 const PREFIX_REGEX = /^ceramic:\/\/|^\/ceramic\//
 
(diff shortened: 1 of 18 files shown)


packages/ceramic-cli/src/ceramic-cli-utils.ts
```diff
@@ -0,0 +1,310 @@
...
+     */
+    static async schemaCreateDoc(content: string, owners: string, onlyGenesis: boolean, isUnique: boolean): Promise<void> {
+        try {
+            const schemaObj = JSON.parse(content)
+            if (!DoctypeUtils.isSchemaValid(schemaObj)) {
+                console.error('Invalid schema')
+                return
+            }
+        } catch (e) {
+            console.error(e)
```
PaulLeCam, line 153: Shouldn't there be a `return` statement here? Otherwise it seems in case the `content` is not valid JSON the doc would still get created, no?

packages/ceramic-cli/src/ceramic-cli-utils.ts
```diff
@@ -0,0 +1,310 @@
...
+        }
+
+        try {
+            const schemaObj = JSON.parse(content)
+            if (!DoctypeUtils.isSchemaValid(schemaObj)) {
+                console.error('Invalid schema')
+                return
+            }
+        } catch (e) {
+            console.error(e)
```
PaulLeCam, line 178: Same as previous comment

packages/ceramic-cli/src/ceramic-cli-utils.ts
simonovic86: nice catch!

packages/ceramic-cli/src/ceramic-cli-utils.ts
simonovic86: 👍

packages/ceramic-cli/src/ceramic-cli-utils.ts
```diff
@@ -0,0 +1,312 @@
+import CeramicClient from "@ceramicnetwork/ceramic-http-client"
+import { CeramicApi, DoctypeUtils } from "@ceramicnetwork/ceramic-common"
+
+const PREFIX_REGEX = /^ceramic:\/\/|^\/ceramic\//
+
+/**
+ * Ceramic CLI utility functions
+ */
+export default class CeramicCliUtils {
+
```
oed, line 10: Nice that you pulled these out. Cli class is very clean now :)

packages/ceramic-cli/src/index.ts
```diff
@@ -1,163 +1,109 @@
...
+schemas
+    .command('change <docId> <new-content>')
+    .option('--owners <owners>', 'Change owner of this document (only 3ID)')
+    .description('Update the content of a schema')
+    .action(async (docId, content, { owners }) => {
+        await CeramicCliUtils.schemaChangeDoc(docId, content, owners)
+    })
+
+schemas
+    .command('validate <schemaDocId> <new-content>')
```
oed, line 102: I think these make sense. So basically it just creates a tile and also validates that it's adding an actual schema?

packages/ceramic-cli/src/ceramic-cli-utils.ts
simonovic86: 🚀

packages/ceramic-cli/src/index.ts
simonovic86: yep, I think it's good for now, as well 👍

PR #199, opened 2020-08-05
Validate schema

simonovic86:
- validate schema
- print error messages in CLI

### DIFF
--- a/packages/ceramic-common/src/utils/doctype-utils.ts
+++ b/packages/ceramic-common/src/utils/doctype-utils.ts
@@ -163,7 +163,8 @@ export class DoctypeUtils {
      * @param schema - Doctype schema
      */
     static isSchemaValid(schema: object): boolean {
-        return this.validator.validateSchema(schema)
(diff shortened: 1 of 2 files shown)


packages/ceramic-common/src/utils/doctype-utils.ts
```diff
@@ -163,7 +163,8 @@ export class DoctypeUtils {
      * @param schema - Doctype schema
      */
     static isSchemaValid(schema: object): boolean {
-        return this.validator.validateSchema(schema)
+        this.validator.compile(schema) // throws an error on invalid schema
```
PaulLeCam, line 166: If it throws an error maybe this could be wrapped in a `try { ... return true } catch { ... return false }` statement so the caller doesn't have to catch the error explicitly?

simonovic86: The error thrown contains information about the errors. That's why its better to throw it.

PR #217, opened 2020-08-17
Changes for IDX lib

PaulLeCam:
There's a breaking change in `did-resolver` that causes the current issue with building `3id-did-resolver`, these changes should fix it.

It also adds dependencies on `dids` and `rpc-utils` in `ceramic-core` to make the types match with the interfaces expected by `idp2p`.


packages/ceramic-common/src/ceramic-api.ts
```diff
@@ -23,25 +25,17 @@ export interface PinApi {
...
-    'result': object;
-}
-
 /**
  * Describes DID provider instance
  */
-export interface DIDProvider {
-    send(jsonReq: object): JsonRpc2Response;
-}
+export type DIDProvider = RPCConnection
```
oed, line 31: +:+1:

PR #219, opened 2020-08-18
`pin ls` arguments are optional. No arguments means list all documents.

ukstv:
Without this change `pin ls` command emits an error on invalid docId.


oed (approved):
lgtm.

PR #226, opened 2020-08-19
Use dids package and idx field in 3ID doc

PaulLeCam:


packages/ceramic-common/src/ceramic-api.ts
```diff
@@ -23,25 +24,17 @@ export interface PinApi {
...
-    send(jsonReq: object): JsonRpc2Response;
-}
+export type { DIDProvider } from 'dids'
 
 /**
  * Describes Ceramic node API
  */
 export interface CeramicApi {
     pin: PinApi;
+    user?: DID;
```
simonovic86, line 37: do we need `user` here?

PaulLeCam: It's used in various places in IDX to check if the user is authenticated, access its `DID` to set the owner when creating documents, resolve the DID document, etc.

oed: yep, this is just a reference to the js-did instance.

packages/ceramic-common/package.json
```diff
@@ -28,7 +28,8 @@
   "dependencies": {
     "@types/json-schema": "^7.0.5",
     "ajv": "^6.12.3",
-    "did-resolver": "^2.1.1"
+    "did-resolver": "^2.1.1",
+    "dids": "^0.2.1"
```
oed, line 32: You should only need ti import this as a devDep as only the types are being used.

oed (commented):
Looks good.

packages/ceramic-common/src/ceramic-api.ts
simonovic86: makes sense! cool!

simonovic86 (approved):

PR #227, opened 2020-08-20
Use `ipfs+connection` string to use IPFS connection from context.

ukstv:


packages/ceramic-core/src/pinning/ipfs-pinning.ts
```diff
@@ -24,17 +24,22 @@ export class IpfsPinning implements Pinning {
...
-        if (ipfsHost === FROM_CONTEXT_HOST) {
-            this.ipfsAddress = FROM_CONTEXT_HOST
+        if (connectionString == 'ipfs+context') {
+            this.ipfsAddress = FROM_CONTEXT
         } else {
+            const url = new URL(connectionString)
+            const ipfsHost = url.hostname
+            const ipfsPort = parseInt(url.port, 10) || 5001
             const protocol = url.protocol
                 .replace('ipfs:', 'http')
```
oed, line 34: I guess we can remove this line?

oed (commented):
Thanks for making this update! `ipfs+context` makes sense.

packages/ceramic-core/src/pinning/ipfs-pinning.ts
ukstv: Ah, true.

oed (approved):

PR #230, opened 2020-08-26
Export CeramicCliUtils

ukstv:
This pull request mainly exports CeramicCliUtils. Additionally, and that was necessary, it adds missing dependencies. For some packages, a dependency was used, but not declared in `package.json`.


packages/ceramic-doctype-account-link/src/account-link-doctype.ts
```diff
@@ -10,6 +10,10 @@ export interface AccountLinkParams extends DocParams {
     content?: object;
 }
 
+function isValidChainId(chainId: string) {
+    return chainId === 'eip155:1' || chainId === 'fil:t' || chainId === 'fil:m'
+}
```
oed, line 15: We should really support all caip-2 chainIDs that are supported by `3id-blockchain-utils`. So I don't think we need to check if it's valid here. We can just accept all of them!

oed (changes requested):
Looks good. Just one thing I think we should remove.

packages/ceramic-doctype-account-link/src/account-link-doctype.ts
ukstv: Yay! Makes total sense.

oed (approved):
Looks good to me! Please have a look @simonovic86

packages/ceramic-core/src/anchor/anchor-service-factory.ts
```diff
@@ -20,6 +20,7 @@ export class AnchorServiceFactory {
      * Get AnchorService instance by type
      */
     public get(): AnchorService {
+        console.log('anc', this._config.anchorServiceUrl)
```
simonovic86, line 23: do we need this `console.log`?

simonovic86 (changes requested):

packages/ceramic-core/src/anchor/anchor-service-factory.ts
ukstv: Sure as hell no. Sorry for that. And thank you for finding this.

simonovic86: awesome! no worries! 😊

simonovic86 (approved):

PR #231, opened 2020-08-26
feat(cli): enable js-ipfs

simonovic86:
- enable `js-ipfs`
- enable `dag-jose`
- enable `dht`


oed:
Looks good to me 👍

packages/ceramic-cli/src/index.ts
```diff
@@ -18,14 +25,39 @@ program
...
             ethereumRpcUrl: ethereumRpc,
             anchorServiceUrl: anchorServiceApi,
             stateStorePath: stateStorePath,
             validateDocs,
             pinning: pinning
-        })
+        }
+
+        if (ipfsApi) {
+            Object.assign(config, {
```
PaulLeCam, line 38: Not an issue but seems this could simply be `config.ipfsHost = ipfsApi` as it's only setting this field, no?

packages/ceramic-cli/src/index.ts
```diff
@@ -18,14 +25,39 @@ program
...
+                ipfsHost: ipfsApi
+            })
+        } else {
+            multiformats.multicodec.add(dagJose)
+            const format = legacy(multiformats, dagJose.name)
+
+            const ipfs = await Ipfs.create({
```
PaulLeCam, line 45: Seems this could simply be set as `config.ipfs = await Ipfs.create({`

PaulLeCam (approved):

packages/ceramic-cli/src/index.ts
simonovic86: 👍

packages/ceramic-cli/src/index.ts
simonovic86: 👍

.circleci/config.yml
```diff
@@ -4,7 +4,7 @@ orbs:
 jobs:
   build-and-test:
     docker:
-      - image: circleci/node:12
+      - image: circleci/node:14
```
PaulLeCam, line 7: Could we run the tests in both node 12 and 14? As 12 is the current LTS I think we should make sure the setup works for it.

packages/ceramic-cli/babel.config.js
```diff
@@ -1,7 +1,19 @@
...
 module.exports = {
-  presets: [
-    ['@babel/preset-env', {targets: {node: 'current'}}],
-    '@babel/preset-typescript',
-  ],
-};
+	presets: [
+		["@babel/preset-env", {
+			targets: {
+				node: 'current',
```
PaulLeCam, line 6: Use `node: 12` maybe to avoid making it specific to the development env?

packages/ceramic-core/src/__tests__/ceramic.test.ts
```diff
@@ -48,12 +39,15 @@ describe('Ceramic integration', () => {
 
   beforeAll(async () => {
     tmpFolder = await tmp.dir({ unsafeCleanup: true })
-    ipfs1 = await Ipfs.create(genIpfsConf(tmpFolder.path, 0))
-    ipfs2 = await Ipfs.create(genIpfsConf(tmpFolder.path, 1))
-    ipfs3 = await Ipfs.create(genIpfsConf(tmpFolder.path, 2))
-    multaddr1 = (await ipfs1.id()).addresses[0].toString()
-    multaddr2 = (await ipfs2.id()).addresses[0].toString()
-    multaddr3 = (await ipfs3.id()).addresses[0].toString()
+    ipfs1 = await IpfsUtils.create(tmpFolder.path, 0)
```
PaulLeCam, line 42: Seems like these calls could be made in parallel which should help run the tests faster, ex:
```ts
async function createIPFS(path, id): Promise<[Ipfs, string]> {
  const ipfs = await IpfsUtils.create(path, id)
  const ipfsID = await ipfs.id()
  return [ipfs, ipfsID.addresses[0].toString()]
}

await Promise.all([1, 2, 3].map(id => createIPFS(tmpFolder.path, id)))
```

oed: nice 👍

simonovic86: yep, it can be done that way 👍

.circleci/config.yml
simonovic86: yep, I wanted to test it with `14`

PaulLeCam: Yes I'm not suggesting necessarily dropping 14 but rather also testing 12 so that CI can help catch issues we might not have in our local environment.

simonovic86: Yeah, we'll do that definitely! I'll refactor the configuration in the next PR. Makes sense?

PaulLeCam: Cool!

simonovic86: I have to examine Jest vm/env in more depth.

simonovic86:
We were able to solve the issue with Jest + ESM for now. I'm closing this ticket.

PR #248, opened 2020-08-31
fix: doctype getters now returns next state

oed:
The doctype class getters now return the pending state in `next` if present. Just as already done in the `document` class previously.

### DIFF
--- a/packages/ceramic-common/src/doctype.ts
+++ b/packages/ceramic-common/src/doctype.ts
@@ -8,7 +8,7 @@ import { DoctypeUtils } from "./utils/doctype-utils"
  * Describes signature status
  */
 export enum SignatureStatus {
-    GENESIS,
+  GENESIS,
     PARTIAL,
     SIGNED
 }
@@ -17,7 +17,7 @@ export enum SignatureStatus {
  * Describes all anchor statuses
  */
 export enum AnchorStatus {
-    NOT_REQUESTED,
+  NOT_REQUESTED,
     PENDING,
     PROCESSING,
     ANCHORED,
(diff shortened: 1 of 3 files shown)


packages/ceramic-common/src/doctype.ts
```diff
@@ -8,7 +8,7 @@ import { DoctypeUtils } from "./utils/doctype-utils"
  * Describes signature status
  */
 export enum SignatureStatus {
-    GENESIS,
+  GENESIS,
```
simonovic86, line 11: can you pls fix the formatting here

packages/ceramic-common/src/doctype.ts
```diff
@@ -28,142 +28,144 @@ export enum AnchorStatus {
  * Describes anchor record
  */
 export interface AnchorRecord {
-    prev: CID; // should be CID type
-    proof: CID; // should be CID type
-    path: string;
+  prev: CID; // should be CID type
```
simonovic86, line 31: is there a reason for tab size change in the doc? it's hard to see the changes you made

simonovic86 (changes requested):

packages/ceramic-common/src/doctype.ts
oed: I just used my editors default. However the eslint configuration that Paul set up (which we should slowly start to adopt everywhere) uses 2 for tabwidth :)
https://github.com/3box/eslint-config-3box

simonovic86: let's create a separate issue for formatting based on the `eslint-config-3box` and reformat the whole project :)

packages/ceramic-common/src/doctype.ts
```diff
@@ -28,142 +28,144 @@ export enum AnchorStatus {
...
-     * @param params - Change parameters
-     * @param opts - Initialization options
-     */
-    abstract change(params: DocParams, opts?: DocOpts): Promise<void>
+  constructor(private _state: DocState, private _context: Context) {
+    super()
+  }
+
+  get id(): string {
+    return DoctypeUtils.createDocIdFromGenesis(this._state.log[0])
```
PaulLeCam, line 109: I usually try to avoid performing computations on getters because it's not obvious what is happening from the consuming side.

For example the consumer could have code using `doctype.id` multiple times and each time it makes this computation, while if the method was called `getId()` the consumer could instead do `const id = doctype.getId()` and avoid calling it multiple times.

An alternative could also be something like this:

```ts
get id(): string {
  if (this._id == null) {
    this._id = DoctypeUtils.createDocIdFromGenesis(this._state.log[0])
  }
  return this._id
}
```
In cases a getter is likely to be used multiple times and the computed value is immutable, it could be cached on the instance on the first access.

PaulLeCam (approved):

packages/ceramic-common/src/doctype.ts
oed: Note that this getter was not changed by this PR.

PaulLeCam: Yes no need to change this now I think, I just commented on it as I noticed it.

simonovic86 (approved):

PR #276, opened 2020-09-07
fix: verify that signatures where made by correct DID

oed:
This PR addresses two issues:
1. Actually verify that the signature was made by the DID that created the document
2. Don't add existing metadata to new records


simonovic86 (approved):
nice 🚀

PR #282, opened 2020-09-07
feat: add read only gateway option

zachferland:


packages/ceramic-cli/src/ceramic-daemon.ts
```diff
@@ -89,6 +84,26 @@ class CeramicDaemon {
...
+    app.get(toApiPath('/pin/ls/ceramic/:cid'), this.listPinned.bind(this))
+    app.get(toApiPath('/pin/ls'), this.listPinned.bind(this))
+
+    if (!gateway) {
+      app.post(toApiPath('/create'), this.createDocFromGenesis.bind(this))
+      app.post(toApiPath('/apply'), this.applyRecord.bind(this))
+      app.get(toApiPath('/pin/add/ceramic/:cid'), this.pinDocument.bind(this))
+      app.get(toApiPath('/pin/rm/ceramic/:cid'), this.unpinDocument.bind(this))
+    } else {
+      app.post(toApiPath('/create'), this.notSupported.bind(this))
```
oed, line 100: `/create` needs to be supported in gateway mode. Otherwise we won't be able to lookup account links.

oed (changes requested):

simonovic86 (approved):
lgtm 👍

packages/ceramic-cli/src/ceramic-daemon.ts
zachferland: ok yeah, but that does dag.put, seems like another read only function needed after? but can just move for now

zachferland: on second glance, its not actually pinning, so this will be good for now

zachferland:
@simonovic86 update this pr, on what looked like some refactors, should be good, but wasnt able to compile yet, may need some help releasing tmrrw, dont know process here, or if you want to do a release

oed (approved):

simonovic86:
> @simonovic86 update this pr, on what looked like some refactors, should be good, but wasnt able to compile yet, may need some help releasing tmrrw, dont know process here, or if you want to do a release

of course! 👍

simonovic86 (approved):

PR #283, opened 2020-09-09
fix: set DID when requesting a signature

oed:
Two things in this PR:
1. set the DID that we want to sign the record
2. expect that the owner of a `three-id` document is a [`did:key`](https://w3c-ccg.github.io/did-method-key/)


PaulLeCam (approved):

simonovic86 (approved):
lgtm 👍

PR #291, opened 2020-09-10
update npm name

drbh:
npm install -g ceramic-cli fails with `'ceramic-cli@latest' is not in the npm registry.`

prepended `@ceramicnetwork/`


oed (approved):
Good catch, thanks!

PR #321, opened 2020-09-23
feat: implement initial key-did-resolver module

simonovic86:
Implement initial key-did-resolver module. It supports `Secp256k1` for now.


packages/key-did-resolver/package.json
```diff
@@ -0,0 +1,59 @@
...
+  "scripts": {
+    "test": "./node_modules/.bin/jest --coverage",
+    "build": "../../node_modules/.bin/tsc -p tsconfig.json",
+    "prepublishOnly": "npm run build",
+    "prebuild": "npm run clean",
+    "lint": "./node_modules/.bin/eslint ./src --ext .js,.jsx,.ts,.tsx",
+    "clean": "rm -rf ./lib"
+  },
+  "dependencies": {
+    "@trust/keyto": "^1.0.1",
```
oed, line 30: This should not be needed.

packages/key-did-resolver/package.json
```diff
@@ -0,0 +1,59 @@
...
+    "bs58": "^4.0.1",
```
oed, line 31: We can use https://github.com/achingbrain/uint8arrays

packages/key-did-resolver/package.json
```diff
@@ -0,0 +1,59 @@
...
+    "canonicalize": "^1.0.3",
+    "secp256k1": "^4.0.2",
+    "base64url": "^3.0.1"
```
oed, line 34: We can use https://github.com/achingbrain/uint8arrays

packages/key-did-resolver/package.json
```diff
@@ -0,0 +1,59 @@
...
+    "prepublishOnly": "npm run build",
+    "prebuild": "npm run clean",
+    "lint": "./node_modules/.bin/eslint ./src --ext .js,.jsx,.ts,.tsx",
+    "clean": "rm -rf ./lib"
+  },
+  "dependencies": {
+    "@trust/keyto": "^1.0.1",
+    "bs58": "^4.0.1",
+    "canonicalize": "^1.0.3",
+    "secp256k1": "^4.0.2",
```
oed, line 33: this should definitely not be needed.

packages/key-did-resolver/package.json
```diff
@@ -0,0 +1,59 @@
...
+    "build": "../../node_modules/.bin/tsc -p tsconfig.json",
+    "prepublishOnly": "npm run build",
+    "prebuild": "npm run clean",
+    "lint": "./node_modules/.bin/eslint ./src --ext .js,.jsx,.ts,.tsx",
+    "clean": "rm -rf ./lib"
+  },
+  "dependencies": {
+    "@trust/keyto": "^1.0.1",
+    "bs58": "^4.0.1",
+    "canonicalize": "^1.0.3",
```
oed, line 32: Should not be needed.

packages/key-did-resolver/src/secp256k1/secp256k1-driver.ts
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Constructs the document based on the method key
+ */
+const keyToDidDoc = (fingerprint: string): any => {
+    const secp256k1KeyPair = Secp256k1KeyPair.fromFingerprint(fingerprint)
+    const did = `did:key:${secp256k1KeyPair.fingerprint()}`
+    const keyId = `#${secp256k1KeyPair.fingerprint()}`
+    return {
+        '@context': ['https://www.w3.org/ns/did/v1', {
+            '@base': did,
```
oed, line 12: What's `@base`?

packages/key-did-resolver/src/index.ts
```diff
@@ -0,0 +1,29 @@
...
+
+interface ResolverRegistry {
+    [index: string]: DIDResolver;
+}
+
+import * as secp256k1 from './secp256k1/secp256k1-driver'
+
+// supported drivers
+const prefixToDriverMap: any = {
+    zQ3s: secp256k1,
```
oed, line 11: We should not assume this is encoded with base58btc.

oed: Let's check which multibase first, then use `varint` package to parse the first varint to see which pubkey type we have.

packages/key-did-resolver/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+### Installation
+```
+$ npm install @ceramicnetwork/key-did-resolver
+```
+
+## Contributing
+We are happy to accept small and large contributions. Make sure to check out the [Ceramic specifications](https://github.com/ceramicnetwork/specs) for details of how the protocol works.
+
+## License
```
PaulLeCam, line 13: Seems to be missing info about the license here?

packages/key-did-resolver/src/index.ts
```diff
@@ -0,0 +1,29 @@
...
+}
+
+export default {
+    getResolver: (): ResolverRegistry => ({
+        'key': async (did: string, parsed: ParsedDID): Promise<DIDDocument | null> => {
+            if (did.indexOf('did:key:') !== 0) {
+                throw new Error('did must be of method did:key.');
+            }
+            const idchar: any = did.split('did:key:').pop();
```
oed, line 20: we can just use id from the parsedDID: https://github.com/decentralized-identity/did-resolver/blob/1631fa91cac0888c039c84a543855193081273e3/src/resolver.ts#L63

packages/key-did-resolver/babel.config.js
```diff
@@ -0,0 +1,10 @@
+// babel.config.js
+module.exports = {
+  presets: [
+    ['@babel/preset-env', {targets: {node: 'current'}}],
```
PaulLeCam, line 4: `node: 12` maybe to avoid potential issues?

packages/key-did-resolver/src/index.ts
```diff
@@ -0,0 +1,29 @@
...
+            const encodedType = idchar.substring(0, 4);
```
oed, line 21: We can use `varint` here to parse the multicodec bytes.

packages/key-did-resolver/src/secp256k1/secp256k1-key-pair.ts
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    /**
+     * Derive instance from options
+     */
+    static async from(options: any): Promise<Secp256k1KeyPair> {
+        let privateKeyBase58 = options.privateKeyBase58
+        let publicKeyBase58 = options.publicKeyBase58
+
+        if (options.privateKeyHex) {
```
PaulLeCam, line 49: Might be worth adding a check if both `privateKeyHex` and `privateKeyJwk` or `publicKeyHex` and `publicKeyJwk` are provided to throw an error so that only one variant should be provided.

I think it should also check if either private key variant is provided and throw an error if not rather than create an invalid instance, and if the private key is provided but not the public key, shouldn't this function derive it?

packages/key-did-resolver/src/secp256k1/secp256k1-driver.ts
```diff
@@ -0,0 +1,44 @@
+import { Secp256k1KeyPair } from "./secp256k1-key-pair"
+
+/**
+ * Constructs the document based on the method key
+ */
+const keyToDidDoc = (fingerprint: string): any => {
```
oed, line 6: Why is this called fingerprint?

packages/key-did-resolver/src/secp256k1/secp256k1-driver.ts
```diff
@@ -0,0 +1,44 @@
+import { Secp256k1KeyPair } from "./secp256k1-key-pair"
```
oed, line 1: Don't get why we need a keypair class here?

packages/key-did-resolver/src/secp256k1/__tests__/__snapshots__/secp256k1-driver.test.ts.snap
```diff
@@ -0,0 +1,43 @@
...
+  "id": "did:key:zQ3shwsCgFanBax6UiaLu1oGvM7vhuqoW88VBUiUTCeHbTeTV",
+  "publicKey": Array [
+    Object {
+      "controller": "did:key:zQ3shwsCgFanBax6UiaLu1oGvM7vhuqoW88VBUiUTCeHbTeTV",
+      "id": "#zQ3shwsCgFanBax6UiaLu1oGvM7vhuqoW88VBUiUTCeHbTeTV",
+      "publicKeyBase58": "2Bv3fQSJiGJsbUUfUdiPsXXZGG5cXqaSeYnieuJcUFAbm",
+      "type": "EcdsaSecp256k1VerificationKey2019",
+    },
+  ],
+}
```
oed, line 42: Have a look at the example here: https://w3c-ccg.github.io/did-method-key/#create

oed (commented):
Seems like this almost produces the right output, but seems a bit overcomplicated I htinl

packages/key-did-resolver/package.json
simonovic86: 👍

packages/key-did-resolver/src/@types/index.d.ts
```diff
@@ -0,0 +1,2 @@
+declare module 'bs58';
+declare module 'uint8arrays'
```
oed, line 2: Paul showed me a better way to do declarations: https://github.com/ceramicnetwork/js-did/blob/master/src/declarations.d.ts

packages/key-did-resolver/src/secp256k1.ts
```diff
@@ -0,0 +1,27 @@
...
+    const did = `did:key:${fingerprint}`
+    const keyId = `${did}#${fingerprint}`
+    return {
+        "@context": "https://w3id.org/did/v1",
+        id: did,
+        publicKey: [{
+            id: keyId,
+            type: 'Secp256k1VerificationKey2018',
+            controller: did,
+            publicKeyHex: u8a.toString(pubKeyBytes, "base16"),
```
oed, line 16: Maybe we should set both `publicKeyHex` and `publicKeyBase58` here for future compatibility?

oed (approved):
Looks great now!

packages/key-did-resolver/src/@types/index.d.ts
simonovic86: oh nice! I'll do that 👍

packages/key-did-resolver/src/secp256k1.ts
simonovic86: makes sense 👍

packages/key-did-resolver/src/secp256k1/secp256k1-key-pair.ts
simonovic86: tnx! removed this piece of code 😊

packages/key-did-resolver/README.md
simonovic86: you're right. I have some separate stories regarding the documentation, that should be part of those 👍

packages/key-did-resolver/babel.config.js
simonovic86: makes sense 👍

PR #323, opened 2020-09-24
feat: remove 3id doctype

simonovic86:
- remove `3id` doctype
- adapt tests
- set debug level to `silent` by default


packages/ceramic-doctype-tile/src/tile-doctype.ts
```diff
@@ -60,11 +60,20 @@ export class TileDoctype extends Doctype {
...
      */
     static async makeGenesis(params: DocParams, context?: Context, opts: DocOpts = {}): Promise<Record<string, any>> {
-        if (!context.did || !context.did.authenticated) {
-            throw new Error('No DID authenticated')
+        const metadata = params.metadata? params.metadata : { owners: [] }
+        const is3id = metadata.tags?.includes('3id')
+        if (is3id) {
+            if (metadata.owners.length === 0) {
+                throw new Error('The owner of the 3ID needs to be specified')
+            }
```
oed, line 68: I don't think we need any special logic for 3IDs.

packages/ceramic-doctype-tile/src/tile-doctype.ts
```diff
@@ -60,11 +60,20 @@ export class TileDoctype extends Doctype {
...
         }
 
-        const metadata = params.metadata? params.metadata : { owners: [] }
+        if (!is3id) {
+            // check for DID and authentication
+            if (!context.did || !context.did.authenticated) {
+                throw new Error('No DID authenticated')
+            }
+        }
```
oed, line 76: Do we really need to make this check here?

packages/ceramic-doctype-tile/src/tile-doctype.ts
```diff
@@ -78,18 +87,23 @@ export class TileDoctype extends Doctype {
 
         const { content } = params
         const record = { doctype: DOCTYPE, data: content, header: metadata, unique }
-        return TileDoctype._signDagJWS(record, context.did, metadata.owners[0])
+        if (!is3id) {
+            return TileDoctype._signDagJWS(record, context.did, metadata.owners[0])
+        }
```
oed, line 92: To sign the genesis record or not should be an option passed. 3ID documents should actually have a signed genesis record (in most cases).

oed (commented):
I realize that there is another change needed in IDW to make all of this work properly. Will see if I can make that change quickly!

packages/ceramic-doctype-tile/src/tile-doctype.ts
simonovic86: agree 👍

packages/ceramic-doctype-tile/src/tile-doctype.ts
simonovic86: agree 👍

packages/ceramic-doctype-tile/src/tile-doctype.ts
simonovic86: now every 3ID genesis record is being signed

PaulLeCam (approved):

PR #348, opened 2020-10-01
Document queue fix

simonovic86:
- replace `async.queue` with `p-queue` in order to handle errors properly
- add configurable topic
- fix document handling
- fix ceramic `close()`
- fix tests


PaulLeCam (approved):

oed (dismissed):
Where we not using p-queue before?

simonovic86:
> Where we not using p-queue before?

not for some time, but I got it back

simonovic86:
@oed it should be ready now

oed (approved):

PR #351, opened 2020-10-02
fix: add gateway to ceramic config and use pinningStorePath for cli

### LINKED ISSUE #350: Pinning store path and gateway options not honored
## Repro

`node packages/ceramic-cli/lib/bin/ceramic.js daemon --gateway --debug`

`node packages/ceramic-cli/lib/bin/ceramic.js daemon --debug --pinning-store-path=$PWD/.pinning.store.1`

``

## Observed

Logs get output but component is `NODE`

```
[2020-10-01T20:00:25.130Z] DEBUG NODE CeramicDaemon: {"request":{"timestamp":1601582425130,"rawHeaders": ...
```

Tries to use .pinning.store/ (but fails because it is already in use with another daemon on a different port)

```
Ceramic API running on port 7007

/Users/val/Code/js-ceramic/node_modules/levelup/lib/levelup.js:119
      return callback(new OpenError(err))
                      ^
Error [OpenError]: IO error: lock /Users/val/Code/js-ceramic/.pinning.store/LOCK: Resource temporarily unavailable

```

## Expected

Should be `DEBUG GATEWAY`

Shouldn't get the lock conflict

valmack:
Closes #350 

### Motivation

- Ran into lockfile issue when trying to run two versions of Ceramic bc the pinning store path argument wasn't working
- Logs from the gateway were not being prefixed with "GATEWAY"

### Changes

- Changed the arg `stateStorePath` to `pinningStorePath`
- Added the `gateway` option to `ceramicConfig`

### DIFF
--- a/packages/ceramic-cli/src/bin/ceramic.ts
+++ b/packages/ceramic-cli/src/bin/ceramic.ts
@@ -14,8 +14,8 @@ program
     .option('--port <int>', 'Port daemon is availabe. Default is 7007')
     .option('--debug', 'Enable debug logging level. Default is false')
     .description('Start the daemon')
-    .action(async ({ ipfsApi, ethereumRpc, anchorServiceApi, validateDocs, pinning, stateStorePath, gateway, port, debug }) => {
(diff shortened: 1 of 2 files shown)


simonovic86 (approved):
looks good! 👍

Issue #349, opened 2020-10-01
Update logger to send error response data


valmack:
It is useful to log error responses so we can track 5XX in metrics, etc

Issue #339, opened 2020-09-30
Dispatcher not decoding messages properly


valmack:
From debugging it seems this line is causing the error below. See: https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/ceramic-core/src/dispatcher.ts#L103

` const { typ, id, cid } = JSON.parse(message.data)`

```
(node:79781) UnhandledPromiseRejectionWarning: SyntaxError: Unexpected token , in JSON at position 3
    at JSON.parse (<anonymous>)
    at Dispatcher.<anonymous> (/Users/val/Code/js-ceramic/packages/ceramic-core/lib/dispatcher.js:96:47)
    at Generator.next (<anonymous>)
    at /Users/val/Code/js-ceramic/packages/ceramic-core/lib/dispatcher.js:8:71
    at new Promise (<anonymous>)
    at __awaiter (/Users/val/Code/js-ceramic/packages/ceramic-core/lib/dispatcher.js:4:12)
    at Dispatcher.handleMessage (/Users/val/Code/js-ceramic/packages/ceramic-core/lib/dispatcher.js:89:16)
    at readMessages (/Users/val/Code/js-ceramic/node_modules/ipfs-http-client/src/pubsub/subscribe.js:77:9)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

### Actual
The message data is received as a buffer `"data":{"0":123,"1":34,"2":116,"3":121,"4":112,"5":34,"6":58,"7":48,"8":44,...`

### Expected
It should be json
```json
{
"typ":0,
"id":"/ceramic/bagcqcera74id723qaqdcrfoigb35juhwjxra4yibx65q5fvcsl7myay4ypnq",
"cid":"bagcqcera4khvqsrclxifb2n5yikt7grh6pldek7cxx4sjbcgxsdpurrrmwtq",
"doctype":"tile"
}
```

valmack:
Looks like this may be fixed but not sure where. No longer seeing the issue

valmack:
I have a patch in this in #352  but will leave it open because we need a better way of handling it

PR #352, opened 2020-10-02
Feat/log files

### LINKED ISSUE #338: Write logger output to files
This will make it easy for people to ingest logs to an aggregator like mtail, promtail, fluentd, etc.

valmack:
Closes #338 #349 

Depends on #351 to properly label gateway logs

### Motivation

- Logging to files, instead of just stdout, allows someone to collect the logs in an aggregation tool and feed them into a monitoring system (for example Promtail/Loki/Grafana)
- Additionally it is valuable to monitor any errors sent from the API and through QA I found the logger used for errors was not structured properly to give error info

### Changes

- Added plugin to the `LoggerProvider` to write logs to files
- Refactored API logging to log errors with req/res metadata
- Removed log level setting in daemon constructor so that plugins do not get overridden. Now the log level is set solely when `CeramicDaemon.create` is called, and this allows plugins to be applied correctly wherever `RootLogger` is used

### Future work
- Updated `handleMessage` in Dispatcher to patch #339 but this needs to be handled better

### DIFF
--- a/packages/ceramic-common/src/logger-provider.ts
+++ b/packages/ceramic-common/src/logger-provider.ts
@@ -85,23 +85,23 @@ class LoggerProvider {
 
     /**
      * Plugin to append log messages to files named after components
-     * @param options Should include `component` name
+     * @notice If no component name is given 'default' will be included in the file name
+     * @param options Should include `component` name string
      */
     static _includeFilePlugin (options: Options): void {
         const originalFactory = log.methodFactory;
 
         log.methodFactory = (methodName: string, logLevel: any, loggerName: string): MethodFactory => {
             const rawMethod = originalFactory(methodName, logLevel, loggerName);
-
             return (...args: any[]): any => {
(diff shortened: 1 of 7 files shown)


packages/ceramic-common/src/logger-provider.ts
```diff
@@ -61,6 +63,7 @@ class LoggerProvider {
         }
 
         LoggerProvider._applyPrefix(options)
+        LoggerProvider._includeFilePlugin(options)
```
simonovic86, line 66: let's make this optional, makes sense?

simonovic86 (approved):
LGTM :) We should think about a rotating file logger as well.

packages/ceramic-common/src/logger-provider.ts
valmack: yeah @simonovic86 with a cli flag?

simonovic86: yep :)

PR #357, opened 2020-10-05
Implement loadDocumentRecords API

simonovic86:
- implement `loadDocumentRecords` API
- implement full serialize/deserialize record functionality
- use `uint8arrays` instead of `Buffer`
- add and adapt tests


PaulLeCam (approved):
LGTM but might be useful to add a test using `loadDocumentRecords()` and `applyRecord()` to make sure the full flow works as expected for the use-case we have?

simonovic86:
> LGTM but might be useful to add a test using `loadDocumentRecords()` and `applyRecord()` to make sure the full flow works as expected for the use-case we have?

makes sense 👍

PR #365, opened 2020-10-06
feat: ceramic docid library

zachferland:
~~TODO~~
~~readme - ref cips - gen docs - example~~
~~simple tests~~
~~not pulling in typed dependencies/defs~~


packages/ceramic-docid/README.md
```diff
@@ -0,0 +1,13 @@
+# Ceramic DocId
+
+> This package contains Ceramic DocId implementation.
+
+### Installation
+```
+$ npm install @ceramicnetwork/ceramic-docid
```
oed, line 7: Can you change the package and folder name to just be `docid` we are changing this for all packages as well later.
So the package would be `@ceramicnetwork/docid`

packages/ceramic-docid/src/doctype-table.json
```diff
@@ -0,0 +1,4 @@
+{
+  "tile": 0,
+  "caip10-link": 1
```
oed, line 3: Should we use hex here for consistency with spec?
```suggestion
  "tile": 0x00,
  "caip10-link": 0x01
```

packages/ceramic-docid/src/index.ts
```diff
@@ -0,0 +1,195 @@
...
+   * @returns {Uint8Array}
+   * @readonly
+   */
+  get bytes (): Uint8Array {
+    const codec = varint.encode(DOCID_CODEC)
+    const subnet = varint.encode(subnets[this.subnet])
+    const doctype = varint.encode(doctypes[this.docType])
+    const bytes = uint8ArrayConcat([
+      codec, subnet, doctype, this._cidBytes
+    ], codec.byteLength, subnet.byteLength, doctype.byteLength, this._cidBytes.byteLength)
```
oed, line 143: Why do you actually need to pass the lengths here?

packages/ceramic-docid/src/index.ts
```diff
@@ -0,0 +1,195 @@
...
+    } else if (typeof doctype === 'string') {
+      // From DocIdString
+      const docIdStr = doctype
+      this._multibaseName = multibase.isEncoded(docIdStr)
+      const bytes = multibase.decode(docIdStr).slice(2)
+      this._initFromBytes(bytes)
+    } else {
+      throw new Error('constructor: Invalid Arguments or CID')
+    }
+  }
```
oed, line 60: Can we change the constructor to be less ambiguous? 
My preference would be:
* Have a static `fromBytes` function
* Have a static `fromString` function
* doctype is either the string name or the number (this allows ppl to use custom doctypes without adding to table

packages/ceramic-docid/src/index.ts
```diff
@@ -0,0 +1,195 @@
...
+      const bytes = doctype 
+      this._multibaseName = typeof cid === 'string' ? cid : 'base32'
+      this._initFromBytes(bytes)
+    } else if (Object.keys(doctypes).includes(doctype)) {
+      // From args
+      if (!doctype) throw new Error('constructor: doctype required')
+      if (!cid) throw new Error('constructor: cid required')
+      this._docType = doctype
+      this._subnet = subnet || 'devnet'
+      this._multibaseName = multibaseName || 'base32'
```
oed, line 48: Spec was updated to default to `base36`:)

packages/ceramic-docid/src/index.ts
```diff
@@ -0,0 +1,195 @@
...
+   *
+   * @param   {DocId}   other
+   * @returns {bool}
+   */
+  equals(other: DocId | Uint8Array | string): boolean {
+    // same cid 
+    // same doctype
+    // maybe ignore subnet
+    return true
+  }
```
oed, line 159: Assume WIP.

packages/ceramic-docid/src/index.ts
```diff
@@ -0,0 +1,195 @@
...
+
+  /**
+   * Encode the DocId into a string.
+   *
+   * @param   {string}   [base=this.multibaseName]   Base encoding to use.
+   * @returns {string}
+   */
+  toString(base: string): string {
+    return this.toBaseEncodedString(base)
+  }
```
oed, line 179: Default to `base36`.

packages/ceramic-docid/src/subnet-table.json
```diff
@@ -0,0 +1,4 @@
+{
+  "devnet": 0,
+  "livenet": 1
```
oed, line 3:
```suggestion
  "devnet": 0x00,
  "livenet": 0x01
```

oed (changes requested):
Looks pretty good, have a few thoughts and preferences :)

packages/ceramic-docid/README.md
zachferland: yup, this makes more sense

packages/ceramic-docid/src/doctype-table.json
zachferland: i don't see why they are not just represented as ints (maybe to represent size as well)? most of ipfs libs use int tables internally instead of the hex, and if going to allow subnets/doctypes by code, not only strings in table, then seems ints would be nicer, not much preference though

packages/ceramic-docid/src/index.ts
zachferland: no, have to be known at start, but function will calculate same way internally, will just remove

packages/ceramic-docid/src/index.ts
zachferland: 👍

packages/ceramic-docid/src/index.ts
zachferland: 👍

packages/ceramic-docid/src/doctype-table.json
oed: ok, that makes sense. We should update the CIPs as well I guess to reflect.

packages/docid/src/index.ts
```diff
@@ -0,0 +1,210 @@
...
+    return DocId.fromBytes(bytes, multibaseName)
+  }
+
+  /**
+   * Get doc type code
+   *
+   * @returns {number}
+   * @readonly
+   */
+  get docType (): number {
```
oed, line 69: Rename to `type`?

packages/docid/src/index.ts
```diff
@@ -0,0 +1,210 @@
...
+    return this._docType
+  }
+
+  /**
+   * Get doc type name
+   *
+   * @returns {string}
+   * @readonly
+   */
+  get docTypeName (): string {
```
oed, line 79: Rename to `typeName`?

packages/docid/src/index.ts
oed: If not we should remove the camelcase: `doctype`.

packages/docid/src/index.ts
```diff
@@ -0,0 +1,210 @@
...
+   * Compare equality with another DocId.
+   *
+   * @param   {DocId}   other
+   * @returns {bool}
+   */
+  equals(other: DocId | Uint8Array | string): boolean {
+    // same cid 
+    // same doctype
+    // maybe ignore subnet
+    return true
```
oed, line 173: Just compare bytes, or string. Should **not** ignore subnet.

packages/docid/src/index.ts
```diff
@@ -0,0 +1,210 @@
...
+   *
+   * @param   {any}     other
+   * @returns {bool}
+   */
+  static isDocId (other: DocId | Uint8Array | string): boolean {
+    // just try taking all parts 
+    // confirm docid codec 
+    // confirm cid 
+    // cacth return false
+    return true
```
oed, line 206: Try to create a DocID instance, return true if it doesn't fail.

oed (commented):
Missing tests?

PaulLeCam:
Small feature request: could DocID also support Ceramic URLs as input/output please? That would be very useful for IDX as most of the APIs use the URL form.

Ideally we could have an extra `toURL()` method that returns `ceramic://${this.toString()}` and an extra check in `fromString()` to strip the starting `ceramic://` string if present?

Thanks!

packages/docid/src/index.ts
```diff
@@ -0,0 +1,210 @@
...
+    return this._cid.codec
+  }
+
+  /**
+   * Get bytes of DocId
+   *
+   * @returns {Uint8Array}
+   * @readonly
+   */
+  get bytes (): Uint8Array {
```
PaulLeCam, line 153: Could we cache this value please? Ex:
```js
get bytes (): Uint8Array {
  if (this._bytes == null) {
    ...
    this._bytes = uint8ArrayConcat(...)
  }
  return this._bytes
}
```

This way if the consumer uses `docid.bytes` multiple times, the value is cached rather than being computed every time.

PaulLeCam:
Might be nice to also implement [Symbol.toPrimitive](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toPrimitive) and maybe [node's logger custom symbol as for CID](https://github.com/multiformats/js-cid/blob/master/src/index.js#L269) for better readability in logs.

packages/docid/README.md
```diff
@@ -0,0 +1,65 @@
...
+cid.type           // 0
+cid.typeName       // 'tile'
+cid.subnet         // 0
+cid.subnetName     // 'subnet
+cid.multibaseName  // 'base36'
+cid.bytes          // Uint8Array(41) [ 206,   1,   0,   0,   1, 133,   1, ...] 
+cid.cid            // CID('bagcqcerakszw2vsov...')
+cid.toString()     
+//k3y52l7mkcvtg023bt9txegccxe1bah8os3naw5asin3baf3l3t54atn0cuy98yws
+cid.toUrl()
```
oed, line 37: Should say `docid` not `cid`?

packages/docid/src/index.ts
```diff
@@ -0,0 +1,311 @@
...
+   *
+   * @returns {String}
+   */
+  [Symbol.for('nodejs.util.inspect.custom')] (): string {
+    return `DocID(${this.toString()})`
+  }
+
+  [Symbol.toPrimitive](hint: string): string | Uint8Array {
+    if (hint === 'string') return this.toString()
+    if (hint === 'default') return this.bytes
```
PaulLeCam, line 284: I think the primitives can only be `string` or `number`, here I don't think it's necessary to use the `hint`, could simply be:
```ts
[Symbol.toPrimitive](): string {
  return this.toString()
}
```

PaulLeCam (commented):
Joel mentioned this morning that DocIDs will also include the doc version, so I think it would be useful to have methods to:

- set a different version by cloning the DocID (not mutating an existing one)
- strip the version (so it should always represent the tip)
- differentiate between "strict" comparison of DocIDs based on the genesis + version and "loose" one on genesis only
- maybe have the strict/loose options for the `to...` methods as well, either as arguments or separate functions?

It would be useful for libs and apps to have these utilities I think to have more info about what specific resources a DocID represents. Thanks!

oed:
If it wasn't clear @PaulLeCam: In most cases the version will not be included in the DocID.

PaulLeCam:
@oed yes I assume most use cases won't need the version but as it's something the DocID would vary on I think the consumers should have clear expectations about it.
For example would `docID(cid1, v0)` equal `docID(cid1, v1)` or not? To me that's where the strict/loose flag would make the comparison explicit.

zachferland:
thanks @PaulLeCam, yeah was thinking/concerned about the same things yesterday. yeah definitely dont want to mutate and need way to get base easily. was thinking of just adding something like docid.baseId which would return another DocID, either it self (if no version) or another DocID with out version. Then that should make these thing explicit without having to define what strict/loose means here. (ie docid1.baseid.equal(docid2.baseId) )

packages/docid/src/index.ts
zachferland: it does have default - https://tc39.es/ecma262/#sec-toprimitive but maybe should also be string, assumed most string context would be explicit and default would be used for like []+docid, so bytes seemed reasonable

PaulLeCam: Yes sorry I meant for the returned type, I'm not sure `Uint8Array` is valid because it's not a JS primitive.
At least my understanding for this method is that a primitive is always `string` or `number`, and the `hint` is there to give contextual information, cf https://javascript.info/object-toprimitive
Maybe it's not even needed to implement this method, assuming the JS engine defaults to using `toString()` though I don't know if it is standard behavior.

zachferland: ah i see what your saying! yeah so just return string seems to make sense and we will make it explicit still

packages/docid/src/index.ts
```diff
@@ -209,11 +195,17 @@ class DocID {
...
     if (this._bytes == null) {
       const codec = varint.encode(DOCID_CODEC)
-      const subnet = varint.encode(this._subnet)
-      const doctype = varint.encode(this._doctype)
-      const versionBytes =  this._version ? this._version.bytes : new Uint8Array(0)
+      const doctype = varint.encode(this.type)
+
+      let versionBytes 
+      if (this.version) {
+        versionBytes = this.cid.equals(this.version) ? varint.encode(0) : this.version.bytes
```
oed, line 202: I guess you could just do `new Uint8Array([0])` instead of using varint.

packages/docid/package.json
```diff
@@ -24,8 +24,12 @@
     "lib": "lib"
   },
   "dependencies": {
-    "cids": "^1.0.1",
+    "cids": "^1.0.0",
```
oed, line 27: Why was cids downgraded?

zachferland:
@simonovic86 added you as a reviewer if you have a chance tmrrw, and/or we maybe just sync on it during my morning 

two tests that were failing, but made changes that may have changed to the intended meaning of them, should be able to see diff for them.

- Ceramic integration › can apply existing records successfully
- Ceramic integration › can propagate multiple update across nodes with common connection

Otherwise rest of test coverage looks good, but maybe some issues with less tested parts, like http-client. Have not extensively tested through any use, just mostly running tests here.

packages/docid/package.json
zachferland: no reason, just matching others

packages/docid/src/index.ts
zachferland: yeah, just used for clarity, but i guess it will never be any other int, so could be

packages/ceramic-core/src/ceramic.ts
```diff
@@ -234,9 +238,7 @@ class Ceramic implements CeramicApi {
    * @param genesisCid
    */
   getDocFromMap(genesisCid: any): Document {
-    const id = DoctypeUtils.createDocIdFromGenesis(genesisCid)
-    const normalizedDocId = DoctypeUtils.normalizeDocId(id)
-    return this._docmap[normalizedDocId]
+    return this._docmap[genesisCid.toString()]
```
oed, line 241: I assume this should have type CID and we should convert it to a docid before returning the map?

packages/ceramic-core/src/ceramic.ts
```diff
@@ -268,8 +270,7 @@ class Ceramic implements CeramicApi {
     }
 
     doc = await Document.create(genesisCid, doctypeHandler, this.dispatcher, this.pinStore, this.context, opts, this._validateDocs);
-    const normalizedId = DoctypeUtils.normalizeDocId(doc.id)
-    this._docmap[normalizedId] = doc
+    this._docmap[doc.id.toString()] = doc
```
oed, line 273: Yup, here we are storing it using the docid, so the previous comment is correct.

packages/ceramic-core/src/document.ts
```diff
@@ -72,7 +73,9 @@ class Document extends EventEmitter {
       validate = true,
   ): Promise<Document> {
     const genesis = await dispatcher.retrieveRecord(genesisCid)
-    const id = DoctypeUtils.createDocIdFromGenesis(genesisCid)
+  
+    const doctype = (await context.ipfs.dag.get(genesis.link)).value.doctype
+    const id = new DocID(doctype, genesisCid)
```
oed, line 78: Should we not change the function signature of this function to get a DocID instead of genesisCID?

oed (changes requested):
Looks good overall. Some changes needed. And I think I found a small bug.
It also looks like CI fails to bootstrap.

simonovic86:
> * can apply existing records successfully



> @simonovic86 added you as a reviewer if you have a chance tmrrw, and/or we maybe just sync on it during my morning
> 
> two tests that were failing, but made changes that may have changed to the intended meaning of them, should be able to see diff for them.
> 
> * Ceramic integration › can apply existing records successfully
> * Ceramic integration › can propagate multiple update across nodes with common connection
> 
> Otherwise rest of test coverage looks good, but maybe some issues with less tested parts, like http-client. Have not extensively tested through any use, just mostly running tests here.

@zachferland should be fine now :)

simonovic86 (approved):

oed:
Plz don't merge before my comments are addressed :)

simonovic86:
> Plz don't merge before my comments are addressed :)

@oed we should block pull request merge if changes are requested

oed:
@simonovic86 is there a setting for that?

packages/ceramic-cli/src/ceramic-cli-utils.ts
```diff
@@ -123,22 +122,25 @@ export class CeramicCliUtils {
      * @param schemaDocId - Optional schema document ID
      */
     static async change(docId: string, content: string, owners: string, schemaDocId?: string): Promise<void> {
-        if (!CeramicCliUtils._validateDocId(docId)) {
+        let id: DocID
+        try {
+            id = DocID.fromString(docId)
+        } catch(e) {
             console.error(`Invalid docId: ${docId}`)
             return
```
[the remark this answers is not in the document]
zachferland: yeah i was pattern matching with existing behavior here, but yeah a console.error and return, should be the same now as descriptive thrown error from docid, will change these 👍

packages/ceramic-cli/src/ceramic-cli-utils.ts
```diff
@@ -154,13 +156,16 @@ export class CeramicCliUtils {
      * @param docId - Document ID
      */
     static async show(docId: string): Promise<void> {
-        if (!CeramicCliUtils._validateDocId(docId)) {
+        let id: DocID
+        try {
+            id = DocID.fromString(docId)
+        } catch(e) {
```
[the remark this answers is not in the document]
zachferland: 👍

packages/ceramic-cli/src/ceramic-daemon.ts
```diff
@@ -205,13 +206,10 @@ class CeramicDaemon {
    * Get document content
    */
   async show (req: Request, res: Response, next: NextFunction): Promise<void> {
-    let docId = ['/ceramic', req.params.cid].join('/')
-    if (req.query.version) {
-      docId = `${docId}?version=${req.query.version}`
-    }
+    const docId = DocID.fromString(req.params.cid, req.query.version ? req.query.version.toString() : undefined)
```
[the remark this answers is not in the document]
zachferland: yup wanted to change that to docid!

still seems like it makes sense to allow a version param here or a docid including version

packages/ceramic-core/src/ceramic.ts
zachferland: yeah the problem is getting the type, you have dag.get instead of just hit cached object

but yeah inconsistent with every where else, would almost leans towards cid, but that would probably make code less readable everywhere else

oed: but docIDs are used elsewhere in this map?

packages/ceramic-core/src/document.ts
zachferland: hmm yeah i could could go either way here, seems find for create to be made from parts, but maybe makes most sense if docid created in func where both type and genesis cid are easily available, and docid passed here

oed: I think we eventually want to remove the doctype property on the genesis record. So this would be required.

PR #378, opened 2020-10-06
Fix and unify loadDocumentRecords

simonovic86:
- fix `loadDocumentRecords` API

### DIFF
--- a/packages/ceramic-common/src/utils/doctype-utils.ts
+++ b/packages/ceramic-common/src/utils/doctype-utils.ts
@@ -1,6 +1,7 @@
 import ajv from "ajv"
 import CID from 'cids'
 import cloneDeep from "lodash.clonedeep"
+import * as u8a from 'uint8arrays'
 
 import { Ipfs } from "ipfs"
 import { AnchorStatus, DocState, Doctype } from "../doctype"
@@ -109,13 +110,7 @@ export class DoctypeUtils {
 
         if (DoctypeUtils.isSignedRecordDTO(cloned)) {
             cloned.jws.link = cloned.jws.link.toString()
-            if (cloned.linkedBlock.data) {
-                // it's a block
-                cloned.linkedBlock = cloned.linkedBlock.data.toString('base64')
-            } else {
-                // it's a buffer
-                cloned.linkedBlock = cloned.linkedBlock.toString('base64')
-            }
+            cloned.linkedBlock = u8a.toString(cloned.linkedBlock, 'base64')
             return cloned
         }
 
@@ -146,7 +141,7 @@ export class DoctypeUtils {
 
         if (DoctypeUtils.isSignedRecordDTO(cloned)) {
             cloned.jws.link = new CID(cloned.jws.link)
-            cloned.linkedBlock = Buffer.from(cloned.linkedBlock, 'base64')
+            cloned.linkedBlock = u8a.fromString(cloned.linkedBlock, 'base64')
             return cloned
         }
 
@@ -260,7 +255,7 @@ export class DoctypeUtils {
             const linkedBlock = await ipfs.block.get(record.link)
             return {
                 jws: record,
-                linkedBlock: linkedBlock.data,
+                linkedBlock: new Uint8Array(linkedBlock.data.buffer),
             }
         }
         return record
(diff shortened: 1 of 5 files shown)


packages/ceramic-common/src/utils/doctype-utils.ts
```diff
@@ -141,7 +146,7 @@ export class DoctypeUtils {
 
         if (DoctypeUtils.isSignedRecordDTO(cloned)) {
             cloned.jws.link = new CID(cloned.jws.link)
-            cloned.linkedBlock = u8a.fromString(cloned.linkedBlock, 'base64')
+            cloned.linkedBlock = Buffer.from(cloned.linkedBlock, 'base64')
```
PaulLeCam, line 149: Any reason to revert to using `Buffer` here please? Seems like using a `Uint8Array` would be a better option considering the changes we're making across all the packages, no?

PaulLeCam (commented):
I'm not familiar with what the logic should be, but might be useful to add an integration test to check the behavior and avoid possible future regressions?

packages/ceramic-common/src/utils/doctype-utils.ts
simonovic86: The API `ipfs.block.get` returns a `Block` instance that wraps a `Buffer`.  Until now we needed to deserialize a signed record DTO not serialize it, as well. That's why I used buffers in serialize/deserialize.

simonovic86:
> I'm not familiar with what the logic should be, but might be useful to add an integration test to check the behavior and avoid possible future regressions?

There is a test for that but the problem was with serialization that is used in test as well. That's why it went unnoticed.

packages/ceramic-common/src/utils/doctype-utils.ts
```diff
@@ -254,7 +254,8 @@ export class DoctypeUtils {
         if (DoctypeUtils.isSignedRecord(record)) {
             const linkedBlock = await ipfs.block.get(record.link)
             return {
-                jws: record, linkedBlock,
+                jws: record,
+                linkedBlock: new Uint8Array(linkedBlock.data.buffer),
```
oed, line 258: D: this might break when they switch away from buffers here. We should be careful when upgrading ipfs I guess.

oed (approved):

packages/ceramic-common/src/utils/doctype-utils.ts
oed: https://github.com/ipld/js-ipld-block/blob/master/src/index.js

oed: New code ☝️ 
I guess we could add a check if data has type Uint8array? if so just use that, otherwise use the buffer.

simonovic86: I agree 👍  They changed a bunch of things in `master` so there might be more places where it can break :/

simonovic86: > New code ☝️
> I guess we could add a check if data has type Uint8array? if so just use that, otherwise use the buffer.

makes sense 👍

packages/ceramic-common/src/utils/doctype-utils.ts
```diff
@@ -252,10 +252,14 @@ export class DoctypeUtils {
...
-            const linkedBlock = await ipfs.block.get(record.link)
+            let linkedBlock = await ipfs.block.get(record.link)
+            if (linkedBlock.data instanceof Buffer) {
+                // transform Buffer into Uint8Array
+                linkedBlock = new Uint8Array(linkedBlock.data.buffer)
+            }
             return {
                 jws: record,
-                linkedBlock: new Uint8Array(linkedBlock.data.buffer),
+                linkedBlock,
```
oed: This logic is incorrect.

oed (changes requested):

packages/ceramic-common/src/utils/doctype-utils.ts
oed: I think.

oed: oh, outdated.

simonovic86: It should be fine for now, tests are passing. We'll see when we need to upgrade IPFS and discard the forked one.

simonovic86: sorry, why do you think it's incorrect?

simonovic86: I'll create another PR since this one is merged.

oed: the logic will definitely fail when upgraded. Should be something like: 
```js
const block = await ipfs.block.get(record.link)
let linkedBlock = block.data instanceof Uint8Array ? block.data : new Uin8Array(block.data.buffer)
```

PR #396, opened 2020-10-08
Feat/log files

### LINKED ISSUE #388: Move file logging from ceramic-common to package that is not browser side
to package that is not browser side like ceramic-cli

Re: "issues raised trying to build idx-web and self.id because there's an import of fs in ceramic-common (in lib/logger-provider.js), could we remove it please an make sure nothing in ceramic-common is node-specific"

valmack:
Closes #388 

### Motivation

- Move fs library out of ceramic-common (see issue above)

### Changes

- Adds `addPlugin` method to LoggerProvider
- Adds `ceramic-plugins` from which `logToFiles` plugin is passed

### DIFF
--- a/packages/ceramic-cli/src/ceramic-daemon.ts
+++ b/packages/ceramic-cli/src/ceramic-daemon.ts
@@ -3,7 +3,7 @@ import express, { Request, Response, NextFunction } from 'express'
 import Ceramic from '@ceramicnetwork/ceramic-core'
 import type { CeramicConfig } from "@ceramicnetwork/ceramic-core";
 import { DoctypeUtils, RootLogger, Logger } from "@ceramicnetwork/ceramic-common"
-import {loggerProviderPlugins} from "./ceramic-plugins"
+import { LogToFiles } from "./ceramic-logger-plugins"
 // @ts-ignore
 import cors from 'cors'
 
@@ -132,7 +132,7 @@ class CeramicDaemon {
     if (opts.logToFiles) {
         ceramicConfig.logToFiles = opts.logToFiles
         ceramicConfig.logToFilesPlugin = {
-            plugin: loggerProviderPlugins.logToFiles,
+            plugin: LogToFiles.main,
             options: {logPath: opts.logPath}
         }
     }
--- a/packages/ceramic-cli/src/ceramic-daemon.ts
+++ b/packages/ceramic-cli/src/ceramic-daemon.ts
@@ -3,6 +3,7 @@ import express, { Request, Response, NextFunction } from 'express'
 import Ceramic from '@ceramicnetwork/ceramic-core'
 import type { CeramicConfig } from "@ceramicnetwork/ceramic-core";
 import { DoctypeUtils, RootLogger, Logger } from "@ceramicnetwork/ceramic-common"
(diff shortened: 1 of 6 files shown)


valmack:
@PaulLeCam added you just to ensure the modules being imported/removed from satisfy your requirement

packages/ceramic-core/src/ceramic.ts
```diff
@@ -117,13 +122,19 @@ class Ceramic implements CeramicApi {
...
     })
 
+    if (config.logToFiles) {
+        LoggerProvider.addPlugin(
+            config.logToFilesPlugin.plugin,
+            loggerOptions,
+            config.logToFilesPlugin.options
+        )
+    }
+
```
valmack, line 137: @simonovic86 I am having the `init` method return the provider's options here because we add the default options and I think it makes sense to have that passed to the plugin when it gets added.

But this paradigm means there is no single source of truth for the state of options between plugins if we were to add more. So I think it would be best if we can make the options immutable from the perspective of plugins.

packages/ceramic-common/src/logger-provider.ts
```diff
@@ -1,8 +1,6 @@
 import chalk from 'chalk'
```
PaulLeCam, line 1: Could chalk also be moved to the logger plugin please? AFAICT it only targets node.

PaulLeCam (commented):
Apart from the `chalk` dependency still in `ceramic-common` LGTM but I'm not familiar with the full flow and logic so better wait for Janko's approval.

packages/ceramic-cli/src/ceramic-daemon.ts
```diff
@@ -3,6 +3,7 @@ import express, { Request, Response, NextFunction } from 'express'
 import Ceramic from '@ceramicnetwork/ceramic-core'
 import type { CeramicConfig } from "@ceramicnetwork/ceramic-core";
 import { DoctypeUtils, RootLogger, Logger } from "@ceramicnetwork/ceramic-common"
+import {loggerProviderPlugins} from "./ceramic-plugins"
```
oed, line 6: Naming it to `ceramic-plugins` seems to generic? Isn't it more of a custom logger that is passed as an option to the ceramic instance?

packages/ceramic-common/src/logger-provider.ts
simonovic86: `chalk` dependency is not really necessary. @valmack feel free to remove it

packages/ceramic-core/src/ceramic.ts
simonovic86: Yeah, I agree. I think that we can use `Object.freeze()` in the `init` method and prevent options from being changed for now. Makes sense?

simonovic86 (commented):
LGTM. I added a couple of comments related to the immutable `opts` and `chalk` dependency.

packages/ceramic-cli/src/ceramic-daemon.ts
valmack: Yeah I was thinking maybe `ceramic-logger-plugins` is better

simonovic86 (approved):

Issue #202, opened 2020-08-07
Ceramic Core module refactor


simonovic86:
- create separate module for `Pinning` implementation(s)
- create separate module for `AnchorService` implementation(s)
- move `MockAnchorService` to tests

oed:
Why was this closed? I thought this ticket was about separating the Pinning and Anchor modules into separate packages?

simonovic86:
It's closed by mistake. Thanks!

simonovic86:
`chore` color is similar to the [issue](https://github.com/3box/devops/issues/114) :)

PR #408, opened 2020-10-16
Ceramic anchoring (squash, in-memory service, pub/sub fixes, etc.)

simonovic86:
* Modify mock anchor service for better testing
    * add option for synchronous `anchoring` 
    * add option for delays in order to test more thoroughly
- test various anchoring cases
- add `nonce` to `TileDoctype` header
- fix inconsistency with `TileDoctype` spec
- fix document registration to `Dispatcher`
- fix issues produced with instant anchoring mock


packages/ceramic-doctype-tile/src/tile-doctype.ts
```diff
@@ -114,6 +126,31 @@ export class TileDoctype extends Doctype {
...
+            if (DoctypeUtils.isAnchorRecord(record)) {
+                return nonce
+            }
+            if (DoctypeUtils.isSignedRecord(record)) {
+                record = (await doctype.context.ipfs.dag.get(record.link)).value
+            }
+            cid = record.prev
+            nonce++
+        }
+        return nonce
```
oed, line 151: This will make the nonce be higher with each new signed record that is added? That is not what I meant with the nonce at all and I'm not sure this actually solves the problem.

It should be something more simple like this:
```suggestion
        // If there hasn't been any update prior to this we should not set the nonce
        if (!doctype.state.next) return
        // Get the current nonce and increment it by one
        return (doctype.state.next.metadata?.nonce || 0)++
```
Assuming that we don't use the `anchor` property as mentioned above. This means that the nonce only takes affect on each record "height" and once a signed record has been anchored, the nonce is reset to 0.

oed (changes requested):
I think I might have not communicated clearly enough how this should work. What is being done now will hit performance pretty bad I think and I'm not sure it will actually solve the problem. Suggesting a simpler solution below.

packages/ceramic-doctype-tile/src/tile-doctype.ts
```diff
@@ -60,7 +67,7 @@ export class TileDoctype extends Doctype {
      * @param opts - Initialization options
      */
     static async makeGenesis(params: DocParams, context?: Context, opts: DocOpts = {}): Promise<Record<string, any>> {
-        const metadata = params.metadata? params.metadata : { owners: [] }
+        const metadata = params.metadata? params.metadata : { owners: [], anchor: { nonce: 0 } }
```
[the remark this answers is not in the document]
oed: Don't think we need a nonce in genesis at all actually.

packages/ceramic-doctype-tile/src/tile-doctype.ts
```diff
@@ -114,6 +126,31 @@ export class TileDoctype extends Doctype {
         return TileDoctype._signDagJWS(record, did, doctype.owners[0])
     }
 
+    /**
+     * Calculates anchor nonce
+     */
+    private static async _calculateNonce(doctype: Doctype): Promise<number> {
+        if (doctype.context.ipfs == null) {
+            return 0 // return nonce 0 if IPFS is not available
+        }
```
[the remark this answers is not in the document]
oed: With the edit below we should be able to just remove these lines.

packages/ceramic-core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -55,6 +56,15 @@ export default class EthereumAnchorService extends AnchorService {
...
     }
 
+    /**
+     * Set Ceramic API instance
+     *
+     * @param ceramic - Ceramic API used for various purposes
+     */
+    set ceramic(ceramic: CeramicApi) {
+        this._ceramic = ceramic
+    }
```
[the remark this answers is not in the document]
simonovic86: It's not used in this implementation but it's part of the interface. It's used in the `InMemoryAnchorService`.

packages/ceramic-doctype-tile/src/tile-doctype.ts
```diff
@@ -105,6 +112,11 @@ export class TileDoctype extends Doctype {
             header.owners = newOwners
         }
 
+        const nonce = await TileDoctype._calculateNonce(doctype)
+        header.anchor = {
+            nonce,
+        }
```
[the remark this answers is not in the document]
simonovic86: I added `anchor` for more information about the `nonce`. I think we didn't actually specify how we're going to set it. I'll remove `anchor`.

packages/ceramic-doctype-tile/src/tile-doctype.ts
simonovic86: Thanks for the suggestions. I'll test this solution now. We should definitely specify the proposals like this in the future (e.g. somewhere in the story, etc).

packages/ceramic-core/src/anchor/ethereum/ethereum-anchor-service.ts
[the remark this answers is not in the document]
oed: Right I have some ideas of how this could be done in a better way once we want to refactor. But should be fine for now 👍

packages/ceramic-doctype-tile/src/tile-doctype.ts
oed: yeah, sorry I thought it was more clear that it was.

packages/ceramic-doctype-tile/src/tile-doctype.ts
```diff
@@ -114,6 +126,20 @@ export class TileDoctype extends Doctype {
...
+     */
+    private static async _calculateNonce(doctype: Doctype): Promise<number> {
+        if (doctype.context.ipfs == null) {
+            return 0 // return nonce 0 if IPFS is not available
+        }
+
+        // If there hasn't been any update prior to this we should not set the nonce
+        if (!doctype.state.next) return
+        // Get the current nonce and increment it by one
+        return (doctype.state.next.metadata?.nonce || 0)++
```
oed, line 140:
```suggestion
        return ++(doctype.state.next.metadata?.nonce || 0)
```

packages/ceramic-doctype-tile/src/tile-doctype.ts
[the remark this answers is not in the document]
oed: Remove these lines

packages/ceramic-core/src/anchor/ethereum/ethereum-anchor-service.ts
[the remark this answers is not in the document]
simonovic86: This is a temporary solution. We have a separate story for the refactor into separate modules [202](https://github.com/ceramicnetwork/js-ceramic/issues/202).

packages/ceramic-doctype-tile/src/tile-doctype.ts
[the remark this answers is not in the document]
simonovic86: I removed the nonce from genesis.

packages/ceramic-doctype-tile/src/tile-doctype.ts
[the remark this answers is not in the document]
simonovic86: 👍

packages/ceramic-doctype-tile/src/tile-doctype.ts
simonovic86: With this change you proposed the `log`s in the tests are the same. I'll take a look of how to implement the optimization you proposed (squashing the updates).

oed: That's great, ty!

packages/ceramic-doctype-tile/src/tile-doctype.ts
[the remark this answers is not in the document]
simonovic86: Removed `anchor` from `header`.

packages/ceramic-core/src/ceramic.ts
```diff
@@ -48,6 +50,8 @@ export interface CeramicConfig {
   gateway?: boolean;
 
   topic?: string;
+
+  [index: string]: any; // allow arbitrary properties
```
PaulLeCam, line 54: Maybe rather than allowing arbitrary config props, we could define the supported interfaces?
For example, we could have something like `anchorService?: EthereumAnchorServiceConfig | InMemoryAnchorServiceConfig`?

packages/ceramic-core/src/__tests__/ceramic-anchor.test.ts
```diff
@@ -0,0 +1,347 @@
...
+    const ceramic2 = await createCeramic(ipfs2, false, topic)
+    const owner = ceramic1.context.did.id
+
+    const doctype = await ceramic1.createDocument(DOCTYPE_TILE, { content: { x: 1 } }, { applyOnly: true })
+
+    await doctype.change({ content: { x: doctype.content.x + 1 }, metadata: { owners: [owner] } }, { applyOnly: false })
+    await doctype.change({ content: { x: doctype.content.x + 1 }, metadata: { owners: [owner] } }, { applyOnly: true })
+
+    let updatePromise = new Promise(resolve => {
+      doctype.on('change', () => {
```
PaulLeCam, line 275: Maybe it's not an issue in this case, but considering the `doctype` is reused and can emit the `change` event multiple times, using `.once()` rather than `.on()` could be better suited for this purpose.

PaulLeCam (commented):
Looks good to me, but I think it should be possible to make the tests a bit faster by making independent async calls in parallel, such as replacing

```ts
const ceramic1 = await createCeramic(ipfs1, true, topic)
const ceramic2 = await createCeramic(ipfs2, false, topic)
```
by
```ts
const [ceramic1, ceramic2] = await Promise.all([
  createCeramic(ipfs1, true, topic),
  createCeramic(ipfs2, false, topic)
])

File: packages/ceramic-core/src/ceramic.ts (hunk 7 above)
Comment (simonovic86) [reply packages/ceramic-core/src/ceramic.ts:54]: Yep, makes sense 👍  I think this is acceptable for now since anchor service(s) will be refactored as part of a separate story soon.

File: packages/ceramic-core/src/__tests__/ceramic-anchor.test.ts (hunk 8 above)
Comment (simonovic86) [reply packages/ceramic-core/src/__tests__/ceramic-anchor.test.ts:275]: Yep, makes sense!

File: packages/ceramic-doctype-tile/src/__tests__/three-id.test.ts [hunk 9]
```diff
@@ -80,15 +80,15 @@ const RECORDS = {
           }
         ],
         "prev": "bafybeig6xv5nwphfmvcnektpnojts33jqcuam7bmye2pb54adnrtccjlsu",
-        "header": {}
+        "header": {
+          "nonce": 0
```
oed, line 84: Possible to avoid setting nonce to 0? It could be implicitly 0 right?

packages/ceramic-doctype-tile/src/tile-doctype.ts
```diff
@@ -105,15 +111,33 @@ export class TileDoctype extends Doctype {
...
         return TileDoctype._signDagJWS(record, did, doctype.owners[0])
     }
 
+    /**
+     * Calculates anchor nonce
+     */
+    private static _calculateNonce(doctype: Doctype): number {
+        // if there hasn't been any update prior to this we should set the nonce to 0
+        if (!doctype.state.next) {
+            return 0
```
oed, line 135: return null or undefined here?

packages/ceramic-doctype-tile/src/tile-doctype.ts
```diff
@@ -105,15 +111,33 @@ export class TileDoctype extends Doctype {
             header.owners = newOwners
         }
 
+        header.nonce = TileDoctype._calculateNonce(doctype)
```
oed, line 114: Only set this if the calculated nonce > 0?

oed (approved):
Looks good. Only one small change I'd make!

packages/ceramic-doctype-tile/src/tile-doctype.ts
simonovic86: I can omit `0` but I set it here to `0` for consistency. `genesis` and `anchor` records don't have `nonce`s. Makes sense?

packages/ceramic-doctype-tile/src/__tests__/three-id.test.ts
simonovic86: yep, just wanted to have it here because of consistency.

packages/ceramic-doctype-tile/src/tile-doctype.ts
oed: Right, seems like it's just adding overhead for no real benefit though?

simonovic86: I wanted to avoid some future places in CAS where I would need to query if `nonce === 0 || nonce === undefined`. I'll revert it for now.

packages/ceramic-doctype-tile/src/__tests__/three-id.test.ts
```diff
@@ -81,7 +81,7 @@ const RECORDS = {
         ],
         "prev": "bafybeig6xv5nwphfmvcnektpnojts33jqcuam7bmye2pb54adnrtccjlsu",
         "header": {
-          "nonce": 0
+          "nonce": undefined
```
oed, line 84: Seems like the nonce is still being set here?

oed (commented):
looks like we're still setting the nonce? I meant that we should not set the property at all. Otherwise this would not be more efficient I guess :P

PaulLeCam (approved):

packages/ceramic-doctype-tile/src/__tests__/three-id.test.ts
simonovic86: then I'll remove `nonce` field whatsoever

simonovic86:
> looks like we're still setting the nonce? I meant that we should not set the property at all. Otherwise this would not be more efficient I guess :P

yep, you're right 👍

PR #409, opened 2020-10-16
Move schema related utils to core

simonovic86:
- move schema related utils to core and free some of the dependencies (`ajv` in particular)


packages/ceramic-cli/src/bin/ceramic.ts
```diff
@@ -91,13 +91,6 @@ schemas
         await CeramicCliUtils.schemaChangeDoc(docId, content, owners)
     })
 
-schemas
```
simonovic86, line 94: I think that this functionality is not necessary for now

packages/ceramic-cli/src/ceramic-cli-utils.ts
```diff
@@ -225,17 +225,7 @@ export class CeramicCliUtils {
...
-            if (!DoctypeUtils.isSchemaValid(schemaObj)) {
-                console.error('Invalid schema')
-                return
-            }
-        } catch (e) {
-            console.error(e.message)
-            return
-        }
-
+        // TODO validate schema on the client side
```
simonovic86, line 228: without client validation this is the same as `createDoc`. I left it here since I want to add client validation as well :)

PaulLeCam (approved):
Nice, thanks!

PR #424, opened 2020-10-26
chore: regenerate docs

simonovic86:


oed (commented):
Nice, I think I made a mistake when I updated the docs last time because now it's not rendering properly here: https://ceramicnetwork.github.io/js-ceramic/

I think it can be changed by creating the .nojekyll file in the docs folder and not docs/api. Could you update that before you merge this?

simonovic86:
> Nice, I think I made a mistake when I updated the docs last time because now it's not rendering properly here: https://ceramicnetwork.github.io/js-ceramic/
> 
> I think it can be changed by creating the .nojekyll file in the docs folder and not docs/api. Could you update that before you merge this?

yep, makes sense! 👍

oed (approved):

PR #426, opened 2020-10-26
feat(core): Rename 'head' to 'tip'

stbrody:


packages/ceramic-common/src/utils/doctype-utils.ts
```diff
@@ -127,7 +127,7 @@ export class DoctypeUtils {
      */
     static makeReadOnly<T extends Doctype>(doctype: T): T {
         doctype.change = (): Promise<void> => {
-            throw new Error('The version of the document is readonly. Checkout the latest HEAD in order to update.')
+            throw new Error('The version of the document is readonly. Checkout the latest TIP in order to update.')
```
stbrody, line 130: I wasn't totally sure if it still made sense to keep 'TIP' capitalized in the places where 'HEAD' was, so let me know if you think 'Tip' would be better here

stbrody:
I didn't update the docs this time, I was thinking it may be easier to wait and push a single docs update in a few days once all the breaking changes are in, but definitely let me know if it would be better to make sure the docs always stay in sync with the code and I can add a docs update commit to this PR.

packages/ceramic-common/src/utils/doctype-utils.ts
oed: I would change it to something like this:
```suggestion
            throw new Error('The version of the document is readonly. Load the document without version to make updates.')
```

oed (approved):
great!

packages/ceramic-common/src/utils/doctype-utils.ts
simonovic86: @oed yep, agree 👍

simonovic86 (approved):
LGTM 🙂

packages/ceramic-common/src/utils/doctype-utils.ts
stbrody: Is the issue that the version specified is readonly, or that *any* specific version is inherently readonly?  Would a better message be: "Historical document versions cannot be modified. Load the document without specifying a version to make updates."?

packages/ceramic-core/src/document.ts
```diff
@@ -314,12 +314,12 @@ class Document extends EventEmitter {
   }
 
   /**
-   * Handles HEAD from the PubSub topic
+   * Handles Tip from the PubSub topic
```
stbrody, line 317: New version updated these comments to use "Tip" instead of "TIP"

oed (approved):
Much better error msgs now :)

Issue #395, opened 2020-10-08
Create Ceramic Clayground


simonovic86:
- create Ceramic sandbox environment

PR #427, opened 2020-10-26
feat(core): Make isUnique default to true

stbrody:


packages/ceramic-cli/src/bin/ceramic.ts
```diff
@@ -25,7 +25,7 @@ program
     .option('--content <content>', 'New document content')
     .option('--only-genesis', 'Only create the genesis object. No anchor will be created')
     .option('--controllers <controllers>', 'Specify a comma-separated list of the controllers of the document. Controllers are the users that are allowed to publish updates to this document. Defaults to current user')
-    .option('--unique', 'Ensure document is unique regardless of content')
+    .option('--unique <unique>', 'Controls whether document is ensured to be unique regardless of content. Defaults to true')
```
stbrody, line 28: I wasn't sure how to test the behavior of the CLI.  I assume there's no automated tests and this will have to be done manually?  I guess that means it's time for me to figure out how to deploy a ceramic node locally and use the CLI with it.  Can you link me to the best starting point docs on that?

packages/ceramic-doctype-tile/src/__tests__/three-id.test.ts
```diff
@@ -49,6 +49,7 @@ const RECORDS = {
         controllers: [
           "did:key:zQ3shwsCgFanBax6UiaLu1oGvM7vhuqoW88VBUiUTCeHbTeTV"
         ],
+        isUnique: false,
```
oed, line 52: Can you edit the Tile doctype handler to make sure that if `isUnique` is set to `false` then the property is **not** included in the header of the genesis record?

packages/ceramic-cli/src/bin/ceramic.ts
simonovic86: Yep, you're right. We're still working on more detailed documentation for the CLI but you can start from [here](https://medium.com/ceramic/building-with-clay-getting-started-with-the-ceramic-network-3d7a8a1e01bf). The complete local environment should be completed as part of the story [395 ](https://github.com/ceramicnetwork/js-ceramic/issues/395).

packages/ceramic-doctype-tile/src/__tests__/three-id.test.ts
stbrody: If we do that, then documents can no longer roundtrip correctly.  I.E. if you copy an existing document and then create a new document with the copied output as the arguments, now you get different behavior (a unique document where the document you copied was global).  Given that users can provide their own metadata objects directly, modifying the input they provide out from under them seems fishy.

I actually kind of feel like either we should *always* include isUnique in the metadata, or *never* include it.  Including it conditionally based on its value seems brittle and conceptually unclear.  Either the uniqueness of a document is something that is inherently important to know about any given document, or it isn't.  I wonder if rather than considering 'isUnique' to be part of a document's metadata, instead it should just be considered an argument to the various functions for creating documents.  We already have the separate 'unique' field which ensures the behavior is correct, conceptually I'm not sure uniqueness is part of an object's inherent description/metadata and think it may just be some syntactic sugar to the document creation process.

packages/ceramic-doctype-tile/src/tile-doctype.ts
```diff
@@ -74,7 +74,7 @@ export class TileDoctype extends Doctype {
         }
 
         let unique: string
-        if (metadata.isUnique) {
+        if (metadata.isUnique || !('isUnique' in metadata)) { // default isUnique to true
```
stbrody, line 77: Making the change at the low level like I did in this PR means that schema document creation behavior changes as well, is that what we want?  It kind of makes sense to me that schema documents would default to being global even if other types of tile documents default to being unique, what do you think @oed @simonovic86?

packages/ceramic-doctype-tile/src/__tests__/three-id.test.ts
oed: Good point. Maybe we can set `unique` to 0 if false? It's a little strange to have two different representations here. Would be nice if it's always one field.

packages/ceramic-doctype-tile/src/tile-doctype.ts
oed: I think it would be weird ro have different behavior for different document content. Would prefer consistent behavior. 
Also don't see an inherent reason for a different behavior for schemas.

packages/ceramic-doctype-tile/src/tile-doctype.ts
```diff
@@ -73,9 +73,19 @@ export class TileDoctype extends Doctype {
             throw new Error('No DID authenticated')
         }
 
+        // TODO: stop checking metadata for 'isUnique' once all dependent libraries are using DocParams to specify it
```
stbrody, line 76: While testing I got burned once again by the circular dependency with identity-wallet.  Since this patch changes the way isUnique is passed as an argument to document creation, [this line](https://github.com/3box/identity-wallet-js/blob/443f9b68f514951c6063f4f991d15cd51e3c645e/src/keyring.ts#L193) in identity wallet needs to be updated to use the new format.  So to work around that, I added extra handling here so that we continue to check the metadata temporarily as a fallback to passing isUnique in docParams.  Once identity wallet is updated, then the metadata fallback code can be removed

stbrody:
Okay, new version is up with corrected CLI behavior, as well as removing isUnique from the document metadata.  @oed @simonovic86, PTAL!

packages/ceramic-cli/src/bin/ceramic.ts
```diff
@@ -77,10 +78,11 @@ schemas
     .command('create <new-content>')
     .option('--only-genesis', 'Only create the genesis object. No anchor will be created')
     .option('--controllers <controllers>', 'Specify a comma-separated list of the controllers of the schema document. Defaults to' + ' current user')
-    .option('--unique', 'Ensure schema document is unique regardless of content')
+    .option('--unique', 'Ensure schema document is unique regardless of content. This is the default behavior.')
+    .option('--not-unique', 'Creating a schema document with identical content to an existing schema document will be a no-op.')
```
oed, line 82: Are both of these really needed? Isn't `--not-unique` enough?

oed (commented):
Good idea to move it out of the metadata.

packages/ceramic-cli/src/bin/ceramic.ts
stbrody: yes, now it would be.  I guess I was thinking about some combination of backwards compatibility and future-proofing against potential future behavior changes.  But I can remove `--unique` if you prefer just having the one

stbrody: If I remove `--unique` do you still think `--not-unique` is the right name for the remaining flag?  I also thought about naming it something that describes what it *is* rather than what it is not, along the lines of `--global` or `--dedupe`, but went with `--not-unique` for symmetry with `--unique`.  I don't feel strongly either way so if you have a preference LMK

stbrody: @oed, thoughts on removing `--unique` and/or renaming `--not-unique`?

oed: Fine with removing `--unique`. Not sure about the global or dedupe names, but fine with renaming if we have a good name. Maybe `--deterministic`?

stbrody: Removed --unique.  I put together a patch to rename the command line flag from `--not-unique` to `--deterministic`, but when I looked at it the lack of symmetry/connection between the command line flag and the `isUnique` DocParam arg seemed weird and likely to confuse users, so I put it back to `--not-unique`, even though I don't love it.

oed: Fair enough. I guess another option could be to rename the `isUnique` option to `deterministic` and flip the boolean logic?

Both are fine by me. Latter requires a small change in IDW, but that's easy!

stbrody: Okay, just put up a new version that changes the `isUnique` bool to a `uniquenessMode` enum, which I like a lot better.  The tests will fail until IDW is updated - I'm happy to make the change to IDW just let me know if you like this approach and I'll go ahead and do so.

oed: Kind of like the `uniquenessMode` less because it requires devs to import an additional type for a thing that could just be a boolean. Seems a bit excessive. Would love to hear thoughts from @PaulLeCam though!

PaulLeCam: I like the `deterministic` name instead of `isUnique`, especially as the default behavior is changed.
About the `uniquenessMode` it's nice for forwards compatibility if we plan on having more than 2 states here, but if we don't have a good use-case of it a boolean would be simpler.

stbrody: Okay, new version up that puts it back to a bool but names it `deterministic`.  Tests are broken dependent on https://github.com/3box/identity-wallet-js/pull/93

oed (approved):

PR #435, opened 2020-10-27
Sync CAS logic with in-memory implementation

simonovic86:
- sync CAS logic with in-memory implementation
- fix CLI `seed` generation
- adapt tests

### DIFF
--- a/packages/ceramic-cli/src/ceramic-cli-utils.ts
+++ b/packages/ceramic-cli/src/ceramic-cli-utils.ts
@@ -278,7 +278,8 @@ export class CeramicCliUtils {
         const cliConfig = await CeramicCliUtils._loadCliConfig()
 
         if (!cliConfig.seed) {
-            cliConfig.seed = CeramicCliUtils._generateSeed()
+            cliConfig.seed = u8a.toString(randomBytes(32))
+            console.log('Identity wallet seed generated')
             await CeramicCliUtils._saveCliConfig(cliConfig)
         }
 
@@ -376,16 +377,6 @@ export class CeramicCliUtils {
 
         await fs.writeFile(fullCliConfigPath, JSON.stringify(cliConfig, null, 2))
         return cliConfig
-    }
-
-    /**
-     * Generate new seed
-     * @private
-     */
-    static _generateSeed(): string {
-        const seed = u8a.toString(randomBytes(32))
-        console.log('Identity wallet seed generated')
-        return seed
     }
 
     /**
--- a/packages/ceramic-cli/src/ceramic-cli-utils.ts
+++ b/packages/ceramic-cli/src/ceramic-cli-utils.ts
@@ -1,6 +1,7 @@
 import os from "os"
 import path from "path"
 import { randomBytes } from '@stablelib/random'
+import * as u8a from 'uint8arrays'
 
 const fs = require('fs').promises
 
@@ -257,7 +258,7 @@ export class CeramicCliUtils {
      */
     static async pinLs(docId?: string): Promise<void> {
        const id = DocID.fromString(docId)
-       
+
         await CeramicCliUtils._runWithCeramic(async (ceramic: CeramicApi) => {
             const pinnedDocIds = []
             const iterator = await ceramic.pin.ls(id)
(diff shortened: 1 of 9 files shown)


packages/ceramic-cli/src/ceramic-cli-utils.ts
```diff
@@ -381,7 +383,7 @@ export class CeramicCliUtils {
      * @private
      */
     static _generateSeed(): string {
-        const seed = '0x' + Buffer.from(randomBytes(32)).toString('hex')
+        const seed = u8a.toString(randomBytes(32))
```
oed, line 386: is this what IDW uses? Should it not be a uint8array?

packages/ceramic-core/src/anchor/memory/in-memory-anchor-service.ts
```diff
@@ -94,43 +121,64 @@ class InMemoryAnchorService extends AnchorService {
...
+    const { payload, signatures } = record
+    const { signature, protected: _protected } = signatures[0]
+
+    const decodedHeader = JSON.parse(base64url.decode(_protected))
+    const { kid } = decodedHeader
+
+    const didDoc = await this._ceramic.context.resolver.resolve(kid)
+    const jws = [_protected, payload, signature].join(".")
+    await didJwt.verifyJWS(jws, didDoc.publicKey)
+    return kid
```
oed, line 179: We likely need to strip the extra data from the DID. Should only be `did:3:bafy...`. Should not include `version-id` or `#<keyfragment`.

packages/ceramic-core/src/anchor/memory/in-memory-anchor-service.ts
```diff
@@ -94,43 +121,64 @@ class InMemoryAnchorService extends AnchorService {
...
+  async verifySignedRecord(record: any): Promise<string> {
+    const { payload, signatures } = record
+    const { signature, protected: _protected } = signatures[0]
+
+    const decodedHeader = JSON.parse(base64url.decode(_protected))
+    const { kid } = decodedHeader
+
+    const didDoc = await this._ceramic.context.resolver.resolve(kid)
+    const jws = [_protected, payload, signature].join(".")
+    await didJwt.verifyJWS(jws, didDoc.publicKey)
```
oed, line 178: you didn't end up using https://ceramicnetwork.github.io/js-did/classes/did.html#verifyjws ?

oed: Fine if current logic is easier 👍

oed (changes requested):
One small change needed :)

packages/ceramic-cli/src/ceramic-cli-utils.ts
simonovic86: this is just because of the serialization - saving it to `config.json`. I think that we can remove this method now 👍

packages/ceramic-core/src/anchor/memory/in-memory-anchor-service.ts
simonovic86: yeah, I did that initially but I changed my mind :) I'll revert it, tnx!

packages/ceramic-core/src/anchor/memory/in-memory-anchor-service.ts
simonovic86: yeah, I had resolvers in the context earlier 👍

PaulLeCam (approved):

packages/ceramic-core/src/anchor/memory/in-memory-anchor-service.ts
```diff
@@ -176,7 +176,7 @@ class InMemoryAnchorService extends AnchorService {
     const didDoc = await this._ceramic.context.resolver.resolve(kid)
     const jws = [_protected, payload, signature].join(".")
     await didJwt.verifyJWS(jws, didDoc.publicKey)
-    return kid
+    return kid.split('?')[0]
```
oed, line 179: we might have strings of the following format here:
`did:3:bafoier...?version-id=baiuehr...#Ab3d2`
and 
`did:key:Abm32...#Abm32`
We need to account for both of them. Safest is probably to use a regex similar to the did-resolver: https://github.com/decentralized-identity/did-resolver/blob/master/src/resolver.ts#L121-L123

simonovic86: I used a similar regex 👍

oed (approved):
sweet

PR #439, opened 2020-10-28
feat: remove log differentiation for gateway

valmack:
### Motivation

Keep log file names consistent across all instances of ceramic. Most people are not running both a node and a gateway so we don't need to add those labels to log file names.

If you do run both you can always set different paths  with the `--log-path` flag or use different filesystems (e.g. running ceramic in containers).

### Changes

- Removed component name from log filename

### DIFF
--- a/packages/ceramic-cli/src/ceramic-logger-plugins.ts
+++ b/packages/ceramic-cli/src/ceramic-logger-plugins.ts
@@ -14,10 +14,9 @@ import {
  */
 export class LogToFiles {
     /**
-     * Modifies `rootLogger` to append log messages to files named after components
-     * @notice If no component name is given 'default' will be included in the file name
+     * Modifies `rootLogger` to append log messages to files
      * @param rootLogger Root logger to use throughout the library
-     * @param loggerOptions Should include `component` name string and `logPath` string
+     * @param loggerOptions Not used by this plugin so should be null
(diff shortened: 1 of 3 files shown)


packages/ceramic-cli/src/ceramic-logger-plugins.ts
```diff
@@ -14,10 +14,9 @@ import {
  */
 export class LogToFiles {
     /**
-     * Modifies `rootLogger` to append log messages to files named after components
-     * @notice If no component name is given 'default' will be included in the file name
+     * Modifies `rootLogger` to append log messages to files
      * @param rootLogger Root logger to use throughout the library
-     * @param loggerOptions Should include `component` name string and `logPath` string
+     * @param loggerOptions null
```
simonovic86, line 19: should we add a description here?

simonovic86 (commented):
LGTM, I just left a minor comment :)

packages/ceramic-cli/src/ceramic-logger-plugins.ts
valmack: Yeah good idea. `loggerOptions` is not being used in this particular plugin so I'll add a note saying so there.

simonovic86 (approved):

PR #445, opened 2020-10-29
Use external modularized pinning backends

ukstv:
Drop right-on dependency on Powergate client


packages/ceramic-core/package.json
```diff
@@ -37,7 +37,9 @@
     "@ethersproject/base64": "^5.0.2",
     "@ethersproject/providers": "^5.0.0-beta.144",
     "@ethersproject/random": "^5.0.2",
-    "@textile/powergate-client": "^0.2.0",
+    "@pinning-aggregation/aggregation": "^2.1.1",
+    "@pinning-aggregation/common": "^2.0.0",
+    "@pinning-aggregation/ipfs-pinning": "^2.0.1",
```
oed, line 42: I guess it would be nice for these deps to be part of this monorepo as well?

oed (commented):
Awesome! I'll let @simonovic86 provide a more in depth review 👍

packages/ceramic-core/package.json
ukstv: Could be.

ukstv: They are the part of the repo now.

simonovic86 (approved):
sweet!

packages/ceramic-core/package.json
oed: Very cool, thanks!

PR #447, opened 2020-10-29
Log file rotation and small improvement

valmack:
### Motivation

1. We don't want to keep log files to for a long time (they will get large and take up space unnecessarily). We also don't want to implement full rotation because it can be complicated and is not needed atm.

2. With loki 2.0, arrays in json log messages are ignored, so we want to use dictionaries as much as possible.

### Changes

1. "Rotate" logs--so we maintain 2 files per filename, the first `*.log` is being actively written to and the second `*.log.old` contains logs from the previous 60 minutes

2. Use `headers` dictionary instead of `rawHeaders` array

### DIFF
--- a/packages/ceramic-cli/src/ceramic-logger-plugins.ts
+++ b/packages/ceramic-cli/src/ceramic-logger-plugins.ts
@@ -9,6 +9,8 @@ import {
     LoggerPluginOptions
 } from "@ceramicnetwork/ceramic-common"
 
+const fsPromises = fs.promises
+
 /**
  * Plugin for the root logger from the `loglevel` library to write logs to files
  */
@@ -35,14 +37,14 @@ export class LogToFiles {
             const rawMethod = originalFactory(methodName, logLevel, loggerName);
             return (...args: any[]): any => {
                 const message = LoggerProvider._interpolate(args)
-                fs.mkdir(basePath, { recursive: true }, (err) => {
+                fs.mkdir(basePath, { recursive: true }, async (err) => {
                     if (err && (err.code != 'EEXIST')) console.warn('WARNING: Can not write logs to files', err)
                     else {
                         const filePrefix = basePath + loggerName.toLowerCase()
                         const filePath = `${filePrefix}.log`
 
-                        LogToFiles._writeStream(filePath, message, 'a')
-                        LogToFiles._writeDocId(filePrefix, message)
+                        await LogToFiles._writeStream(filePath, message, 'a')
+                        await LogToFiles._writeDocId(filePrefix, message)
                     }
                 })
                 rawMethod(...args)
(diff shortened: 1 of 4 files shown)


packages/ceramic-cli/src/ceramic-logger-plugins.ts
```diff
@@ -83,4 +88,23 @@ export class LogToFiles {
             }
         }
     }
+
+    /**
+     * Returns true if it has been `MINUTES_TO_EXPIRATION` minutes since the file was created
+     * @param filePath Full path of file
+     */
+    private static _isExpired (filePath: string): boolean {
+        const { birthtime } = fs.statSync(filePath)
```
simonovic86, line 97: let's use promises here as well. Instead of `fs` sync functions you can import `fs` like this:

```
const fs = require('fs').promises
```

packages/ceramic-cli/src/ceramic-logger-plugins.ts
```diff
@@ -83,4 +88,23 @@ export class LogToFiles {
...
+        return (minutesSinceBirth >= LogToFiles.MINUTES_TO_EXPIRATION)
+    }
+
+    /**
+     * Renames the file given, appending `.old` to the file name
+     * @notice If a file with the new name already exists, it is overwritten
+     * @param filePath Full path of file
+     */
+    private static _rotate (filePath: string): void {
+        fs.renameSync(filePath, `${filePath}.old`)
```
simonovic86, line 108: same as above 🙂

simonovic86 (commented):
LGTM 👍  added a couple of minor comments

packages/ceramic-cli/src/ceramic-logger-plugins.ts
valmack: Good idea!

valmack:
@simonovic86 do you think we need test cases for this? and if so what's the best way to approach it? not sure if it makes sense to actually write to a file during test suite run

simonovic86:
> @simonovic86 do you think we need test cases for this? and if so what's the best way to approach it? not sure if it makes sense to actually write to a file during test suite run

@valmack let's add one simple test just to validate the swap of files. We can use `tmp-promise` like we do in some other tests. One of the examples is [here](https://github.com/ceramicnetwork/js-ceramic/blob/749558711ad55d00191d76e6d728909ecdf44757/packages/ceramic-core/src/__tests__/ceramic.test.ts#L3). Makes sense?

simonovic86 (approved):

valmack:
@simonovic86 actually ended up making the tests a bit more involved so we don't have to write to disk at all. But if it doesn't seem like a good fit I can simplify it with tmp-promise

simonovic86:
> @simonovic86 actually ended up making the tests a bit more involved so we don't have to write to disk at all. But if it doesn't seem like a good fit I can simplify it with tmp-promise

nice! no, this is definitely better! 🙃

PR #452, opened 2020-10-30
Fix pinning dependency

simonovic86:
- fix pinning dependency
- minor refactor (spaces, etc)
- add `references` to root config


PaulLeCam (approved):
Might be useful to start using Prettier sooner than later so that code formatting can be consistent across projects and people?

simonovic86:
> Might be useful to start using Prettier sooner than later so that code formatting can be consistent across projects and people?

Yep, agree 👍

oed (approved):

PR #455, opened 2020-11-02
IPFS version bump

simonovic86:
- bump following dependencies:
    * `IPFS` related deps
    * `typescript`
    * `lerna`
    * `@typescript-eslint/eslint-plugin`
    * `@typescript-eslint/parser`
- remove custom `IPFS` types
- fix lint


packages/ceramic-cli/src/declarations.d.ts
```diff
@@ -3,3 +3,5 @@ declare module 'uint8arrays' {
     export function fromString(s: string, enc?: string): Uint8Array
     export function concat(bs: Array<Uint8Array>): Uint8Array
 }
+
+export type { IPFSAPI as IPFSApi } from 'ipfs-core/dist/src/components'
```
oed, line 7: This seems a bit annoying :/

oed (approved):
Makes sense.
Seems like a new thing with ipfs is the differentiation between the `ipfs` and the `ipfs-core` packages: https://github.com/ipfs/js-ipfs#getting-started-1

Do you know what the difference is and what we should use?

packages/ceramic-cli/src/declarations.d.ts
simonovic86: they are still working out some issues :| here's a [link](https://github.com/ipfs/js-ipfs/issues/3358)

simonovic86:
> Makes sense.
> Seems like a new thing with ipfs is the differentiation between the `ipfs` and the `ipfs-core` packages: https://github.com/ipfs/js-ipfs#getting-started-1
> 
> Do you know what the difference is and what we should use?

@oed nice! I haven't noticed that. It seems that we can use `ipfs-core` instead of `ipfs` if we don't need a command line daemon 🙂

simonovic86:
> > Makes sense.
> > Seems like a new thing with ipfs is the differentiation between the `ipfs` and the `ipfs-core` packages: https://github.com/ipfs/js-ipfs#getting-started-1
> > Do you know what the difference is and what we should use?
> 
> @oed nice! I haven't noticed that. It seems that we can use `ipfs-core` instead of `ipfs` if we don't need a command line daemon 🙂

I'll squash this PR now and wait for the [issue](https://github.com/ipfs/js-ipfs/issues/3358) to resolve.

PR #459, opened 2020-11-03
feat(tile): Fix metadata handling and test schema enforcement behavior

stbrody:


oed (commented):
Looks like a good starting point. Let's discuss tomorrow what the behavior should be. 
Should this be merged before then?

packages/ceramic-doctype-tile/src/tile-doctype-handler.ts
```diff
@@ -133,14 +133,12 @@ export class TileDoctypeHandler implements DoctypeHandler<TileDoctype> {
             }
         }
 
-        let nextMetadata = nextState.metadata
-        if (payload.header?.controllers) {
-            nextState.next.metadata = { ...nextMetadata, controllers: payload.header.controllers }
-            nextMetadata = nextState.next.metadata
+        if (payload.header) {
+            nextState.next.metadata = {...nextState.metadata, ...payload.header}
```
simonovic86, line 137: @stbrody is there a reason why you "deleted" the [commit](https://github.com/ceramicnetwork/js-ceramic/commit/b39a768b3b8a7cf1d1ff7d625a6fe5b166808821) that I pushed yesterday that does this? Let's not do that because of the conflicts and merge fails.

simonovic86 (approved):
LGTM. I left a comment in the PR.

simonovic86 (commented):
LGTM. I left a comment in the PR.

simonovic86 (approved):

simonovic86:
> Looks like a good starting point. Let's discuss tomorrow what the behavior should be.
> Should this be merged before then?

Let's not merge this until we sync tomorrow. Makes sense?

packages/ceramic-doctype-tile/src/tile-doctype-handler.ts
stbrody: Sorry about that, I wanted to separate the commit that added tests from the one that fixed the bug so I redid your commit since it included test changes as well, then I force-pushed to the branch on github.  I thought that would be fine since there was no PR associated with it yet.

FYI, the line that this comment is on is slightly different here than it was in your version.  The original version used Object.extend, but that modified not just `nextState.next.metadata` but also `nextState.metadata` which seemed a bigger change and wasn't necessary to fix the issue.  Let me know if you think that `nextState.metadata` *should* also have the payload header folded in as well.

simonovic86: @stbrody nope, I think this is cool 👍

PR #461, opened 2020-11-04
Apply standard flow for applyRecord

simonovic86:
- apply standard flow for `applyRecord`

### DIFF
--- a/packages/ceramic-core/src/document.ts
+++ b/packages/ceramic-core/src/document.ts
@@ -24,6 +24,14 @@ import DocID from '@ceramicnetwork/docid'
 import { PinStore } from './store/pin-store';
 
 /**
+ * Provides a way to call _handleTip in a synchronous manner
+ */
+interface SyncCallback {
+  resolve: Function;
+  reject: Function;
+}
+
+/**
  * Document handles the update logic of the Doctype instance
  */
 class Document extends EventEmitter {
@@ -38,16 +46,15 @@ class Document extends EventEmitter {
   public readonly id: DocID
   public readonly version: CID
 
-  private logger: Logger
+  private _logger: Logger
+  private _isProcessing: boolean
 
-  private isProcessing: boolean
-
-  constructor (id: DocID, public dispatcher: Dispatcher, public pinStore: PinStore) {
+  constructor (id: DocID, public dispatcher: Dispatcher, public pinStore: PinStore, private _validate: boolean) {
     super()
     this.id = id
     this.version = id.version
 
-    this.logger = RootLogger.getLogger(Document.name)
+    this._logger = RootLogger.getLogger(Document.name)
 
     this._applyQueue = new PQueue({ concurrency: 1 })
     this._genesisCid = id.cid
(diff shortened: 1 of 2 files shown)


packages/ceramic-core/src/document.ts
```diff
@@ -234,41 +242,20 @@ class Document extends EventEmitter {
...
-
-    if (validate) {
-      const schema = await Document.loadSchemaById(this._context.api, state.metadata.schema)
-      if (schema) {
-        Utils.validate(state.next.content, schema)
-      }
-    }
-
-    this._doctype.state = state
+    const blockPromise = new Promise((resolve, reject) => {
```
PaulLeCam, line 249: Any details as to why there is this callback logic here please? Any reason the following wouldn't work?
```ts
await this._handleTip(cid)
await this._updateStateIfPinned()
await this._applyOpts(opts)
```

simonovic86: I need to catch the error - ```await this._handleTip(cid)``` just logs it.

PaulLeCam: If the error need to be caught by the caller then I think `_handleTip` should throw it, rather than having to use this kind of workaround that makes the logic harder to follow.

simonovic86: the error needs to be caught just sometimes - on client interaction. This method is also binded as THE method that will process messages from pubsub as well. That's why we don't need to process all the errors from it.

```
this.on('update', this._handleTip.bind(this))
```

PaulLeCam: Maybe the handler bound on the `update` event could be another method wrapping the `_handleTip` call in a `try {} catch {}` statement then so the error handling logic is always up to the consumer?

simonovic86: Yep, that's one of the solutions. I didn't want to refactor that in this PR.

simonovic86: @PaulLeCam refactored this as well 👍

PaulLeCam: Nice!

packages/ceramic-core/src/document.ts
```diff
@@ -677,15 +692,15 @@ class Document extends EventEmitter {
    * Gracefully closes the document instance.
    */
   async close (): Promise<void> {
-    this.off('update', this._handleTip.bind(this))
+    this.off('update', this._update.bind(this))
```
PaulLeCam, line 695: I don't think that works, every time the function is bound it's going to return a new bound function, so the callback attached is not actually removed. 
Instead the `_update` method itself could be bound, ex:
```ts
_update = async (cid: CID): Promise<void> => {
  try {
    await this._handleTip(cid)
  } catch (e) {
    this._logger.error(e)
  } finally {
    this._isProcessing = false
  }
}
```
and
```ts
this.on('update', this._update)
this.off('update', this._update)
```

simonovic86: This is the old code so I didn't want to go deep into. So, what you're saying is to remove `.bind(this)` for `on` and `off`, both?

oed (approved):
Makes sense overall.

packages/ceramic-core/src/document.ts
simonovic86: @PaulLeCam removed binding 👍

PR #467, opened 2020-11-04
feat(caip10-link): Make caip10-link records use 'data' instead of 'content', in accordance with the spec

stbrody:


oed (approved):
Great!

PR #468, opened 2020-11-05
Serialize Doc ID

simonovic86:
- serialize Doc ID on `create`


oed (approved):
Any other places in the daemon where this should happen? Should the http client convert the string to DocID again?

simonovic86:
> Any other places in the daemon where this should happen? Should the http client convert the string to DocID again?

I think we're good. The thing is that this is already fixed on `develop` but we haven't released it yet 🙂

PR #470, opened 2020-11-05
fix(document): Enforce schema when loading genesis record

stbrody:


stbrody:
Okay, I feel like I have a much firmer grasp of what's going on now, and have updated the tests here to reflect our current behavior more accurately.  We did have a bug where we weren't enforcing the schema on the genesis record when loading documents, but that's actually not why I was able to load non-conforming documents successfully in my original versions of the test.  That was working just because we have an in-memory document cache that we fetch documents from if we've seen them before.  Removing the document from the cache lets us test the actual behavior when loading a document, which is that `loadDocument` returns right after loading the genesis record, without applying any of the log.  The log application happens asynchronously off the pubsub channel, which does correctly do schema enforcement and discards the conflicting updates.  By that point, however, `loadDocument()` has already returned, so there's really no way to make `loadDocument` throw an exception if the document is non-conformant (unless that non-conformance applies to the genesis record, with the fix for that in this CR).  Instead what happens is you just get a document with the genesis record only, and no future updates ever become visible.  That is indeed probably the correct behavior, so I've updated the tests to test that more explicit and directly.

PR #472, opened 2020-11-05
fix(document): Enforce schema when loading genesis record

stbrody:


packages/core/src/__tests__/ceramic-api.test.ts
```diff
@@ -377,74 +377,6 @@ describe('Ceramic API', () => {
...
-      expect(doc.content).toEqual({ a: 'x' })
-
-      // Update schema so that existing doc no longer conforms
-      const updatedSchema = cloneDeep(stringMapSchema)
-      updatedSchema.additionalProperties.type = "number"
-      await schemaDoc.change({content: updatedSchema})
-      await syncDoc(schemaDoc)
-
-      // Test that we can load the existing document without issue
-      const doc2 = await ceramic.loadDocument(doc.id)
```
stbrody, line 421: The whole point of this test was to test the behavior when loading a document that doesn't conform to its schema, but this check here isn't actually telling us anything useful since it's not going through the "real" document loading process, but instead just fetching it out of the local cache [here](https://github.com/ceramicnetwork/js-ceramic/blob/28da783f6c6ca29dc925152ce4c3c5fc6e3c0bdd/packages/core/src/ceramic.ts#L368).

To do this test properly it would need to be re-written as an integration test with two ceramic instances, one that creates the non-conforming document and a second that loads it.  I'm happy to do that if you think it's worthwhile to add, but in the interest of time I decided to just focus on unit testing the core Document/applyRecord behavior and trust that it continues to work when being driven by the pubsub network

packages/core/src/__tests__/document.test.ts
```diff
@@ -326,6 +339,66 @@ describe('Document', () => {
...
+      }
+      // Create a document that isn't conforming to the schema
+      const doc = await create(docParams, ceramicWithoutSchemaValidation, context)
+      await anchorUpdate(doc)
+
+      expect(doc.content).toEqual({stuff:1})
+      expect(doc.metadata.schema).toEqual(schemaDoc.id.toString())
+
+      try {
+        await Document.load(doc.id, findHandler, dispatcher, pinStore, context, {skipWait:true})
```
stbrody, line 395: This is the actual test for the core fix in this PR.  Without the change to Document.load, this test fails.

stbrody:
I feel like I have a much better grasp of how this all works now.  My previous statement "we aren't enforcing schema validation when loading documents" was only partially correct.  We did indeed have a bug of not enforcing the schema when loading the genesis record, which is what this ticket fixes.  But for non-genesis update records, we do actually enforce the schema.  The reason I wasn't seeing that in my testing was due to [this document cache](https://github.com/ceramicnetwork/js-ceramic/blob/28da783f6c6ca29dc925152ce4c3c5fc6e3c0bdd/packages/core/src/ceramic.ts#L368) which causes my test code to not actually truly reload the document.  The actual behavior when loading a document where one of the intermediary update records doesn't conform to the schema as specified at the time of that update is that that signedRecord just gets thrown out silently.  This means that `loadDocument` will not throw an error if there's an update in the loaded document's log that is invalid due to a schema violation.  Instead, you'll just be left with an earlier version of the document from before the violation was introduced.

stbrody:
Hmm... I'm not sure why circleci failed to run the tests, they pass for me locally.  Any ideas how to go about looking into that?  @oed @simonovic86?

It really would be nice if our test system logged what test was being run **before** running it, instead of only after...

oed:
Yeah, idk why circleci is not showing which test result in an error 🤔

simonovic86:
> Hmm... I'm not sure why circleci failed to run the tests, they pass for me locally. Any ideas how to go about looking into that? @oed @simonovic86?
> 
> It really would be nice if our test system logged what test was being run **before** running it, instead of only after...

It seems that some of the tests are failing in a random manner. I guess that's related with how we're using the `InMemoryAnchorService`. I suggest that we call the anchoring explicitly (like in `ceramic-anchor.test.ts`) and then wait for the document to sync. Makes sense? :)

packages/core/src/document.ts
```diff
@@ -145,6 +145,13 @@ class Document extends EventEmitter {
       doc._doctype.state = await doc._doctypeHandler.applyRecord(record, doc._genesisCid, context)
     }
 
+    if (validate) {
```
simonovic86, line 148: yep, that's it 👍

simonovic86 (commented):
LGTM. I would only suggest that we explicitly call the anchoring and then wait for the doc to sync. I guess that would solve issues with random test fails.

packages/core/src/__tests__/ceramic-api.test.ts
```diff
@@ -38,12 +38,14 @@ const createIPFS =(overrideConfig: Record<string, unknown> = {}): Promise<IPFSAp
  * document is anchored
  * @param doc
  */
-const syncDoc = async (doc: any): Promise<void> => {
-  await new Promise(resolve => {
+const anchorDoc = async (ceramic: Ceramic, doc: any): Promise<void> => {
+  const p = new Promise(resolve => {
```
stbrody, line 42: I did this a little differently than how it works in ceramic-anchor-test, but I think this way is more correct?  If the concern is that the doc could get anchored before we set up the change listener, then I think the way ceramic-anchor-test does it (`anchor()` then immediately `syncDoc()`) makes the race window smaller, but doesn't completely eliminate it, whereas this way does by ensuring we're already listening for a change before requesting an anchor.  Does that make sense to you @simonovic86?  Should I update ceramic-anchor-test to do the same?

packages/core/src/__tests__/ceramic-anchor.test.ts
```diff
@@ -50,18 +50,27 @@ const createCeramic = async (ipfs: IPFSApi, anchorManual: boolean, topic: string
...
 }
 
-const anchor = async (ceramic: Ceramic): Promise<void> => {
-  await ceramic.context.anchorService.anchor()
-}
-
-const syncDoc = async (doctype: Doctype): Promise<void> => {
-  await new Promise(resolve => {
-    doctype.on('change', () => {
+const registerChangeListener = function (doc: Doctype): Promise<void> {
```
stbrody, line 53: I hit some weirdness trying to get this to work.  I originally had this as an async function, but then waiting on the returned promise in anchorDoc would hang. Removing the 'return' keyword after making it async would get it to stop hanging, but then it would fail on the next line in every test because the document log would be one entry smaller than expected.  I definitely think I'm hitting up against some rough edges of the concurrency/execution model of Node.js.  Would you be open to taking a look at this together tomorrow morning @simonovic86?

stbrody: Okay nevermind, I more or less see the problem. Making registerChangeListener async makes the compiler think that saying `await registerChangeListener(doc)` returns `void`, rather than `Promise<void>` like it should be.  I tried to change the return type to `Promise<Promise<void>>` but that doesn't seem to help, the compiler seems to flatten the stack of Promises. I even found a [bug report](https://github.com/microsoft/TypeScript/issues/27711) about it.  Anyway I think the right move is just to leave the function synchronous, as I've done here.  So I think this should be good as-is, pending your review @simonovic86

simonovic86: yep, this is OK 👍

packages/core/src/__tests__/ceramic-api.test.ts
simonovic86: feel free to add a test into `ceramic.test.ts` test suite which is already spinning a couple of instances (ipfs+ceramic). It currently serves as a **basic** integration test for testing propagation between multiple instances. On the other side, it's not crucial to add it as part of this PR, we can open another story for that 👍

simonovic86 (approved):
lgtm 🤘

PR #477, opened 2020-11-07
feat(tile): Enforce that schemas must always be assigned with a version

stbrody:


oed (approved):
Looks good. Ideally the metadata about versions and latest version should be encoded in the log, but that's a separate issue.

PR #486, opened 2020-11-10
feat(core): Remove doctype from genesis record

stbrody:


packages/core/src/ceramic.ts
```diff
@@ -287,32 +288,32 @@ class Ceramic implements CeramicApi {
 
   /**
    * Creates doctype from genesis record
+   * @param doctype - Document type
    * @param genesis - Genesis CID
    * @param opts - Initialization options
    */
-  async createDocumentFromGenesis<T extends Doctype>(genesis: any, opts: DocOpts = {}): Promise<T> {
-    const doc = await this._createDocFromGenesis(genesis, opts)
+  async createDocumentFromGenesis<T extends Doctype>(doctype: string, genesis: any, opts: DocOpts = {}): Promise<T> {
```
stbrody, line 295: It irks me a little that this, `findHandler`, and `createDocument` all need the doctype passed in as a string argument when they're already specialized on the Doctype, but my knowledge of typescript generics and inheritance still is still pretty basic and it didn't seem worth it for me to wade into that can of worms to try to figure out how to remove the string argument at this time.  Seems like it should be theoretically possible to get the type name from the type specialization though.

packages/http-client/src/document.ts
```diff
@@ -27,19 +27,9 @@ class Document extends Doctype {
     return new DocID(this.state.doctype, this.state.log[0])
   }
 
-  static async create (apiUrl: string, doctype: string, params: DocParams, context: Context, opts: DocOpts = {}): Promise<Document> {
```
stbrody, line 30: This whole function was dead code, best as I can tell.   I don't even think the `/create` endpoint in the daemon would be able to handle arguments passed the way they are being done here.

packages/core/src/ceramic.ts
```diff
@@ -179,12 +179,13 @@ class Ceramic implements CeramicApi {
...
    */
-  findHandler<T extends DoctypeHandler<Doctype>>(genesisRecord: Record<string, any>): T {
-    if (genesisRecord.doctype in this._doctypeHandlers) {
-      return this._doctypeHandlers[genesisRecord.doctype] as T
+  findHandler<T extends DoctypeHandler<Doctype>>(doctype: string, genesisRecord: Record<string, any>): T {
+    if (doctype in this._doctypeHandlers) {
+      return this._doctypeHandlers[doctype] as T
     } else if (genesisRecord['@context'] === "https://w3id.org/did/v1") {
       return this._doctypeHandlers['3id'] as T
     }
```
oed, line 191: I think this method can be removed since we would only get the handler from the doctype now. 
`genesisRecord['@context'] === "https://w3id.org/did/v1"` was an old approach to 3ID backwards compatibility which can be removed now.

packages/core/src/ceramic.ts
oed: I don't think you actually need to specify the generic when you call this method?

stbrody: interesting.  What type gets returned in that case?

oed: Types doesn't exist runtime, but compile time I guess it's just the generic `Doctype`.

packages/core/src/ceramic.ts
```diff
@@ -365,8 +347,13 @@ class Ceramic implements CeramicApi {
...
     const docIdStr = docId.toString()
 
+    const doctypeHandler = this._doctypeHandlers[docId.typeName]
+    if (!doctypeHandler) {
+      throw new Error(docId.typeName + " is not a valid doctype")
+    }
+
     if (!this._docmap[docIdStr]) {
-      this._docmap[docIdStr] = await Document.load(docId, this.findHandler.bind(this), this.dispatcher, this.pinStore, this.context, opts)
+      this._docmap[docIdStr] = await Document.load(docId, doctypeHandler, this.dispatcher, this.pinStore, this.context, opts)
```
oed, line 356: Nice! Now we don't need to pass the find function 👍

oed (approved):
lgtm

PR #487, opened 2020-11-11
chore: use learna pacages with typedoc and add pages plugin

oed:
This PR enables the lerna plugin for typedoc which much better separates the api docs into modules. It also includes the pages plugin which allows us to have multiple markdown pages within typedoc.

Ideally we should move the generated doc to be deployed to S3 separate from the repo using a github action each time we push code to master, but that's a separate story ;()


docs-src/doctype-starter-doc.md
```diff
@@ -0,0 +1,362 @@
...
+ abstract class Doctype extends EventEmitter {  
+    constructor(private _state: DocState, private _context: Context) {  
+        super()  
+    }  
+  
+    get id(): string {  
+        return DoctypeUtils.createDocId(this.state.log[0])  
+    }  
+  
+    get doctype(): string {
```
stbrody, line 36: Unrelated to this CR, but I really think we need to either rename the Doctype class or rename every place we refer to the string name of a doctype.  I'd love to rename the class but I can't actually think of a better name for it.  So maybe the thing to do is change everyplace where we're referring to the type name to be "doctypename" (doctypeName? doctype_name?)

Anyway this is totally an aside, just something that's been bugging me

docs-src/doctype-starter-doc.md
```diff
@@ -0,0 +1,362 @@
...
+The method `change` is more or less sugar coding for the developer since all the operations can be implemented using the `CeramicAPI` interface. For example the snipped below describes `TileDoctype` included in the Ceramic node out-of-the-box.
+
+```typescript
+const DOCTYPE = 'tile'  
+  
+/**  
+ * Tile doctype parameters 
+ */
+export interface TileParams {  
+	content: object;
```
stbrody, line 84: inconsistent tabbing throughout

docs-src/doctype-starter-doc.md
```diff
@@ -0,0 +1,362 @@
...
+One interface worth noting is the `CeramicAPI` interface which lets the developer communicate with the Ceramic node in two ways:
+- **directly**: the node is the process running on the same machine
+- **indirectly**: the node is running somewhere else and the communication is done over the HTTP protocol.
+	
+# Doctype
+The interface is listed in the code snipped below.
+
+```typescript
+/**  
+ * Describes common doctype attributes
```
stbrody, line 25: Instead of copying code and embedding it into the docs where it can quickly get out of sync, should we just be linking to the relevant code in github, perhaps pinned to the last stable release version?

docs-src/doctype-starter-doc.md
oed: Not sure we need this property anymore since you can get it from the DocID?

docs-src/doctype-starter-doc.md
oed: Good point, mainly copied this document over. I think we can now actually link the correct code in the docsite

oed:
@stbrody pushed an update that uses symbol links in the doctype guide.

stbrody (approved):

PR #488, opened 2020-11-11
Enable IPFS DHT functionality

simonovic86:
- enable IPFS DHT functionality

### DIFF
--- a/packages/ceramic-cli/src/ceramic-cli-utils.ts
+++ b/packages/ceramic-cli/src/ceramic-cli-utils.ts
@@ -25,7 +25,7 @@ const DEFAULT_CLI_CONFIG_FILE = 'config.json'
 export const DEFAULT_PINNING_STORE_PATH = ".pinning.store"
 const DEFAULT_CLI_CONFIG_PATH = path.join(os.homedir(), '.ceramic')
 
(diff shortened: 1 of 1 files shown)


packages/ceramic-cli/src/ceramic-cli-utils.ts
```diff
@@ -25,6 +25,8 @@ const DEFAULT_CLI_CONFIG_FILE = 'config.json'
 export const DEFAULT_PINNING_STORE_PATH = ".pinning.store"
 const DEFAULT_CLI_CONFIG_PATH = path.join(os.homedir(), '.ceramic')
 
+const IPFS_DHT_CLIENT_MODE = process.env.DHT_CLIENT_MODE === 'true'
```
PaulLeCam, line 28: I don't know what the client mode is but assuming it's more suited for short-running nodes, shouldn't we make it the default?
I'd imagine most users of the CLI would start the daemon for temporary use when testing their app rather than as a long-running server, no?

simonovic86: @PaulLeCam yep, makes more sense 👍

PaulLeCam (approved):

PR #491, opened 2020-11-12
feat(core): Interpret anchorServiceURL as base URL of the CAS, without the full path to the API endpoint

stbrody:


packages/cli/src/ceramic-daemon.ts
```diff
@@ -11,7 +11,7 @@ import * as core from "express-serve-static-core"
 const DEFAULT_PORT = 7007
 const toApiPath = (ending: string): string => '/api/v0' + ending
 
-const DEFAULT_ANCHOR_SERVICE_URL = "https://cas.3box.io:8081/api/v0/requests"
+const DEFAULT_ANCHOR_SERVICE_URL = "https://cas.3box.io:8081"
```
oed, line 14: 👍

packages/cli/src/bin/ceramic.ts
```diff
@@ -6,7 +6,7 @@ program
     .command('daemon')
     .option('--ipfs-api <url>', 'The ipfs http api to use')
     .option('--ethereum-rpc <url>', 'The Ethereum RPC URL used for communicating with Ethereum blockchain')
-    .option('--anchor-service-api <url>', 'The anchor service URL to use')
+    .option('--anchor-service-url <url>', 'The anchor service URL to use')
```
oed, line 9: Is there a reason you renamed this? Seems like the param name should make this clear enough.

valmack: @oed Actually I had been feeling like these flags should be more descriptive.. "ipfs-api-url" and "anchor-service-api-url" leave little room for doubt.

oed: But isn't it redundant when the parameter name is `<url>` anyway?

valmack: Yeah not a huge deal either way but a bit easier to remember. For me I may forget what the flag accepts (e.g. port number, full url, version?) so I would have to take an extra step and do `--help` to be sure.

stbrody: I also prefer including 'url' in the name so it's clearer.  I'd go with either `anchor-service-url` or `anchor-service-api-url` personally.  But I also don't feel strongly either way and am happy to go with whatever you think is best @oed

oed: As long as all options are consistent I'm fine either way @stbrody. Right now `-url` is only on `anchor-service`.

stbrody: You're right, if we change this we should change all the other flag names too.  Maybe it's not actually worth it.  New version up that puts the flag name back the way it was

oed (approved):

PR #493, opened 2020-11-12
Feat/chainid in metadata

stbrody:


packages/doctype-tile/src/tile-doctype.ts
```diff
@@ -64,14 +67,20 @@ export class TileDoctype extends Doctype {
...
 
         // check for DID and authentication
         if (!context.did || !context.did.authenticated) {
             throw new Error('No DID authenticated')
         }
 
+        if ('chainId' in metadata) {
+            throw new Error('Cannot manually specify chainId')
+        }
+        const chainId = await context.anchorService?.getChainId()
```
stbrody, line 81: So, I got most of the way through creating this commit, but am hitting problems due to how the http-client is implemented.  The http client [calls this function directly](https://github.com/ceramicnetwork/js-ceramic/blob/abe87fd834dc85334f1e449acae39655cb7fe1f6/packages/http-client/src/ceramic-http-client.ts#L73) and then [serializes the resulting record](https://github.com/ceramicnetwork/js-ceramic/blob/abe87fd834dc85334f1e449acae39655cb7fe1f6/packages/http-client/src/document.ts#L43) over to the daemon, which in turn puts it into IPFS and publishes the new CID.  The problem with the HTTP client generating the genesis record itself is that it won't have the same `context` as the daemon.  Notably, the http client doesn't know anything about the anchorService, and thus can't fetch the chainId from it.

The HTTP client delegates most of its functionality over to the daemon, except for the task of actually generating the records.  It has to generate the records itself so that it can apply the proper signatures to them.

The only solution I can think of to deal with this is to change the Ceramic core's [_createDocFromGenesis](https://github.com/ceramicnetwork/js-ceramic/blob/abe87fd834dc85334f1e449acae39655cb7fe1f6/packages/core/src/ceramic.ts#L305) modify the provided genesis record to add the chainId before sending it to the dispatcher to be stored on IPFS.  I don't love that since it means that the record that ultimately gets stored is no longer the same as the one returned from DocType.makeGenesis, which may be surprising behavior to some users.  But I really can't think of any other way to do it that fetches the chainid from the anchor service configuration.  The only other approach I can think of is to have the user provide the chainid themselves every time they create a document, which doesn't seem great either.

I'd be curious for any thoughts you have here @oed @simonovic86

stbrody: Oh actually, modifying the record in _createDocFromGenesis won't work either since by that point the record has already been signed.

I guess the HTTP client could send a separate request to the ceramic daemon for the sole purpose of finding the chainId that the daemon is using, then the chainId could be passed into DocType.makeGenesis rather than having makeGenesis fetch the chainId itself.  I can give that a try tomorrow if you agree that that's the right approach @oed.

PR #494, opened 2020-11-13
ref: Remove identity-wallet dep, use did:key

oed:
This PR removes `identity-wallet` as a dependency and replaces it with `key-did-provider-ed25519`.


packages/cli/src/ceramic-cli-utils.ts
```diff
@@ -329,9 +329,8 @@ export class CeramicCliUtils {
         }
 
         const seed = u8a.fromString(cliConfig.seed)
```
simonovic86, line 331: I guess we need to update the `seed` generation as well, right?

```js
if (!cliConfig.seed) {
  cliConfig.seed = u8a.toString(randomBytes(32))
  console.log('Identity wallet seed generated')
  await CeramicCliUtils._saveCliConfig(cliConfig)
}
```

simonovic86 (commented):
nice! I left a comment related to the CLI and non-existing `seed`

PaulLeCam (approved):

packages/cli/src/ceramic-cli-utils.ts
oed: if it's 32 bytes it should be correct @simonovic86

oed: Everywhere in the tests we had 64 bytes for some reason 🤷

simonovic86: nice! 🙂

simonovic86 (approved):

PR #496, opened 2020-11-15
ref: Store version information in DocState

### LINKED ISSUE #495: Rename Document.currentVersionDocID to Document.versionId

oed:
This PR adds version information to the `DocState` which eliminates the need for extra calls to get the versions of a document.

Changes:
- Remove `listVersions` method
- Rename `currentVersionDocID` -> `versionId`
- Add `allVersionIds` property to Doctype class
- Remove `versions` endpoint on daemon
- Remove `lastAnchored` property from `DocState`

Fixes #495


simonovic86 (approved):
nice!

packages/common/src/doctype.ts
```diff
@@ -70,6 +70,10 @@ export interface DocNext {
     metadata?: DocMetadata;
 }
 
+export interface LogEntry {
+  cid: CID
+  isVersion?: boolean
```
stbrody, line 75: Great improvement making the log array contain structured objects rather than just the CID!

I wonder if instead of an isVersion bool we should store something representing the type of Record (GenesisRecord, SignedRecord, AnchorRecord) this entry represents.  Then when filtering the log for entries that are also versions we filter for Genesis or Anchor Records.  Gives us more information in the log, and pushes the logic for deciding which entries are versions from where the log entries are built to where we are actually requesting the set of versions, which puts all the code for dealing with versions in one place.

packages/common/src/doctype.ts
```diff
@@ -70,6 +70,10 @@ export interface DocNext {
     metadata?: DocMetadata;
 }
 
+export interface LogEntry {
```
stbrody, line 73: Should we call this LogRecord?  We seem to call these things Records everywhere else.

stbrody (commented):
Overall this looks like a great cleanup!  Made a couple of small suggestions in comments.

packages/common/src/doctype.ts
oed: I think we then risk overloading the term "Record" a bit. Usually we use the term to refer to the content of the record that is stored in IPFS. In this log entry we're just storing the CID of the record + metadata.

packages/common/src/doctype.ts
oed: This makes sense. I'll make an update.

stbrody (approved):
LGTM

PR #508, opened 2020-11-17
feat(docid): base36 toUrl

zachferland:
re @PaulLeCam 

Changes docid.toUrl() to not require a base, and now always returns base36 so that always urls safe


oed (approved):

PaulLeCam:
Nice, thanks!

simonovic86 (approved):

PR #509, opened 2020-11-17
feat(core): Add chainId to genesis header

stbrody:


packages/http-client/src/ceramic-http-client.ts
```diff
@@ -131,6 +132,17 @@ class CeramicClient implements CeramicApi {
     }
   }
 
+  async getChainId(): Promise<string> {
```
oed, line 135: I think we should change this to something like this: 
```suggestion
  async getSupportedChains(): Promise<Array<string>>
```
This way we don't need to introduce a breaking change when nodes can support multiple anchor services.

packages/common/src/anchor-service.ts
```diff
@@ -29,4 +29,9 @@ export abstract class AnchorService extends EventEmitter {
      */
     abstract validateChainInclusion(anchorProof: AnchorProof): Promise<void>;
 
+    /**
+     * @returns a string representing the CAIP-2 chain ID of the configured blockchain.
+     */
+    abstract getChainId(): Promise<string>;
```
oed, line 35: I guess we need to change it here as well :)

oed (changes requested):
Looks good, one change I'd like in here.

packages/http-client/src/ceramic-http-client.ts
stbrody: Good idea!  New version up that does this @oed.  Also updated the CAS side to also return an array in https://github.com/ceramicnetwork/ceramic-anchor-service/pull/151

oed (approved):

simonovic86 (approved):
lgtm!

PR #511, opened 2020-11-17
feat: 3id resolve v0 and v1 3ids

zachferland:
Resolves v0 3ids through api, likely in future will want make sure all existing did docs are available in ipfs elsewhere (ie our ceramic ipfs node), and ipfs instance can be passed from ceramic instance already passed to resolver. Likely would not be reliable to resolve through existing 3box ipfs node, as have to directly connect and ipfs ceramic/3box versions diverging.

### DIFF
--- a/packages/3id-did-resolver/package.json
+++ b/packages/3id-did-resolver/package.json
@@ -27,10 +27,8 @@
     "clean": "rm -rf ./lib"
   },
   "dependencies": {
-    "3id-resolver": "^1.0.1",
     "@ceramicnetwork/common": "^0.12.2",
     "@ceramicnetwork/docid": "^0.2.0",
-    "bs58": "^4.0.1",
     "cross-fetch": "^3.0.6"
   },
   "devDependencies": {
--- a/packages/3id-did-resolver/package.json
+++ b/packages/3id-did-resolver/package.json
@@ -29,7 +29,9 @@
   "dependencies": {
     "@ceramicnetwork/common": "^0.12.2",
     "@ceramicnetwork/docid": "^0.2.0",
-    "cross-fetch": "^3.0.6"
+    "cids": "^1.0.2",
+    "cross-fetch": "^3.0.6",
+    "uint8arrays": "^1.1.0"
   },
   "devDependencies": {
     "@babel/core": "^7.9.0",
--- a/packages/3id-did-resolver/package.json
+++ b/packages/3id-did-resolver/package.json
@@ -27,8 +27,11 @@
     "clean": "rm -rf ./lib"
   },
   "dependencies": {
+    "3id-resolver": "^1.0.1",
     "@ceramicnetwork/common": "^0.12.2",
-    "bs58": "^4.0.1"
+    "@ceramicnetwork/docid": "^0.2.0",
+    "bs58": "^4.0.1",
+    "cross-fetch": "^3.0.6"
   },
   "devDependencies": {
     "@babel/core": "^7.9.0",
--- a/packages/3id-did-resolver/src/__tests__/__snapshots__/index.test.ts.snap
+++ b/packages/3id-did-resolver/src/__tests__/__snapshots__/index.test.ts.snap
@@ -52,13 +52,13 @@ Object {
     Object {
       "controller": "did:3:k2t6wyfsu4pg0t2n4j8ms3s33xsgqjhtto04mvq8w5a2v5xo48idyz38l7ydki",
       "id": "did:3:k2t6wyfsu4pg0t2n4j8ms3s33xsgqjhtto04mvq8w5a2v5xo48idyz38l7ydki#signing",
-      "publicKeyHex": "03fff18103bdca9458e73cca5f0a0f64964312d686712f2190bf5b4794c6e29924",
+      "publicKeyBase58": "2Bv3fQSJiGJsbUUfUdiPsXXZGG5cXqaSeYnieuJcUFAbm",
       "type": "Secp256k1VerificationKey2018",
     },
     Object {
       "controller": "did:3:k2t6wyfsu4pg0t2n4j8ms3s33xsgqjhtto04mvq8w5a2v5xo48idyz38l7ydki",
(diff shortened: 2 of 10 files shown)


zachferland:
hmm may change this check if cid to resolve v0, otherwise this breaks tests everywhere we are using invalid 3id strings to mock 3ids

packages/3id-did-resolver/package.json
```diff
@@ -27,8 +27,11 @@
     "clean": "rm -rf ./lib"
   },
   "dependencies": {
+    "3id-resolver": "^1.0.1",
```
oed, line 30: Can we please not use this?

zachferland:
"sub" did resolution not added from 3id-resolver given our migration path

packages/3id-did-resolver/src/index.ts
```diff
@@ -88,13 +93,53 @@ export function wrapDocument(content: any, did: string): DIDDocument {
...
+  const legacyDoc = await LegacyResolver(didId)
+  if (!legacyDoc) return null
+  if (version === '0') return legacyDoc
+
+  const keyEntry = legacyDoc.publicKey.findIndex(e => e.id.endsWith('managementKey'))
+  const managementKey = legacyDoc.publicKey[keyEntry].ethereumAddress 
+  if (!managementKey) return null
+
+  try {
+    const content =  { metadata: { controllers: [`did:key:${managementKey}`], tags: ['3id'] } }
```
oed, line 121: We can't have a `did:key` that's an ethereumAddress. Therefore in IDW you can see that we use the `signingKey` as the `managementKey` for the genesis record.

packages/3id-did-resolver/src/legacyResolver.ts
```diff
@@ -0,0 +1,82 @@
...
+      doc.authentication = JSON.parse(JSON.stringify(doc.authentication).replace(re, documentCid))
+    }
+    if (doc.service) {
+      doc.service = JSON.parse(JSON.stringify(doc.service).replace(re, documentCid))
+    }
+  }
+  if (doc.previousDocument) {
+    // make CID human readable
+    doc.previousDocument = { '/': doc.previousDocument.toString() }
+  }
```
oed, line 62: This can be removed as it's never set.

packages/3id-did-resolver/src/legacyResolver.ts
```diff
@@ -0,0 +1,82 @@
...
+    doc.id = doc.id.replace(re, documentCid)
+    if (doc.publicKey) {
+      doc.publicKey = JSON.parse(JSON.stringify(doc.publicKey).replace(re, documentCid))
+    }
+    if (doc.authentication) {
+      doc.authentication = JSON.parse(JSON.stringify(doc.authentication).replace(re, documentCid))
+    }
+    if (doc.service) {
+      doc.service = JSON.parse(JSON.stringify(doc.service).replace(re, documentCid))
+    }
```
oed, line 57: Don't think we ever used services here, so can probably be removed.

oed (changes requested):
Looks good overall, one main change needed. I also wonder how we can get the v0 docs migrated to our new ipfs infra?

packages/3id-did-resolver/src/index.ts
oed, line 121: We should also change it from tags to family: 
```suggestion
    const content =  { metadata: { controllers: [`did:key:${managementKey}`], family: '3id' } }
```
See here: https://github.com/3box/identity-wallet-js/pull/97

packages/3id-did-resolver/src/index.ts
```diff
@@ -26,14 +35,14 @@ export function wrapDocument(content: any, did: string): DIDDocument {
...
-      const keyBuf = bs58.decode(keyValue.slice(1))
+      const keyBuf = u8a.fromString(keyValue.slice(1), 'base58btc')
       if (keyBuf[0] === 0xe7) { // it's secp256k1
         diddoc.publicKey.push({
           id: `${did}#${keyName}`,
           type: 'Secp256k1VerificationKey2018',
           controller: did,
           // remove multicodec varint and encode to hex
-          publicKeyHex: keyBuf.slice(2).toString('hex')
+          publicKeyHex: u8a.toString(keyBuf.slice(2), 'base16')
```
oed, line 45: Can we change this to be `publicKeyBase58`? did-jwt is updated to support it!

packages/3id-did-resolver/src/index.ts
```diff
@@ -45,14 +54,14 @@ export function wrapDocument(content: any, did: string): DIDDocument {
           id: `${did}#${keyName}`,
           type: 'Curve25519EncryptionPublicKey',
           controller: did,
-          publicKeyBase64: keyBuf.slice(2).toString('base64')
+          publicKeyBase64: u8a.toString(keyBuf.slice(2), 'base64')
```
oed, line 57: `publicKeyBase58` here as well!

packages/3id-did-resolver/src/__tests__/__snapshots__/index.test.ts.snap
```diff
@@ -58,7 +58,7 @@ Object {
     Object {
       "controller": "did:3:k2t6wyfsu4pg0t2n4j8ms3s33xsgqjhtto04mvq8w5a2v5xo48idyz38l7ydki",
       "id": "did:3:k2t6wyfsu4pg0t2n4j8ms3s33xsgqjhtto04mvq8w5a2v5xo48idyz38l7ydki#encryption",
-      "publicKeyBase64": "N2/yXOxq8FCX4zzht8JQA3ftVRl+6C39CeU82eFB4Ug=",
+      "publicKeyBase64": "N2/yXOxq8FCX4zzht8JQA3ftVRl+6C39CeU82eFB4Ug",
```
oed, line 61: The reason `=` is gone is because you used `base64` instead of `base64pad`.

oed (approved):
Looks great. Just one small nitpick.

oed:
Also fix CI :)

oed (approved):
Pushed some small changes to how the 3IDv0 document is created. Also updated the mocks to return the same format of documents as created from IDW.

PR #512, opened 2020-11-17
feat(core): Bring conflict resolution approach into alignment with spec

stbrody:


packages/core/src/document.ts
```diff
@@ -392,17 +392,66 @@ class Document extends EventEmitter {
...
+        // If they have the same block timestamp fall through to fallback mechanism
+      }
+    }
+
+    // If we got this far, that means that we don't have sufficient information to make a good
+    // decision about which log to choose.  The most common way this can happen is that neither log
+    // is anchored, although it can also happen if both are anchored but in the same blockNumber or
+    // blockTimestamp. At this point, the decision of which log to take is arbitrary, but we want it
+    // to still be deterministic. Therefore, we take the log whose first entry has the lowest CID.
+    return state1.log[0] > state2.log[0] // TODO Is it safe to compare raw CIDs? Do I need to call toString() on them?
```
stbrody, line 441: [As best as I can tell](https://stackoverflow.com/questions/36926463/why-is-an-object-greater-less-than-or-equal-to-a-different-object) this should be comparing the results of calling `toString()` on each CID, but I don't trust my understanding of javascript well enough to say if this is something I can trust and rely on.  Can someone confirm the behavior here?

stbrody:
Sorry, I sent this out prematurely, please don't actually review yet @simonovic86 @oed.  There seems to be an issue in how updates are applied that this change exposes.  I need to track that down before this can be reviewed for real

packages/core/src/document.ts
```diff
@@ -378,17 +378,77 @@ class Document extends EventEmitter {
...
+    } else if (nonce2 < nonce1) {
+      return false
+    }
+
+    // If we got this far, that means that we don't have sufficient information to make a good
+    // decision about which log to choose.  The most common way this can happen is that neither log
+    // is anchored, although it can also happen if both are anchored but in the same blockNumber or
+    // blockTimestamp. At this point, the decision of which log to take is arbitrary, but we want it
+    // to still be deterministic. Therefore, we take the log whose first entry has the lowest CID.
+    return state1.log[0].cid > state2.log[0].cid // TODO Is it safe to compare raw CIDs? Do I need to call toString() on them?
```
stbrody, line 438: [As best as I can tell](https://stackoverflow.com/questions/36926463/why-is-an-object-greater-less-than-or-equal-to-a-different-object) this should be comparing the results of calling `toString()` on each CID, but I don't trust my understanding of javascript well enough to say if this is something I can trust and rely on.  Can someone confirm the behavior here?

packages/core/src/document.ts
```diff
@@ -378,17 +378,77 @@ class Document extends EventEmitter {
...
+          return true
+        }
+        // If they have the same block timestamp fall through to fallback mechanism
+      }
+    }
+
+    // The anchor status is the same between both logs.  If either log has a 'nonce' that means
+    // multiple updates have been squashed into a single anchor window, so prefer the log with the
+    // higher nonce (which indicates it has had more writes).
+    const nonce1 = (state1.next?.metadata.nonce ?? state1.metadata.nonce) || 0
```
stbrody, line 425: This now leaks the concept of a 'nonce' outside of the tile-doctype for the first time.  Longer term I imagine all conflict resolution behavior will move into the specific doctype classes and out of Document.

stbrody (commented):
Okay, this should be ready for review now @oed @simonovic86.  The issue was related to squashing behavior - I forgot to consider nonces as part of conflict resolution in my first version

packages/core/src/document.ts
oed: I guess it would be safer to do `cid.bytes`?

stbrody: Done

packages/core/src/__tests__/document.test.ts
```diff
@@ -11,6 +11,7 @@ import { TileDoctype, TileParams, TileDoctypeHandler } from "@ceramicnetwork/doc
 import { PinStore } from "../store/pin-store";
 import { LevelStateStore } from "../store/level-state-store";
 import { DID } from "dids"
+import { sha256 } from 'js-sha256'
```
oed, line 14: Not super important yet, but we should be using `@stablelib/sha256` everywhere.

packages/core/src/document.ts
```diff
@@ -378,17 +378,77 @@ class Document extends EventEmitter {
...
+          return true
+        }
+        // If they have the same block number fall through to fallback mechanism
+      } else {
+        // The logs are anchored in different blockchains, compare block timestamps to decide which to take
+        if (proof1.blockTimestamp < proof2.blockTimestamp) {
+          return false
+        } else if (proof2.blockTimestamp < proof1.blockTimestamp) {
+          return true
+        }
```
oed, line 417: Is this needed since we have the check that it's anchored in the right blockchain already?

packages/core/src/document.ts
```diff
@@ -378,17 +378,77 @@ class Document extends EventEmitter {
...
+      // compare anchor proofs if both states are anchored
+      const { anchorProof: proof1 } = state1
+      const { anchorProof: proof2 } = state2
+
+      if (proof1.chainId == proof2.chainId) {
+        // The logs are anchored in the same blockchain, compare block heights to decide which to take
+        if (proof1.blockNumber < proof2.blockNumber) {
+          return false
+        } else if (proof2.blockNumber < proof1.blockNumber) {
+          return true
```
oed, line 408:
```suggestion
        if (proof1.blockNumber !== proof2.blockNumber) {
          return proof2.blockNumber < proof1.blockNumber
```
Is this simpler?

oed (approved):
Looks good, have a few nitpicks

packages/core/src/document.ts
stbrody: I guess not for now, though we may need it in the future.  Probably makes sense to leave out until then though.  Removed.
We'll need to update the spec too since it talks about this case explicitly.

packages/core/src/__tests__/document.test.ts
stbrody: Ack.  Decided to leave it as-is in the interest of time since this is test code and there is other test code already working this way (which I copied here).  We can change it all in one fell swoop when needed.

packages/core/src/document.ts
stbrody: Nope, doesn't work because that would catch the case where the block numbers are equal, which is totally possible if the conflicting records were sent to different CAS instances.  Without using the fallback method of picking the lower CID, you could wind up with diverging histories in this case as which record gets chosen would depend on which record each node heard about first.

oed: The first if-statement I suggested explicitly checks that the block numbers are not equal, so this should not be the case. Or am I missing something here?

simonovic86 (approved):
lgtm!

PR #522, opened 2020-11-20
IPFS bump [HOTFIX]

simonovic86:
- bump `IPFS` and related dependencies
- fix lint errors


packages/ceramic-cli/src/ceramic-daemon.ts
```diff
@@ -34,8 +33,8 @@ export interface CreateOpts {
 }
 
 interface HttpLog {
-  request: object;
-  response?: object;
+  request: Record<string, unknown>;
```
PaulLeCam, line 36: Could use the Request/Response types from Express maybe if it's matching?

packages/ceramic-cli/src/ceramic-cli-utils.ts
```diff
@@ -3,6 +3,7 @@ import path from "path"
 import { randomBytes } from '@stablelib/random'
 import * as u8a from 'uint8arrays'
 
+// eslint-disable-next-line @typescript-eslint/no-var-requires
 const fs = require('fs').promises
```
PaulLeCam, line 7: `import { promises as fs } from 'fs'`?

PaulLeCam (approved):
Great being able to use the main IPFS package again!

packages/ceramic-cli/src/ceramic-cli-utils.ts
simonovic86: 👍

packages/ceramic-cli/src/ceramic-daemon.ts
simonovic86: nope, those are different objects

PR #524, opened 2020-11-20
feat: allow any ipfs url to be passed

valmack:
### Motivation

Currently you will get an error trying to run this if ipfs is not running or running on different host and/or port. Also it is not straightforward to add additional commands like --debug, etc.

### Changes

- Remove env var for ipfs url and use entrypoint instead of cmd


simonovic86 (approved):
lgtm! should we release a new version?

valmack:
Closing in favor of https://github.com/ceramicnetwork/js-ceramic/pull/526

PR #526, opened 2020-11-20
Dockerfiles

valmack:
### Motivation

Currently you will get an error trying to run this if ipfs is not running or running on different host and/or port. Also it is not straightforward to add additional commands like --debug, etc.

### Changes
- Remove env var for ipfs url and use entrypoint instead of cmd
- Small name change for updated package names


valmack:
Will be good to hotfix this into master so we have the update deployed to dockerhub

Dockerfile.daemon
```diff
@@ -11,12 +9,12 @@ RUN npm install
...
-COPY packages/common/package.json packages/common/package-lock.json ./packages/common/
-COPY packages/http-client/package.json packages/cli/package-lock.json ./packages/http-client/
-COPY packages/doctype-tile/package.json packages/doctype-tile/package-lock.json ./packages/doctype-tile/
-COPY packages/doctype-caip10-link/package.json packages/doctype-caip10-link/package-lock.json ./packages/doctype-caip10-link/
+COPY packages/ceramic-cli/package.json packages/ceramic-cli/package-lock.json ./packages/ceramic-cli/
+COPY packages/ceramic-core/package.json packages/ceramic-cli/package-lock.json ./packages/ceramic-core/
+COPY packages/ceramic-common/package.json packages/ceramic-common/package-lock.json ./packages/ceramic-common/
+COPY packages/ceramic-http-client/package.json packages/ceramic-cli/package-lock.json ./packages/ceramic-http-client/
+COPY packages/ceramic-doctype-tile/package.json packages/ceramic-doctype-tile/package-lock.json ./packages/ceramic-doctype-tile/
+COPY packages/ceramic-doctype-account-link/package.json packages/ceramic-doctype-account-link/package-lock.json ./packages/ceramic-doctype-account-link/
```
oed, line 17: These changes shouldn't be here. Is this PR based on develop or master?

oed (commented):
Hm, why are there two dockerfiles here actually?

Dockerfile.daemon
valmack: I modified this one to use the package names for the master branch and created an additional dockerfile to use the package names that are on develop.

oed: Right, we want to make a release to master today so these package names will be removed.

oed (commented):
lgtm 👍

PR #527, opened 2020-11-20
feat(core): Break up and rename DocOpts args

stbrody:


packages/core/src/document.ts
```diff
@@ -258,10 +258,16 @@ class Document extends EventEmitter {
...
+    const waitForSync = opts.waitForSync ?? true
+    if (anchor) {
       await this.anchor()
-      this._publishTip()
-    } else if (!opts.skipWait) {
+    }
+    if (publish) {
+      await this._publishTip()
+    }
+    if (waitForSync) {
```
stbrody, line 270: This introduces a minor behavior change by making waiting for sync the default behavior for all document writes.  Currently we only wait for sync if we skipped anchoring and publishing which seems backwards to me, and is why I changed the behavior here.  Most of the test fallout in this CR adding `{waitForSync:false}` everywhere is because of this behavior change.  Although the real issue with those tests is that they depend on the timing of when anchors are performed, but aren't controlling that themselves.  I filed https://github.com/ceramicnetwork/js-ceramic/issues/528 to fix this in the tests

oed: So if you call `change` on a doc will it wait by default? I assume it shouldn't right?

oed (commented):
Looks good overall.

packages/core/src/document.ts
stbrody: So you're thinking we should only wait by default for load operations?  That makes sense, but probably makes sense to encode that default into the change() method rather than have it here.  I'll try that now

stbrody:
Okay, new version up that's merged to latest `develop` and includes the change to defaults for write operations.  PTAL @simonovic86 @oed

packages/cli/src/__tests__/ceramic-daemon.test.ts
```diff
@@ -96,12 +100,14 @@ describe('Ceramic interop: core <> http-client', () => {
 
     it('properly creates document', async () => {
         const doc1 = await core.createDocument(DOCTYPE_TILE, { content: { test: 123 } }, {
-            applyOnly: true,
-            skipWait: true
+            anchor: false,
+            publish: false,
+            sync: false,
```
oed, line 105: This looks very clean 👍

oed (approved):

simonovic86 (approved):
lgtm!

PR #529, opened 2020-11-20
feat(core) Ensure Context is available everywhere

stbrody:
This was already true almost everywhere, so this turned out to be a pretty small change thankfully.  Only thing that didn't have access to Context already was a couple of static methods on Document.  The rest is just code cleanup


simonovic86 (approved):
nice!

PR #532, opened 2020-11-23
ref: update daemon api and http-client

oed:
Now there are four main names of endpoints:
* records - get all records of a document
* document - interact with a document
* pins - interact with the pinset
* node - get information about the node

### DIFF
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -143,22 +143,21 @@ class CeramicDaemon {
   }
 
   registerAPIPaths (app: core.Express, gateway: boolean): void {
-    app.get(toApiPath('/records/ceramic/:docid'), this.records.bind(this))
-    app.get(toApiPath('/show/ceramic/:docid'), this.show.bind(this))
-    app.get(toApiPath('/state/ceramic/:docid'), this.state.bind(this))
-    app.get(toApiPath('/pin/ls/ceramic/:docid'), this.listPinned.bind(this))
-    app.get(toApiPath('/pin/ls'), this.listPinned.bind(this))
-    app.get(toApiPath('/supported_chains'), this.getSupportedChains.bind(this))
-    app.post(toApiPath('/create'), this.createDocFromGenesis.bind(this))
+    app.get(toApiPath('/records/:docid'), this.records.bind(this))
+    app.put(toApiPath('/document'), this.createDocFromGenesis.bind(this))
+    app.get(toApiPath('/document/:docid'), this.state.bind(this))
+    app.get(toApiPath('/pins/:docid'), this.listPinned.bind(this))
+    app.get(toApiPath('/pins'), this.listPinned.bind(this))
+    app.get(toApiPath('/node/chains'), this.getSupportedChains.bind(this))
 
     if (!gateway) {
-      app.post(toApiPath('/apply'), this.applyRecord.bind(this))
-      app.get(toApiPath('/pin/add/ceramic/:docid'), this.pinDocument.bind(this))
-      app.get(toApiPath('/pin/rm/ceramic/:docid'), this.unpinDocument.bind(this))
+      app.post(toApiPath('/document'), this.applyRecord.bind(this))
+      app.post(toApiPath('/pins/:docid'), this.pinDocument.bind(this))
+      app.delete(toApiPath('/pins/:docid'), this.unpinDocument.bind(this))
     } else {
-      app.post(toApiPath('/apply'),  this._notSupported.bind(this))
-      app.get(toApiPath('/pin/add/ceramic/:docid'),  this._notSupported.bind(this))
-      app.get(toApiPath('/pin/rm/ceramic/:docid'),  this._notSupported.bind(this))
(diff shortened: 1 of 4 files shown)


packages/cli/src/ceramic-daemon.ts
```diff
@@ -143,22 +143,21 @@ class CeramicDaemon {
...
   registerAPIPaths (app: core.Express, gateway: boolean): void {
-    app.get(toApiPath('/records/ceramic/:docid'), this.records.bind(this))
-    app.get(toApiPath('/show/ceramic/:docid'), this.show.bind(this))
-    app.get(toApiPath('/state/ceramic/:docid'), this.state.bind(this))
-    app.get(toApiPath('/pin/ls/ceramic/:docid'), this.listPinned.bind(this))
-    app.get(toApiPath('/pin/ls'), this.listPinned.bind(this))
-    app.get(toApiPath('/supported_chains'), this.getSupportedChains.bind(this))
-    app.post(toApiPath('/create'), this.createDocFromGenesis.bind(this))
+    app.get(toApiPath('/records/:docid'), this.records.bind(this))
+    app.put(toApiPath('/document'), this.createDocFromGenesis.bind(this))
```
stbrody, line 147: Dropping the 'ceramic' from all these API urls makes sense to me, but I'm not sure why we'd want to consolidate all the APIs in a way that removes semantic meaning from the URLs?  Why is `/document` better than `/document/create`?  The latter makes it more immediately obvious what is happening from just the URL being called.  Also having the same URL endpoints do different things based on the HTTP method called seems like a potential source of confusion with no real benefit?  Why not separate `/document/create` and `/document/apply_record`?

stbrody, line 147: This is minor, but my understanding is that HTTP PUT methods are supposed to be idempotent, which createDocFromGenesis won't be for Tile doctypes unless `unique:false` is explicitly provided.  Not sure how much the choice of HTTP method really matters anyway, but figured I'd call that out.

stbrody (commented):
Code change looks correct, and the overall aim of cleaning up and simplifying these API endpoints sounds good to me, I just have a couple of comments about what the ideal endpoint URLs actually are.  Take my comments with a grain of salt though as I don't have any experience creating REST APIs so maybe there are conventions/best practices I'm not aware of.

packages/cli/src/ceramic-daemon.ts
```diff
@@ -143,22 +143,21 @@ class CeramicDaemon {
...
 
   registerAPIPaths (app: core.Express, gateway: boolean): void {
-    app.get(toApiPath('/records/ceramic/:docid'), this.records.bind(this))
-    app.get(toApiPath('/show/ceramic/:docid'), this.show.bind(this))
-    app.get(toApiPath('/state/ceramic/:docid'), this.state.bind(this))
-    app.get(toApiPath('/pin/ls/ceramic/:docid'), this.listPinned.bind(this))
-    app.get(toApiPath('/pin/ls'), this.listPinned.bind(this))
-    app.get(toApiPath('/supported_chains'), this.getSupportedChains.bind(this))
-    app.post(toApiPath('/create'), this.createDocFromGenesis.bind(this))
+    app.get(toApiPath('/records/:docid'), this.records.bind(this))
```
simonovic86, line 146: can we rename the `records` method to `getRecords` or similar?

packages/cli/src/ceramic-daemon.ts
```diff
@@ -143,22 +143,21 @@ class CeramicDaemon {
...
+    app.put(toApiPath('/document'), this.createDocFromGenesis.bind(this))
+    app.get(toApiPath('/document/:docid'), this.state.bind(this))
+    app.get(toApiPath('/pins/:docid'), this.listPinned.bind(this))
+    app.get(toApiPath('/pins'), this.listPinned.bind(this))
+    app.get(toApiPath('/node/chains'), this.getSupportedChains.bind(this))
```
simonovic86, line 151: Should we refactor `/node/chains` into `/node` or `/node/config`? In that way we don't need to change the API if we want to add more info instead of supported chains.

simonovic86 (commented):
LGTM. I added a couple of comments.

packages/http-client/src/document.ts
```diff
@@ -5,7 +5,7 @@ import {
 import { fetchJson, typeDocID } from './utils'
 import DocID from '@ceramicnetwork/docid'
 
-const docIdUrl = (docId: DocID ): string => `/ceramic/${docId.toString()}`
+const docIdUrl = (docId: DocID ): string => `/${docId.toString()}`
```
simonovic86, line 8: `docIdUrl` is not being used anywhere as far as I can see

packages/cli/src/ceramic-daemon.ts
oed: @simonovic86 recommended that this was the way to go.

oed: Addressed this by changing applyRecord to just use `records`.

simonovic86 (approved):
👍

simonovic86 (approved):
LGTM!

Issue #499, opened 2020-11-16
Make http-client poll interval configurable


oed:
As a global configuration to begin with.

Or include disable

PR #533, opened 2020-11-23
Ceramic client config

simonovic86:
- add Ceramic client configuration
- replace `setInterval` with `setTimeout`
- disable Ceramic client document polling by default - issue [499](https://github.com/ceramicnetwork/js-ceramic/issues/499)
- fix `applyRecord` functionality - issue [534](https://github.com/ceramicnetwork/js-ceramic/issues/534)

### DIFF
--- a/packages/cli/src/__tests__/ceramic-daemon.test.ts
+++ b/packages/cli/src/__tests__/ceramic-daemon.test.ts
@@ -21,6 +21,7 @@ const waitChange = (doc: EventEmitter, count = 1): Promise<void> => {
         doc.on('change', () => {
             if (++c === count) {
                 resolve()
+                doc.removeAllListeners('change')
             }
         })
     })
--- a/packages/cli/src/__tests__/ceramic-daemon.test.ts
+++ b/packages/cli/src/__tests__/ceramic-daemon.test.ts
@@ -80,7 +80,7 @@ describe('Ceramic interop: core <> http-client', () => {
         core._doctypeHandlers['tile'] = doctypeHandler
 
         daemon = new CeramicDaemon(core, { port })
-        client = new CeramicClient(apiUrl, { docSyncEnabled: true, docSyncInterval: 2000 })
+        client = new CeramicClient(apiUrl, { docSyncEnabled: true })
 
         const provider = new Ed25519Provider(seed)
         await core.setDIDProvider(provider)
--- a/packages/cli/src/__tests__/ceramic-daemon.test.ts
+++ b/packages/cli/src/__tests__/ceramic-daemon.test.ts
@@ -15,7 +15,16 @@ import basicsImport from 'multiformats/cjs/src/basics-import.js'
 import legacy from 'multiformats/cjs/src/legacy.js'
 
 const seed = u8a.fromString('6e34b2e1a9624113d81ece8a8a22e6e97f0e145c25c1d4d2d0e62753b4060c83', 'base16')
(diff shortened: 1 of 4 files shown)


packages/http-client/src/ceramic-http-client.ts
```diff
@@ -7,21 +7,43 @@ import { TileDoctypeHandler } from "@ceramicnetwork/doctype-tile"
...
-const CERAMIC_HOST = 'http://localhost:7007'
 const API_PATH = '/api/v0'
+const CERAMIC_HOST = 'http://localhost:7007'
+
+/**
+ * Default Ceramic client configuration
+ */
+export const DEFAULT_CLIENT_CONFIG: CeramicClientConfig = {
+  docSyncEnabled: false,
+  docSyncInterval: 1000,
```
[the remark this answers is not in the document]
simonovic86: 👍

packages/cli/src/__tests__/ceramic-daemon.test.ts
```diff
@@ -15,7 +15,16 @@ import basicsImport from 'multiformats/cjs/src/basics-import.js'
 import legacy from 'multiformats/cjs/src/legacy.js'
 
 const seed = u8a.fromString('6e34b2e1a9624113d81ece8a8a22e6e97f0e145c25c1d4d2d0e62753b4060c83', 'base16')
-const anchorUpdate = (doc: EventEmitter): Promise<void> => new Promise(resolve => doc.on('change', resolve))
+const waitChange = (doc: EventEmitter, count = 1): Promise<void> => {
+    return new Promise(resolve => {
+        let c = 0
+        doc.on('change', () => {
+            if (++c === count) {
+                resolve()
```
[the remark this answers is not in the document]
simonovic86: 👍

oed (approved):
lgtm

packages/http-client/src/ceramic-http-client.ts
```diff
@@ -7,21 +7,43 @@ import { TileDoctypeHandler } from "@ceramicnetwork/doctype-tile"
...
-const CERAMIC_HOST = 'http://localhost:7007'
 const API_PATH = '/api/v0'
+const CERAMIC_HOST = 'http://localhost:7007'
+
+/**
+ * Default Ceramic client configuration
+ */
+export const DEFAULT_CLIENT_CONFIG: CeramicClientConfig = {
+  docSyncEnabled: false,
+  docSyncInterval: 5000,
```
stbrody, line 18: What about `docSyncIntervalMs`?  I'm not a fan of numeric arguments that are ambiguous about what unit they take.  Time arguments are especially prone to bugs I've found.  In a language with a stronger type system I like to use the type system to handle this (i.e. take a Milliseconds type that does not support any implicit conversions from numeric constants), but in lieu of that, encoding the unit into the name is the next best option I believe.

I know this PR was already merged, but since this argument is user-facing I figured it was still worth raising in case we want to put together a small change before releasing and getting locked in.  @oed @simonovic86 @PaulLeCam

simonovic86: @stbrody hm, I guess it's enough that we can document the unit for the variable. I've opened the story for documenting the http client [537](https://github.com/ceramicnetwork/js-ceramic/issues/537).

stbrody: @simonovic86 why not rename the field to include the unit?  Then it becomes self-documenting and makes it immediately obvious at all call sites what the value represents

PR #539, opened 2020-11-24
feat(core): Use new format for UPDATE pubsub messages

stbrody:


packages/core/src/dispatcher.ts
```diff
@@ -32,6 +34,24 @@ interface LogMessage {
   message?: Record<string, unknown>;
 }
 
+/**
+ * Format for UPDATE messages on the pub/sub topic. Uses the 'runtypes' library to provide type
+ * checking at run time when parsing messages off the pub/sub topic.
+ */
+const UpdateMessage = runtypes.Record({
```
stbrody, line 41: It turns out `runtypes` doesn't have a way to enforce that an object has keys of a certain type but without knowing their field names, which I need to define the format for RESPONSE messages.  So I'm probably going to remove the use of runtypes in a later commit.  I might try using jsonSchema to do the validation instead, but that will come in a separate review.  So for now don't worry too much about the way that the message validation is implemented, although the tests and basic ideas will continue to exist.

stbrody:
Closing in favor of https://github.com/ceramicnetwork/js-ceramic/pull/542

PR #540, opened 2020-11-24
feat(core): Use new format for QUERY pubsub messages

stbrody:


packages/core/src/dispatcher.ts
```diff
@@ -85,11 +107,20 @@ export default class Dispatcher extends EventEmitter {
...
     this._ipfs.pubsub.publish(this.topic, JSON.stringify(payload))
     this._log({ peer: this._peerId, event: 'published', topic: this.topic, message: payload })
   }
 
+  async _buildQueryMessage(document: Document): Promise<Record<string, any>> {
+    const message = { typ: MsgType.QUERY, doc: document.id.baseID.toString() }
+
+    // Add 'id' to message.  'id' is the multihash of the rest of the query object canonicalized
+    // using dag-cbor
+    const id: Uint8Array = dagCBOR.util.serialize(message)
```
stbrody, line 120: This is based on this line from CIP-71: `unique identifier of the query, multihash of the query object without the id property canonicalized using dag-cbor`
This was my guess as to what that meant, please let me know if this wasn't what you had in mind @oed

packages/core/src/dispatcher.ts
```diff
@@ -213,12 +244,18 @@ export default class Dispatcher extends EventEmitter {
...
+    // Runtime check that the message adheres to the expected format for QUERY messages
+    QueryMessage.check(message)
+
+    const { doc } = message
+    if (!this._documents[doc]) {
       return
     }
 
-    this._documents[id].emit('tipreq')
+    // TODO: Should we validate that the 'id' field is the correct hash of the rest of the message?
```
stbrody, line 255: I'm wondering if it's worth it to validate that the 'id' property is what we expect it to be.  I'm leaning towards no, at least for now, as we don't actually *rely* on the id format for protocol correctness, it's just an optimization to allow deduping query messages.  Curious your thoughts on this though @oed @simonovic86

stbrody:
This review builds on https://github.com/ceramicnetwork/js-ceramic/pull/539

stbrody:
Closing in favor of https://github.com/ceramicnetwork/js-ceramic/pull/542

PR #541, opened 2020-11-24
feat(core): Use RESPONSE message to reply to QUERY messages on pubsub topic

stbrody:

PR #542, opened 2020-11-25
feat(core): Update pubsub messages to those defined in CIP-71

stbrody:


stbrody:
This CR is a unified view of https://github.com/ceramicnetwork/js-ceramic/pull/539, https://github.com/ceramicnetwork/js-ceramic/pull/540, and https://github.com/ceramicnetwork/js-ceramic/pull/541, in case it's easier to view in unified form.  If you review this version there's no need to review those @oed @simonovic86.  This version also removes the message input validation since it seemed like the approach I was using wouldn't work for `RESPONSE` messages.  I'll add a follow-up story to add message validation.

packages/core/src/dispatcher.ts
```diff
@@ -64,12 +70,26 @@ export default class Dispatcher extends EventEmitter {
...
     this._ipfs.pubsub.publish(this.topic, JSON.stringify(payload))
     this._log({ peer: this._peerId, event: 'published', topic: this.topic, message: payload })
   }
 
+  async _buildQueryMessage(document: Document): Promise<Record<string, any>> {
+    const message = { typ: MsgType.QUERY, doc: document.id.baseID.toString() }
+
+    // Add 'id' to message.  'id' is the multihash of the rest of the query object canonicalized
+    // using dag-cbor
+    const id: Uint8Array = dagCBOR.util.serialize(message)
```
stbrody, line 89: This is based on this line from CIP-71: `unique identifier of the query, multihash of the query object without the id property canonicalized using dag-cbor`
This was my guess as to what that meant, please let me know if this wasn't what you had in mind @oed

packages/core/src/dispatcher.ts
```diff
@@ -134,37 +153,107 @@ export default class Dispatcher extends EventEmitter {
...
+   */
+  async _handleQueryMessage(message: any): Promise<void> {
+    // TODO Add validation the message adheres to the proper format.
+
+    const { doc: docId, id } = message
+    if (!this._documents[docId]) {
+      return
+    }
+
+    // TODO: Should we validate that the 'id' field is the correct hash of the rest of the message?
```
stbrody, line 224: I'm wondering if it's worth it to validate that the 'id' property is what we expect it to be, when receiving a QUERY message over the pubsub topic.  I'm leaning towards no, at least for now, as we don't actually *rely* on the id format for protocol correctness, it's just an optimization to allow deduping query messages.  Curious your thoughts on this though @oed @simonovic86

stbrody (commented):
Copying a couple of comments over from the earlier separated-out PRs

packages/core/src/dispatcher.ts
oed: This is exactly right.

oed, line 89: Please compute a multihash of the serialized message. Now the `id` is the entire message serialized as dag-cbor.

packages/core/src/dispatcher.ts
oed: yeah, I agree we dont' need to.

oed (changes requested):
One change needed. Looks good otherwise.

packages/core/src/dispatcher.ts
stbrody: Okay new version is up.  Is this what you had in mind @oed?

packages/core/src/dispatcher.ts
```diff
@@ -63,12 +71,42 @@ export default class Dispatcher extends EventEmitter {
...
+  async _hashMessage(message: any) : Promise<Uint8Array> {
+    // DAG-CBOR encoding
+    let id: Uint8Array = dagCBOR.util.serialize(message)
+
+    // SHA-256 hash
+    id = sha256.hash(id)
+
+    // Multihash encoding
+    const buf = Buffer.from(id)
+    return multihashes.encode(buf, 'sha2-256')
```
oed, line 107: multihashes doesn't support Uint8Arrays?

oed (approved):
Looks good 👍

packages/core/src/dispatcher.ts
stbrody: Doesn't look like it, I get a compilation error if I try to pass in the Uint8Array directly ☹️

PR #544, opened 2020-11-25
chore: bump to dids@1.0.0

oed:
Bump DIDs to the latest version. Also bumps `key-did-provider-ed25519`.

In addition we also set the resolvers to the DID instance, in both http-client and core.


simonovic86 (approved):
lgtm!

PR #549, opened 2020-11-26
Fix resolver construction

simonovic86:


packages/http-client/src/ceramic-http-client.ts
```diff
@@ -56,7 +56,7 @@ export default class CeramicClient implements CeramicApi {
     const keyDidResolver = KeyDidResolver.getResolver()
     const threeIdResolver = ThreeIdResolver.getResolver(this)
     this.context.resolver = new Resolver({
-      ...config.didResolver, ...threeIdResolver, ...keyDidResolver,
+      ...this._config.didResolver, ...threeIdResolver, ...keyDidResolver,
```
oed, line 59: The config variable should be available here.

oed: It's passed to the constructor

oed: Oh, is it if it's unset?

oed (approved):
I see now. Makes sense 👍

packages/http-client/src/ceramic-http-client.ts
simonovic86: yep 🙂

PR #551, opened 2020-11-26
Multiple anchoring services

simonovic86:
- support multiple anchoring services
- `note`: this story is just one step to [202](https://github.com/ceramicnetwork/js-ceramic/issues/202)

### DIFF
--- a/packages/common/src/context.ts
+++ b/packages/common/src/context.ts
@@ -13,7 +13,7 @@ export interface Context {
     ipfs?: IpfsApi; // an ipfs instance
     resolver?: Resolver; // a DID resolver instance
     provider?: DIDProvider; // a DID provider (3ID provider initially)
-    anchorService?: AnchorService;
+    anchorServices?: Record<string, Array<AnchorService>>; // mapping from chainId(s) to the AnchorService instance(s)
 
     api?: CeramicApi; // the self reference to the Ceramic API
 }
--- a/packages/common/src/context.ts
+++ b/packages/common/src/context.ts
@@ -14,6 +14,7 @@ export interface Context {
     resolver?: Resolver; // a DID resolver instance
     provider?: DIDProvider; // a DID provider (3ID provider initially)
     anchorServices?: Record<string, Array<AnchorService>>; // mapping from chainId(s) to the AnchorService instance(s)
+    preferredChainId?: string; // preferred chain ID for the anchoring service
 
     api?: CeramicApi; // the self reference to the Ceramic API
 }
--- a/packages/common/src/context.ts
+++ b/packages/common/src/context.ts
@@ -14,7 +14,15 @@ export interface Context {
     resolver?: Resolver; // a DID resolver instance
     provider?: DIDProvider; // a DID provider (3ID provider initially)
     anchorServices?: Record<string, Array<AnchorService>>; // mapping from chainId(s) to the AnchorService instance(s)
-    preferredChainId?: string; // preferred chain ID for the anchoring service
+    chainInfo?: ChainInfo; // chain IDs information
 
     api?: CeramicApi; // the self reference to the Ceramic API
+}
+
+/**
+ * Chain IDs information
+ */
+export interface ChainInfo {
+    preferredChain?: string
+    supportedChains?: Array<string>
 }
--- a/packages/core/src/__tests__/document.test.ts
+++ b/packages/core/src/__tests__/document.test.ts
@@ -175,7 +175,9 @@ describe('Document', () => {
       const resolver = new Resolver({ ...threeIdResolver })
       context = {
         did: user,
-        anchorService,
+        anchorServices: {
+          'inmemory:12345': [anchorService],
+        },
         ipfs: dispatcher._ipfs,
         resolver,
         provider: null,
@@ -347,7 +349,7 @@ describe('Document', () => {
       // sometime the tests are very fast
       await new Promise(resolve => setTimeout(resolve, 1))
       // TODO - better mock for anchors
-      
+
       const conflictingNewContent = { asdf: 2342 }
       updateRec = await TileDoctype._makeRecord(doc2.doctype, user, conflictingNewContent, doc2.controllers)
       await doc2.applyRecord(updateRec)
@@ -618,9 +620,7 @@ describe('Document', () => {
     beforeEach(() => {
       dispatcher = Dispatcher(true)
       anchorService = new InMemoryAnchorService({})
-      anchorService.ceramic = {
-        dispatcher,
-      }
+
       user = new DID()
       user.createJWS = jest.fn(async () => {
         // fake jws
@@ -646,7 +646,9 @@ describe('Document', () => {
       const resolver = new Resolver({ ...threeIdResolver })
       context = {
         did: user,
-        anchorService,
+        anchorServices: {
+          'inmemory:12345': [anchorService],
+        },
         ipfs: dispatcher._ipfs,
         resolver,
         provider: null,
--- a/packages/core/src/__tests__/document.test.ts
+++ b/packages/core/src/__tests__/document.test.ts
@@ -176,7 +176,7 @@ describe('Document', () => {
       context = {
         did: user,
         anchorServices: {
-          'eip155:3': [anchorService],
+          'inmemory:12345': [anchorService],
         },
         ipfs: dispatcher._ipfs,
         resolver,
@@ -206,19 +206,19 @@ describe('Document', () => {
       expect(doc.state.anchorStatus).toEqual(AnchorStatus.PENDING)
       await anchorUpdate(doc)
       expect(doc.state.anchorStatus).not.toEqual(AnchorStatus.NOT_REQUESTED)
-      expect(doc.state.metadata.chainId).toEqual("eip155:3")
+      expect(doc.state.metadata.chainId).toEqual("inmemory:12345")
     })
 
     it("Can specify chainId if it matches the configured anchor service's chainid", async () => {
-      const metadata = { controllers, tags: ['3id'], chainId: 'eip155:3' }
+      const metadata = { controllers, tags: ['3id'], chainId: 'inmemory:12345' }
       const doc = await create({ content: initialContent, metadata }, ceramic, context)
-      expect(doc.state.metadata.chainId).toEqual("eip155:3")
+      expect(doc.state.metadata.chainId).toEqual("inmemory:12345")
     })
 
     it("Cannot specify chainId if it is different than the configured anchor service's chainid", async () => {
       const metadata = { controllers, tags: ['3id'], chainId: 'newchain' }
       const createPromise = create({ content: initialContent, metadata }, ceramic, context)
-      await expect(createPromise).rejects.toThrow("Requested chainId 'newchain' is not supported. Supported chains are: 'eip155:3'")
+      await expect(createPromise).rejects.toThrow("Requested chainId 'newchain' is not supported. Supported chains are: 'inmemory:12345'")
     })
 
     it('is loaded correctly', async () => {
@@ -647,7 +647,7 @@ describe('Document', () => {
       context = {
         did: user,
         anchorServices: {
-          'eip155:3': [anchorService],
+          'inmemory:12345': [anchorService],
         },
         ipfs: dispatcher._ipfs,
         resolver,
--- a/packages/core/src/__tests__/document.test.ts
+++ b/packages/core/src/__tests__/document.test.ts
@@ -181,7 +181,10 @@ describe('Document', () => {
         ipfs: dispatcher._ipfs,
         resolver,
         provider: null,
-        preferredChainId: 'inmemory:12345'
+        chainInfo: {
+          preferredChain: 'inmemory:12345',
+          supportedChains: ['inmemory:12345']
+        }
       }
 
       anchorService.ceramic = {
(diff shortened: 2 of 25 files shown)


PaulLeCam (approved):

packages/core/src/utils.ts
```diff
@@ -58,5 +61,16 @@ export default class Utils {
         }
     }
 
-
+    /**
+     * Gets the default anchoring service.
+     *
+     * The default service anchors on Ropsten. If it's not specified, return the in-memory one.
+     *
+     * Note: only the first anchor service is returned from the list. If there are multiple services for the same
```
stbrody, line 69: In that case, why keep a list at all?  Let's only store a single anchor service per chainId for now and simplify the code.  If/when we support multiple anchor services for the same chainId, we can change it then.

packages/common/src/context.ts
```diff
@@ -13,7 +13,7 @@ export interface Context {
     ipfs?: IpfsApi; // an ipfs instance
     resolver?: Resolver; // a DID resolver instance
     provider?: DIDProvider; // a DID provider (3ID provider initially)
-    anchorService?: AnchorService;
+    anchorServices?: Record<string, Array<AnchorService>>; // mapping from chainId(s) to the AnchorService instance(s)
```
stbrody, line 16: Since we only support one anchor service per chainId currently, I think we'd be better off not trying store an array of AnchorServices here.  We're just complicating our code to handle a case that we might not implement for some time, if ever.

packages/core/src/__tests__/document.test.ts
```diff
@@ -18,7 +18,7 @@ import ThreeIdResolver from '@ceramicnetwork/3id-did-resolver'
 
 jest.mock('../store/level-state-store')
 
-import InMemoryAnchorService from "../anchor/memory/in-memory-anchor-service"
+import InMemoryAnchorService, { IN_MEMORY_ANCHOR_SERVICE_CHAIN_ID } from "../anchor/memory/in-memory-anchor-service"
```
stbrody, line 21: Including `IN_MEMORY_ANCHOR_SERVICE_CHAIN_ID` but not using it in this file.

packages/core/src/ceramic.ts
```diff
@@ -361,23 +379,19 @@ class Ceramic implements CeramicApi {
...
-      return this._supportedChains
-    }
+    const chainIds = Object.keys(this.context.anchorServices)
 
-    // Fetch the chainId from the anchor service and cache the result
-    if (!this.context.anchorService) {
-      throw new Error("No anchor service configured")
+    const index = chainIds.indexOf(DEFAULT_ANCHOR_SERVICE_CHAIN_ID);
+    if (index > -1) {
+      // set default chainId to be the first one
```
stbrody, line 388: Instead of doing this every time getSupportedChains() is called, maybe we could do it once in the constructor and store the properly ordered list of supported chainIds?

stbrody (commented):
A few suggestions, overall looks good!

packages/core/src/utils.ts
simonovic86: Right now, anchor service can return a list of chain IDs which it supports. That's why it's a list.

packages/common/src/context.ts
simonovic86: Same as above.

packages/core/src/__tests__/document.test.ts
simonovic86: 👍

packages/core/src/ceramic.ts
simonovic86: yep, we can do that 👍

packages/core/src/utils.ts
stbrody: One anchor service can support multiple chainIds, but one js-ceramic instance can't support multiple anchor services for the *same* chainId

simonovic86: We haven't defined that still. There will be more anchoring services on the network which can be passed to the `js-ceramic` instance. Some of them will support same chainIds but different anchoring intervals, reputation, etc. This way we can support more anchoring services and choose which one to use based on the other properties.

oed: There might be special case anchor services for DAOs and such, but I don't see a huge reason to make our code more complex now to account for a maybe. Don't feel super strongly though.

packages/doctype-tile/src/tile-doctype.ts
```diff
@@ -76,11 +76,11 @@ export class TileDoctype extends Doctype {
...
 
-        const supported_chains = await context.api.getSupportedChains()
-        if ('chainId' in metadata && !supported_chains.includes(metadata.chainId)) {
-            throw new Error("Requested chainId '" + metadata.chainId + "' is not supported. Supported chains are: '" + supported_chains.join("', '") + "'")
+        const supportedChains = await context.api.getSupportedChains()
+        if ('chainId' in metadata && !supportedChains.includes(metadata.chainId)) {
+            throw new Error("Requested chainId '" + metadata.chainId + "' is not supported. Supported chains are: '" + supportedChains.join("', '") + "'")
         }
-        metadata.chainId = metadata.chainId ?? supported_chains[0]
+        metadata.chainId = metadata.chainId ?? supportedChains[0]
```
oed, line 83: Can we have a hardcoded chainId default (`eip155:3` for now) that you can change with a config? Right now it's unclear if this will be inmemory or ethereum if an ethereum service is configured.

oed (commented):
Only one thing I think we should change.

packages/core/src/utils.ts
stbrody: Yep, that was my thinking as well.  For public-facing APIs it makes sense to try to be as future proof as possible, but for internal-only code I've found it's generally not worth it to try to code for features that don't exist yet as more often than not by the time you get around to actually implementing the feature you wind up having to redo everything anyway.

simonovic86: The change is in the `context` which is public facing. Also, I think this change is really not complicating our code, to be honest. 🙂

packages/doctype-tile/src/tile-doctype.ts
simonovic86: yep 👍

packages/core/src/anchor/memory/in-memory-anchor-service.ts
```diff
@@ -9,7 +9,10 @@ import type Dispatcher from '../../dispatcher'
 import Ceramic, { CeramicConfig } from "../../ceramic"
 
 const DID_MATCHER = '^(did:([a-zA-Z0-9_]+):([a-zA-Z0-9_.-]+(:[a-zA-Z0-9_.-]+)*)((;[a-zA-Z0-9_.:%-]+=[a-zA-Z0-9_.:%-]*)*)(/[^#?]*)?)([?][^#]*)?(#.*)?';
-const CHAIN_ID = 'inmemory:12345'
+
+// In-memory CAS is used in testing only and will not be included if there are other anchoring services. That's why
+// its chain ID is equal to Ropsten chain ID.
+export const IN_MEMORY_ANCHOR_SERVICE_CHAIN_ID = 'eip155:3'
```
oed, line 15: This doesn't make sense. The in memory anchor service should have it's own unique chainID otherwise it may cause confusion or even bugs / other issues.

packages/doctype-tile/src/tile-doctype.ts
oed: This still grabs the chainId from an array which we don't neccesarily control the order over. This can cause unknown errors. Also we likely need to do something similar for the `caip10-link` doctype.

oed (changes requested):

packages/core/src/anchor/memory/in-memory-anchor-service.ts
simonovic86: This "anchoring" service is used just for testing. It was never meant to be used in production.

packages/doctype-tile/src/tile-doctype.ts
simonovic86: We are ordering the list on the `Ceramic` side. And yes, this should be possibly done in `caip10-link` doctype.

What was your suggestion about hardcoding the `eip155:3` in these places?

packages/core/src/ceramic.ts
```diff
@@ -170,20 +158,26 @@ class Ceramic implements CeramicApi {
...
       for (const chainId of supportedChains) {
         anchorServices[chainId] = anchorServices[chainId] ? [...anchorServices[chainId], ethereumService] : [ethereumService]
       }
+      preferredChain = ROPSTEN_CHAIN_ID
     }
 
     const context: Context = {
       ipfs,
       anchorServices,
+      chainInfo: {
```
stbrody, line 177: Instead of putting this into the Context, I think it'd be cleaner to just put it into a private member of this Ceramic instance

packages/doctype-caip10-link/src/caip10-link-doctype.ts
```diff
@@ -65,11 +65,14 @@ export class Caip10LinkDoctype extends Doctype {
             throw new Error('Exactly one controller must be specified')
         }
 
-        const supported_anchor_chains = await context.api.getSupportedChains()
-        if ('chainId' in metadata && !supported_anchor_chains.includes(metadata.chainId)) {
-            throw new Error("Requested chainId '" + metadata.chainId + "' is not supported. Supported chains are: " + supported_anchor_chains.toString())
+        if (!('chainId' in metadata) || metadata.chainId == null) {
```
stbrody, line 68: This could just be `if (!metadata.chainId) {`

packages/core/src/utils.ts
stbrody: I actually don't think it is?  AFAIK, the main public-facing APIs that correspond primarily to CRUD operations on Documents (plus Ceramic.create() to get started) are:
* `Ceramic.create()`
* `Ceramic.createDocument()`
* `Ceramic.createDocumentFromGenesis()`
* `Ceramic.loadDocument()`
* `Document.change()`

None of those take Context.  I believe Context is an internal-only object for hanging process-wide globals off of, whose contents are derived from the `CeramicConfig` passed to `Ceramic.create()`.  I think we should strive to keep Context out of any APIs that we expect end user developers to use.  If it were to every wind up in a function that users use, it would hopefully only be as a black box that is passed through into lower-level functions.

Anyway back to the original convo: IMO any code written for a feature that doesn't exist yet is complication by definition.  But in this case it's certainly not a huge difference either way so if you feel strongly that we should store an array for each chainId then that's fine

packages/core/src/ceramic.ts
```diff
@@ -175,9 +196,17 @@ class Ceramic implements CeramicApi {
...
     }
 
+    await ceramic.init()
     return ceramic
   }
 
+  /**
+   * Initialize the Ceramic API
+   */
+  async init(): Promise<void> {
```
stbrody, line 206: Real question: what's the benefit of having a separate init() method from the constructor in typescript?  I almost added this myself during an earlier commit I was working on as it is a very common pattern in C++, but in C++ it's bad practice to do i/o within constructors as throwing exceptions from constructors has weird semantics.  So in the C++ world, having a separate initializer method that can do i/o after the constructor has finished is very useful.  But in typescript we seem to do tons of i/o within our constructors all over the place, so it's less clear to me what the advantage of a separate initializer method is?

CC @PaulLeCam in case he has any insight here

packages/core/src/utils.ts
simonovic86: I think you're not looking at the right APIs. One such is a static method [makeGenesis](https://github.com/ceramicnetwork/js-ceramic/blob/f23ff938e92e2f835f613e8e3162fe7795194ebf/packages/common/src/doctype.ts#L210). On the last iteration that we had @oed we decided that we're going to put `preferredChain` in the `context`, right?

Also, every `Doctype` instance has a `context` instance by default.

packages/core/src/ceramic.ts
simonovic86: If you do that how would you call `makeGenesis` in the `TileDoctype` for example (`client` side)? You need to find out about the preferred chain, right? In that case you should request info by hitting the API for `getChainsInfo` and find out, is that right?

packages/core/src/ceramic.ts
simonovic86: You cannot `await` in the `constructor`. One of the two common patterns here are `init` and `builder`. I've used `init` in many places but I believe that there are some workarounds as well. See link [how-to-write-an-async-class-constructor-in-typescript-javascript](https://medium.com/better-programming/how-to-write-an-async-class-constructor-in-typescript-javascript-7d7e8325c35e).

packages/core/src/ceramic.ts
stbrody: Yep, exactly.  Anywhere you have access to a Context you have access to a CeramicAPI (and vice versa), so it doesn't actually change where this information is available, it's just changing the abstraction layer/separation of concerns

packages/core/src/utils.ts
stbrody: I hadn't imagined any end-user developer would ever call `TileDoctype.makeGenesis` directly.  I'd imagined they would call `CeramicApi.createDocumentFromGenesis` instead.

I do think this conversation illustrates that we should probably spend some time clarifying which of our APIs are designed to be used by developers building on Ceramic and which are only for developers *of* Ceramic.  I filed https://github.com/ceramicnetwork/js-ceramic/issues/554 for this

packages/core/src/ceramic.ts
stbrody: Ah there it is, that makes sense.  That article you linked is paywalled, but the explanation that init() methods are important for making async function calls makes sense.  Thanks!

simonovic86: np! btw, you can bypass the paywall using the **incognito** mode 🙂

simonovic86:
closing this PR since it's not valid anymore. The story [202](https://github.com/ceramicnetwork/js-ceramic/issues/202) is still valid.

PR #552, opened 2020-11-27
Chore/docs deploy

oed:
This PR removes the docs folder from this repo and instead uploads its content to s3.

TODO:
- [x] Point `docs.ceramic.network` to http://js-ceramic-docs.s3-website-us-west-1.amazonaws.com/ (@valmack can you help me with this?)
- [x] Update readme to point to `docs.ceramic.network`


valmack:
@oed do you mind if we change bucket name to docs.ceramic.network? Trying to keep it consistent with other s3 static sites

valmack:
Will also add an s3 bucket for request logging but that does not need to be in the pipeline

valmack (approved):
Nice. Added the `docs.ceramic.network` bucket to `us-east-1` so we can update those now and should be good to go.

PR #564, opened 2020-11-30
fix(doctype-tile): properly handle unsigned genesis records

oed:
* Allow both `null` and `undefined` in genesis record
* Properly apply the first signed record if genesis record has no content


simonovic86 (approved):
lgtm!

PR #574, opened 2020-11-30
feat(core): Remove chainId from document metadata and stop enforcing of chainId when applying records

stbrody:


stbrody:
FYI @PaulLeCam

oed (approved):
Lgtm.
Also seem like we never implemented the chainid logic for `caip10-link`s 🤔

stbrody:
> Also seem like we never implemented the chainid logic for `caip10-link`s

@oed what do you mean?  This PR includes changes to `caip10-link-doctype.ts` and `caip10-link-doctype-handler.ts` that is removing the chainId logic that was previously added...

oed:
Err, seems like I somehow missed that 😵

PR #576, opened 2020-11-30
feat(core): Add ceramic network configurations

stbrody:
We support 4 networks to start: 'mainnet', 'testnet-clay', 'local', and 'inmemory'.


packages/cli/src/bin/ceramic.ts
```diff
@@ -15,9 +15,10 @@ program
     .option('--debug', 'Enable debug logging level. Default is false')
     .option('--log-to-files', 'If debug is true, write logs to files. Default is false')
     .option('--log-path <dir>', 'Store logs in this directory. Defaults to "/usr/local/var/log/ceramic"')
+    .option('--ceramic-network <name>', 'Name of the ceramic network to connect to. Defaults to "inmemory"')
```
oed, line 18: Can we rename this to something like `--network` or `--network-name`? Having ceramic in there seems redundant. Also is there a way from the CLI to get more info about which networks are available for use?

packages/core/src/__tests__/ceramic.test.ts
```diff
@@ -122,11 +123,42 @@ describe('Ceramic integration', () => {
     await tmpFolder.cleanup()
   })
 
+  it('can create Ceramic instance on default network', async () => {
+    const stateStorePath = await tmp.tmpName()
+    const ceramic = await Ceramic.create(ipfs1, {stateStorePath})
+    await delay(1000)
```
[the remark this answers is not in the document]
oed: My guess is that it has something to do with `ceramic.close()` below and that things are not getting shut down correctly.

oed (commented):
Looks good, one small name change I'd make.

packages/cli/src/bin/ceramic.ts
oed:
```suggestion
    .option('--network <name>', 'Name of the ceramic network to connect to. Defaults to "inmemory"')
```

packages/core/src/__tests__/ceramic.test.ts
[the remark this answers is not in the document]
PaulLeCam: It's likely some listener such as a socket or file handler not being closed, https://jestjs.io/docs/en/cli#--detectopenhandles might help debug this

oed:
Pushed a commit that changes the CLI to use `testnet-clay` by default. From discussion with Paul and Janko: it's important that the cli is on the same network as 3id-connect by default.

packages/core/src/__tests__/ceramic.test.ts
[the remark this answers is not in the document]
simonovic86: @oed @stbrody this is not related to the `ceramic.close()` but with IPFS instance start/stop(s). I have to investigate this issue more but I'll approve this PR since it's a blocker.

simonovic86 (approved):
lgtm

packages/cli/src/bin/ceramic.ts
stbrody: > Can we rename this to something like `--network` or `--network-name`? Having ceramic in there seems redundant.

I used "ceramic network" over "network" in a few places throughout this PR (such as the class `CeramicNetworkOptions` because I was worried about confusion with lower level networking configuration.  I think this flag you changed is the only user-facing one, but I could see us in the future exposing more config options for the lower level networking behavior and I worry a little about this flag name getting confusing.  But I also agree that `ceramic-network` feels redundant.  If in the future we wind up with `--network=mainnet --networkProtocol=UDP` or something like that, I'm not sure how confusing it will actually be.  Hopefully we can name such hypothetical future options clearly enough to disambiguate them from this flag.  This is all to say I think I'm fine with the change you made, even if I have some minor reservations.

> Also is there a way from the CLI to get more info about which networks are available for use?
If you provide the flag without a network name or with an invalid network name then the set of supported networks does get printed.  I didn't put the list here originally so there'd be fewer places that need to be updated if we add a network.  But it's probably better to include them here anyway so its immediately obvious to developers who run `--help`

oed: Thanks for clarifying. I think you are right about just making future flags more specific.

PR #578, opened 2020-12-01
Log Async

valmack:
### Motivation

I noticed some warnings in our hosted node about files not being rotated so looking into this I found that using `fs.promises` alone is not the best way to handle writing to files because it can lead to memory leaks and race conditions if we are doing a lot of logging.

### Changes

- Use `Writable` stream to handle writing logs to files -- In most cases this should work fine but in the event that the stream buffer gets full or is busy, it will prevent additional writes until it is flushed/finished, which may cause some logs to be dropped.

### DIFF
--- a/packages/cli/src/__mocks__/fs.ts
+++ b/packages/cli/src/__mocks__/fs.ts
@@ -63,6 +63,26 @@ function createWriteStream(filePath: string, options: { flags: string }): MockSt
     return new MockStream(filePath, options.flags)
 }
 
+function writeFile(filePath: string, data: any, options: { flag: string }): Promise<void> {
+    const prevFile = mockFs[filePath] || {
+        name: filePath,
+        birthtime: new Date(),
+        message: ''
+    }
+
+    if (options.flag === 'w') {
+        mockFs[filePath] = {
+            ...prevFile,
+            message: data
+        }
+    } else if (options.flag === 'a') {
+        mockFs[filePath] = {
+            ...prevFile,
+            message: prevFile.message.concat(data)
+        }
+    }
+}
+
 async function rename(filePath: string, nextFilePath: string): Promise<void> {
     const file = mockFs[filePath]
     delete mockFs[filePath]
@@ -78,5 +98,6 @@ fs.__getMockFs = __getMockFs
 fs.createWriteStream = createWriteStream
 fs.promises.rename = rename
 fs.promises.stat = stat
+fs.promises.writeFile = writeFile
 
 module.exports = fs
--- a/packages/cli/src/__mocks__/fs.ts
+++ b/packages/cli/src/__mocks__/fs.ts
@@ -7,6 +7,7 @@ interface MockFile {
     birthtime: Date;
     message: string;
     size: number;
+    __blocking: boolean;
 }
 
 interface MockFs {
@@ -24,12 +25,15 @@ class MockStream {
         this.writeFlag = writeFlag
     }
 
-    write(message: string): void {
+    on(event: string, cb: Function): void {}
+
+    write(message: string): boolean {
         const prevFile = mockFs[this.filePath] || {
             name: this.filePath,
             birthtime: new Date(),
             message: '',
-            size: 0
+            size: 0,
+            __blocking: false
         }
 
         if (this.writeFlag === 'w') {
@@ -43,6 +47,11 @@ class MockStream {
                 message: prevFile.message.concat(message)
             }
         }
+        setTimeout(() => {
+            // delay
+        }, 1000);
+        if (mockFs[this.filePath].__blocking) return false // Can not write again
+        return true
     }
 
     end(): void {
@@ -65,27 +74,6 @@ function createWriteStream(filePath: string, options: { flags: string }): MockSt
     return new MockStream(filePath, options.flags)
 }
 
-function writeFile(filePath: string, data: any, options: { flag: string }): Promise<void> {
-    const prevFile = mockFs[filePath] || {
-        name: filePath,
-        birthtime: new Date(),
-        message: '',
-        size: 0
-    }
-
-    if (options.flag === 'w') {
-        mockFs[filePath] = {
-            ...prevFile,
-            message: data
-        }
-    } else if (options.flag === 'a') {
-        mockFs[filePath] = {
-            ...prevFile,
-            message: prevFile.message.concat(data)
-        }
-    }
-}
-
 async function rename(filePath: string, nextFilePath: string): Promise<void> {
     const file = mockFs[filePath]
     delete mockFs[filePath]
@@ -93,7 +81,7 @@ async function rename(filePath: string, nextFilePath: string): Promise<void> {
 }
 
 async function stat(filePath: string): Promise<MockFile> {
-    return mockFs[filePath]
+    return Promise.resolve(mockFs[filePath])
 }
 
 fs.__initMockFs = __initMockFs
@@ -101,6 +89,5 @@ fs.__getMockFs = __getMockFs
 fs.createWriteStream = createWriteStream
 fs.promises.rename = rename
 fs.promises.stat = stat
-fs.promises.writeFile = writeFile
 
 module.exports = fs
--- a/packages/cli/src/__mocks__/fs.ts
+++ b/packages/cli/src/__mocks__/fs.ts
@@ -46,9 +46,6 @@ class MockStream {
                 message: prevFile.message.concat(message)
             }
         }
-        setTimeout(() => {
-            // delay
-        }, 1000);
         if (mockFs[this.filePath].__blocking) return false // Can not write again
         return true
     }
--- a/packages/cli/src/__mocks__/fs.ts
+++ b/packages/cli/src/__mocks__/fs.ts
@@ -24,8 +24,7 @@ class MockStream {
         this.filePath = filePath
         this.writeFlag = writeFlag
     }
-
-    on(event: string, cb: Function): void {}
+    on(event: string, cb: any): void { /* eslint-disable-line */ }
 
     write(message: string): boolean {
         const prevFile = mockFs[this.filePath] || {
@@ -54,9 +53,7 @@ class MockStream {
         return true
     }
 
-    end(): void {
-        // pass
-    }
+    end(): void { /* eslint-disable-line */ }
 }
 
 function __initMockFs(files: Array<MockFile>): void {
--- a/packages/cli/src/__mocks__/fs.ts
+++ b/packages/cli/src/__mocks__/fs.ts
@@ -6,6 +6,7 @@ interface MockFile {
     name: string;
     birthtime: Date;
     message: string;
+    size: number;
 }
 
 interface MockFs {
@@ -27,7 +28,8 @@ class MockStream {
         const prevFile = mockFs[this.filePath] || {
             name: this.filePath,
             birthtime: new Date(),
-            message: ''
+            message: '',
+            size: 0
         }
 
         if (this.writeFlag === 'w') {
@@ -67,7 +69,8 @@ function writeFile(filePath: string, data: any, options: { flag: string }): Prom
     const prevFile = mockFs[filePath] || {
         name: filePath,
         birthtime: new Date(),
-        message: ''
+        message: '',
+        size: 0
     }
 
     if (options.flag === 'w') {
--- a/packages/cli/src/__tests__/ceramic-logger-plugins.test.ts
+++ b/packages/cli/src/__tests__/ceramic-logger-plugins.test.ts
@@ -1,4 +1,4 @@
-import { LogToFiles } from '../ceramic-logger-plugins'
+import { BlockedFiles, LogToFiles } from '../ceramic-logger-plugins'
 
 jest.mock('fs')
 import fs from 'fs'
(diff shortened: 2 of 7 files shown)


valmack:
@PaulLeCam added you because I thought you might know about how streams work in Node

packages/cli/src/__mocks__/fs.ts
```diff
@@ -63,6 +63,26 @@ function createWriteStream(filePath: string, options: { flags: string }): MockSt
     return new MockStream(filePath, options.flags)
 }
 
+function writeFile(filePath: string, data: any, options: { flag: string }): Promise<void> {
+    const prevFile = mockFs[filePath] || {
+        name: filePath,
+        birthtime: new Date(),
```
oed, line 69: Haha, love this variable naming!

oed (approved):

PaulLeCam:
Sorry I'm not familiar at all with high-frequency I/O but I think some existing logging libraries might already have solutions for this?
I like using RxJS to help manipulate data streams and doing flow control but I don't know if that's a good option here.

oed:
@stbrody want to give this a final review?

packages/cli/src/ceramic-logger-plugins.ts
```diff
@@ -86,28 +113,125 @@ export class LogToFiles {
...
+        })
+        stream.on('drain', () => {
+            blockedFiles[filePath] = false
+            return
+        })
+        stream.on('finish', () => {
+            blockedFiles[filePath] = false
+            return
+        })
+        blockedFiles[filePath] = !stream.write(util.format(message) + '\n', () => {
```
stbrody, line 175: If stream.write() returns false, is that definitely because the stream is blocked (the file system write buffer is full)?  Or do we need to check a return code or something to see why exactly the write failed?

packages/cli/src/ceramic-logger-plugins.ts
```diff
@@ -54,29 +73,37 @@ export class LogToFiles {
...
-        stream.write(util.format(message) + '\n')
-        stream.end()
+    private static async _write (
+        blockedFiles: BlockedFiles,
+        basePath: string,
+        filePrefix: string,
+        message: string
+    ): Promise<void> {
+        const filePath = `${basePath}${filePrefix}.log`
+        await LogToFiles._shouldRotate(filePath) && await LogToFiles._rotate(filePath)
```
stbrody, line 89: hmm... stat-ing a file (which _shouldRotate does internally to get the file size and creation time) is a relatively expensive syscall.  Calling this on every single line logged may cause us problems down the road.  Instead of doing this inline to logging, can we have a periodic background task that checks if all log files need rotating and rotates any that do?

packages/cli/src/ceramic-logger-plugins.ts
```diff
@@ -117,7 +241,7 @@ export class LogToFiles {
      */
     private static async _rotate (filePath: string): Promise<void> {
         try {
-          await fsPromises.rename(filePath, `${filePath}.old`)
+            await fsPromises.rename(filePath, `${filePath}.old`)
```
stbrody, line 244: what happens after the second rotation?  Don't you wind up overriding your old `<filename>.old` log file?  I think we'd be better off encoding a timestamp into the file name for historical log files, so we can keep around all historical logs until they are manually archived.

packages/cli/src/ceramic-logger-plugins.ts
```diff
@@ -86,28 +113,125 @@ export class LogToFiles {
...
+        writeFlag: string
+    ): Promise<WriteResult> {
+        if (blockedFiles[filePath]) {
+            console.warn(`Stream busy for ${filePath}. Some logs may be dropped.`)
+            return { blocked: blockedFiles[filePath], filePath }
+        }
+
+        blockedFiles[filePath] = true
+
+        const stream = fs.createWriteStream(filePath, { flags: writeFlag })
```
stbrody, line 162: Hmm... I don't see a reference to 'fs' in our package.json file.  Do you know which specific package this is we're using?  I'm trying to find docs as to whether `fs.createWriteStream` creates a new file descriptor each time, or if it keeps open a longer lived file handle under the hood.

I think this most likely creates and destroys a fd every time we log a line, which may cause us some perf issues down the road, but doing the work to keep a file handle alive longer is non-trivial and probably not worth it right now.  Just something to keep an eye on going forward as a place for future optimization.

packages/cli/src/ceramic-logger-plugins.ts
valmack: No it is because stream is blocked and we need to wait for it to emit drain before writing again.

packages/cli/src/ceramic-logger-plugins.ts
valmack: Yes to keep this simple we are overwriting. These logs do get archived by Loki after we ship them so we don't need to add that complexity here.

packages/cli/src/ceramic-logger-plugins.ts
valmack: Fs is native to Node.js. It creates a new fd each time, so yes it does create and destroy, which we want for now to ensure we don't have fds lying around open.

valmack: For context: https://nodejs.org/api/fs.html#fs_fs_createwritestream_path_options

packages/cli/src/ceramic-logger-plugins.ts
valmack: Thanks for pointing this out. Alternatively we can track the file size/created time in some state and then rotate when it exceeds the threshold, that way we only have to do the stat when the process starts.

But I'd actually prefer using something like this: https://github.com/iccicci/rotating-file-stream
I believe that is what they are doing as well.

packages/cli/src/__mocks__/fs.ts
```diff
@@ -41,11 +46,14 @@ class MockStream {
                 message: prevFile.message.concat(message)
             }
         }
+        setTimeout(() => {
```
PaulLeCam, line 49: What is the intent here please?
I don't think adding the timer here is going to block the thread, in case that was the expected behavior?

packages/cli/src/__tests__/ceramic-logger-plugins.test.ts
```diff
@@ -10,105 +10,181 @@ describe('Ceramic Logger Plugins', () => {
...
+    const BLOCKING = 4
 
     describe('LogToFiles', () => {
         let mockFiles: Array<any>
         let mockFs: any
+        let blockedFiles: BlockedFiles
+        let basePath: string
 
         beforeEach(async () => {
+            basePath = '/usr/local/var/log/'
```
PaulLeCam, line 28: Does this work on Windows?

packages/cli/src/ceramic-logger-plugins.ts
```diff
@@ -86,28 +113,125 @@ export class LogToFiles {
...
+        return false
+    }
+
+    /**
+     * Returns true if it has been `MINUTES_TO_EXPIRATION` minutes since the file was created
+     * @notice Returns false if unable to get file creation datetime
+     * @param fileStats `fs.Stats` of file
+     * @returns If the file is expired
+     */
+    private static async _isExpired (fileStats: fs.Stats): Promise<boolean> {
```
PaulLeCam, line 217: Doesn't need to be `async`

packages/cli/src/ceramic-logger-plugins.ts
```diff
@@ -86,28 +113,125 @@ export class LogToFiles {
...
+        const minutesSinceBirth = (Date.now() - birthtime.getTime()) / (1000 * 60)
+        return minutesSinceBirth >= LogToFiles.MINUTES_TO_EXPIRATION
+    }
+
+    /**
+     * Returns true if file size is >= `MAX_FILE_BYTES`
+     * @param fileStats `fs.Stats` of file
+     * @returns If the file is at max size
+     */
+    private static async _isMaxSize (fileStats: fs.Stats): Promise<boolean> {
```
PaulLeCam, line 232: Doesn't need to be `async`

packages/cli/src/ceramic-logger-plugins.ts
```diff
@@ -86,28 +113,125 @@ export class LogToFiles {
...
-            if (err.code == 'ENOENT') return false
-            else console.error(err)
+            if (err.code == 'ENOENT') { // File not found
+                return false
+            }
+            console.error(err)
+            return true
+        }
+
+        if (!fileStats) {
```
PaulLeCam, line 200: Looks like `LogToFiles._isExpired` and `LogToFiles._isMaxSize` don't need to be async so this logic could be reduced to `return !fileStats || LogToFiles._isExpired(fileStats) || LogToFiles._isMaxSize(fileStats)`

packages/cli/src/ceramic-logger-plugins.ts
stbrody: Ooh yeah, if we can *not* have to write this ourselves that's definitely the way to go :sweat_smile:

packages/cli/src/ceramic-logger-plugins.ts
stbrody: okay great!  I assume it throws if there's another type of error?  What happens in that case?  I think aborting the process in the case of a logging error is probably fine and the way to go?

packages/cli/src/ceramic-logger-plugins.ts
stbrody: How quickly/frequently do they get shipped/backed up?  Is there a risk if we fill a file too quickly that we might rotate over it before it's backed up?

packages/cli/src/ceramic-logger-plugins.ts
stbrody: (I'm fine with deferring this to future optimization though)

packages/cli/src/__tests__/ceramic-logger-plugins.test.ts
valmack: i have to change this path anyway because it doesn't work with docker anymore. Will look into something that should work independent of os

packages/cli/src/__mocks__/fs.ts
valmack: I think it should when using `jest.useFakeTimers` https://jestjs.io/docs/en/timer-mocks

packages/cli/src/ceramic-logger-plugins.ts
valmack: Unfortunately for now I haven't found a way around this. At least working with docker, the attached volume doesn't get the writes until the __fd is closed__. So even when I use `stream.cork`, `stream.uncork` which is _supposed_ to buffer and flush, docker does see anything until fd is closed/process is destroyed.

packages/cli/src/ceramic-logger-plugins.ts
valmack: Yeah i think so actually. will add that to the on error event.

packages/cli/src/ceramic-logger-plugins.ts
valmack: they are getting shipped immediately (on every write). so i don't think there should be an issue of a file getting filled too quickly.

packages/cli/src/ceramic-logger-plugins.ts
valmack: oh thank you! forgot to clean that up

packages/cli/src/__mocks__/fs.ts
PaulLeCam: Isn't the behavior there relying on the `callback` being called?
I haven't used these timers features from Jest so maybe it works as-is, all good if that's the case!

packages/cli/src/ceramic-logger-plugins.ts
stbrody: Thinking about this more, that solution only works for our infra, not for others running their own js-ceramic nodes.  Does it add a lot of complexity to encode a timestamp into the file name?  That's how I've usually seem log rotation handled so that we never override an existing log.  That leaves more flexibility for different deployments to use different archiving strategies.

valmack: Hmm... i mean we can add a time stamp but that will won't be proper rotation. Logs will just grow infinitely. We would need compression at least too and some removal strategy. It's a bit of a can of worms. Using the [package I mentioned](https://www.npmjs.com/package/rotating-file-stream) will help to take care of this so I'll add it as a story for this sprint. For this pr though I'd rather not because it is currently broken :sweat_smile:

packages/cli/src/__mocks__/fs.ts
valmack: Oh you are right. I think I just had it left over from some debugging but it's not needed. Thanks!

packages/cli/src/ceramic-logger-plugins.ts
valmack: @stbrody seem ok to you?

stbrody: Yeah, the point is separation of concerns.  js-ceramic shouldn't mandate any particular archival strategy, and we shouldn't be performing log cleanup internally as we could wind up deleting logs that are still needed.  From the js-ceramic perspective, yes, that would mean that the size of the logs grow unbounded, but there could be external processes to archive and remove old logs so we don't fill up the storage volume.  I do feel kind of strongly that js-ceramic shouldn't be deleting logs out from under our users, but I don't think it needs to be a release blocker.  If you want to merge this and move on building our infrastructure with this log rotation strategy for now, that's fine, but I do think we'll want to revisit this not too long after launch so that users who are running their own nodes have flexibility in how they handle log archival.

valmack: Agree!

PR #580, opened 2020-12-01
feat(core): Add default anchor service for 'testnet-clay'

stbrody:


oed (approved):
Nice, we likely have to update this when we have the new deployments fyi.

PR #584, opened 2020-12-01
chore(core): Fix flaky test by waiting for ipfs nodes to start up

stbrody:


packages/core/src/__tests__/ceramic.test.ts
```diff
@@ -96,12 +96,17 @@ describe('Ceramic integration', () => {
...
+        return await getPort({port: getPort.makeRange(start, start + offset)})
     }
 
     ([port1, port2, port3] = await Promise.all([p1Start, p2Start, p3Start].map(start => findPort(start, pOffset))));
     ([ipfs1, ipfs2, ipfs3] = await Promise.all([port1, port2, port3].map(port => createIPFS(buildConfig(tmpFolder.path, port)))));
-    ([p1Start, p2Start, p3Start] = [p1Start, p2Start, p3Start].map(start => start + pOffset))
+    ([p1Start, p2Start, p3Start] = [p1Start, p2Start, p3Start].map(start => start + 1))
+
+    // Wait for all IPFS nodes to be fully started up.
+    await ipfs1.config.getAll();
```
stbrody, line 103: These three lines are what actually fixes the tests once the calls to `delay(1000)` are removed.  The rest is just cleanup.

How I got to this line is interesting.  I believe that there may very well be a bug in js-ipfs where `IPFSApi.start()` doesn't actually properly wait for every part of the startup process to completely finish.  I set about trying to verify this hypothesis in the js-ipfs repo.  I tried to write a test to repro it, but I couldn't figure out some details of their test environment set up, and it didn't seem worthwhile to spend a ton of time on doing so.  But while I was in there I noticed that everywhere I saw their tests calling `IPFSApi.start()` it was always immediately followed by a call to `<node>.config.getAll()`.  This was because the tests always used the returned config to assert something, but it got me wondering if that call always being there for testing was actually masking the underlying race in `IPFSApi.start()`.  I decided to give it a try here and lo and behold it worked.  Now, that might be simply because adding this call adds runtime to the test, similar to my old calls to `delay(1000)` and that's what makes things work rather than this actually properly waiting for the ipfs node startup to finish under the hood.

I don't love the idea of committing this without understanding why it works, but I also don't know much time it makes sense to spend on this issue right now...

stbrody: An alternative would be to replace these three `config.getAll()` lines with a single `delay(1000)` here instead, which also seems to make the tests pass.  So we're just explicitly giving some extra time to the IPFS node to finish starting up.  The downside is that then the tests are definitely still racy - if startup takes longer than a second then the tests might still fail, but at least we know why it fails and what's happening, instead of relying on this magic incantation I found by spelunking in the IPFS code base  :laughing:

simonovic86: nice! I guess it does the trick for now 🙂

simonovic86 (approved):
nice! lgtm

oed (approved):

PR #587, opened 2020-12-02
fix(doctype-tile): bug in squashing logic was mitigated

oed:
The squashed record should not depend on the state of the previous record that added to `next` but is no longer part of the document log.


simonovic86 (approved):
lgtm!

stbrody (approved):

stbrody (changes requested):
Actually, I retract my approval - can we get a test somewhere that would have exposed this?  This is a pretty subtle bug - I'm curious how you even noticed it - so getting a regression test in place now seems valuable

oed:
@stbrody added a test

packages/doctype-tile/src/__tests__/tile-doctype.test.ts
```diff
@@ -267,6 +267,41 @@ describe('TileDoctypeHandler', () => {
     expect(state).toMatchSnapshot()
   })
 
+  it.only('squashes record correctly', async () => {
```
stbrody, line 270: I'm not the most familiar with jest, but from a quick google it seems like using `it.only` here will cause the other tests in this file not to run?  @oed

packages/doctype-tile/src/__tests__/tile-doctype.test.ts
```diff
@@ -267,6 +267,41 @@ describe('TileDoctypeHandler', () => {
...
+    const signedRecord2 = await TileDoctype._makeRecord(doctype, did, { other: { obj2: 'fefe' } })
+
+    await context.ipfs.dag.put(signedRecord2, FAKE_CID_3)
+    const sPayload2 = dagCBOR.util.deserialize(signedRecord2.linkedBlock)
+    await context.ipfs.dag.put(sPayload2, signedRecord2.jws.link)
+
+    // apply signed
+    const state2 = await tileDoctypeHandler.applyRecord(signedRecord2.jws, FAKE_CID_3, context, deepCopy(genesisState))
+
+    expect(state2).toMatchSnapshot()
```
stbrody, line 302: sigh, I get why the snapshots are useful and convenient, but I gotta admit I'm not a huge fan.  It means that it's totally unclear in the actual test what the desired behavior really is.  What goes wrong without your fix?  Is the content of the document wrong, or just some of the log structure is less ideal in some way?  Since I thought this bug was a blocker to something else, I assume that there's a real behavior change that's observable to the end user in some way?  Can we actually test that behavior explicitly here?  Can be in addition to the snapshot check, which is still useful for at least noticing when data formats change, but doesn't actually tell us if the change is good or bad.

packages/doctype-tile/src/__tests__/tile-doctype.test.ts
oed: err, yeah this should have been removed -.-

oed: It's useful when you just want to run a test while coding.

packages/doctype-tile/src/__tests__/tile-doctype.test.ts
oed: The test in this case crashes without the fix so wasn't exactly clear how to show this here. Without the fix, line 300 would throw an error.

stbrody: Oh interesting.  So was squashing just totally broken and totally untested?

packages/doctype-tile/src/__tests__/tile-doctype.test.ts
stbrody: kinda thing we could have a lint rule for, once we have a linting system in place

packages/doctype-tile/src/__tests__/tile-doctype.test.ts
oed: I think it worked in some cases but not all.

PR #588, opened 2020-12-03
fix(doctype-tile): don't require auth when creating empty genesis records

oed:


packages/doctype-tile/src/tile-doctype.ts
```diff
@@ -65,29 +61,19 @@ export class TileDoctype extends Doctype {
...
-        } else {
-            // If 'deterministic' is undefined, default to creating document uniquely
-            unique = base64Encode(randomBytes(12))
+        // If 'deterministic' is undefined, default to creating document uniquely
+        const unique = params.deterministic ? '0' : base64Encode(randomBytes(12))
+
+        const metadata = params.metadata ? params.metadata : { controllers: [] }
+        if (!metadata.controllers || metadata.controllers.length === 0) {
+            if (params.content) {
+                metadata.controllers = [context.did?.id]
```
simonovic86, line 70: if the DID is not defined this will be `[undefined]`, right?

simonovic86 (changes requested):
There might be a potential bug if I'm not mistaken.

packages/doctype-tile/src/tile-doctype.ts
oed: If DID is not defined then the `_signDagJWS` function will throw an error.

simonovic86: yeah, but that's happening later in the code. I think we should throw an `No controllers specified` error instead.

oed: Why does that matter? I think this should throw `No DID Authenticated` error regardless.

simonovic86: `metadata.controllers` will be `[undefined]` until the `_signDagJWS` call

oed: If `context.did` is undefined `_signDagJWS` will throw before that.

oed: Why does that matter?

oed: I fail to see how this could ever be a problem?

simonovic86: Instead of 
```
metadata.controllers = [context.did?.id]
``` 
we could write

```
metadata.controllers  = context.did? [context.did.id] : []
```

and controllers list will never be `[undefined]`. If someone later tries to do something with `[undefined]` there could be an issue. I'm just saying. If you think that's not an issue than OK.

oed: I don't understand the problem you are trying to solve?

simonovic86:
```
if (params.content) {
    metadata.controllers = [context.did?.id]
} else {
    throw new Error('No controllers specified')
}
```

I'm saying that we can throw an error if the `metadata.controllers = [undefined]` as well.

oed: But why? Seems like it just adds more code without solving any problem?

PaulLeCam (commented):
LGTM but I agree with @simonovic86's comment on avoiding having `controllers: [undefined]`, even if `_signDagJWS` throws an error I think it's safer to throw errors as close to the relevant logic as possible, if nothing else it makes it easier to understand the code.

packages/doctype-tile/src/tile-doctype.ts
simonovic86: If someone needs to validate or do whatever with the controllers list **before** the `_signDagJWS` call there will be an issue.

oed:
Ok @simonovic86 @PaulLeCam, figured out a clean way to implement the desired behaviour :)

simonovic86 (approved):
lgtm

PR #591, opened 2020-12-03
Feat/multiquery ceramic api

zachferland:
~~In progress~~

re earlier (not changed here yet) 

does seem to make sense to implement a `multiQuery' function in core along with loadLinkedDocuments 
```
async multiQuery(queries: StateQuery):  Promise<Record<string, Doctype>>
```

then it will be easier to test there and the http daemon handler can just mostly deal with (de)serialize as it should instead of the extra logic included there now

http client can then implement multiQuery as well 

also with both func in core, it can be optimized when handling many queries, especially if there is likely to be any redundant paths between docids/queries (ie docid1 included already in path of docid2 and then they may have subpaths overlapping as well)

### DIFF
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -1,7 +1,7 @@
 import express, { Request, Response, NextFunction } from 'express'
 import Ceramic from '@ceramicnetwork/core'
 import type { CeramicConfig} from "@ceramicnetwork/core"
-import { DoctypeUtils, RootLogger, Logger, IpfsApi, Doctype  } from "@ceramicnetwork/common"
+import { DoctypeUtils, RootLogger, Logger, IpfsApi, Doctype, MultiQuery  } from "@ceramicnetwork/common"
 import { LogToFiles } from "./ceramic-logger-plugins"
 import DocID from "@ceramicnetwork/docid"
 import cors from 'cors'
@@ -39,12 +39,7 @@ interface HttpLog {
   response?: Record<string, unknown>;
 }
 
-interface MultiQuery {
-  docid: string
-  paths?: Array<string>
-}
-
-interface StatesQuery {
+interface MultiQueries {
   queries: Array<MultiQuery>
 }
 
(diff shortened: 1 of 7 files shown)


packages/cli/src/ceramic-daemon.ts
```diff
@@ -143,6 +152,7 @@ class CeramicDaemon {
   registerAPIPaths (app: core.Express, gateway: boolean): void {
     app.get(toApiPath('/records/:docid'), this.records.bind(this))
     app.post(toApiPath('/documents'), this.createDocFromGenesis.bind(this))
+    app.post(toApiPath('/states'), this.multiQuery.bind(this))
```
oed, line 155: Maybe we should call this endpoint `multiqueries` since we no longer have a `state` endpoint?
cc @simonovic86 any thoughts on naming here?

oed (commented):
Looks good. I agree that it would make sure to add the method you suggest.
Also, don't forget to implement the http-client :)

packages/cli/src/ceramic-daemon.ts
simonovic86: yep, `multiqueries` is more appropriate 👍

packages/cli/src/ceramic-daemon.ts
```diff
@@ -256,6 +266,43 @@ class CeramicDaemon {
...
+  async multiQuery (req: Request, res: Response, next: NextFunction): Promise<void> {
+    const { queries } = <StatesQuery> req.body
+    try {
+
+      const promiseList = queries.map(query => {
+        try {
+          const docId = DocID.fromString(query.docid)
+          return this.ceramic.loadLinkedDocuments(docId, query.paths)
+        } catch (e) {
+          return Promise.resolve({})
```
simonovic86, line 281: let's add error logging here

simonovic86 (approved):
nice! added a couple of comments

packages/cli/src/ceramic-daemon.ts
oed: Probably `multiquery` is better?

simonovic86: I think `multiqueries` makes more sense:
- we are creating one multiquery in for multiqueries resource
- we used plural before for all the resources
- we can refactor it to make it async and query the multiquery itself (GET `/multiqueries/:id`)

zachferland:
Lastly need to handle errors for loading any docs, and timeouts. Since often will be used optimistically, any docs that cant be resolved should not block return of rest, and should likely have shorter timeout than else where. Errors for individual docs could be maybe returned in some form, or can just be excluded from the response. 

@simonovic86 think timeouts were discussed before for loadDocument? was something going to be implemented there or is it just up to lib/func consumer to implement if needed?

simonovic86:
> Lastly need to handle errors for loading any docs, and timeouts. Since often will be used optimistically, any docs that cant be resolved should not block return of rest, and should likely have shorter timeout than else where. Errors for individual docs could be maybe returned in some form, or can just be excluded from the response.
> 
> @simonovic86 think timeouts were discussed before for loadDocument? was something going to be implemented there or is it just up to lib/func consumer to implement if needed?

@zachferland I guess that `timeout`s should be left for the consumer side. Any thoughts @oed ?

zachferland:
yeah that would make sense to me, then multiquery it self should still implement some timeout, so that one doc does not block all others from returning

PR #594, opened 2020-12-04
feat(document): Unify implementations of Document.load and Document.loadVersion

stbrody:
Unifies Document.load and Document.loadVersion by removing the latter and making the former load a specific version if the DocID contains one.  This also adds functionality to load a Document at a specific tip.  That functionality is not yet exposed through the CeramicAPI however, that will come in a follow-up review.


packages/core/src/document.ts
```diff
@@ -172,46 +151,87 @@ class Document extends EventEmitter {
...
 
   /**
-   * Loads a specific version of the Doctype
-   *
-   * @param doc - Document instance
-   * @param version - Document version
-   * @param validate - Validate content against schema
+   * Loads a specific version of the document
+   * @param id - base document id
+   * @param version - version of the document being requested
```
oed, line 161: Can we make version as part of the `id`. Would make more sense since the DocID class includes version anyway. Right now it would be confusing if you have a versioned DocID  and  want to load it.

oed (commented):
Can we make `load` and `loadVersion` into one function? Seems like there shouldn't be any difference between them (assuming DocID includes the version, which the `load` function need to be able to handle anyway)

Also we should add an optional param `tip` in case you want to load a document at a specific tip.

packages/core/src/document.ts
```diff
@@ -113,49 +114,122 @@ class Document extends EventEmitter {
...
+   * or apply any additional records past the genesis record.
+   * @param id - Document id
+   * @param handler
+   * @param dispatcher
+   * @param pinStore
+   * @param context
+   * @param validate
+   * @private
    */
-  async validate(): Promise<void> {
```
stbrody, line 150: This function was totally unused and was already a source of bugs via confusion between `this.validate` (the function) and `this._validate` (the boolean member variable).  Given validation should be happening as part of document loading anyway, I just went ahead and removed this function entirely

stbrody:
Okay @oed a new version is up that removed `Document.loadVersion` and puts its functionality into `Document.load`.  I also add a new argument to `Document.load` that can be used to request the Document as-of any valid log record, regardless of whether its a valid version (i.e. it's an anchor record) or not.  That functionality to load at a 'tip' isn't added to CeramicAPI and thus isn't accessible via the HTTP client yet, as I think there are some UX questions about how we want to do that and this PR was getting complex enough already, I figured it made sense to get this out and do a separate review for exposing the new functionality to the HTTP api.

packages/core/src/document.ts
```diff
@@ -164,48 +238,12 @@ class Document extends EventEmitter {
    * @param version - Document version
    */
   async loadVersion<T extends Doctype>(version: CID): Promise<T> {
-    const doc = await Document.loadVersion<T>(this, version, this._validate)
+    const versionId = DocID.fromOther(this.id, version)
+    const opts: DocOpts = {anchor: false, publish: false, sync: false}
+    const doc = await Document.load<T>(versionId, this._doctypeHandler as DoctypeHandler<T>, this.dispatcher, this.pinStore, this._context, opts, this._validate)
```
oed, line 243: Do we want to keep this method? A bit confusing that there are still two  ways of loading a version of a doc?

oed (approved):
Looks good. Just one question on keeping the `loadVersion` method on Document instances.

packages/cli/src/ceramic-daemon.ts
```diff
@@ -208,9 +208,8 @@ class CeramicDaemon {
    * Get document state
    */
   async state (req: Request, res: Response, next: NextFunction): Promise<void> {
-    const docId = DocID.fromString(req.params.docid, req.query.version ? req.query.version.toString() : undefined)
     try {
-      const doc = await this.ceramic.loadDocument(docId)
+      const doc = await this.ceramic.loadDocument(req.params.docid)
```
simonovic86, line 212: is there a reason why you removed `DocID.fromString(req.params.docid)`? If we do it like this we're going to pass a `string` to `loadDocument`.

simonovic86: also, we should definitely wrap `const docId = DocID.fromString(req.params.docid)` which is used in many methods here since it can throw an error. We should call `next(err)` then. Maybe we can do that as part of another story.

packages/core/src/__tests__/document.test.ts
```diff
@@ -277,35 +277,76 @@ describe('Document', () => {
...
       expect(docV1.controllers).toEqual(controllers)
       expect(docV1.content).toEqual(initialContent)
       expect(docV1.state.anchorStatus).toEqual(AnchorStatus.ANCHORED)
 
-      // try to call doctype.change
+      // try to call doctype.change on doc that's tied to a specific version
       try {
         await docV1.doctype.change({ content: doc.content, controllers: doc.controllers })
-        throw new Error('Should not be able to fetch not anchored version')
+        throw new Error('Should not be able to change document that was loaded at a specific version')
```
simonovic86, line 310: can we use `jest`s `fail` here: `fail('Should not be able to change document that was loaded at a specific version')`

packages/core/src/__tests__/document.test.ts
```diff
@@ -277,35 +277,76 @@ describe('Document', () => {
...
+
+      // doc.loadVersion is a convenience alias for Document.load with a specific version
+      const docV1Again = await doc.loadVersion(doc.doctype.state.log[1].cid)
+      expect(docV1Again.state).toEqual(docV1.state)
+
+      // Cannot load historical signed record when loading a version
+      try {
+        const invalidVersionId = DocID.fromOther(doc.id, doc.doctype.state.log[2].cid)
+        await Document.load(invalidVersionId, doctypeHandler, dispatcher, pinStore, context)
+        throw new Error('Should not be able to fetch not anchored version')
```
simonovic86, line 323: same as above: `fail('Should not be able to fetch not anchored version')`

packages/core/src/__tests__/document.test.ts
```diff
@@ -277,35 +277,76 @@ describe('Document', () => {
...
+      } catch (e) {
+        expect(e.message).toContain('does not refer to a valid version, which must correspond to an anchor record')
+      }
+
+      // Cannot load signed record that hasn't been anchored when loading a version
+      try {
+        const invalidVersionId = DocID.fromOther(doc.id, doc.doctype.state.log[4].cid)
+        await Document.load(invalidVersionId, doctypeHandler, dispatcher, pinStore, context)
+        throw new Error('Should not be able to fetch not anchored version')
```
simonovic86, line 332: `fail('Should not be able to fetch not anchored version')`

packages/core/src/__tests__/document.test.ts
```diff
@@ -412,7 +453,7 @@ describe('Document', () => {
 
       try {
         await Document.load(doc.id, doctypeHandler, dispatcher, pinStore, context, { sync: false })
-        throw new Error('Should not be able to assign a schema to a document that does not conform')
+        throw new Error("Should not be able to load a document that doesn't conform to its schema")
```
simonovic86, line 456: `fail('Should not be able to load a document that doesn't conform to its schema')`

simonovic86 (commented):
great cleanup! added a couple of comments

stbrody (commented):
Okay, new version is up that removed `doc.loadVersion` and also removes the functionality for loading a document at a tip, as that will come in a follow up commit under https://github.com/ceramicnetwork/js-ceramic/issues/612

packages/core/src/ceramic.ts
```diff
@@ -488,15 +488,20 @@ class Ceramic implements CeramicApi {
...
     if (!doctypeHandler) {
       throw new Error(docId.typeName + " is not a valid doctype")
     }
-
-    if (!this._docmap[docIdStr]) {
-      this._docmap[docIdStr] = await Document.load(docId, doctypeHandler, this.dispatcher, this.pinStore, this.context, opts)
+    const doc = await Document.load(docId, doctypeHandler, this.dispatcher, this.pinStore, this.context, opts)
+    if (!docId.version) {
+      // Only cache document if we're loading the baseId (i.e. the current version)
+      this._docmap[docIdStr] = doc
```
oed, line 502: In the future it would be nice if we could cache the state object, but not the instance of a doc.

packages/core/src/document.ts
```diff
@@ -113,97 +113,113 @@ class Document extends EventEmitter {
...
+   */
+  static async _syncDocumentToCurrent(doc: Document, pinStore: PinStore, opts: DocOpts): Promise<Document> {
+    // TODO: Assert that doc contains only the genesis record
+    const id = doc.id
+
+    // Update document state to cached state if any
     const isStored = await pinStore.stateStore.exists(id)
     if (isStored) {
       doc._doctype.state = await pinStore.stateStore.load(id)
     }
```
oed, line 146: I think this (`isStored`) also needs to happen in the `_syncDocumentToVersion` method.

oed (changes requested):
Looks like there is one required change here.

packages/core/src/__tests__/document.test.ts
simonovic86: yep, there are more places in the codebase definitely where we can change that

packages/cli/src/ceramic-daemon.ts
simonovic86: 👍

packages/core/src/ceramic.ts
stbrody: agreed

packages/core/src/document.ts
stbrody: Well, as written I don't think that works as there's no guarantee that the cached state is on the same branch of history as the version being requested.  I think we can do this as part of https://github.com/ceramicnetwork/js-ceramic/issues/614 though, and I just added a comment there with an approach that I think will work that we can do now.  But I'd like to do that as a follow-up since I have a stack of commits built on top of this one

oed (approved):
FYI commit message should be: `feat(<affected-package-name>):`. This will make the lerna semver publishing tool happy. If multiple packages just do `feat:`.

PR #596, opened 2020-12-06
fix(3id-did-resovler): resolve legacy 3IDs properly

oed:
Fixes three things when resolving legacy docs:
* Document creation should be deterministic
* Properly set the `controller` of the ceramic document
* Use correct DID string when wrapping document


oed:
Not sure why build is failing. Haven't changed anything that could cause this I think 🤔 
cc @simonovic86

packages/3id-did-resolver/src/index.ts
```diff
@@ -92,23 +92,21 @@ const legacyResolve = async (ceramic: Ceramic, didId: string, version?: string):
   const legacyDoc = wrapDocument(legacyPublicKeys, `did:3:${didId}`)
   if (version === '0') return legacyDoc
 
-  const managementKey = `did:key:${legacyPublicKeys.publicKeys.signing}`
-
   try {
-    const content =  { metadata: { controllers: [managementKey], family: '3id' } }
+    const content =  { deterministic: true, metadata: { controllers: [legacyPublicKeys.keyDid], family: '3id' } }
```
zachferland, line 96: in ceramic how does the deterministic flag in content work now?

zachferland (approved):

packages/3id-did-resolver/src/index.ts
oed: If it not set a random number will be added to the genesis record.

simonovic86: Can we rename `content` to `params` since it's a type of `DocParams`? @zachferland `deterministic` flag is not part of the `content` but of `DocParams`.

packages/3id-did-resolver/src/index.ts
```diff
@@ -92,23 +92,21 @@ const legacyResolve = async (ceramic: Ceramic, didId: string, version?: string):
...
     const doc = await ceramic.createDocument('tile', content, { applyOnly: true })
-    const didDoc = await resolve(ceramic, doc.id.toString(), version)
+    const didDoc = await resolve(ceramic, doc.id.toString(), version, didId)
```
simonovic86, line 98: We can inline variable here

simonovic86 (approved):
lgtm! added a couple of comments

simonovic86:
> Not sure why build is failing. Haven't changed anything that could cause this I think 🤔
> cc @simonovic86

@oed hm, I'll take a look now

packages/3id-did-resolver/src/index.ts
oed: inline what variable?

simonovic86: a minor refactor, just `return resolve(ceramic, doc.id.toString(), version, didId)`

oed: Aha, yep. Didn't get it because your comment only showed the line above the return statement.

oed: Actually we don't want to return this promise since we want to catch any error that happens while it executes.

simonovic86: np! actually, it's my bad. It should be `return await resolve(ceramic, doc.id.toString(), version, didId)` 🙂

Issue #548, opened 2020-11-25
Add message format validation to incoming pub/sub messages


stbrody:
We should enforce that pub/sub messages match the expected schema for the message type.  Easiest way to implement is probably to define a few json-schemas for the different message types and check that the incoming message conforms to the json-schema in the dispatcher after figuring out the message type from its `typ` field.

PR #601, opened 2020-12-07
Record types

simonovic86:
- add more strict record types
- revert to `delay` in `ceramic.test.ts`


packages/common/src/doctype.ts
```diff
@@ -18,56 +18,85 @@ export enum AnchorStatus {
...
-    proof: CID; // should be CID type
-    path: string;
+export interface RecordHeader {
+    controllers: Array<string>
+    schema?: string
+    tags?: Array<string>
+
+    unique?: string
+
+    [index: string]: any // allow support for future changes
```
PaulLeCam, line 28: Shouldn't future changes update this interface anyways?

packages/common/src/doctype.ts
```diff
@@ -18,56 +18,85 @@ export enum AnchorStatus {
...
 export interface DocParams {
-    metadata?: DocMetadata;
+    metadata?: DocMetadata
     // deterministic is a tri-state. True means means always create the document deterministically,
     // false means always force the document to be unique, undefined means use the default behavior
-    deterministic?: boolean;
+    deterministic?: boolean
 
-    [index: string]: any; // allow arbitrary properties
+    [index: string]: any // allow arbitrary properties
```
PaulLeCam, line 90: Do we want this? Having explicit props would make it easier to know what is supported here.

packages/common/src/utils/doctype-utils.ts
```diff
@@ -2,8 +2,16 @@ import CID from 'cids'
...
+import {
+    AnchorRecord,
+    CeramicRecord,
+    DocMetadata,
+    IpfsApi,
+    RecordHeader,
+    SignedRecord, SignedRecordDTO
+} from "../index"
 import { AnchorStatus, DocState, Doctype, LogEntry } from "../doctype"
+import { DagJWS, DagJWSResult } from "dids/lib"
```
PaulLeCam, line 14: Should import from `dids` rather than `dids/lib` here

PaulLeCam (changes requested):
Nice, thanks for making these changes!
I just spotted one issue with the `dids` import.

packages/common/src/utils/doctype-utils.ts
simonovic86: yep, tnx! 👍

packages/common/src/doctype.ts
simonovic86: I agree. I'll remove it

packages/common/src/doctype.ts
simonovic86: makes sense! I'll update it

packages/common/src/doctype.ts
simonovic86: actually, `DocParams` can have arbitrary properties depending on the doctype (`content` for tile, etc.)

packages/common/src/doctype.ts
simonovic86: I guess we have to leave this for now because of the `nonce` in `TileDoctype`.

packages/common/src/doctype.ts
PaulLeCam: Maybe that could be handled dynamically then? Something like
```ts
DocParams<T> = {
  deterministic?: boolean
  metadata?: DocMetadata
  doctype: T
}
```
This way we could still have an explicit key and type for the doctype-specific params?

simonovic86: @PaulLeCam hm, not sure if that will solve the problem. For example, `TileParams`:

```
interface TileParams extends DocParams {
    content?: Record<string, unknown>;
}
```

packages/common/src/doctype.ts
oed: yeah, this should be flexible

packages/common/src/doctype.ts
PaulLeCam: OK I see, I think we could use method overrides then?
```ts
createDocument('caip10-link', Caip10LinkParams): ...
createDocument('tile', TileParams): ...
...
```

packages/core/src/document.ts
```diff
@@ -124,13 +123,6 @@ class Document extends EventEmitter {
 
     const record = await dispatcher.retrieveRecord(doc._genesisCid)
 
-    let payload
-    if (DoctypeUtils.isSignedRecord(record)) {
-      payload = await dispatcher.retrieveRecord(record.link)
-    } else {
-      payload = record
-    }
```
oed, line 132: Why is this no longer needed?

oed (commented):
Makes sense.

packages/common/src/doctype.ts
simonovic86: The `createDocument` in the API should become something like:

```
createDocument<T extends Doctype, D extends DocParams>(doctype: string, params: D, opts?: DocOpts): Promise<T>;
```

Maybe I didn't get it 🤔  Do you have something else in mind?

packages/core/src/document.ts
simonovic86: I guess it has been refactored before since payload is not being used anymore.

packages/common/src/doctype.ts
PaulLeCam: The issue with `<T extends Doctype, D extends DocParams>` is that `T = TileDoctype` and `D = Caip10LinkParams` would be valid type-wise even though it can create runtime errors, so I think that should be avoided.
With the code I suggested above, having `createDocument('tile', Caip10LinkParams)` would be a compilation error.

PaulLeCam (approved):

packages/common/src/doctype.ts
```diff
@@ -18,56 +18,85 @@ export enum AnchorStatus {
...
+
+    [index: string]: any // allow support for future changes
+}
+
+export type GenesisHeader = RecordHeader
+
+export type GenesisRecord = {
+    header: GenesisHeader
+    data?: any
+    unique?: string
```
stbrody, line 36: unique is duplicated - here and in RecordHeader.  We probably only need the one here, not the one in the header?

packages/common/src/doctype.ts
```diff
@@ -18,56 +18,85 @@ export enum AnchorStatus {
...
+}
+
+export interface UnsignedRecord {
+    id: CID
```
stbrody, line 40: what CID is this?  Is this the records own CID?  That couldn't ever be in the record itself...

EDIT: I believe this is actually the CID of the genesis record.  Basically a proxy for the DocID.  Can we just add a quick comment clarifying that?

packages/common/src/doctype.ts
```diff
@@ -18,56 +18,85 @@ export enum AnchorStatus {
...
-    root: CID;
+    chainId: string,
+    blockNumber: number,
+    blockTimestamp: number,
+    txHash: CID,
+    root: CID,
+}
+
+export interface AnchorRecord {
+    id: CID,
```
stbrody, line 55: Same question as UnsignedRecord - what CID is this?

packages/common/src/doctype.ts
```diff
@@ -18,56 +18,85 @@ export enum AnchorStatus {
...
+    prev: CID,
+    refs?: Array<CID>
```
stbrody, line 57: This refs field isn't actually used anywhere yet, right?  Can we remove it for now until we actually implement logarithmic document syncing?

packages/common/src/utils/doctype-utils.ts
```diff
@@ -143,39 +151,51 @@ export class DoctypeUtils {
      * @param record - Record value
      * @param ipfs - IPFS instance
      */
-    static async convertRecordToDTO(record: any, ipfs: IpfsApi): Promise<any> {
+    static async convertRecordToDTO(record: CeramicRecord, ipfs: IpfsApi): Promise<CeramicRecord> {
```
stbrody, line 154: maybe I'm being overly pedantic here, but I don't actually think any of these new types are DTOs as they don't actually handle serialization/deserialization off the network.  Because typescript's type checking is so loose, even with this change it's possible to load an object off the network and put it into a CeramicRecord even if it doesn't actually conform to the proper format for any of the CeramicRecord types.  Ideally we would perform some type checking on every object we read off the network, but we don't do that yet.  I already filed https://github.com/ceramicnetwork/js-ceramic/issues/548 for the pub-sub messages, we should probably have a corresponding ticket for the CeramicRecords we load off of IPFS.

packages/core/src/__tests__/ceramic.test.ts
```diff
@@ -93,17 +97,12 @@ describe('Ceramic integration', () => {
     }
 
     const findPort = async (start: number, offset: number): Promise<number> => {
-        return await getPort({port: getPort.makeRange(start, start + offset)})
+      return await getPort({port: getPort.makeRange(start + 1, start + offset)})
```
stbrody, line 100: why do we need to add 1 to the start port?

packages/core/src/__tests__/ceramic.test.ts
```diff
@@ -93,17 +97,12 @@ describe('Ceramic integration', () => {
...
     }
 
     ([port1, port2, port3] = await Promise.all([p1Start, p2Start, p3Start].map(start => findPort(start, pOffset))));
     ([ipfs1, ipfs2, ipfs3] = await Promise.all([port1, port2, port3].map(port => createIPFS(buildConfig(tmpFolder.path, port)))));
-    ([p1Start, p2Start, p3Start] = [p1Start, p2Start, p3Start].map(start => start + 1))
-
-    // Wait for all IPFS nodes to be fully started up.
-    await ipfs1.config.getAll();
```
stbrody, line 104: What happens if you replace these calls to getAll() with a single call to delay(1000), and remove the delay calls from the test bodies?  Would be nice to handle this in the generic beforeEach part if possible so it doesn't have to be remember to be done in each individual test.

packages/core/src/document.ts
```diff
@@ -100,7 +99,7 @@ class Document extends EventEmitter {
   /**
    * Loads the Doctype by id
    * @param id - Document ID
-   * @param findHandler - find handler fn
+   * @param handler - find handler
```
stbrody, line 102: I don't think the word 'find' means anything here, let's just drop it

packages/doctype-caip10-link/src/caip10-link-doctype-handler.ts
```diff
@@ -40,12 +48,12 @@ export class Caip10LinkDoctypeHandler implements DoctypeHandler<Caip10LinkDoctyp
...
      * @param state - Document state
      */
-    async applyRecord(record: any, cid: CID, context: Context, state?: DocState): Promise<DocState> {
+    async applyRecord(record: CeramicRecord, cid: CID, context: Context, state?: DocState): Promise<DocState> {
         if (state == null) {
             return this._applyGenesis(record, cid)
         }
 
-        if (record.proof) {
+        if ((record as AnchorRecord).proof) {
```
stbrody, line 56: Can we use the `isAnchorRecord` static function from doctype-utils for this?

stbrody (commented):
Phew, you turned this around so quick I didn't get a chance to review before it was merged!

Overall I think this is a great cleanup!  I have a few minor comments that could be done in a small follow-up commit.  I think the main thing I want to point out though, since this ticket was in the "security fixes" epic, is that none of this actually adds any additional safety at runtime.  Typescript does no type checking at runtime, so when we fetch records out of IPFS, putting them into a `CeramicRecord` type doesn't actually ensure that the data conforms to any of the valid `CeramicRecord` data layouts.  If we want actual input validation on what we're taking off the wire, that would need additional work.

PR #604, opened 2020-12-08
Record maximum size

simonovic86:
- add restrictions for record size: the default limit is **500KB**
- refactor tests

### DIFF
--- a/packages/core/src/__tests__/ceramic.test.ts
+++ b/packages/core/src/__tests__/ceramic.test.ts
@@ -18,15 +18,15 @@ const seed = u8a.fromString('6e34b2e1a9624113d81ece8a8a22e6e97f0e145c25c1d4d2d0e
  * @param overrideConfig - IPFS config for override
  */
 const createIPFS = async (overrideConfig: Record<string, unknown> = {}): Promise<IpfsApi> => {
-    basicsImport.multicodec.add(dagJose)
-    const format = legacy(basicsImport, dagJose.name)
+  basicsImport.multicodec.add(dagJose)
+  const format = legacy(basicsImport, dagJose.name)
 
-    const config = {
-        ipld: { formats: [format] },
-    }
+  const config = {
+    ipld: { formats: [format] },
+  }
 
-    Object.assign(config, overrideConfig)
-    return IPFS.create(config)
+  Object.assign(config, overrideConfig)
+  return IPFS.create(config)
 }
 
 /**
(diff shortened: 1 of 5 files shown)


packages/core/src/dispatcher.ts
```diff
@@ -134,9 +141,22 @@ export default class Dispatcher extends EventEmitter {
    * @param cid - Record CID
    */
   async retrieveRecord (cid: CID | string): Promise<any> {
+    await this._restrictRecordSize(cid)
     return cloneDeep((await this._ipfs.dag.get(cid)).value)
```
oed, line 145: this won't work before we do `ipfs.dag.get` since we won't have the block in the blockstore yet.

oed (changes requested):
2 mB seems quite large? Maybe something like 500 kB is better?

One issue below needs fixing.

packages/core/src/__tests__/ceramic.test.ts
```diff
@@ -17,274 +15,118 @@ const seed = u8a.fromString('6e34b2e1a9624113d81ece8a8a22e6e97f0e145c25c1d4d2d0e
 
 /**
  * Create an IPFS instance
- * @param overrideConfig - IPFS config for override
+ * @param overrideConfig - IFPS config for override
```
PaulLeCam, line 18: This change seems like a typo

packages/core/src/__tests__/ceramic.test.ts
```diff
@@ -17,274 +15,118 @@ const seed = u8a.fromString('6e34b2e1a9624113d81ece8a8a22e6e97f0e145c25c1d4d2d0e
...
-    const doctype3 = await ceramic3.createDocument<TileDoctype>(DOCTYPE_TILE, {
-      content: { test: 321 },
-      metadata: { controllers: [controller], tags: ['3id'] },
-      deterministic: true,
-    }, {
-      anchor: false, publish: false, sync: false
+    afterAll(async () => {
+        await ipfs.stop(() => console.log('IPFS stopped'))
+        await tmpFolder.cleanup()
+        await new Promise(resolve => setTimeout(() => resolve(), 1000))
```
PaulLeCam, line 80: I think it's a bit problematic if tests need these kinds of tricks to run, this may hide underlying problems that will show up in consuming apps and libs.

PaulLeCam (commented):
Using 256 kB as the max size we could restrict to using a single block, which might help with optimizations maybe?

packages/core/src/dispatcher.ts
simonovic86: makes sense!

packages/core/src/__tests__/ceramic.test.ts
simonovic86: nice catch 👍

packages/core/src/__tests__/ceramic.test.ts
simonovic86: these are the same tests that we tried to solve with `ipfs.getConfig()`. IPFS instances are not being stopped correctly. I'll take a look at it in more depth.

packages/core/src/dispatcher.ts
simonovic86: does it make sense to get it and check it's size? I'm not sure if we can skip the `get` part 🤔

oed: yeah Don't think it's possible to check size without getting the data from the network.

simonovic86: 👍

simonovic86:
closing this PR in place of [607](https://github.com/ceramicnetwork/js-ceramic/pull/607)

PR #605, opened 2020-12-08
Rename SignedRecordDTO to SignedRecordContainer

simonovic86:
- rename `SignedRecordDTO` to `SignedRecordContainer`

### DIFF
--- a/packages/common/src/utils/doctype-utils.ts
+++ b/packages/common/src/utils/doctype-utils.ts
@@ -25,7 +25,7 @@ export class DoctypeUtils {
     static serializeRecord(record: any): any {
         const cloned = cloneDeep(record)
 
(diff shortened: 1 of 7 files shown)


packages/common/src/utils/doctype-utils.ts
```diff
@@ -180,7 +180,7 @@ export class DoctypeUtils {
      * @param record - Record
      */
     static isSignedRecordDTO(record: CeramicRecord): boolean {
```
oed, line 182: maybe rename this method?

oed (approved):
lgtm

packages/common/src/utils/doctype-utils.ts
simonovic86: 👍

PR #607, opened 2020-12-08
Record size restriction

simonovic86:
- add restrictions for record size: the default limit is **256KB**
- add tests

### DIFF
--- a/packages/core/src/dispatcher.ts
+++ b/packages/core/src/dispatcher.ts
@@ -10,8 +10,8 @@ import { DoctypeUtils, RootLogger, Logger, IpfsApi } from "@ceramicnetwork/commo
 import { TextDecoder } from 'util'
 import DocID from "@ceramicnetwork/docid";
 
-const DEFAULT_GET_TIMEOUT = 30000 // 30 seconds
-const DEFAULT_MAX_IPFS_OBJECT_SIZE = 500000 // 500 KB
+const IPFS_GET_TIMEOUT = 30000 // 30 seconds
+const IPFS_MAX_RECORD_SIZE = 256000 // 256 KB
 
 /**
  * Ceramic Pub/Sub message type.
@@ -141,20 +141,20 @@ export default class Dispatcher extends EventEmitter {
    * @param cid - Record CID
    */
   async retrieveRecord (cid: CID | string): Promise<any> {
-    const record = await this._ipfs.dag.get(cid, { timeout: DEFAULT_GET_TIMEOUT })
+    const record = await this._ipfs.dag.get(cid, { timeout: IPFS_GET_TIMEOUT })
     await this._restrictRecordSize(cid)
     return cloneDeep(record.value)
   }
 
   /**
-   * Restricts record size to DEFAULT_MAX_IPFS_OBJECT_SIZE
-   * @param cid -
+   * Restricts record size to IPFS_MAX_RECORD_SIZE
+   * @param cid - Record CID
    * @private
    */
   async _restrictRecordSize(cid: CID | string): Promise<void> {
-    const stat = await this._ipfs.block.stat(cid, { timeout: DEFAULT_GET_TIMEOUT })
-    if (stat.size > DEFAULT_MAX_IPFS_OBJECT_SIZE) {
-      throw new Error(`${cid.toString()} record size ${stat.size} exceeds the maximum block size of ${DEFAULT_MAX_IPFS_OBJECT_SIZE}`)
+    const stat = await this._ipfs.block.stat(cid, { timeout: IPFS_GET_TIMEOUT })
+    if (stat.size > IPFS_MAX_RECORD_SIZE) {
+      throw new Error(`${cid.toString()} record size ${stat.size} exceeds the maximum block size of ${IPFS_MAX_RECORD_SIZE}`)
     }
   }
 
--- a/packages/core/src/dispatcher.ts
+++ b/packages/core/src/dispatcher.ts
@@ -10,6 +10,9 @@ import { DoctypeUtils, RootLogger, Logger, IpfsApi } from "@ceramicnetwork/commo
 import { TextDecoder } from 'util'
 import DocID from "@ceramicnetwork/docid";
 
+const DEFAULT_GET_TIMEOUT = 30000 // 30 seconds
+const DEFAULT_MAX_IPFS_OBJECT_SIZE = 500000 // 500 KB
+
 /**
  * Ceramic Pub/Sub message type.
  */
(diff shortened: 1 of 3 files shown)


PaulLeCam (approved):

packages/core/src/dispatcher.ts
```diff
@@ -10,6 +10,9 @@ import { DoctypeUtils, RootLogger, Logger, IpfsApi } from "@ceramicnetwork/commo
 import { TextDecoder } from 'util'
 import DocID from "@ceramicnetwork/docid";
 
+const DEFAULT_GET_TIMEOUT = 30000 // 30 seconds
+const DEFAULT_MAX_IPFS_OBJECT_SIZE = 500000 // 500 KB
```
stbrody, line 14: The word "default" in my mind implies that it can be overridden.  Given that there's no way to override these currently, they just are what is used, I'd vote to drop the word "default" from the names

packages/core/src/dispatcher.ts
```diff
@@ -134,7 +141,21 @@ export default class Dispatcher extends EventEmitter {
...
    */
   async retrieveRecord (cid: CID | string): Promise<any> {
-    return cloneDeep((await this._ipfs.dag.get(cid)).value)
+    const record = await this._ipfs.dag.get(cid, { timeout: DEFAULT_GET_TIMEOUT })
+    await this._restrictRecordSize(cid)
+    return cloneDeep(record.value)
+  }
+
+  /**
+   * Restricts record size to DEFAULT_MAX_IPFS_OBJECT_SIZE (2MB)
```
stbrody, line 150:
```suggestion
   * Restricts record size to DEFAULT_MAX_IPFS_OBJECT_SIZE
```

packages/core/src/dispatcher.ts
```diff
@@ -134,7 +141,21 @@ export default class Dispatcher extends EventEmitter {
    * @param cid - Record CID
    */
   async retrieveRecord (cid: CID | string): Promise<any> {
-    return cloneDeep((await this._ipfs.dag.get(cid)).value)
+    const record = await this._ipfs.dag.get(cid, { timeout: DEFAULT_GET_TIMEOUT })
+    await this._restrictRecordSize(cid)
```
stbrody, line 145: Would be nice to do this before loading the record, so we don't bother to load a very large record we're just going to throw out.

packages/core/src/dispatcher.ts
```diff
@@ -123,9 +126,13 @@ export default class Dispatcher extends EventEmitter {
       const cid = await this._ipfs.dag.put(jws, { format: 'dag-jose', hashAlg: 'sha2-256' })
       // put the payload into the ipfs dag
       await this._ipfs.block.put(linkedBlock, { cid: jws.link.toString() })
+      await this._restrictRecordSize(jws.link.toString())
```
stbrody, line 129: Could we enforce the size limit before storing the record in IPFS?  Otherwise we spend a lot of time uploading a large file over the network for nothing.  This would however require calculating the object size in memory rather than relying on IPFS's `stat` method...

packages/core/src/dispatcher.ts
```diff
@@ -123,9 +126,13 @@ export default class Dispatcher extends EventEmitter {
...
+      await this._restrictRecordSize(cid)
```
stbrody, line 130: This is an aside and doesn't need to be done now, but figured I'd mention it while I'm thinking of it: longer term I think it'd be nice to put an interface in front of IPFS, like you did in the CAS.  Then the logic for enforcing record size could live there so we don't have to duplicate it every time we put or fetch an object from IPFS

packages/core/src/dispatcher.ts
```diff
@@ -134,7 +141,21 @@ export default class Dispatcher extends EventEmitter {
    * @param cid - Record CID
    */
   async retrieveRecord (cid: CID | string): Promise<any> {
-    return cloneDeep((await this._ipfs.dag.get(cid)).value)
+    const record = await this._ipfs.dag.get(cid, { timeout: DEFAULT_GET_TIMEOUT })
```
stbrody, line 144: Will this work if `cid` is a string, or do we need to normalize it to be a CID object?

packages/core/src/dispatcher.ts
simonovic86: I don't think it's possible to check size without getting the data from the network.

packages/core/src/dispatcher.ts
simonovic86: I guess we can use IPLD for calculating the size of the record but I have to check that.

packages/core/src/dispatcher.ts
simonovic86: Yeah, it works both ways: `CID` | `string`.

packages/core/src/dispatcher.ts
simonovic86: Yeah, that makes sense. We could decouple the `Dispatcher` from the IPFS service.

packages/core/src/dispatcher.ts
oed: Can we make the max size an even number? @PaulLeCam suggested 256 KB earlier.

packages/core/src/dispatcher.ts
oed: Yeah, unless we want to throw a zero-knowledge proof into the IPFS implementation 😛

packages/core/src/dispatcher.ts
oed: @simonovic86 the `cid` in this snippet will never be to large fyi. The payload is always the CID that is represented in the `link` property.

packages/core/src/dispatcher.ts
simonovic86: Yep 👍

packages/core/src/dispatcher.ts
simonovic86: yeah, I know. Just wanted to be sure 🙂

packages/core/src/dispatcher.ts
stbrody: ah, I saw that 'stat' method on IPFS and thought that it was a query to the IPFS network for someone who owns the contents to return the 'stat' info.  But I guess it's actually asking the local IPFS cache for 'stat' info on an already-loaded CID?

EDIT: Nothing in the [docs](https://github.com/ipfs/js-ipfs/blob/master/docs/core-api/BLOCK.md#ipfsblockstatcid-options) suggest that you have to load the document before you can access it via the `ipfs.block` api... Is that just a case of the IPFS docs being bad?

oed: I just assumed that this was the case. Not completely certain though.

stbrody: maybe worth a manual test to see what the behavior is @simonovic86?

simonovic86: yeah, it makes sense to do that. I created a [620](https://github.com/ceramicnetwork/js-ceramic/issues/620) story for it.

packages/core/src/dispatcher.ts
simonovic86: Let's investigate this as part of [621](https://github.com/ceramicnetwork/js-ceramic/issues/621) story.

PR #608, opened 2020-12-08
fix(doctype-tile): unbreak tile makeGenesis metadata

oed:
Main painpoint was that `family` could not be set. But we should not restrict what could be set in the metadata header.

### DIFF
--- a/packages/doctype-tile/src/tile-doctype.ts
+++ b/packages/doctype-tile/src/tile-doctype.ts
@@ -70,7 +70,7 @@ export class TileDoctype extends Doctype {
         // If 'deterministic' is undefined, default to creating document uniquely
         const unique = params.deterministic ? '0' : base64Encode(randomBytes(12))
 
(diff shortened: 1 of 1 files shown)


simonovic86 (approved):
lgtm

stbrody (commented):
Do we need to do the same thing to caip10-link-doctype?

oed:
@stbrody I don't think this was present there.

stbrody:
> @stbrody I don't think this was present there.

@oed https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/doctype-caip10-link/src/caip10-link-doctype.ts#L77

oed:
Oh, weird. I looked for it there but couldn't find it. Will make a follow up PR to deal with this (and tighten up what is allowed to be done with caip10-links.

oed:
See here @stbrody https://github.com/ceramicnetwork/js-ceramic/pull/619

PR #617, opened 2020-12-08
feat(core): Allow loading signed records as document versions

stbrody:


stbrody:
CC @PaulLeCam since this will cause fallout to 3id-did-provider.  It will have to switch to using the new anchorVersionIds() method instead of allVersionIds()

oed (approved):
Looks good.

PR #618, opened 2020-12-08
feat(core): Rename document "versions" to document "commits"

stbrody:


packages/3id-did-resolver/src/index.ts
```diff
@@ -70,31 +70,31 @@ const isLegacyDid = (didId: string): boolean => {
   }
 }
 
-const getVersion = (query = ''): string | null => {
-  const versionParam = query.split('&').find(e => e.includes('version-id'))
-  return versionParam ? versionParam.split('=')[1] : null
+const getCommit = (query = ''): string | null => {
```
stbrody, line 73: This file is probably the most relevant to you @PaulLeCam

packages/core/src/__tests__/__validation__/__schemas__/collection-reference.json
```diff
@@ -21,7 +21,7 @@
   "definitions": {
     "CeramicDocId": {
       "type": "string",
-      "pattern": "^ceramic://.+(\\?version=.+)?"
+      "pattern": "^ceramic://.+(\\?commit=.+)?"
```
PaulLeCam, line 24: I think it would be useful to discuss this change further, I don't think this concept of commit matters much to consumer beyond 3id-did-provider, while a concept of a version (specific contents) would.
I don't think commits should be a concern for the schemas we use.

packages/3id-did-resolver/src/index.ts
```diff
@@ -70,31 +70,31 @@ const isLegacyDid = (didId: string): boolean => {
...
+  const commitParam = query.split('&').find(e => e.includes('commit-id'))
```
oed, line 74: Please keep this as `version-id` as this is part of the DID-core spec.

packages/core/src/__tests__/__validation__/__schemas__/collection-reference.json
oed: Not sure why the query param is still left in the schema here. No longer needed as commits/versions are part of the DocID.
```suggestion
      "pattern": "^ceramic://.+"
```

packages/core/src/__tests__/__validation__/__schemas__/doc-id-doc-id-map.json
```diff
@@ -11,7 +11,7 @@
   "definitions": {
     "CeramicDocId": {
       "type": "string",
-      "pattern": "^ceramic://.+(\\?version=.+)?"
+      "pattern": "^ceramic://.+(\\?commit=.+)?"
```
oed, line 14: Remove query param.

packages/core/src/__tests__/__validation__/__schemas__/doc-id-map.json
```diff
@@ -8,7 +8,7 @@
   "definitions": {
     "CeramicDocId": {
       "type": "string",
-      "pattern": "^ceramic://.+(\\?version=.+)?"
+      "pattern": "^ceramic://.+(\\?commit=.+)?"
```
oed, line 11: Remove query param.

packages/core/src/__tests__/__validation__/__schemas__/collection-reference.json
oed: We can discuss more, but the main reason `commit` might make more sense than `version` is that the concept of a version only really makes sense if there is one canonical log. In the future when we support CRDT doctypes that can have multiple tips, the concept of a version doesn't make much sense. 
The concept of a `commit` should be familiar to most developers, and I do think it makes sense for schemas as well.

oed (commented):
Looks good so far, seems like we need to discuss this more.

packages/core/src/__tests__/__validation__/__schemas__/collection-reference.json
PaulLeCam: I agree `commit` is familiar as in "I commit this change and I get a hash I can use as reference", but if I understand correctly commits would be performed by the anchor service, no?

oed: Ah no. Commits would be all updates to a document. Signed records and anchor records.

oed: See here https://github.com/ceramicnetwork/js-ceramic/pull/617 @PaulLeCam

stbrody:
New version up

PaulLeCam (approved):
LGTM, so just to confirm once this change is published IDX needs to use `doc.commitId` wherever `doc.versionId` is currently used, right?

oed (approved):

PR #634, opened 2020-12-10
fix(core): Fail loading document at a commit if it is rejected by conflict resolution with known state of the document

stbrody:


packages/core/src/document.ts
```diff
@@ -151,36 +149,38 @@ class Document extends EventEmitter {
...
+   * @param id - DocID of the document including the requested commit
+   * @param doc - Most current version of the document that we know about
+   * @private
    */
-  static async _syncDocumentToCommit<T extends Doctype> (
-      doc: Document,
-      tip: CID,
-      dispatcher: Dispatcher): Promise<Document> {
-    // TODO: Assert that doc contains only the genesis record
+  static async loadAtCommit<T extends Doctype> (
```
stbrody, line 162: Unfortunately we need this separate load path in Document exposed in order for the caching in the Ceramic instance to work properly

oed: Ok, that's unfortunate. Hopefully we the Document class itself is not fully needed in the future in the same way anyway.

oed (commented):
Looks good, but failing tests.

simonovic86 (approved):
lgtm!

PR #637, opened 2020-12-11
Healthcheck

valmack:
### Motivation

We are using `/pins` for healthchecks right now which gives inaccurate metrics and is not that useful to actually know the health of the node.

### Changes

- Add simple `/healthcheck` endpoint (returns 503 if cpu is over 0.7 or memory is over 0.7, 200 otherwise)
- Add cli option for max healthy cpu and memory


zachferland:
great, maybe just add thresholds as env vars instead, so they can be paramatized for deployments and not need releases here

and threshold could have a bit higher defaults, closer to 0.3 (70% cpu) and similar or bit higher for memory, might not be same here but saw significant slow ups anywhere close to these levels in past

simonovic86 (approved):
lgtm!

zachferland (approved):

oed:
@valmack can you fix the merge conflicts? Then I'll merge this.

PR #640, opened 2020-12-11
Remove squashing

simonovic86:
- remove automatic squashing
- remove `nonce` from doctype metadata and record header
- modify `in-memory` anchor service

### DIFF
--- a/packages/common/src/doctype.ts
+++ b/packages/common/src/doctype.ts
@@ -74,6 +74,8 @@ export interface DocMetadata {
     controllers: Array<string>
     schema?: string
     tags?: Array<string>
+
+    [index: string]: any // allow arbitrary properties
 }
 
 /**
--- a/packages/common/src/doctype.ts
+++ b/packages/common/src/doctype.ts
@@ -74,8 +74,6 @@ export interface DocMetadata {
     controllers: Array<string>
     schema?: string
     tags?: Array<string>
-
-    [index: string]: any // allow arbitrary properties (nonce, etc.)
 }
 
 /**
--- a/packages/core/src/__tests__/ceramic-anchor.test.ts
+++ b/packages/core/src/__tests__/ceramic-anchor.test.ts
@@ -73,7 +73,6 @@ describe('Ceramic anchoring', () => {
   let ipfs3: IpfsApi;
   let multaddr1: string;
   let multaddr2: string;
-  let multaddr3: string;
   let tmpFolder: any;
 
   const DOCTYPE_TILE = 'tile'
@@ -109,7 +108,6 @@ describe('Ceramic anchoring', () => {
 
     multaddr1 = (await ipfs1.id()).addresses[0].toString()
     multaddr2 = (await ipfs2.id()).addresses[0].toString()
-    multaddr3 = (await ipfs3.id()).addresses[0].toString()
   })
 
   afterEach(async () => {
--- a/packages/core/src/__tests__/ceramic-anchor.test.ts
+++ b/packages/core/src/__tests__/ceramic-anchor.test.ts
@@ -1,6 +1,5 @@
 import Ceramic from '../ceramic'
 import { Ed25519Provider } from 'key-did-provider-ed25519'
-import { TileDoctype } from "@ceramicnetwork/doctype-tile"
 import { AnchorStatus, Doctype, IpfsApi } from "@ceramicnetwork/common"
 import tmp from 'tmp-promise'
 import IPFS from 'ipfs-core'
--- a/packages/core/src/__tests__/ceramic-anchor.test.ts
+++ b/packages/core/src/__tests__/ceramic-anchor.test.ts
@@ -319,7 +319,7 @@ describe('Ceramic anchoring', () => {
     await ceramic2.close()
   })
 
-  it('test the same state anchored twice, first one wins', async () => {
+  it('test the same doc anchored twice (same Ceramic instance), first one wins', async () => {
     await ipfs2.swarm.connect(multaddr1)
 
     const [ceramic1, ceramic2] = await Promise.all([
@@ -346,7 +346,7 @@ describe('Ceramic anchoring', () => {
     await ceramic2.close()
   })
 
-  it('test the same state anchored twice, last one wins (in-memory anchor service behavior)', async () => {
+  it('test the same doc anchored twice (different Ceramic instances), first one wins)', async () => {
     await ipfs3.swarm.connect(multaddr1)
     await ipfs3.swarm.connect(multaddr2)
 
@@ -375,14 +375,14 @@ describe('Ceramic anchoring', () => {
     await handle1
     await handle2
 
-    expect(doctype1.content).toEqual({ x: 5 })
+    expect(doctype1.content).toEqual({ x: 7 })
     expect(doctype1.state.log.length).toEqual(3)
 
-    expect(doctype2.content).toEqual({ x: 5 })
+    expect(doctype2.content).toEqual({ x: 7 })
     expect(doctype2.state.log.length).toEqual(3)
 
     const doctype3 = await ceramic3.loadDocument(doctype1.id)
-    expect(doctype3.content).toEqual({ x: 5 })
+    expect(doctype3.content).toEqual({ x: 7 })
     expect(doctype3.state.log.length).toEqual(3)
 
     await ceramic1.close()
--- a/packages/core/src/__tests__/ceramic-anchor.test.ts
+++ b/packages/core/src/__tests__/ceramic-anchor.test.ts
@@ -1,5 +1,6 @@
 import Ceramic from '../ceramic'
 import { Ed25519Provider } from 'key-did-provider-ed25519'
+import { TileDoctype } from "@ceramicnetwork/doctype-tile"
 import { AnchorStatus, Doctype, IpfsApi } from "@ceramicnetwork/common"
 import tmp from 'tmp-promise'
 import IPFS from 'ipfs-core'
@@ -318,7 +319,7 @@ describe('Ceramic anchoring', () => {
     await ceramic2.close()
   })
 
-  it('test the same state anchored twice, last one wins', async () => {
+  it('test the same state anchored twice, last one wins (in-memory anchor service behavior)', async () => {
     await ipfs3.swarm.connect(multaddr1)
     await ipfs3.swarm.connect(multaddr2)
 
--- a/packages/core/src/__tests__/ceramic-anchor.test.ts
+++ b/packages/core/src/__tests__/ceramic-anchor.test.ts
@@ -10,6 +10,7 @@ import getPort from 'get-port'
 import dagJose from 'dag-jose'
 import basicsImport from 'multiformats/cjs/src/basics-import.js'
 import legacy from 'multiformats/cjs/src/legacy.js'
+import { TileDoctype } from "@ceramicnetwork/doctype-tile"
 
 jest.mock('../store/level-state-store')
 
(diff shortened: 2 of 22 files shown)


packages/common/src/doctype.ts
```diff
@@ -74,8 +74,6 @@ export interface DocMetadata {
     controllers: Array<string>
     schema?: string
     tags?: Array<string>
-
-    [index: string]: any // allow arbitrary properties (nonce, etc.)
```
oed, line 78: plz don't remove this. We want to support other things in the metadata (like family for example).

oed (commented):
lgtm, why draft thought?

packages/common/src/doctype.ts
simonovic86: 👍

simonovic86:
> lgtm, why draft thought?

I wanted to take another look at it and possibly optimize some things and tests. I guess it's ready for review 🙂

packages/core/src/__tests__/document.test.ts
```diff
@@ -488,7 +489,7 @@ describe('Document', () => {
       })
     })
 
-    it("Neither log is anchored, no nonces", async () => {
```
stbrody, line 491: instead of `neither log is anchored, no nonces`, this should be `neither log is anchored, logs of the same length`

packages/core/src/__tests__/ceramic-anchor.test.ts
```diff
@@ -314,30 +318,47 @@ describe('Ceramic anchoring', () => {
     await ceramic2.close()
   })
 
-  it('test the same state anchored twice, first one wins', async () => {
-    await ipfs2.swarm.connect(multaddr1)
+  it('test the same state anchored twice, last one wins', async () => {
```
stbrody, line 321: these changes completely change what this test is testing.  I'm actually not sure why removing nonces should have had any affect on this test at all?  Can we put it back the way it was?

packages/core/src/anchor/memory/in-memory-anchor-service.ts
```diff
@@ -97,31 +102,78 @@ class InMemoryAnchorService extends AnchorService {
...
-
-        let currentNonce
-        if (DoctypeUtils.isSignedRecord(record)) {
-          const payload = (await this._ceramic.ipfs.dag.get(record.link)).value
-          currentNonce = payload.header?.nonce || 0
-        } else {
-          currentNonce = record.header?.nonce || 0
+      let candidates: Candidate[] = validCandidates[compositeKey]
+      candidates = await Promise.all(candidates.map(async c => {
+        c.partialLog.push(...await this._loadCommitHistory(c.cid))
```
stbrody, line 107: Can we move loading the commit history to right after line 94 when we first build the Candidate object?

packages/core/src/anchor/memory/in-memory-anchor-service.ts
```diff
@@ -97,31 +102,78 @@ class InMemoryAnchorService extends AnchorService {
...
+  /**
+   * Load last "depth" commits
+   *
+   * Note: this method will be replaced once CAS becomes aware of documents, not just individual commits
+   *
+   * @param commitId - Start CID
+   * @param depth - Number of history commits
+   * @private
+   */
+  async _loadCommitHistory(commitId: CID, depth = 10): Promise<CID[]> {
```
stbrody, line 148: why 10?  Seems like we should load the entire history, not stop at some arbitrary point.

stbrody (changes requested):
As written this breaks some of the consensus properties of the system.  I left comments throughout, definitely let me know if anything's unclear or if there are any questions about to proceed!

packages/core/src/anchor/memory/in-memory-anchor-service.ts
```diff
@@ -97,31 +102,78 @@ class InMemoryAnchorService extends AnchorService {
...
-          const payload = (await this._ceramic.ipfs.dag.get(record.link)).value
-          currentNonce = payload.header?.nonce || 0
-        } else {
-          currentNonce = record.header?.nonce || 0
+      let candidates: Candidate[] = validCandidates[compositeKey]
+      candidates = await Promise.all(candidates.map(async c => {
+        c.partialLog.push(...await this._loadCommitHistory(c.cid))
+        return c
+      }))
+      candidates.sort((c1, c2) => c2.timestamp - c1.timestamp)
```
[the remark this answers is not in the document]
simonovic86: Yes, we can use timestamps in the **in-memory** anchor service as a heuristic as it is stated in the comments. The `in-memory` anchor service is a mock anchor service used for testing only and it is not meant to be used in production.

Since it is an in-memory anchor service and the data is mocked (`SAMPLE_ETH_TX_HASH`, etc) the timestamps and log check will do the trick until the issue [221](https://github.com/ceramicnetwork/ceramic-anchor-service/issues/221) is resolved.

Feel free to introduce the changes planned for the [221](https://github.com/ceramicnetwork/ceramic-anchor-service/issues/221) into the in-memory service to reflect the real behavior.

simonovic86:
> As written this breaks some of the consensus properties of the system. The problem is that there was an implicit assumption in our existing conflict resolution strategy that the diverging logs would only contain a single update due to squashing, and we use nonce to determine which log is logically longer. Now that squashing is gone that assumption is broken, and we need to actively consider log length as part of conflict resolution.
> 
> I left comments throughout, definitely let me know if anything's unclear or if there are any questions about to proceed!

This is not breaking the consensus properties since squashing and the `nonce` property was never part of the protocol to begin with. It only worked for `Tile` doctypes and other doctypes were not affected by it.

I will create another story for addressing the conflict resolution in general.

packages/core/src/document.ts
```diff
@@ -418,17 +418,6 @@ class Document extends EventEmitter {
       // If they have the same block number fall through to fallback mechanism
     }
 
-    // The anchor status is the same between both logs.  If either log has a 'nonce' that means
```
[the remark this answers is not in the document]
simonovic86: This needs to be implemented as part of the [649](https://github.com/ceramicnetwork/js-ceramic/issues/649) story.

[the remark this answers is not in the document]
simonovic86: reflected in the [657](https://github.com/ceramicnetwork/js-ceramic/pull/657)

packages/core/src/__tests__/document.test.ts
```diff
@@ -614,33 +597,6 @@ describe('Document', () => {
       expect(await Document._pickLogToAccept(state1, state2)).toEqual(false)
       expect(await Document._pickLogToAccept(state2, state1)).toEqual(true)
     })
-
-    it("Both logs anchored in same blockchains in the same block, one has nonce", async () => {
```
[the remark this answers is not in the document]
simonovic86: reflected in the PR [657](https://github.com/ceramicnetwork/js-ceramic/pull/657)

packages/core/src/__tests__/document.test.ts
```diff
@@ -501,29 +502,11 @@ describe('Document', () => {
...
       }
 
-      // When neither log is anchored and there's no nonces we should pick the log whose first
-      // entry has the smaller CID.
+      // When neither log is anchored we should pick the log whose first entry has the smaller CID.
       expect(await Document._pickLogToAccept(state1, state2)).toEqual(false)
       expect(await Document._pickLogToAccept(state2, state1)).toEqual(true)
     })
 
-    it("Neither log is anchored, different nonces", async () => {
```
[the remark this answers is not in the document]
simonovic86: reflected in the PR [657](https://github.com/ceramicnetwork/js-ceramic/pull/657)

packages/core/src/__tests__/document.test.ts
simonovic86: reflected in the PR [657](https://github.com/ceramicnetwork/js-ceramic/pull/657)

packages/core/src/__tests__/ceramic-anchor.test.ts
simonovic86: 👍

packages/core/src/anchor/memory/in-memory-anchor-service.ts
simonovic86: 👍

packages/core/src/anchor/memory/in-memory-anchor-service.ts
simonovic86: 👍

packages/core/src/__tests__/ceramic-anchor.test.ts
stbrody: Which test duplicates this one?  This look like a basic conflict resolution test, the only one that exists at this high a level that I see.

packages/core/src/anchor/memory/in-memory-anchor-service.ts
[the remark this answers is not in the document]
stbrody: Issue 221 is for changes to the ceramic-anchor-service repo.  Nothing we're doing here should be dependent on that work.  We want the in-memory anchor service to mirror a real anchor service as much as possible, so while we're in here setting the conflict resolution for js-ceramic, we should make the in-memory anchor service reflect the conflict resolution strategy we are moving towards.  Having the in-memory anchor service have this special behavior then leaks into our tests and affects our ability to test conflict resolution properly

packages/core/src/__tests__/ceramic-anchor.test.ts
```diff
@@ -314,30 +318,47 @@ describe('Ceramic anchoring', () => {
     await ceramic2.close()
   })
 
-  it('test the same state anchored twice, first one wins', async () => {
```
stbrody, line 317: As mentioned in my other comment, I still think this test is useful to include and would rather not remove it

packages/core/src/__tests__/ceramic-anchor.test.ts
```diff
@@ -314,30 +318,47 @@ describe('Ceramic anchoring', () => {
...
+    await handle1
+    await handle2
+
+    expect(doctype1.content).toEqual({ x: 5 })
     expect(doctype1.state.log.length).toEqual(3)
 
-    const doctype2 = await ceramic2.loadDocument(doctype1.id)
-    expect(doctype1.content).toEqual(doctype2.content)
-    expect(doctype1.state.log.length).toEqual(doctype2.state.log.length)
+    expect(doctype2.content).toEqual({ x: 5 })
```
stbrody, line 353: This is specifically relying on the behavior of the in-memory anchor service to pick which logs to anchor based on timestamp rather than based on deterministic conflict resolution strategies like we'll actually use in real anchor services. If we're going to write a test of conflict resolution behavior, it should be based on the behaviors we actually want and expect

packages/core/src/anchor/memory/in-memory-anchor-service.ts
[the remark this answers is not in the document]
simonovic86: timestamps are used just as heuristic, the sort can be removed and the behavior is the same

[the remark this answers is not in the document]
simonovic86: timestamps removed

packages/core/src/anchor/memory/in-memory-anchor-service.ts
```diff
@@ -99,29 +104,68 @@ class InMemoryAnchorService extends AnchorService {
...
-
-        let currentNonce
-        if (DoctypeUtils.isSignedRecord(record)) {
-          const payload = (await this._ceramic.ipfs.dag.get(record.link)).value
-          currentNonce = payload.header?.nonce || 0
-        } else {
-          currentNonce = record.header?.nonce || 0
+      // naive implementation for finding the "valid" commit for the document.
+      // the decision making is going to change once the anchoring service becomes aware of Doctypes
+      let selected: Candidate
```
stbrody, line 109: So, I still don't think this is exactly right for how the InMemoryAnchorService should be handling conflict resolution, but rather than continuing to hold up this PR, I went ahead and filed https://github.com/ceramicnetwork/js-ceramic/issues/664 so we can get this out the door and we can circle back to the inmemory anchor service after

stbrody (approved):
LGTM, thank you for your patience during review process!

PR #657, opened 2020-12-14
Change conflict resolution strategy

simonovic86:
- change the conflict resolution strategy to address the case when anchor states are the same for both logs. In that case, we should compare compare log lengths and choose the one with longer length.

### DIFF
--- a/packages/core/src/document.ts
+++ b/packages/core/src/document.ts
@@ -418,17 +418,6 @@ class Document extends EventEmitter {
       // If they have the same block number fall through to fallback mechanism
     }
 
-    // The anchor status is the same between both logs.  If either log has a 'nonce' that means
-    // multiple updates have been squashed into a single anchor window, so prefer the log with the
-    // higher nonce (which indicates it has had more writes).
-    const nonce1 = (state1.next?.metadata.nonce ?? state1.metadata.nonce) || 0
-    const nonce2 = (state2.next?.metadata.nonce ?? state2.metadata.nonce) || 0
-    if (nonce1 < nonce2) {
-      return true
-    } else if (nonce2 < nonce1) {
-      return false
-    }
-
     // If we got this far, that means that we don't have sufficient information to make a good
     // decision about which log to choose.  The most common way this can happen is that neither log
     // is anchored, although it can also happen if both are anchored but in the same blockNumber or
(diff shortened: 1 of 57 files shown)


stbrody:
Hmm, is there any way to see this and https://github.com/ceramicnetwork/js-ceramic/pull/640 together as one PR @simonovic86?  As-is it's hard to see the overall change as that PR removes some tests/behaviors and this PR adds them back in.  I think it'd be easier to review if I could just see the total net change from where we are today with squashing to where we are after squashing is removed and conflict resolution works under the new system

simonovic86:
> Hmm, is there any way to see this and #640 together as one PR @simonovic86? As-is it's hard to see the overall change as that PR removes some tests/behaviors and this PR adds them back in. I think it'd be easier to review if I could just see the total net change from where we are today with squashing to where we are after squashing is removed and conflict resolution works under the new system

Yes, you can always compare the `feat/doc-conflict-resolution` branch with `develop`.

packages/core/src/document.ts
```diff
@@ -418,6 +418,13 @@ class Document extends EventEmitter {
       // If they have the same block number fall through to fallback mechanism
     }
 
+    // The anchor states are the same for both logs. Compare log lengths and choose the one with longer length.
+    if (state1.log.length > state2.log.length) {
```
stbrody, line 422: I think this case is backwards. We're supposed to return true when selecting state2's log, and false when selecting state1's log

simonovic86: 👍

packages/core/src/document.ts
```diff
@@ -384,14 +384,14 @@ class Document extends EventEmitter {
   /**
    * Given two different DocStates representing two different conflicting histories of the same
    * document, pick which commit to accept, in accordance with our conflict resolution strategy
-   * @param state1 - first log's state
-   * @param state2 - second log's state
-   * @returns true if state2's log should be taken, or false if state1's log should be taken
```
stbrody, line 389: Hmm... So I very intentionally didn't include any reference to 'local' or 'remote' when I wrote this, as an earlier version of this logic always preferred the remote log in the case of a tie and it's important that this behavior be entirely deterministic. So I think it's important that this function be written in a way that it is just comparing two random equally valid logs and not have any information about which came from where.

I do agree, however, that the name of the function and it returning a bool makes it hard to know from reading calling code what it's supposed to be doing.  I think perhaps the thing to do is change this to instead of returning a bool, return the DocState that it chose.  Feel free to do that here, or file a ticket and I can take care of it in a separate commit.  Either way though, can we revert the change to the function and variable names?

stbrody (changes requested):
Almost good to go, just want to revert the function name change for now

stbrody (approved):

PR #660, opened 2020-12-14
JUST FOR SPENCER'S VIEW: Unified PR of squashing and conflict resolution changes

stbrody:
This is a rollup of https://github.com/ceramicnetwork/js-ceramic/pull/640 and https://github.com/ceramicnetwork/js-ceramic/pull/657 designed to prevent a unified view for review purposes.  This PR however won't be merged, it is those other PRs backing it that will be the ones that get merged

### DIFF
--- a/packages/core/src/__tests__/ceramic-anchor.test.ts
+++ b/packages/core/src/__tests__/ceramic-anchor.test.ts
@@ -73,7 +73,6 @@ describe('Ceramic anchoring', () => {
   let ipfs3: IpfsApi;
   let multaddr1: string;
   let multaddr2: string;
-  let multaddr3: string;
   let tmpFolder: any;
 
   const DOCTYPE_TILE = 'tile'
@@ -109,7 +108,6 @@ describe('Ceramic anchoring', () => {
 
     multaddr1 = (await ipfs1.id()).addresses[0].toString()
     multaddr2 = (await ipfs2.id()).addresses[0].toString()
-    multaddr3 = (await ipfs3.id()).addresses[0].toString()
   })
 
   afterEach(async () => {
--- a/packages/core/src/__tests__/ceramic-anchor.test.ts
+++ b/packages/core/src/__tests__/ceramic-anchor.test.ts
@@ -10,6 +10,7 @@ import getPort from 'get-port'
 import dagJose from 'dag-jose'
 import basicsImport from 'multiformats/cjs/src/basics-import.js'
 import legacy from 'multiformats/cjs/src/legacy.js'
+import { TileDoctype } from "@ceramicnetwork/doctype-tile"
 
 jest.mock('../store/level-state-store')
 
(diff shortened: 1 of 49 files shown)


packages/core/src/__tests__/ceramic-anchor.test.ts
```diff
@@ -314,30 +318,47 @@ describe('Ceramic anchoring', () => {
     await ceramic2.close()
   })
 
-  it('test the same state anchored twice, first one wins', async () => {
```
stbrody, line 317: I still think the original version of this test is useful to have and don't see why we are removing it

simonovic86:
@stbrody Why is this PR created? We talked about it today and decided that the conflict resolution PR will be based on the previous one for squashing. Those are two different things and there's a reason why I did that. Pls don't do this anymore.

packages/core/src/__tests__/ceramic-anchor.test.ts
```diff
@@ -314,30 +318,47 @@ describe('Ceramic anchoring', () => {
...
+    await handle1
+    await handle2
+
+    expect(doctype1.content).toEqual({ x: 5 })
     expect(doctype1.state.log.length).toEqual(3)
 
-    const doctype2 = await ceramic2.loadDocument(doctype1.id)
-    expect(doctype1.content).toEqual(doctype2.content)
-    expect(doctype1.state.log.length).toEqual(doctype2.state.log.length)
+    expect(doctype2.content).toEqual({ x: 5 })
```
stbrody, line 353: This is specifically relying on the behavior of the in-memory anchor service to pick which logs to anchor based on timestamp rather than based on deterministic conflict resolution strategies like we'll actually use in real anchor services.  If we're going to write a test of conflict resolution behavior, it should be based on the behaviors we actually want and expect

packages/core/src/document.ts
```diff
@@ -418,14 +418,10 @@ class Document extends EventEmitter {
...
     }
 
-    // The anchor status is the same between both logs.  If either log has a 'nonce' that means
-    // multiple updates have been squashed into a single anchor window, so prefer the log with the
-    // higher nonce (which indicates it has had more writes).
-    const nonce1 = (state1.next?.metadata.nonce ?? state1.metadata.nonce) || 0
-    const nonce2 = (state2.next?.metadata.nonce ?? state2.metadata.nonce) || 0
-    if (nonce1 < nonce2) {
+    // The anchor states are the same for both logs. Compare log lengths and choose the one with longer length.
+    if (state1.log.length > state2.log.length) {
```
stbrody, line 422: I think this case is backwards.  We're supposed to return true when selecting state2's log, and false when selecting state1's log

stbrody (commented):
Publishing comments so I can keep track for myself of what comments have already been made on the other two backing PRs, but all conversation and changes will happen on the original PRs, not here

PR #666, opened 2020-12-16
fix(doctype-caip10-link): replay attack vector was mitigated

oed:


oed:
Not sure what's up with the build here. Works locally.

oed:
Sorry yeah. The attack is that before this fix someone could take an old *linkProof* and republish it as an update to the document to change which DID the caip10-link points to. This is because the proof itself doesn't include the `prev` pointer. So if I have a caip10-link that first pointed to DID1, then update it to point to DID2, an attacker could change it back to DID1. We mitigate this by checking that the timestamps are in order.

stbrody:
Interesting. I just re-read CIP-7 and have a better understanding now.  The timestamp in the proof just comes from the signer's system time right?  Not from a blockchain transaction timestamp.  So that means if the user who creates the original link has a clock that's far in the future when they sign the original link message, then replay attacks are still possible until the linked blockchain's timestamp catches up to the timestamp of the original proof, right?


Could we encode the 'prev' link into the proof?  CIP-7 talks about how the entire payload of the signed records is not included in the signature - I guess that's basically what I'm suggesting here.  The CIP says this is to improve user experience at the cost of security - can you explain what the user experience problem would be?

Either way, this seems good enough to mitigate the current issue, even if it doesn't totally solve it in all cases.  LGTM

stbrody (approved):

oed:
>  So that means if the user who creates the original link has a clock that's far in the future when they sign the original link message, then replay attacks are still possible until the linked blockchain's timestamp catches up to the timestamp of the original proof, right?

Correct. However it's unclear why a user would do this willingly. Perhaps an attack where someone modifies your system time?

> can you explain what the user experience problem would be?

Just more random characters that shows up when you want to create the link.

I agree that ideally we'd want to include prev (etc) at some point.

simonovic86 (approved):
lgtm!

stbrody:
> Correct. However it's unclear why a user would do this willingly. Perhaps an attack where someone modifies your system time?

Could also happen accidentally.  It means that clock synchronization on **client** machines is now an important part of our security model.  We should make sure to document that clearly somewhere @michaelsena

oed:
Yeah, good point. This would require a larger refactor of the way we link accounts in general, but definitely worth considering in the future.

PR #668, opened 2020-12-16
chore(http-client): optimize bundle size

oed:
This is an attempt at trying to minimize the build size of the http-client. The http-client shouldn't need the doctypeHandlers, just the doctypes themselves. The caip10-link handlers bundle size is quite large so the hope is that this will reduce the package  size dramatically.

Still need to confirm this though. Could potentially release an alpha to test it on bundlephobia.

Would love your thoughts @PaulLeCam!

### DIFF
--- a/packages/http-client/src/ceramic-http-client.ts
+++ b/packages/http-client/src/ceramic-http-client.ts
@@ -4,6 +4,7 @@ import Document from './document'
 import { DID } from 'dids'
 import {
   Doctype,
+  DoctypeConstructor,
   DoctypeHandler,
   DocOpts,
   DocParams,
@@ -15,8 +16,8 @@ import {
   DoctypeUtils, 
   MultiQuery
 } from "@ceramicnetwork/common"
-import { TileDoctypeHandler } from "@ceramicnetwork/doctype-tile"
-import { Caip10LinkDoctypeHandler } from "@ceramicnetwork/doctype-caip10-link"
+import { TileDoctype } from "@ceramicnetwork/doctype-tile"
+import { Caip10LinkDoctype } from "@ceramicnetwork/doctype-caip10-link"
 import DocID from '@ceramicnetwork/docid'
 import ThreeIdResolver from '@ceramicnetwork/3id-did-resolver'
 import KeyDidResolver from '@ceramicnetwork/key-did-resolver'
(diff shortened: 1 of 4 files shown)


PaulLeCam:
Looks good, I think the main thing to see the effect would be to make sure the bundling process properly does tree-shaking and dead code elimination so any unnecessary dependency is dropped?

oed:
Yeah, I'm hoping for tree-shaking.

packages/http-client/src/ceramic-http-client.ts
```diff
@@ -176,15 +177,7 @@ export default class CeramicClient implements CeramicApi {
   }
 
   addDoctypeHandler<T extends Doctype>(doctypeHandler: DoctypeHandler<T>): void {
```
stbrody, line 179: This function doesn't seem to actually be called anywhere in our codebase.  If this is just for developers to use to load their custom doctypes, can we rename it and change it to take the DocType directly instead of the DoctypeHandler?

packages/http-client/src/ceramic-http-client.ts
```diff
@@ -136,7 +137,7 @@ export default class CeramicClient implements CeramicApi {
     if (!this._docmap[docIdStr]) {
       this._docmap[docIdStr] = await Document.load(docId, this._apiUrl, this.context, this._config)
     }
-    this._docmap[docIdStr].doctypeHandler = this.findDoctypeHandler(this._docmap[docIdStr].state.doctype)
+    this._docmap[docIdStr].doctypeConstructor = this._doctypes[this._docmap[docIdStr].state.doctype]
```
stbrody, line 140: This dance between doctype and document makes me sad.  It seems like it should be possible to have this set when the document is first constructed rather than up here in the client.  Anyway that's a bigger change and doesn't need to happen now, this is good cleanup either way

packages/http-client/src/ceramic-http-client.ts
oed: Exactly. Well that would break the symmetry between core and http-client. In core the handler is needed because updates need to use the handler to apply records.

packages/http-client/src/ceramic-http-client.ts
oed: Me too :/

packages/http-client/src/ceramic-http-client.ts
stbrody: ah right, makes sense

stbrody (approved):
lgtm once tests are passing

oed:
Hm, unfortunately it doesn't seem like the tree shaking is working very effectively. So not sure if this is worth it. Thoughts @PaulLeCam ?
https://bundlephobia.com/result?p=@ceramicnetwork/http-client@0.8.9-test.0

oed:
stale, closing.

PR #669, opened 2020-12-16
feat(core): Update default CAS URL to point to clay infra

stbrody:


stbrody:
@valmack, we should try to make sure that that URL is valid and usable before releasing this, even if it's still pointing to old infra and needs to be redirected shortly thereafter.

packages/cli/src/ceramic-daemon.ts
```diff
@@ -11,7 +11,7 @@ const DEFAULT_PORT = 7007
 const DEFAULT_NETWORK = 'testnet-clay'
 const toApiPath = (ending: string): string => '/api/v0' + ending
 
-const DEFAULT_ANCHOR_SERVICE_URL = "https://cas.3box.io:8081"
+const DEFAULT_ANCHOR_SERVICE_URL = "https://cas-clay.3box.io:8081"
```
valmack, line 14: should be `https://cas-clay.3boxlabs.com`

valmack (commented):
Will point it to dev until new one is yp

packages/cli/src/ceramic-daemon.ts
stbrody: Done

valmack (approved):

PR #670, opened 2020-12-16
chore(core): Update conflict resolution in InMemoryAnchorService

stbrody:
This PR should not be considered a blocker to releasing tomorrow

### DIFF
--- a/packages/core/src/__tests__/ceramic-anchor.test.ts
+++ b/packages/core/src/__tests__/ceramic-anchor.test.ts
@@ -319,33 +319,6 @@ describe('Ceramic anchoring', () => {
     await ceramic2.close()
   })
 
-  it('test the same doc anchored twice (same Ceramic instance), first one wins', async () => {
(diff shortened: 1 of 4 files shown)


packages/core/src/__tests__/ceramic-anchor.test.ts
```diff
@@ -319,33 +319,6 @@ describe('Ceramic anchoring', () => {
     await ceramic2.close()
   })
 
-  it('test the same doc anchored twice (same Ceramic instance), first one wins', async () => {
```
stbrody, line 322: You were totally right about this test Janko, my mistake.  Because they share a ceramic instance they learn about each other's updates instantly and there's no real conflict resolution behavior being tested.  Sorry I didn't follow what you were getting at on the other review!

simonovic86: nice, cool!

simonovic86 (approved):
lgtm

PR #672, opened 2020-12-16
feat(core): Cleanup and fix conflict resolution

stbrody:
This does 2 things. First it cleans up the conflict resolution code so _pickLogToAccept returns which log it chose rather than a boolean, to increase code readability.  It also changes the conflict resolution fallback to compare the CIDs of the last log entries rather than the first, to match what the CAS does.

This PR should not be considered a blocker to releasing tomorrow

### DIFF
--- a/packages/core/src/__tests__/document.test.ts
+++ b/packages/core/src/__tests__/document.test.ts
@@ -516,10 +516,10 @@ describe('Document', () => {
         metadata: {},
       }
 
-      // When neither log is anchored and log lengths are the same we should pick the log whose first entry has the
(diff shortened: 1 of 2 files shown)


simonovic86 (approved):

stbrody:
Hmm... looks like this got merged into the `chore/in-memory-anchor-service-cleanup` branch but not into `develop`

stbrody:
> Hmm... looks like this got merged into the `chore/in-memory-anchor-service-cleanup` branch but not into `develop`

Nevermind, the code changes did get committed, they just got combined into a single commit along with the changes from https://github.com/ceramicnetwork/js-ceramic/pull/670

PR #704, opened 2020-12-22
Fix gateway endpoints

### LINKED ISSUE #703: Gateway mode should not include createDocFromGenesis
This endpoint is available when running ceramic as a gateway
```app.post(toApiPath('/documents'), this.createDocFromGenesis.bind(this))```

See https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/cli/src/ceramic-daemon.ts#L160-L179

But gateway mode should be readonly and `createDocFromGenesis` allows a doc to be created and potentially anchored

### PR DESCRIPTION
**Pull request by valmack**
Closes #703 

### Motivation

In gateway mode ceramic should not be able to publish or anchor documents

### Changes

- Replaced handler for `/documents` endpoint when in gateway mode to use `createDocumentFromGenesis` with publish and anchor set to false

### DIFF
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -243,7 +243,7 @@ class CeramicDaemon {
    */
   async createReadOnlyDocFromGenesis (req: Request, res: Response, next: NextFunction): Promise<void> {
     const { doctype, genesis, docOpts } = req.body
-    const readOnlyDocOpts = {...docOpts, anchor: false, publish: false}
+    const readOnlyDocOpts = { ...docOpts, anchor: false, publish: false }
     try {
       const doc = await this.ceramic.createDocumentFromGenesis(doctype, DoctypeUtils.deserializeRecord(genesis), readOnlyDocOpts)
       res.json({ docId: doc.id.toString(), state: DoctypeUtils.serializeState(doc.state) })
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -159,7 +159,6 @@ class CeramicDaemon {
 
   registerAPIPaths (app: core.Express, gateway: boolean): void {
     app.get(toApiPath('/records/:docid'), this.records.bind(this))
-    app.post(toApiPath('/documents'), this.createDocFromGenesis.bind(this))
     app.post(toApiPath('/multiqueries'), this.multiQuery.bind(this))
     app.get(toApiPath('/documents/:docid'), this.state.bind(this))
     app.get(toApiPath('/pins/:docid'), this.listPinned.bind(this))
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages/cli/src/ceramic-daemon.ts [hunk 1]
```diff
@@ -237,6 +238,21 @@ class CeramicDaemon {
     next()
   }
 
+  /**
+   * Create read-only document from genesis record
+   */
+  async createReadOnlyDocFromGenesis (req: Request, res: Response, next: NextFunction): Promise<void> {
```
stbrody, line 244: maybe call this `loadDocumentFromGenesis`?

packages/cli/src/ceramic-daemon.ts
```diff
@@ -237,6 +238,21 @@ class CeramicDaemon {
     next()
   }
 
+  /**
+   * Create read-only document from genesis record
```
stbrody, line 242:
```suggestion
   * Loads a document when the docID is unknown, but you have the genesis contents.
```

stbrody (approved):
LGTM

packages/cli/src/ceramic-daemon.ts
valmack: I was thinking that expect we already have `ceramic.loadDocument` which takes in a docId and internally does `Document.load`. So it might be confusing to use the name load here because it internall maps to `ceramic.createDocumentFromGenesis` and `Document.create`

oed (approved):
Nice, this  keeps the ability to load from genesis!

packages/cli/src/ceramic-daemon.ts
stbrody: Yeah, the fact that it uses different functions internally doesn't bother me, as that's an implementation detail.  What bothers me slightly more is that it doesn't return `null` if the doc doesn't already exist, like loadDocument does.  I wish we had a client-side way to get the CID for some content without having to *actually* publish that content to IPFS.  I wonder if we could find the hashing code in IPFS and expose it if it isn't already.  Anyway that's probably not worth exploring right now.

I think behavior-wise what you wrote here is fine for now, the only question is the name.  I'd probably lean towards something like `loadDocumentFromGenesis` or `loadDocFromGenesisContent`, but add a TODO comment that it shouldn't publish the genesis record to IPFS if it isn't already there.  But I don't feel strongly here

valmack: Will push this through for now but you are definitely right--the current functionality does `ipfs.dag.put` so it technically is creating something.

For that reason I'd rather the function name be more accurate so we know what this is doing (if it were public facing like the API I'd agree with you but this is hidden behind `/documents` anyway).

I'll create an additional story and add a TODO to modify this behavior

packages/cli/src/ceramic-daemon.ts
valmack: Yes this is more clear

PR #706, opened 2020-12-23
fix: surface express errors properly

### LINKED ISSUE #705: Errors in js-ceramic always respond with 200
Express is not surfacing errors due to bug in our code

See https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/cli/src/ceramic-daemon.ts#L106

```js
res.json({ error: err.message }) // sends a response with status 200 right away
next(err) // never gets called
```

valmack:
Closes #705 

### Motivation

When a request is made to the daemon that causes an error the server should respond with an error status code and message. Right not it always responds with 200

### Changes

- Moved custom error handler to be the last Express handler
  - per the [express docs](https://expressjs.com/en/guide/error-handling.html): "you define error-handling middleware last, after other app.use() and routes calls;"
- Removed the `res.json` call so Express can handle the error properly

### DIFF
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -70,13 +70,6 @@ class CeramicDaemon {
     this.registerAPIPaths(app, opts.gateway)
 
     if (this.debug) {
-      app.use((err: Error, req: Request, res: Response, next: NextFunction): void => {
-        const requestStart = Date.now()
-        const httpLog = this._buildHttpLog(requestStart, req, res)
-        const logString = JSON.stringify(httpLog)
-        this.logger.error(logString)
-        next(err)
-      })
       app.use((req: Request, res: Response, next: NextFunction): void => {
         const requestStart = Date.now()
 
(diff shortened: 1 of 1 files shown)


oed:
what does this change do? Do we not respond with 200 on errors anymore?

valmack:
> what does this change do? Do we not respond with 200 on errors anymore?

@oed Yes. By default if we get an error it will respond with 500 and send the error message as the response body

oed:
@valmack The daemon should always reply with 200. The reason for this is to remove expected errors from the console in web browsers. Browsers always log http errors even if you catch them (see https://stackoverflow.com/questions/45581361/how-can-i-catch-a-404-error-in-javascript). This is undesirable when building webapps, because sometimes we expect errors. 

There might be room for some nuance here though. Some errors should probably not be 200 if they are not expected.

valmack:
@oed I think it is fine for expected errors to respond with 200 but I don't understand why we would want the daemon to _always_ be 200. Then we don't have an easy way to know when an error is expected or not.

If we expect an error we should handle it and then send a response with whatever code we deem appropriate, usually a 4xx but could be 2xx. But by default when an error is not handled it should be surfaced and return with 500.

Maybe I'm missing something here? I don't see how the linked example applies.

oed:
No I think you are right @valmack. After this PR, which errors result in 200 and which ones does not?

simonovic86 (approved):
lgtm!

valmack:
@oed Right now we are wrapping most api operations in a try catch and then just passing the error through. With this PR as it is, pretty much all errors will result in a 500. Do you or @simonovic86 have a sense of what errors would be expected?

PR #714, opened 2020-12-23
fix(core): Don't update document state before applying anchor record

stbrody:


packages/core/src/document.ts
```diff
@@ -612,9 +612,6 @@ class Document extends EventEmitter {
           return
         }
         case 'COMPLETED': {
-          const state = this._doctype.state
```
stbrody, line 615: With this as it was, an HTTP client that is polling for updates to a document can see and emit an update with a new state that says it's anchored, but doesn't actually include the anchor record in its log

stbrody: This isn't needed because `_handleTip` will take care of updating the state appropriately

oed (approved):
Not sure how you caught this, but seems reasonable.

simonovic86 (approved):
lgtm

PR #719, opened 2020-12-25
fix(http-client): reload cached documents

oed:
Also includes a commit that fixes the package name used for `key-did-resolver`. 

Packages published as `next`:
```
 - @ceramicnetwork/3id-did-resolver@0.6.2-alpha.2
 - @ceramicnetwork/cli@0.18.1-alpha.2
 - @ceramicnetwork/common@0.15.2-alpha.2
 - @ceramicnetwork/core@0.17.2-alpha.2
 - @ceramicnetwork/docid@0.4.2-alpha.2
 - @ceramicnetwork/doctype-caip10-link@0.12.4-alpha.2
 - @ceramicnetwork/doctype-tile@0.12.6-alpha.2
 - @ceramicnetwork/http-client@0.8.10-alpha.2
 - key-did-resolver@0.2.4-alpha.2
 - @ceramicnetwork/pinning-aggregation@0.1.7-alpha.2
 - @ceramicnetwork/pinning-ipfs-backend@0.1.7-alpha.2
 - @ceramicnetwork/pinning-powergate-backend@0.1.7-alpha.2
 ```

### DIFF
--- a/packages/http-client/src/ceramic-http-client.ts
+++ b/packages/http-client/src/ceramic-http-client.ts
@@ -47,6 +47,13 @@ export interface CeramicClientConfig {
  */
 export default class CeramicClient implements CeramicApi {
   private readonly _apiUrl: string
+  /**
+   * _docmap stores handles to Documents that been handed out. This allows us
+   * to update the state within the Document object when we learn about changes
+   * to the document. This means that client code with Document references
+   * always have access to the most recent known-about version, without needing
+   * to explicitly re-load the document.
+   */
   private readonly _docmap: Record<string, Document>
   private _supportedChains: Array<string>
 
@@ -125,10 +132,7 @@ export default class CeramicClient implements CeramicApi {
     const docIdStr = doc.id.toString()
     if (!this._docmap[docIdStr]) {
       this._docmap[docIdStr] = doc
-    } else if (
-      JSON.stringify(DoctypeUtils.serializeState(doc.state)) !==
-      JSON.stringify(DoctypeUtils.serializeState(this._docmap[docIdStr].state))
-    ) {
+    } else if (DoctypeUtils.statesEqual(doc.state, this._docmap[docIdStr].state)) {
       this._docmap[docIdStr].state = doc.state
       this._docmap[docIdStr].emit('change')
     }
--- a/packages/http-client/src/ceramic-http-client.ts
+++ b/packages/http-client/src/ceramic-http-client.ts
@@ -127,6 +127,7 @@ export default class CeramicClient implements CeramicApi {
       this._docmap[docIdStr] = doc
     } else {
       this._docmap[docIdStr].state = doc.state
+      this._docmap[docIdStr].emit('change')
     }
     this._docmap[docIdStr].doctypeHandler = this.findDoctypeHandler(this._docmap[docIdStr].state.doctype)
     return this._docmap[docIdStr] as unknown as T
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: packages/http-client/src/ceramic-http-client.ts [hunk 1]
```diff
@@ -125,6 +125,8 @@ export default class CeramicClient implements CeramicApi {
     const docIdStr = doc.id.toString()
     if (!this._docmap[docIdStr]) {
       this._docmap[docIdStr] = doc
+    } else {
+      this._docmap[docIdStr].state = doc.state
```
stbrody, line 129: what's the difference between setting the entire doc here vs just the state?

packages/http-client/src/ceramic-http-client.ts
```diff
@@ -135,6 +137,8 @@ export default class CeramicClient implements CeramicApi {
     const docIdStr = docId.toString()
     if (!this._docmap[docIdStr]) {
       this._docmap[docIdStr] = await Document.load(docId, this._apiUrl, this.context, this._config)
+    } else {
+      this._docmap[docIdStr]._syncState()
```
stbrody, line 141: At this point is there even any reason to be caching the documents at all?

Also I think you need `await`

packages/http-client/src/ceramic-http-client.ts
oed: If you have a pointer to this doc elsewhere you will get the updated version.

packages/http-client/src/ceramic-http-client.ts
oed: same as above. Nice catch with `await`.

packages/http-client/src/ceramic-http-client.ts
oed: Added the emitting of the event now (same as _syncState).

packages/http-client/src/ceramic-http-client.ts
stbrody: Ah interesting, so this isn't really a cache at all, it's just a way to hold on to the handle to Documents that have been handed out so we can update the state out from under them.  Interesting.  Would be nice to write that down somewhere, can we add a comment to _docmap?

packages/http-client/src/ceramic-http-client.ts
stbrody: hmm... _syncState only emits a change event if the state is actually different than the cached version.  Here we emit the event even if nothing changed

packages/http-client/src/ceramic-http-client.ts
oed: What would you include in that comment?

packages/http-client/src/ceramic-http-client.ts
oed: Hm, you're right. I'll see if there is an easy fix.

packages/http-client/src/ceramic-http-client.ts
stbrody: How about:
`_docmap stores handles to Documents that been handed out.  This allows us to update the state within the Document object when we learn about changes to the document.  This means that client code with Document references always have access to the most recent known-about version, without needing to explicitly re-load the document.`

stbrody (approved):

packages/http-client/src/ceramic-http-client.ts
```diff
@@ -125,6 +125,12 @@ export default class CeramicClient implements CeramicApi {
     const docIdStr = doc.id.toString()
     if (!this._docmap[docIdStr]) {
       this._docmap[docIdStr] = doc
+    } else if (
+      JSON.stringify(DoctypeUtils.serializeState(doc.state)) !==
```
simonovic86, line 129: can we move this comparison to `DoctypeUtils` as well? Something like `compareStates` or similar

simonovic86 (approved):
lgtm! added just one comment

packages/http-client/src/ceramic-http-client.ts
oed: Fixed :)

packages/http-client/src/ceramic-http-client.ts
oed: Thanks, added.

packages/http-client/src/ceramic-http-client.ts
simonovic86: 🤘

PR #721, opened 2020-12-28
Docs LRU cache

simonovic86:
- implement `DocCache` and add it to `core` and `http-client`
- add tests


simonovic86:
closing it because it's still in progress

PR #722, opened 2020-12-28
Docs LRU cache

simonovic86:
- implement doctypes LRU cache: `DocCache` and replace doc maps in `core` and `http-client` with it
- add `DocStateHolder` interface: the interface is temporary since we're going to refactor `Document`(s)
- add tests
- minor refactor

### DIFF
--- a/packages/common/src/doc-cache.ts
+++ b/packages/common/src/doc-cache.ts
@@ -1,15 +1,15 @@
 import { LRUMap } from 'lru_map'
 import DocID from "@ceramicnetwork/docid"
 
-import Document from "./document"
+import { DocStateHolder } from "./doctype"
 
 /**
  * Encapsulates document caching (base, commits)
  */
-class DocumentCache {
+export class DocCache {
     private readonly _cacheCommits: boolean
-    private readonly _baseCache: LRUMap<string, Document>
-    private readonly _commitCache: LRUMap<string, Document>
+    private readonly _baseCache: LRUMap<string, DocStateHolder>
+    private readonly _commitCache: LRUMap<string, DocStateHolder>
 
     constructor(limit, cacheCommits = true) {
         this._cacheCommits = cacheCommits
@@ -18,10 +18,10 @@ class DocumentCache {
     }
 
     /**
-     * Set document to cache
-     * @param doc - Document instance
+     * Sets to cache
+     * @param doc - DocStateHolder instance
      */
-    set(doc: Document): void {
+    set(doc: DocStateHolder): void {
         this._baseCache.set(doc.id.baseID.toString(), doc)
         if (this._cacheCommits && doc.id.commit) {
             this._commitCache.set(doc.id.toString(), doc)
@@ -29,16 +29,22 @@ class DocumentCache {
     }
 
     /**
-     * Get document from cache
-     * @param docId - Document ID
+     * Gets from cache
+     * @param docId - DocId instance
      */
-    get(docId: DocID): Document {
+    get(docId: DocID): DocStateHolder {
         const doc = this._baseCache.get(docId.toString())
         if (doc) {
             return doc
         }
         return this._cacheCommits? this._commitCache.get(docId.toString()): null
     }
-}
 
-export default DocumentCache
+    /**
+     * Doc exists?
+     * @param docId - DocId instance
+     */
+    has(docId: DocID): boolean {
+        return this.get(docId) != null
+    }
+}
--- a/packages/common/src/doc-cache.ts
+++ b/packages/common/src/doc-cache.ts
@@ -14,7 +14,9 @@ export class DocCache {
     constructor(limit, cacheCommits = true) {
         this._cacheCommits = cacheCommits
         this._baseCache = new LRUMap(limit)
-        this._commitCache = new LRUMap(limit)
+
+        // use the same 'limit' if cacheCommits is enabled
+        this._commitCache = this._cacheCommits? new LRUMap(limit) : new LRUMap(0)
     }
 
     /**
--- a/packages/common/src/doc-cache.ts
+++ b/packages/common/src/doc-cache.ts
@@ -8,15 +8,15 @@ import { DocStateHolder } from "./doctype"
  */
 export class DocCache {
     private readonly _cacheCommits: boolean
-    private readonly _baseCache: LRUMap<string, DocStateHolder>
-    private readonly _commitCache: LRUMap<string, DocStateHolder>
+    private readonly _baseDocCache: LRUMap<string, DocStateHolder>
+    private readonly _commitDocCache: LRUMap<string, DocStateHolder>
 
     constructor(limit, cacheCommits = true) {
         this._cacheCommits = cacheCommits
-        this._baseCache = new LRUMap(limit)
+        this._baseDocCache = new LRUMap(limit)
 
         // use the same 'limit' if cacheCommits is enabled
-        this._commitCache = this._cacheCommits? new LRUMap(limit) : new LRUMap(0)
+        this._commitDocCache = this._cacheCommits? new LRUMap(limit) : new LRUMap(0)
     }
 
     /**
@@ -25,10 +25,9 @@ export class DocCache {
      */
     set(doc: DocStateHolder): void {
         if (doc.id.commit == null) {
-            this._baseCache.set(doc.id.baseID.toString(), doc)
-        }
-        if (this._cacheCommits && doc.id.commit) {
-            this._commitCache.set(doc.id.toString(), doc)
+            this._baseDocCache.set(doc.id.baseID.toString(), doc)
+        } else if (this._cacheCommits) {
+            this._commitDocCache.set(doc.id.toString(), doc)
         }
     }
 
@@ -37,11 +36,11 @@ export class DocCache {
      * @param docId - DocId instance
      */
     get(docId: DocID): DocStateHolder {
-        const doc = this._baseCache.get(docId.toString())
+        const doc = this._baseDocCache.get(docId.toString())
         if (doc) {
             return doc
         }
-        return this._cacheCommits? this._commitCache.get(docId.toString()): null
+        return this._cacheCommits? this._commitDocCache.get(docId.toString()): null
     }
 
     /**
@@ -57,15 +56,15 @@ export class DocCache {
      * @param applyFn - Function to apply
      */
     applyToAll(applyFn: (d: DocStateHolder) => void): void {
-        this._baseCache.forEach((d) => applyFn(d))
-        this._commitCache.forEach((d) => applyFn(d))
+        this._baseDocCache.forEach((d) => applyFn(d))
+        this._commitDocCache.forEach((d) => applyFn(d))
     }
 
     /**
      * Clears cache
      */
     clear(): void {
-        this._baseCache.clear();
-        this._commitCache.clear();
+        this._baseDocCache.clear();
+        this._commitDocCache.clear();
     }
 }
--- a/packages/common/src/doc-cache.ts
+++ b/packages/common/src/doc-cache.ts
@@ -3,58 +3,30 @@ import DocID from "@ceramicnetwork/docid"
 
 import { DocStateHolder } from "./doctype"
 
-export type OnEvictFunction = (doc: DocStateHolder) => Promise<void>
-
 /**
  * Encapsulates document caching (base, commits)
  */
 export class DocCache {
     private readonly _cacheCommits: boolean
 
+    private _pinnedDocCache: Record<string, DocStateHolder>
     private readonly _baseDocCache: LRUMap<string, DocStateHolder>
     private readonly _commitDocCache: LRUMap<string, DocStateHolder>
-    private readonly _pinnedDocCache: Record<string, DocStateHolder>
 
-    constructor(limit, onEvictFn: OnEvictFunction, cacheCommits = true) {
+    constructor(limit, cacheCommits = true) {
         this._cacheCommits = cacheCommits
-        this._baseDocCache = this._initDocLRUCache(limit, onEvictFn)
+        this._baseDocCache = new LRUMap<string, DocStateHolder>(limit)
         this._pinnedDocCache = {}
 
         // use the same 'limit' if cacheCommits is enabled
-        this._commitDocCache = this._cacheCommits? this._initDocLRUCache(limit, onEvictFn) : new LRUMap(0)
-    }
-
-    /**
-     * Initialize single LRU cache
-     * @private
-     */
-    _initDocLRUCache(limit: number, onEvictFn: OnEvictFunction) {
-        const cache = new LRUMap<string, DocStateHolder>(limit)
-        if (onEvictFn) {
-            cache.shift = function() {
-                const entryArr = LRUMap.prototype.shift.call(cache)
-                if (entryArr) {
-                    onEvictFn(entryArr[1]) // TODO handle async call
-                }
-                return entryArr
-            }
-        }
-        return cache
+        this._commitDocCache = this._cacheCommits? new LRUMap<string, DocStateHolder>(limit) : null
     }
 
     /**
      * Puts to cache
      * @param doc - DocStateHolder instance
-     * @param isPinned - Is document pinned?
      */
-    put(doc: DocStateHolder, isPinned = false): void {
-        if (isPinned) {
-            this._pinnedDocCache[doc.id.toString()] = doc;
-            this._baseDocCache.delete(doc.id.baseID.toString())
-            this._commitDocCache.delete(doc.id.baseID.toString())
-            return;
-        }
-
+    put(doc: DocStateHolder): void {
         if (doc.id.commit == null) {
             this._baseDocCache.set(doc.id.baseID.toString(), doc)
             return;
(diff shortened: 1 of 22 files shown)


packages/common/src/doctype.ts
```diff
@@ -137,10 +137,21 @@ export interface DocOpts {
...
+/**
+ * Describes object which stores DocState.
+ *
+ * Note: the interface should be removed once we refactor documents.
+ *
+ */
+export interface DocStateHolder {
+    id: DocID;
+    state: DocState;
+}
```
oed, line 149: Nice, this will be useful in the future 👍

packages/common/src/doc-cache.ts
```diff
@@ -0,0 +1,69 @@
...
+    /**
+     * Sets to cache
+     * @param doc - DocStateHolder instance
+     */
+    set(doc: DocStateHolder): void {
+        this._baseCache.set(doc.id.baseID.toString(), doc)
+        if (this._cacheCommits && doc.id.commit) {
+            this._commitCache.set(doc.id.toString(), doc)
+        }
+    }
```
oed, line 31: Not sure about the logic in this method. Seems like if I cache an older commit it will also be cached as the `_baseCache`?

oed: I guess the baseCache should only be set if `doc.id.commit` is not set?

packages/core/src/ceramic.ts
```diff
@@ -27,11 +34,11 @@ import { PathTrie, TrieNode, promiseTimeout } from './utils'
 
 import EthereumAnchorService from "./anchor/ethereum/ethereum-anchor-service"
 import InMemoryAnchorService from "./anchor/memory/in-memory-anchor-service"
-import { PinningBackendStatic } from "@ceramicnetwork/common"
-
 
 import { randomUint32 } from '@stablelib/random'
 
+const DEFAULT_BASE_DOC_CACHE_LIMIT = 100; // number of base docs stored in the cache
```
oed, line 40: This seems extremely small. The nodes on our infra needs to keep track of documents for all of our users > 15K, and each user will have multiple documents.

packages/core/src/ceramic.ts
```diff
@@ -331,19 +344,7 @@ class Ceramic implements CeramicApi {
...
-   * Puts document into cache if not already there
-   * @param doc - document to cache
-   * @private
-   */
-  private _cacheDocIfNeeded(doc: Document) {
-    const docIdStr = doc.id.toString()
-    if (!this._docmap[docIdStr]) {
-      this._docmap[docIdStr] = doc
-    }
+    return this._docCache.get(docId) as Document
```
oed, line 347: The DocCache class only caches the DocStateHolder type. I assume we need to create an instance of the document here?

oed (approved):
This is interesting. I think with a little more work we can actually use this code to remove the need to keep instances of the `Document` class at all!

One problem I see with the current implementation in that it will GC pinned documents, which is not what we want.

packages/common/src/doc-cache.ts
simonovic86: yep, my bad 👍

packages/core/src/ceramic.ts
simonovic86: I'll increase it to `500`. I added an option for setting a cache limit as well

packages/core/src/ceramic.ts
simonovic86: hm, `Document` also implements `DocStateHolder`

packages/core/src/ceramic.ts
```diff
@@ -59,6 +66,9 @@ export interface CeramicConfig {
   networkName?: string;
   pubsubTopic?: string;
 
+  docBaseCacheLimit?: number;
```
stbrody, line 69: Can we call this just `docCacheLimit` since it applies to both base and commit caches?

packages/core/src/ceramic.ts
```diff
@@ -92,19 +102,22 @@ const tryDocId = (id: string): DocID | null => {
...
+  private readonly _docCache: DocCache
+
   // TODO: Make the constructor private and force the use of Ceramic.create() everywhere
   constructor (public dispatcher: Dispatcher,
                public pinStore: PinStore,
                context: Context,
                private _networkOptions: CeramicNetworkOptions,
-               private _validateDocs: boolean = true) {
-    this._docmap = {}
+               private _validateDocs: boolean = true,
```
stbrody, line 117: We're probably getting close to the point where it would make sense to have a CeramicCreateOptions struct that we pass in, rather than an ever growing list of optional positional args.  Doesn't have to happen in this PR though

stbrody: Oh actually, we basically have that already via the `CeramicConfig` passed to `Ceramic.create`.  The real thing to do is probably what the `TODO` above says, make this constructor private and force everything to go through Ceramic.create.  That would let us put all "default" behaviors in `Ceramic.create` and make all the constructor args here non-optional

packages/http-client/src/ceramic-http-client.ts
```diff
@@ -54,7 +57,7 @@ export default class CeramicClient implements CeramicApi {
    * always have access to the most recent known-about version, without needing
    * to explicitly re-load the document.
    */
-  private readonly _docmap: Record<string, Document>
+  private readonly _docCache: DocCache
```
stbrody, line 60: So this is interesting.  Now that we limit the size of the cache here, we limit the behavior described in the comment.  Now it's possible for application code to hold onto a handle to a Document, but stop receiving updates to it once the internal cache fills up and evicts it.  Ideally we'd have a way to know whether a Document handle is still alive outside of our internal uses and prevent evicting it from our internal cache if so.  I'm not actually sure if something like this is possible in javascript though, since there isn't any real concept of a "destructor", or any other way (that I know of) to trigger code to run when an object reference goes out of scope...  @PaulLeCam, do you know if there's any conventional way to solve problems like this in javascript?

packages/http-client/src/ceramic-http-client.ts
```diff
@@ -129,27 +132,32 @@ export default class CeramicClient implements CeramicApi {
...
-    } else if (!DoctypeUtils.statesEqual(doc.state, this._docmap[docIdStr].state)) {
-      this._docmap[docIdStr].state = doc.state
-      this._docmap[docIdStr].emit('change')
+
+    let docFromCache = this._docCache.get(doc.id) as Document
+    if (docFromCache == null) {
+      this._docCache.set(doc)
+      docFromCache = doc
+    } else if (!DoctypeUtils.statesEqual(doc.state, docFromCache.state)) {
+      docFromCache.state = doc.state
```
stbrody, line 141: does the call to `_docCache.get()` above reset the doc's location in the LRU cache order?  If not, we need to do something to do so explicitly here.

packages/core/src/ceramic.ts
stbrody: But will `DocStateHolder as Document` invoke the `Document` constructor?  I don't think it does, which means that other members of `Document` such as `_logger`, `_applyQueue`, `_genesisCid`, etc won't be properly initialized

packages/common/src/doc-cache.ts
```diff
@@ -0,0 +1,71 @@
...
+
+        // use the same 'limit' if cacheCommits is enabled
+        this._commitCache = this._cacheCommits? new LRUMap(limit) : new LRUMap(0)
+    }
+
+    /**
+     * Sets to cache
+     * @param doc - DocStateHolder instance
+     */
+    set(doc: DocStateHolder): void {
```
stbrody, line 26: definitely a nitpick, but I'd probably call this `put` rather than `set` since it's not a KV-map setting a value to a specific key

oed:
> One problem I see with the current implementation in that it will GC pinned documents, which is not what we want.

This still needs to be addressed.

simonovic86:
> > One problem I see with the current implementation in that it will GC pinned documents, which is not what we want.
> 
> This still needs to be addressed.

oh yeah, I missed this one 👍

packages/common/src/doc-cache.ts
simonovic86: yeah, I wanted to call it `put` as well. The `LRUMap` uses `get` which was strange. I'll rename it

packages/core/src/__tests__/ceramic.test.ts
```diff
@@ -52,6 +55,8 @@ const createCeramic = async (ipfs: IpfsApi, anchorOnRequest = false): Promise<Ce
 }
 
 const anchor = async (ceramic: Ceramic): Promise<void> => {
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
```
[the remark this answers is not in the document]
simonovic86: The `anchor` method doesn't exist in the `AnchorService` interface in this case.

packages/core/src/__tests__/ceramic.test.ts
```diff
@@ -287,4 +292,74 @@ describe('Ceramic integration', () => {
     await ceramic1.close()
     await ceramic2.close()
   })
+
+  it('can utilize documents cache', async () => {
```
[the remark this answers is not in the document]
simonovic86: 👍

packages/core/src/ceramic.ts
simonovic86: 👍

packages/http-client/src/ceramic-http-client.ts
simonovic86: Yes, it registers the recent use.

packages/core/src/ceramic.ts
simonovic86: Why do we need to invoke the constructor?

packages/core/src/__tests__/ceramic.test.ts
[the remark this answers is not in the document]
stbrody: Can we add it then?

packages/http-client/src/ceramic-http-client.ts
stbrody: We discussed this after standup and believe the right way to solve this is to have a single Ceramic-wide EventEmitter for document updates, rather than one per document, that Document instances would subscribe to and only process the events that are relevant to it.

packages/core/src/ceramic.ts
stbrody: Wrote that because I thought the cache just had the state, so we needed to construct a real Document instance.  Now I understand that the cache actually does store full Document objects, so this `as` is just a cast to make the type checker happy

packages/core/src/__tests__/ceramic.test.ts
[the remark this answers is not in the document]
simonovic86: No, we can't do that since `anchor` is not part of the `AnchorService` API. The method anchors immediately and exists in the in-memory anchor service only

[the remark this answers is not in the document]
stbrody: Got it, that makes sense.  Thanks for explaining!

packages/common/src/doc-cache.ts
```diff
@@ -13,13 +15,31 @@ export class DocCache {
...
+     * Initialize single LRU cache
+     * @private
+     */
+    _initDocLRUCache(limit: number, onEvictFn: OnEvictFunction) {
+        const cache = new LRUMap<string, DocStateHolder>(limit)
+        if (onEvictFn) {
+            cache.shift = function() {
+                const entryArr = LRUMap.prototype.shift.call(cache)
+                if (entryArr) {
+                    onEvictFn(entryArr[1]) // TODO handle async call
```
stbrody, line 37: what does this TODO mean?

packages/http-client/src/ceramic-http-client.ts
```diff
@@ -66,11 +67,21 @@ export default class CeramicClient implements CeramicApi {
   private readonly _config: CeramicClientConfig
   public readonly _doctypeHandlers: Record<string, DoctypeHandler<Doctype>>
 
-  constructor (apiHost: string = CERAMIC_HOST, config?: CeramicClientConfig) {
-    this._config = Object.assign(DEFAULT_CLIENT_CONFIG, config ? config : {})
+  constructor (apiHost: string = CERAMIC_HOST, config: CeramicClientConfig = {}) {
+    this._config = config
+
+    // fill missing configuration from default
+    Object.keys(DEFAULT_CLIENT_CONFIG).forEach((key) => {
```
stbrody, line 74: could this be written as 
`this._config = { ...DEFAULT_CLIENT_CONFIG, ...config }`?

packages/core/src/ceramic.ts
```diff
@@ -117,12 +117,18 @@ class Ceramic implements CeramicApi {
...
-    this._docCache = new DocCache(docCacheLimit, cacheDocumentCommits)
     this._doctypeHandlers = {
       'tile': new TileDoctypeHandler(),
       'caip10-link': new Caip10LinkDoctypeHandler()
     }
 
+    // on evict the Document is being unregistered from the Dispatcher and removed from cache
+    // also, the event 'evicted' is fired so the client can catch the eviction and possibly load the doc again
+    this._docCache = new DocCache(docCacheLimit, async (d: Document) => {
+      d.doctype.emit('evicted');
```
stbrody, line 128: I thought we decided on the call earlier that we weren't going to add this event, and we'll just temporarily let Documents silently stop receiving updates if they get evicted from the cache, until we can do the fuller/better thing that requires the refactoring

packages/common/src/doc-cache.ts
```diff
@@ -0,0 +1,129 @@
...
+        }
+        return cache
+    }
+
+    /**
+     * Puts to cache
+     * @param doc - DocStateHolder instance
+     * @param isPinned - Is document pinned?
+     */
+    put(doc: DocStateHolder, isPinned = false): void {
```
stbrody, line 50: Can we remove the default value for this?  It seems useful to force developers to always have to explicitly think about whether the doc is pinned or not before putting it into the cache.

EDIT: Seeing how this is used, it seems to only ever be called with `isPinned` as true as part of the explicit pinning/unpinning workflow.  Could we actually separate this to a separate method, rather than an arg to `put`?  We could have a `pin` and an `unpin` method on the cache directly.  I think that makes it clearer what's happening here, and would also address my comment below about the behavior of `del` by making it clear that `unpin` doesn't actually delete a document from the cache, it just moves that document to the regular part of the cache where it is subject to eventual eviction down the road.

packages/common/src/doc-cache.ts
```diff
@@ -0,0 +1,129 @@
...
+        return this._cacheCommits? this._commitDocCache.get(docId.toString()): null
+    }
+
+    /**
+     * Delete from cache
+     * @param docId - DocId instance
+     * @param pinnedOnly - Delete from pinned only cache
+     */
+    del(docId: DocID, pinnedOnly = false): void {
+        if (pinnedOnly) {
```
stbrody, line 92: if `pinnedOnly` is false, should we delete the doc from all 3 caches?  Currently if `pinnedOnly` is false, but the document does indeed exist in the pinned cache, then it doesn't get deleted

packages/common/src/doc-cache.ts
```diff
@@ -0,0 +1,129 @@
...
+        this._baseDocCache.forEach((d) => applyFn(d))
+        this._commitDocCache.forEach((d) => applyFn(d))
+
+        Object.entries(this._pinnedDocCache).forEach(([, d]) => applyFn(d))
+    }
+
+    /**
+     * Clears cache
+     */
+    clear(): void {
```
stbrody, line 125: need to clear pinned cache too

packages/core/src/__tests__/ceramic.test.ts
```diff
@@ -287,4 +292,135 @@ describe('Ceramic integration', () => {
...
```
stbrody, line 296: phew.  This test is LONG and hard to follow.  Can you break it up into some smaller tests, and possibly add some comments between phases of individual tests, so it's easier to follow what is happening here?

packages/core/src/ceramic.ts
```diff
@@ -27,11 +34,11 @@ import { PathTrie, TrieNode, promiseTimeout } from './utils'
 
 import EthereumAnchorService from "./anchor/ethereum/ethereum-anchor-service"
 import InMemoryAnchorService from "./anchor/memory/in-memory-anchor-service"
-import { PinningBackendStatic } from "@ceramicnetwork/common"
-
 
 import { randomUint32 } from '@stablelib/random'
 
+const DEFAULT_BASE_DOC_CACHE_LIMIT = 500; // number of base docs stored in the cache
```
stbrody, line 40: Can we call this `DEFAULT_DOC_CACHE_LIMIT` (dropping the word "base") since it applies to both the base and commit cache?

packages/core/src/ceramic.ts
simonovic86: created a [740](https://github.com/ceramicnetwork/js-ceramic/issues/740) story for that

packages/common/src/doc-cache.ts
simonovic86: it means that async invocation has to be handled (`.catch()`, `.then()`). This is removed now

packages/core/src/ceramic.ts
simonovic86: the PR was created before the call. It is removed now

packages/core/src/__tests__/ceramic.test.ts
simonovic86: refactored and added more tests

packages/http-client/src/ceramic-http-client.ts
```diff
@@ -63,11 +66,11 @@ export default class CeramicClient implements CeramicApi {
...
   public readonly _doctypeHandlers: Record<string, DoctypeHandler<Doctype>>
 
-  constructor (apiHost: string = CERAMIC_HOST, config?: CeramicClientConfig) {
-    this._config = Object.assign(DEFAULT_CLIENT_CONFIG, config ? config : {})
+  constructor (apiHost: string = CERAMIC_HOST, config: CeramicClientConfig = {}) {
+    this._config = { ...DEFAULT_CLIENT_CONFIG, ...config }
 
     this._apiUrl = combineURLs(apiHost, API_PATH)
-    this._docmap = {}
+    this._docCache = new DocCache(config.docCacheLimit, true)
```
stbrody, line 73: I think we should probably add `cacheDocCommits` to CeramicClientConfig so it can be configured on the http-client side as well.  I actually think the http client probably should not cache doc commits by default, unlike the daemon side.  It makes sense for the daemon to cache commits, since it'll be getting queries from the CAS for documents at specific commits, but the http client won't be, so we should only cache commits if the user explicitly asks us to do so.

stbrody (approved):
LGTM!

PR #725, opened 2020-12-30
File logging updates

valmack:
### Motivation

The way we log doc ids changed so we needed to update how we are parsing that when logging to files. Also we changed the way we parse logs in promtail and want to keep log formats consistent.

### Changes

- Use logfmt format when logging to files
- Update docid logging
- Remove file birthtime warning (not necessary and too much noise)


oed (approved):
lgtm!

simonovic86 (approved):
nice!

PR #726, opened 2020-12-30
chore(cli): Add documentation for how to use the http api endpoints

oed:
Adds documentation for the http-api endpoints.


docs-src/http-api.md
```diff
@@ -0,0 +1,530 @@
...
+    "anchorScheduledFor": "12/30/2020, 1:45:00 PM"
+  }
+}
+```
+
+
+
+# Records
+
+The records endpoint provides lower level access to the data structure of a Ceramic document. It is also the enpoint that has to be used in order to update a document, by adding a new record.
```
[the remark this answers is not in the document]
stbrody: a nitpick, but saying "has to" makes it sound like a negative.  "is" just indicates that this is the reality, without any judgement associated

docs-src/http-api.md
```diff
@@ -0,0 +1,530 @@
...
+      }
+    }
+  ]
+}
+```
+
+## Apply a record to a document
+*`⚠️ This method is not available if gateway mode is enabled.`*
+
+In order to modify a document we apply a record to its docment log. This record usually contains a signature over the *json-patch* of the document contents. The record also needs to contain pointers to the previous record and other metadata. You can read more about this in the [Ceramic Specification](https://github.com/ceramicnetwork/ceramic/blob/master/SPECIFICATION.md#document-records). Different document types may have different formats for their records. If you want to see an example implementation for how to construct a record you can have a look at the implementation of the [[TileDoctype]].
```
stbrody, line 294:
```suggestion
In order to modify a document we apply a record to its docment log. This record usually contains a signature over a *json-patch* diff describing a modification to the document contents. The record also needs to contain pointers to the previous record and other metadata. You can read more about this in the [Ceramic Specification](https://github.com/ceramicnetwork/ceramic/blob/master/SPECIFICATION.md#document-records). Different document types may have different formats for their records. If you want to see an example implementation for how to construct a record you can have a look at the implementation of the [[TileDoctype]].
```

docs-src/http-api.md
```diff
@@ -0,0 +1,530 @@
...
+
+### Request
+`POST /api/v0/records`
+
+Here, `:docid` should be replaced by the string representation of the DocID of the document that is being requested.
```
stbrody, line 299: This whole section looks like it was copy and pasted from elsewhere without being updated

docs-src/http-api.md
```diff
@@ -0,0 +1,530 @@
...
+
+The pins api endpoint can be used to manipulate the pin set. The pin set is all of the documents that a node maintains the state of. Any document opened by the node that is not pinned will eventually be garbage collected from the node.
+
+## Add a document to the pin set
+This method adds the document with the given [[DocID]] to the pin set. 
+
+*`⚠️ This method is not available if gateway mode is enabled.`*
+
+### Request
+`POST /api/v0/pins/:docid`
```
stbrody, line 389: huh, this does indeed match what we have in code, but is surprising to me.  Since this is a POST request not a GET, shouldn't the docid be encoded as an arg to the `pins` endpoint, rather than encoded in the URL?

CC @simonovic86

docs-src/http-api.md
```diff
@@ -0,0 +1,530 @@
...
+
+Here, `:docid` should be replaced by the string representation of the DocID of the document that is being requested.
+
+### Response
+
+* `docId` - the [[DocID]] of the document which was pinned, string
+* `isPinned` - whether the document was pinned, boolean - `true`
```
stbrody, line 396: another comment less about the docs than about the actual impl, but isn't this field redundant and unnecessary?  It can never come back as `false` currently, because if there's a problem we return a non-ok http status response.

@simonovic86

docs-src/http-api.md
```diff
@@ -0,0 +1,530 @@
...
+}
+```
+
+## Remove a document from the pin set
+This method removes the document with the given [[DocID]] from the pin set.
+
+*`⚠️ This method is not available if gateway mode is enabled.`*
+
+### Request
+`DELETE /api/v0/pins/:docid`
```
stbrody, line 415: same comment as above about encoding the docid in the request vs in the url

docs-src/http-api.md
```diff
@@ -0,0 +1,530 @@
...
+
+Here, `:docid` should be replaced by the string representation of the DocID of the document that is being requested.
+
+### Response
+
+* `docId` - the [[DocID]] of the document which was unpinned, string
+* `isPinned` - whether the document was pinned, boolean - `false`
```
stbrody, line 422: this doesn't actually return whether the document was pinned previously.  It is just always `false` if the unpin command was successful

docs-src/http-api.md
```diff
@@ -0,0 +1,530 @@
...
+
+### Request
+
+`GET /api/v0/pins/:docid`
+
+Here, `:docid` should be replaced by the string representation of the DocID of the document that is being requested.
+
+### Response
+
+* `pinnedDocIds` - an array of [[DocID]] strings that are in the pin set
```
stbrody, line 446:
```suggestion
* `pinnedDocIds` - an array containing the specified [[DocID]] string if that document is pinned, or an empty array if that document is not pinned
```

docs-src/http-api.md
```diff
@@ -0,0 +1,530 @@
...
+{
+  "docId": "k2t6wyfsu4pg2qvoorchoj23e8hf3eiis4w7bucllxkmlk91sjgluuag5syphl",
+  "isPinned": false
+}
+```
+
+## Check if a document is in the pin set
+This method is used to check if a particular document is in the pin set.
+
+### Request
```
stbrody, line 438: a nitpick, but I find the flow of this document easier to read if the "list all documents in the pin set" section comes before the "check if document is in the pin set" section

docs-src/http-api.md
```diff
@@ -0,0 +1,530 @@
...
+```bash
+curl http://localhost:7007/api/v0/pins/k2t6wyfsu4pg2qvoorchoj23e8hf3eiis4w7bucllxkmlk91sjgluuag5syphl -X DELETE
+
+{
+  "docId": "k2t6wyfsu4pg2qvoorchoj23e8hf3eiis4w7bucllxkmlk91sjgluuag5syphl",
+  "isPinned": false
+}
+```
+
+## Check if a document is in the pin set
```
stbrody, line 435:
```suggestion
## Check if a specific document is in the pin set
```

stbrody (commented):
Nice!  Lots of little comments, but overall looks pretty good

docs-src/http-api.md
```diff
@@ -0,0 +1,530 @@
...
+      "cid": "bagcqceramof2xi7kh6qblirzkbc7yulcjcticlcob6uvdrx3bexgks37ilva",
+      "type": 0
+    }],
+    "anchorScheduledFor": "12/15/2020, 2:45:00 PM"
+  }
+}
+```
+
+## Create a document from genesis record
+Create a new document, or load a document from it's genesis content (these are equivalent actions). The genesis content may be signed (a DagJWS for the `tile` doctype), or unsigned in some cases.
```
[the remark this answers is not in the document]
oed: > If we had a way to get a CID from some content without actually publishing that content to IPFS

This is possible (technically), don't see how that changes this statement though?

docs-src/http-api.md
oed: Yeah, Agree that it could be removed.

docs-src/http-api.md
oed: good point.

oed:
Thanks for your detailed feedback @stbrody.
I agree that some of the stuff in the api are not optimal. But don't want to change them now as we are in a feature freeze.

docs-src/http-api.md
```diff
@@ -0,0 +1,530 @@
+The Ceramic HTTP API allows you to create and update documents on a remote Ceramic node. These endpoints allows you to manually make requests to a Ceramic node. This is also a good place to start if you want to build a Ceramic client in a new language. 
+
+*`ℹ️ Not all of these methods are exposed if the node runs in gateway mode.`*
+
+# Documents
+The documents endpoint is used to create new documents, load documents from their [[DocID]] or from their genesis content.
```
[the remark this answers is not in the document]
oed: fixed.

stbrody:
Woah, that test failure is one I haven't seen before...

docs-src/http-api.md
```diff
@@ -0,0 +1,527 @@
+The Ceramic HTTP API allows you to create and update documents on a remote Ceramic node. These endpoints allows you to manually make requests to a Ceramic node. This is also a good place to start if you want to build a Ceramic client in a new language.
```
stbrody, line 1:
```suggestion
The Ceramic HTTP API allows you to create, update, and query documents on a remote Ceramic node. These endpoints allows you to manually make requests to a Ceramic node. This is also a good place to start if you want to build a Ceramic client in a new language. 
```

docs-src/http-api.md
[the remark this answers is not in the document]
stbrody: Then they wouldn't be equivalent actions anymore.  It would be possible to load a document from its genesis contents without creating the document if it doesn't exist

docs-src/http-api.md
```diff
@@ -0,0 +1,527 @@
...
+In order to modify a document we apply a record to its docment log. This record usually contains a signature over a *json-patch* diff describing a modification to the document contents. The record also needs to contain pointers to the previous record and other metadata. You can read more about this in the [Ceramic Specification](https://github.com/ceramicnetwork/ceramic/blob/master/SPECIFICATION.md#document-records). Different document types may have different formats for their records. If you want to see an example implementation for how to construct a record you can have a look at the implementation of the [[TileDoctype]].
+
+### Request
+`POST /api/v0/records`
+
+The request body should contain the following fields:
+
+- `docId` - the name of the doctype to use, string
+- `record` - the genesis content of the document (will differ per doctype)
+- `docOpts` - options for the document creation [[DocOpts]]
```
stbrody, line 303:
```suggestion
- `docOpts` - options for the document update [[DocOpts]] (optional)
```

docs-src/http-api.md
```diff
@@ -0,0 +1,527 @@
...
+
+In order to modify a document we apply a record to its docment log. This record usually contains a signature over a *json-patch* diff describing a modification to the document contents. The record also needs to contain pointers to the previous record and other metadata. You can read more about this in the [Ceramic Specification](https://github.com/ceramicnetwork/ceramic/blob/master/SPECIFICATION.md#document-records). Different document types may have different formats for their records. If you want to see an example implementation for how to construct a record you can have a look at the implementation of the [[TileDoctype]].
+
+### Request
+`POST /api/v0/records`
+
+The request body should contain the following fields:
+
+- `docId` - the name of the doctype to use, string
+- `record` - the genesis content of the document (will differ per doctype)
```
stbrody, line 302:
```suggestion
- `record` - the content of the record to apply (will differ per doctype)
```

docs-src/http-api.md
stbrody: let's just remove the mention of `isPinned` from the docs so we can remove it later.  We should also add a sentence to clarify that an OK http response indicates the pin/unpin operation was successful

oed:
@stbrody I think this is ready for merging unless you have other comments

stbrody (approved):

PR #728, opened 2020-12-30
fix(core): Cannot call ipfs.block.stat on an IPLD path

stbrody:


packages/core/src/document.ts
```diff
@@ -555,29 +555,29 @@ class Document extends EventEmitter {
    * @private
    */
   async _verifyAnchorRecord (record: AnchorRecord): Promise<AnchorProof> {
-    const proofRecord =  await this.dispatcher.retrieveRecord(record.proof)
+    const proofCID = record.proof
+    const proof =  await this.dispatcher.retrieveFromIPFS(proofCID)
 
-    let prevRootPathRecord
+    let prevCIDViaMerkleTree
```
stbrody, line 561: I struggled to understand what this function was doing at first.  This name is slightly more clear in my head at least, but definitely let me know if you disagree and think either the original name is clearer, or have a suggestion for some other name that would be clearer

oed (approved):
lgtm, once test pass :)

PR #735, opened 2020-12-31
Hotfix pubsub subscribe

valmack:
### Motivation

Our ipfs node on infra is going down periodically (reason tbd) and when it reboots we stop receiving pubsub messages.
Created an issue on js-ipfs repo here https://github.com/ipfs/js-ipfs/issues/3465 -- the error handler for pubsub is broken in Node.js

### Changes

- Check if we are subscribed every 60 seconds and resubscribe if we are not
- Also updated infra https://github.com/3box/ceramic-infra/commit/164ba63a21b82d5aba6da27dee457079930eb292 to not subscribe from the IPFS API node itself because it was causing this issue https://github.com/ipfs/js-ipfs/issues/3467
- Small change to check that document context includes `anchorService` before removing listeners (which was breaking tests)

### DIFF
--- a/packages/core/src/__tests__/dispatcher.test.ts
+++ b/packages/core/src/__tests__/dispatcher.test.ts
@@ -4,6 +4,7 @@ import Document from "../document"
 import { TileDoctype } from "@ceramicnetwork/doctype-tile"
 import DocID from "@ceramicnetwork/docid";
 
+const TOPIC = '/ceramic'
 const FAKE_CID = new CID('bafybeig6xv5nwphfmvcnektpnojts33jqcuam7bmye2pb54adnrtccjlsu')
 const FAKE_CID2 = new CID('bafybeig6xv5nwphfmvcnektpnojts44jqcuam7bmye2pb54adnrtccjlsu')
 const FAKE_DOC_ID = "kjzl6cwe1jw147dvq16zluojmraqvwdmbh61dx9e0c59i344lcrsgqfohexp60s"
@@ -12,6 +13,10 @@ const ipfs = {
   pubsub: {
     subscribe: jest.fn(),
     unsubscribe: jest.fn(),
+    ls: jest.fn(() => new Promise((resolve, reject) => {
+        resolve([TOPIC])
+        reject()
+    })),
     publish: jest.fn()
   },
   dag: {
@@ -23,7 +28,6 @@ const ipfs = {
   },
   id: (): any => ({ id: 'ipfsid' })
 }
-const TOPIC = '/ceramic'
 
 class TileDoctypeMock extends TileDoctype {
   get doctype() {
--- a/packages/core/src/__tests__/dispatcher.test.ts
+++ b/packages/core/src/__tests__/dispatcher.test.ts
@@ -13,9 +13,6 @@ const ipfs = {
   pubsub: {
     subscribe: jest.fn(),
     unsubscribe: jest.fn(),
-    ls: jest.fn(() => new Promise((resolve) => {
-        resolve([TOPIC])
-    })),
     publish: jest.fn()
   },
   dag: {
--- a/packages/core/src/__tests__/dispatcher.test.ts
+++ b/packages/core/src/__tests__/dispatcher.test.ts
@@ -13,9 +13,8 @@ const ipfs = {
   pubsub: {
     subscribe: jest.fn(),
     unsubscribe: jest.fn(),
-    ls: jest.fn(() => new Promise((resolve, reject) => {
+    ls: jest.fn(() => new Promise((resolve) => {
         resolve([TOPIC])
-        reject()
     })),
     publish: jest.fn()
   },
--- a/packages/core/src/__tests__/dispatcher.test.ts
+++ b/packages/core/src/__tests__/dispatcher.test.ts
@@ -60,7 +60,7 @@ describe('Dispatcher', () => {
 
   it('is constructed correctly', async () => {
     expect(dispatcher._documents).toEqual({})
-    expect(ipfs.pubsub.subscribe).toHaveBeenCalledWith(TOPIC, expect.anything())
+    expect(ipfs.pubsub.subscribe).toHaveBeenCalledWith(TOPIC, expect.anything(), expect.anything())
   })
 
   it('closes correctly', async () => {
--- a/packages/core/src/__tests__/dispatcher.test.ts
+++ b/packages/core/src/__tests__/dispatcher.test.ts
@@ -50,6 +50,10 @@ describe('Dispatcher', () => {
     await dispatcher.init()
   })
 
+  afterEach(async () => {
+    await dispatcher.close()
+  })
+
   it('is constructed correctly', async () => {
     expect(dispatcher._documents).toEqual({})
     expect(ipfs.pubsub.subscribe).toHaveBeenCalledWith(TOPIC, expect.anything())
@@ -62,8 +66,16 @@ describe('Dispatcher', () => {
   })
 
   it('makes registration correctly', async () => {
-    const doc = new Document(DocID.fromString(FAKE_DOC_ID), dispatcher, null)
-    doc._doctype = new TileDoctypeMock()
+    const doc = new Document(
+      DocID.fromString(FAKE_DOC_ID),
+      dispatcher,
+      null,
+      false,
+      {},
+      null,
+      null
+    )
+    doc['_doctype'] = new TileDoctypeMock(null, {})
     await dispatcher.register(doc)
 
     const publishArgs = ipfs.pubsub.publish.mock.calls[0]
@@ -94,8 +106,16 @@ describe('Dispatcher', () => {
   })
 
   it('handle message correctly', async () => {
-    const doc = new Document(DocID.fromString(FAKE_DOC_ID), dispatcher, null)
-    doc._doctype = new TileDoctypeMock()
+    const doc = new Document(
+      DocID.fromString(FAKE_DOC_ID),
+      dispatcher,
+      null,
+      false,
+      {},
+      null,
+      null
+    )
+    doc['_doctype'] = new TileDoctypeMock(null, {})
     await dispatcher.register(doc)
 
     // Store the query ID sent when the doc is registered so we can use it as the response ID later
--- a/packages/core/src/dispatcher.ts
+++ b/packages/core/src/dispatcher.ts
@@ -65,7 +65,7 @@ export default class Dispatcher extends EventEmitter {
   }
 
   async _subscribe(): Promise<void> {
-    let alreadySubscribed = this._isSubscribed
+    const alreadySubscribed = this._isSubscribed
 
     try {
       !alreadySubscribed && await this._ipfs.pubsub.subscribe(this.topic, this.handleMessage.bind(this))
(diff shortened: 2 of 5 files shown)


valmack:
This is actually breaking a lot of tests.. 
I think (1) because `pubsub.ls` is not mocked and (2) because it creates a new Node.js Timer with `setInterval` and requires `dispatcher.close` to be called to clean it up.

Because this is a hotfix anyway I'll see if I can modify some of the code to avoid a lot of test refactoring for now

packages/core/src/dispatcher.ts
```diff
@@ -58,8 +60,75 @@ export default class Dispatcher extends EventEmitter {
...
+  }
+
+  /**
+   * Subscribes IPFS pubsub to `this.topic` and logs an `subscribe` event.
+   */
+  async _subscribe(): Promise<void> {
+    if (this._isSubscribed) {
+      console.log(`Pubsub subscribe skipped. Already subscribed to topic ${this.topic}`)
+    } else {
+      const requestTimeout = 1000 * 5 // five seconds
```
oed, line 74: Can we make this a global constant?

packages/core/src/dispatcher.ts
```diff
@@ -58,8 +60,75 @@ export default class Dispatcher extends EventEmitter {
...
+  /**
+   * Sets `this._iSubscribed` if the `this.topic` is listed as an IPFS pubsub
+   * subscription.
+   * @returns A tuple of `this._isSubscribed` and an error if one was caught.
+   */
+  async _confirmIsSubscribed(): Promise<{isSubscribed: boolean, error: Error | null}> {
+    let isSubscribed = true
+    let error = null
+
+    const requestTimeout = 1000 * 5 // five seconds
```
oed, line 118: Use same global constant here.

packages/core/src/dispatcher.ts
```diff
@@ -58,8 +60,75 @@ export default class Dispatcher extends EventEmitter {
...
+      }
+    }
+  }
+
+  /**
+   * Periodically checks that IPFS pubsub is subscribed to `this.topic` and
+   * attempts to subscribe if not.
+   */
+  _resubscribeOnDisconnect(): void {
+    const intervalDelay = 1000 * 60 // one minute
```
oed, line 101: This should also be a global constant.

oed (commented):
Few comments.

packages/core/src/__tests__/dispatcher.test.ts
```diff
@@ -12,6 +13,10 @@ const ipfs = {
   pubsub: {
     subscribe: jest.fn(),
     unsubscribe: jest.fn(),
+    ls: jest.fn(() => new Promise((resolve, reject) => {
+        resolve([TOPIC])
+        reject()
```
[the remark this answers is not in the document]
valmack: doesn't need reject. but will remove this anyway

packages/core/src/dispatcher.ts
valmack: yup

packages/core/src/dispatcher.ts
```diff
@@ -58,8 +62,70 @@ export default class Dispatcher extends EventEmitter {
    */
   async init(): Promise<void> {
     this._peerId = this._peerId || (await this._ipfs.id()).id
-    await this._ipfs.pubsub.subscribe(this.topic, this.handleMessage.bind(this))
-    this._log({ peer: this._peerId, event: 'subscribed', topic: this.topic })
+    await this._subscribe()
+    !TESTING && this._resubscribeOnDisconnect()
```
stbrody, line 66: Can we just make this an if-statement?  Just eliminates some cognitive load in interpreting these inline expressions IMO

packages/core/src/dispatcher.ts
```diff
@@ -58,8 +62,70 @@ export default class Dispatcher extends EventEmitter {
...
+  }
+
+  /**
+   * Subscribes IPFS pubsub to `this.topic` and logs an `subscribe` event.
+   */
+  async _subscribe(): Promise<void> {
+    if (this._isSubscribed) {
+      console.log(`Pubsub subscribe skipped. Already subscribed to topic ${this.topic}`)
```
stbrody, line 74: If we do an early return here we can avoid the `else` block and extra indentation layer below

packages/core/src/dispatcher.ts
```diff
@@ -58,8 +62,70 @@ export default class Dispatcher extends EventEmitter {
...
+    } else {
+      try {
+        await this._ipfs.pubsub.subscribe(
+          this.topic,
+          this.handleMessage.bind(this),
+          {timeout: !TESTING && IPFS_GET_TIMEOUT}
```
stbrody, line 80:
```suggestion
          {timeout: TESTING ? null : IPFS_GET_TIMEOUT}
```

packages/core/src/dispatcher.ts
```diff
@@ -58,8 +62,70 @@ export default class Dispatcher extends EventEmitter {
...
+        const { isSubscribed, error } = await this._confirmIsSubscribed()
+        if (isSubscribed) {
+          this._log({peer: this._peerId, event: 'subscribed', topic: this.topic })
+        } else if (error) {
+          throw error
+        } else {
+          throw new Error(`Pubsub subscribe failed for topic ${this.topic}`)
+        }
+      } catch (error) {
+        console.error(error)
```
stbrody, line 92: Can you add a comment `// TODO use logger`?  I've already left a few TODOs throughout the codebase with places to log once we have the logging framework in place, so that'll make sure I remember to include this in my logging pass

packages/core/src/dispatcher.ts
```diff
@@ -58,8 +62,70 @@ export default class Dispatcher extends EventEmitter {
...
+   * Sets `this._iSubscribed` if the `this.topic` is listed as an IPFS pubsub
+   * subscription.
+   * @returns A tuple of `this._isSubscribed` and an error if one was caught.
+   */
+  async _confirmIsSubscribed(): Promise<{isSubscribed: boolean, error: Error | null}> {
+    let isSubscribed = true
+    let error = null
+
+    try {
+      const subscriptions = await this._ipfs.pubsub.ls({ timeout: IPFS_GET_TIMEOUT })
```
stbrody, line 118: is there any harm to attempting to subscribe when you're already subscribed?  If not, then we could just always try to subscribe, and we could remove this entire method and keep the logic pretty simple

packages/core/src/__tests__/dispatcher.test.ts
```diff
@@ -62,8 +69,16 @@ describe('Dispatcher', () => {
   })
 
   it('makes registration correctly', async () => {
-    const doc = new Document(DocID.fromString(FAKE_DOC_ID), dispatcher, null)
-    doc._doctype = new TileDoctypeMock()
+    const doc = new Document(
```
stbrody, line 72: How did this PR affect this?  Why did this have to change?

stbrody (commented):
A few small nitpicks, but my main question is if we can avoid the logic that checks *if* we are subscribed and just always try to subscribe unconditionally.

packages/core/src/dispatcher.ts
```diff
@@ -12,6 +12,8 @@ import DocID from "@ceramicnetwork/docid";
 
 const IPFS_GET_TIMEOUT = 30000 // 30 seconds
 const IPFS_MAX_RECORD_SIZE = 256000 // 256 KB
+const IPFS_RESUBSCRIBE_INTERVAL_DELAY = 1000 * 60 // 1 minute
+const TESTING = process.env.NODE_ENV == 'test'
```
oed, line 16: Is this set automatically by jest?

packages/core/src/dispatcher.ts
oed: I think it might throw an error saying that the particular listener is already added, so could be tested. But the way it's done now seems fine imo, it will only be one more call if we're not subscribed.

packages/core/src/dispatcher.ts
valmack: yes

packages/core/src/__tests__/dispatcher.test.ts
valmack: Throws a bunch of console errors

packages/core/src/dispatcher.ts
valmack: Sure

valmack:
@stbrody yeah as joel mentioned it throws an error that you are already subscribed. But agree it's probably worth just doing that in case the `ls` returns unexpected results (which it does in certain conditions).

packages/core/src/dispatcher.ts
```diff
@@ -45,6 +47,8 @@ export default class Dispatcher extends EventEmitter {
 
   private logger: Logger
   private _isRunning = true
+  private _messageHandler = this.handleMessage.bind(this)
```
oed, line 50: What is `this` here? Seems like it's outside the scope of the class instance.

oed (commented):
Looks good, just one Q

packages/core/src/dispatcher.ts
valmack: just check and it does seem this is properly scoped to the instance. i.e. state set in the constructor is accessible when `handleMessage` gets called.

but I'd probably prefer something even simpler which would be to remove this in favor of an arrow function which achieves the same thing

```js
handleMessage = async (message: any): Promise<void> => {
```

packages/core/src/dispatcher.ts
```diff
@@ -63,71 +63,39 @@ export default class Dispatcher extends EventEmitter {
...
+    try {
+      await this._ipfs.pubsub.subscribe(
+        this.topic,
+        this._messageHandler,
+        {timeout: TESTING ? null : IPFS_GET_TIMEOUT}
+      )
+      this._log({peer: this._peerId, event: 'subscribed', topic: this.topic })
+    } catch (error) {
+      // TODO: use logger
+      console.error(error)
```
stbrody, line 85: Can we check the error to see if it's an "already subscribed" error and not log it, or log it at the debug log level?  Otherwise users might get concerned seeing lots of errors in their console that are actually expected

valmack: don't want to tie logic to the error message in case it changes and there isn't a code produced. seems ok to do it this way for now and this is temporary anyway

stbrody (approved):
Nice!  This looks much cleaner.  LGTM mod one comment about making sure the console doesn't fill with errors when everything is actually working as expected

packages/core/src/dispatcher.ts
stbrody: do we know *how* temporary?  I'd rather depend on the error message and deal with it if the message changes and that causes errors to slip out, rather than definitely have spurious errors filling our logs and causing concern among our users during their very first interaction with running a ceramic node

valmack: can revisit this when we add the new logger

stbrody: how will the new logger help?  And why is tying logic to the error message so bad, other than being brittle?  I feel pretty strongly that we shouldn't launch Clay with the log filling with errors that aren't real errors.  I think that will look pretty poor to our community

stbrody: oh, I just saw that you did add an if-statement checking the log message, but we still log the error unconditionally.  I think the `console.error` line needs to be moved into an else-block

valmack: updated

PR #741, opened 2021-01-05
fix(core): respect pinned status on createDocument call

### LINKED ISSUE #717: Restarting ceramic daemon causes unpinned IDX docs with updates to lose their updates
- Create a ceramic doc (do not pin manually)
- Update ceramic doc (do not pin manually)
- Check state of doc (see that the next content is populated correctly)
- Shut down daemon
- Start up daemon
- Check state of doc (see that the next content is no longer populated).

Pinning the updated ceramic doc manually fixes this problem.

This populates to IDX (maybe IDX does pin updates?):

- IDX set a doc
- IDX get doc (it's there)
- restart daemon
- IDX get doc (`null`)

I do have the `autopin` option set to `true` when instantiating IDX instances (I believe this is default behavior anyways).

oed:
This fixes #717, namely that when loading a document from the createDocument method it's pinned state was not being respected.

### DIFF
--- a/packages/core/src/__tests__/ceramic-pinning.test.ts
+++ b/packages/core/src/__tests__/ceramic-pinning.test.ts
@@ -1,0 +1,158 @@
+import IPFS from 'ipfs-core'
+import Ceramic from '../ceramic'
+import { Ed25519Provider } from 'key-did-provider-ed25519'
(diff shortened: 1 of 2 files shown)


simonovic86 (approved):
lgtm!

PR #744, opened 2021-01-06
CORS

valmack:
### Motivation

We want the ability to set CORS headers so that only clients on whitelisted domains can make requests to the Ceramic api.

### Changes

- Added a flag `--cors-allowed-origins` which accepts a space separated list of values to use as a whitelist
  - e.g. `".3boxlabs.com$ ://3boxlabs.com$"` will whitelist all subdomains and the base domain for 3boxlabs.com

### DIFF
--- a/packages/cli/src/bin/ceramic.ts
+++ b/packages/cli/src/bin/ceramic.ts
@@ -18,7 +18,7 @@ program
     .option('--network <name>', 'Name of the ceramic network to connect to. One of: "mainnet", "testnet-clay", "local", or "inmemory". Defaults to "testnet-clay"')
     .option('--max-healthy-cpu <decimal>', 'Fraction of total CPU usage considered healthy. Defaults to 0.7')
     .option('--max-healthy-memory <decimal>', 'Fraction of total memory usage considered healthy. Defaults to 0.7')
-    .option('--cors-allowed-origins <array>', 'List of strings and/or regex expressions to set for Access-Control-Allow-Origin . Defaults to all: ["*"]')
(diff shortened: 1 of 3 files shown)


packages/cli/src/bin/ceramic.ts
```diff
@@ -18,6 +18,7 @@ program
     .option('--network <name>', 'Name of the ceramic network to connect to. One of: "mainnet", "testnet-clay", "local", or "inmemory". Defaults to "testnet-clay"')
     .option('--max-healthy-cpu <decimal>', 'Fraction of total CPU usage considered healthy. Defaults to 0.7')
     .option('--max-healthy-memory <decimal>', 'Fraction of total memory usage considered healthy. Defaults to 0.7')
+    .option('--cors-allowed-origins <list>', 'Space-separated list of strings and/or regex expressions to set for Access-Control-Allow-Origin . Defaults to all: "*"')
```
oed, line 21: Can we make this comma separated instead? I think that's more clear.

oed (commented):
Looks good, just one nitpick

packages/cli/src/bin/ceramic.ts
valmack: Hmm I actually feel like space separated is more clear/simple because there is only one way to do it, versus using commas you can have more than one delimiter (`"foo,bar"` or `"foo, bar"`). Also easier to parse in code so less prone to parsing error.

Not huge either way but I'd probably prefer space separated lists as much as possible.

oed (approved):

PR #750, opened 2021-01-06
fix: log api errors with status and message

valmack:
### Motivation

Status code in the logs were always 200 when the actual response sent back a 5xx.

### Changes

- Send response first then produce the http log so that `res.statusCode` and `res.statusMessage` are accurate
- Also only log usage stats when healthcheck fails to reduce noise of healthcheck logs

### DIFF
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -95,12 +95,15 @@ class CeramicDaemon {
       })
     }
 
+    // next is required in function signature
     app.use((err: Error, req: Request, res: Response, next: NextFunction): void => {
-      const requestStart = Date.now()
-      const httpLog = this._buildHttpLog(requestStart, req, res)
-      const logString = JSON.stringify(httpLog)
-      this.logger.error(logString)
-      next(err)
+      this.logger.error(err)
+      if (res.statusCode < 300) { // 2xx indicates error has not yet been handled
+        res.status(500)
+      }
+      res.send(err.message)
+      // TODO: Get real request start
+      this.logger.error(JSON.stringify(this._buildHttpLog(Date.now(), req, res)))
     })
 
     const port = opts.port || DEFAULT_PORT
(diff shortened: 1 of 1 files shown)


packages/cli/src/ceramic-daemon.ts
```diff
@@ -212,12 +215,12 @@ class CeramicDaemon {
     const stats = `maxHealthyCpu=${this.maxHealthyCpu} cpuUsage=${cpuUsage} freeCpu=${freeCpu} maxHealthyMemory=${this.maxHealthyMemory} memoryUsage=${memUsage} freeMemory=${freeMemory}`
 
     if (cpuUsage > this.maxHealthyCpu || memUsage > this.maxHealthyMemory) {
-      this.logger.error('Ceramic failed a healthcheck')
-      res.status(503).send('Insufficient resources')
+      this.logger.debug(stats)
```
stbrody, line 218: hmm... it still feels useful to log an error if we actually fail a healthcheck I think?  I think this should probably still be logger.error, with a message that we failed a healthcheck, *and* with the stats included.

packages/cli/src/ceramic-daemon.ts
```diff
@@ -95,12 +95,15 @@ class CeramicDaemon {
...
+    // next is required in function signature
     app.use((err: Error, req: Request, res: Response, next: NextFunction): void => {
-      const requestStart = Date.now()
-      const httpLog = this._buildHttpLog(requestStart, req, res)
-      const logString = JSON.stringify(httpLog)
-      this.logger.error(logString)
-      next(err)
+      this.logger.error(err)
+      if (res.statusCode < 300) { // 2xx indicates error has not yet been handled
+        res.status(500)
```
stbrody, line 102: I guess I don't really understand enough about how HTTP status codes are used.  Why wouldn't we just use `res.statusCode` here?  Aren't we losing information by not using the statusCode and just always using 500?

stbrody (commented):
I think this review may be too far out of area of expertise for me to give meaningful input.  I left a comment about a log message, but I think I'm going to defer to Joel to review the changes to the actual HTTP handling

packages/cli/src/ceramic-daemon.ts
oed: I assume this handler will only get called if there is an error elsewhere. Then if the status code has already been set to 200 we should change it to 500 in order to reflect that there was an error.

oed (commented):
This seems good to me if my assumption below is correct :)

packages/cli/src/ceramic-daemon.ts
valmack: Yes exactly @oed. Also @stbrody this is where the "Ceramic failed a healthcheck" will get logged

packages/cli/src/ceramic-daemon.ts
valmack: Other comment addresses this. so yes the message will get logged as an error and the stats will get logged as debug.

PR #752, opened 2021-01-07
fix(cli,http-client): Fix pin API in CLI and http client

stbrody:
Not sure if the commit name works for our semantic versioning system.  How are we supposed to handle a fix to multiple packages?  @oed

### DIFF
--- a/packages/http-client/src/ceramic-http-client.ts
+++ b/packages/http-client/src/ceramic-http-client.ts
@@ -96,7 +96,8 @@ export default class CeramicClient implements CeramicApi {
   _initPinApi(): PinApi {
     return {
       add: async (docId: DocID): Promise<void> => {
-        const doc = await fetchJson(this._apiUrl + '/pins' + `/${docId.toString()}`, { method: 'post' })
+        await fetchJson(this._apiUrl + '/pins' + `/${docId.toString()}`, { method: 'post' })
+        const doc = await this.loadDocument(docId)
         this._docCache.pin(doc)
       },
       rm: async (docId: DocID): Promise<void> => {
@@ -105,7 +106,7 @@ export default class CeramicClient implements CeramicApi {
       },
       ls: async (docId?: DocID): Promise<AsyncIterable<string>> => {
         let url = this._apiUrl + '/pins'
-        if (docId !== undefined) {
+        if (docId) {
           url += `/${docId.toString()}`
         }
         const result = await fetchJson(url)
--- a/packages/cli/src/__tests__/ceramic-daemon.test.ts
+++ b/packages/cli/src/__tests__/ceramic-daemon.test.ts
@@ -81,7 +81,8 @@ describe('Ceramic interop: core <> http-client', () => {
         // performed yet by the time the test checks.  To eliminate this race condition we should set
         // anchorOnRequest to false in the config for the InMemoryAnchorService and anchor manually
         // throughout the tests.
-        core = await Ceramic.create(ipfs, {pubsubTopic: topic})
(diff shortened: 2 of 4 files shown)


packages/http-client/src/ceramic-http-client.ts
```diff
@@ -96,7 +96,8 @@ export default class CeramicClient implements CeramicApi {
   _initPinApi(): PinApi {
     return {
       add: async (docId: DocID): Promise<void> => {
-        const doc = await fetchJson(this._apiUrl + '/pins' + `/${docId.toString()}`, { method: 'post' })
+        await fetchJson(this._apiUrl + '/pins' + `/${docId.toString()}`, { method: 'post' })
+        const doc = await this.loadDocument(docId)
         this._docCache.pin(doc)
```
oed, line 101: Don't think we should fetch the document here.

oed (commented):
Not sure how to properly name the commit. Tried digging a bit (I think this is what we are using: https://github.com/atlassian/lerna-semantic-release) but couldn't figure it out.

Just one small change that I'd make.

packages/http-client/src/ceramic-http-client.ts
stbrody: If we don't then there's no way to pin documents to the local cache in the http client.  We're going to have to rethink all of this with the doc state refactor, but right now the http client is very much built around the `_docCache`, and this is the only way to get a document to be pinned within that cache

oed: What's the benefit of having the document in the cache here though? When you call loadDocument later, it's going to make a http request for the new state anyway.

stbrody: Well, we could argue whether the HTTP client should have a docCache at all, given that we refetch the state from the daemon every time the document is loaded, regardless of whether the doc is in the cache or not.
The main use of the docCache in the http client then is so that handed out Documents can get updated in the background when new updates come in.  Of course, that only happens if polling is enabled, which is off by default.  So maybe we should only use the docCache if polling is enabled?

In either case, this is tangential to this review.  As long as we have the docCache in place, we should use and maintain it in the way it was intended.  If we want to remove or rethink the role of the docCache in the HTTP api we can do so, but should do so explicitly and under a separate ticket.

oed: I just don't get why we make another network request here, when the dev will have to make another one when they call loadDocument anyway.

stbrody: I guess we could try to do a larger change to make the http client "remember" that a document is pinned by its docId, but without actually putting the document into the pin cache until the first time the document is actually loaded.  But again, that seems like a larger change that should be handled separately.  Right now, if we don't do this load, then the local pin cache is broken and useless.  So either we should say that the ability to pin documents into the local cache isn't important and we should remove it outright, or we should make it work and think about future improvements to make it more efficient.  Just removing this one line but otherwise doing nothing else leaves us in a weird in-between state where we have this cache, and a bunch of logic around maintaining the cache, but none of it actually does anything.  We need to pick a direction one way or the other.

stbrody: Okay, I went ahead and filed [a story](https://github.com/ceramicnetwork/js-ceramic/issues/761) and put up [a PR](https://github.com/ceramicnetwork/js-ceramic/pull/762] to remove the ability to pin documents to the local cache, that builds on top of the changes from this PR.  That is the smallest possible unit of work that removes the extraneous document load.

oed (approved):

PR #758, opened 2021-01-10
Doctype/handler modularisation

### LINKED ISSUE #742: Separate doctype packages into handlers and doctype classes
As discussed on core devs call. Currently we have two main classes within the doctype packages. The Doctype itself, and the doctype-handler. We want to separate these into two different packages in order to minimize the package size of the http-client.

ukstv:


packages/core/package.json
```diff
@@ -32,12 +32,17 @@
...
     "@ceramicnetwork/doctype-tile": "^0.12.8",
+    "@ceramicnetwork/doctype-tile-handler": "^0.12.8",
     "@ceramicnetwork/pinning-aggregation": "^0.1.9",
     "@ceramicnetwork/pinning-ipfs-backend": "^0.1.9",
     "@ethersproject/base64": "^5.0.2",
     "@ethersproject/providers": "^5.0.0-beta.144",
     "@ethersproject/random": "^5.0.2",
+    "@stablelib/random": "^1.0.0",
+    "@stablelib/sha256": "^1.0.0",
+    "ipld-dag-cbor": "^0.17.0",
```
oed, line 45: Why were these added? I don't see them being used anywhere in your code changes?

packages/doctype-tile-handler/src/tile-doctype-handler.ts
```diff
@@ -1,13 +1,13 @@
-import CID from 'cids'
+import type CID from 'cids'
 
 import * as didJwt from 'did-jwt'
-import base64url from 'base64url'
-import { PublicKey } from 'did-resolver'
+import * as uint8arrays from 'uint8arrays'
```
oed, line 4: Thanks for removing the `base64url` dep!

packages/http-client/src/ceramic-http-client.ts
```diff
@@ -198,15 +198,16 @@ export default class CeramicClient implements CeramicApi {
...
-  findDoctypeHandler<T extends Doctype>(doctype: string): DoctypeHandler<T> {
-    const doctypeHandler = this._doctypeHandlers[doctype]
-    if (doctypeHandler == null) {
-      throw new Error(`Failed to find doctype handler for doctype ${doctype}`)
+  findDoctypeConstructor<T extends Doctype>(doctype: string) {
+    const constructor = this._doctypeConstructors[doctype]
+    if (constructor) {
+      return constructor as DoctypeConstructor<T>
+    } else {
+      throw new Error(`Failed to find doctype constructor for doctype ${doctype}`)
```
oed, line 209: We can just remove this method as it's not being used.

tsconfig.json
```diff
@@ -32,6 +32,7 @@
    { "path": "./packages/common" },
    { "path": "./packages/doctype-caip10-link" },
    { "path": "./packages/doctype-tile" },
+   { "path": "./packages/doctype-tile-handler" },
```
oed, line 35: Should `doctype-caip10-link-handler` not be added here as well?

packages/cli/package.json
```diff
@@ -44,6 +44,7 @@
     "@ceramicnetwork/core": "^0.17.7",
     "@ceramicnetwork/docid": "^0.4.2",
     "@ceramicnetwork/doctype-tile": "^0.12.8",
+    "@ceramicnetwork/doctype-tile-handler": "^0.12.8",
```
oed, line 47: This should be a dev dep?

oed (commented):
Looks good so far. Left a few comments.

packages/core/package.json
ukstv: These are not new dependencies. These are old and used ones, but undeclared.

@stablelib/random: https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/core/src/ceramic.ts#L38
@stablelib/sha256: https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/core/src/dispatcher.ts#L6
ipld-dag-cbor: https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/core/src/dispatcher.ts#L4

Basically, all of the library consumers were at a mercy of npm flaws that allows one to use a dependency required by some unrelated package in this one.

packages/http-client/src/ceramic-http-client.ts
ukstv: But it is used :) See: https://github.com/ceramicnetwork/js-ceramic/pull/758/files#diff-1e1d2be8d1b9f18852db4000834507647dc77def6e5447c3b133610b44046312R131 and below.

tsconfig.json
ukstv: True.

packages/cli/package.json
ukstv: True.

packages/core/package.json
oed: ok, thanks. So it was just not included previously which you fixed

packages/http-client/src/ceramic-http-client.ts
oed: Right, but we can simply use a map instead. See https://github.com/ceramicnetwork/js-ceramic/pull/668

ukstv: I believe there is an implicit assumption here: one could _only_ have `tile` and `caip10-link` doctypes. Alternative formulation: it is _impossible_ to have other doctypes.

Using map/record is fine, as long as it always follows a happy path. If not, error message regarding null objects and absent methods could be quite hard to understand. This thing happens to construct meaningful error.

So, could you confirm, `caip10-link` and `tile` are the only expected doctypes, and the user have no right to add own doctypes?

oed: Ok, right might be a good idea to keep it because of the error being thrown. 
It's expected that users add their own doctyes.

packages/cli/package.json
```diff
@@ -78,7 +78,8 @@
     "eslint": "^6.8.0",
     "eslint-plugin-jest": "^23.8.2",
     "jest": "^26.6.3",
-    "tmp-promise": "^2.0.2"
+    "tmp-promise": "^2.0.2",
+    "@ceramicnetwork/doctype-tile-handler": "^0.12.8"
```
stbrody, line 82: nitpick: can we keep these in lexographic order?  So this line should be between the `@babel` deps and the `@types` deps

packages/core/package.json
```diff
@@ -55,6 +60,7 @@
     "typestub-multihashes": "^0.0.4"
   },
   "devDependencies": {
+    "uint8arrays": "^1.1.0",
```
stbrody, line 63: nit: keep dependencies in lexographic order

packages/core/src/__tests__/document.test.ts
```diff
@@ -84,7 +85,7 @@ jest.mock('../dispatcher', () => {
       retrieveRecord: jest.fn(cid => {
         return recs[cid.toString()]
       }),
-      retrieveFromIPFS: jest.fn((cid, path) => {
+      retrieveFromIPFS: jest.fn((cid) => {
```
stbrody, line 88: The function signature being mocked here does take two arguments, the second being `path`.  Since the mock implementation here doesn't actually use the second arg, maybe call it `pathUnused` or something like that?  But I think it makes sense to keep the second arg here so it matches the signature of the function being mocked

packages/doctype-caip10-link-handler/README.md
```diff
@@ -0,0 +1,15 @@
+# Ceramic Account Link doctype handler
+
+> This package contains the Caip-10 Link doctype handler implementation.
+
+### Installation
+```shell
+npm install @ceramicnetwork/doctype-caip10-link
```
stbrody, line 7:
```suggestion
npm install @ceramicnetwork/doctype-caip10-link-handler
```

packages/doctype-caip10-link-handler/package.json
```diff
@@ -0,0 +1,46 @@
...
+  "scripts": {
+    "test": "../../node_modules/.bin/jest --coverage",
+    "build": "../../node_modules/.bin/tsc -p tsconfig.json",
+    "prepublishOnly": "npm run build",
+    "prebuild": "npm run clean",
+    "lint": "../../node_modules/.bin/eslint ./src --ext .js,.jsx,.ts,.tsx",
+    "clean": "rm -rf ./lib"
+  },
+  "dependencies": {
+    "@ceramicnetwork/doctype-caip10-link": "^0.12.6",
```
stbrody, line 29: lexographic sort

packages/doctype-caip10-link-handler/src/__tests__/caip10-link-doctype-handler.test.ts
```diff
@@ -0,0 +1,264 @@
+import { Caip10LinkDoctypeHandler } from '../caip10-link-doctype-handler'
```
stbrody, line 1: I'm just assuming that this file is identical to the old version and this is just a file rename.  If anything actually changed in this file let me know and I can review that piece

packages/doctype-caip10-link/src/caip10-link-doctype.ts
```diff
@@ -8,7 +8,7 @@ import {
     CeramicRecord
 } from "@ceramicnetwork/common"
 
-const DOCTYPE = 'caip10-link'
+export const DOCTYPE = 'caip10-link'
```
stbrody, line 11: Now that we're exposing this, can we call it `DOCTYPE_NAME`?

packages/doctype-tile-handler/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "build": "../../node_modules/.bin/tsc -p tsconfig.json",
+    "prepublishOnly": "npm run build",
+    "prebuild": "npm run clean",
+    "lint": "../../node_modules/.bin/eslint ./src --ext .js,.jsx,.ts,.tsx",
+    "clean": "rm -rf ./lib"
+  },
+  "dependencies": {
+    "@ceramicnetwork/common": "^0.15.4",
+    "@ceramicnetwork/doctype-tile": "^0.12.8",
+    "uint8arrays": "^2.0.5",
```
stbrody, line 30: lexographic sort

packages/doctype-tile-handler/src/__tests__/three-id.test.ts
```diff
@@ -0,0 +1,314 @@
+import CID from 'cids'
```
stbrody, line 1: same as before, I'm not reviewing this file because I assume it's just code motion with no changes. If anything meaningful changed please LMK

packages/doctype-tile-handler/src/__tests__/tile-doctype.test.ts
```diff
@@ -0,0 +1,372 @@
+import CID from 'cids'
```
stbrody, line 1: same, if anything real changed, LMK

packages/doctype-tile/src/tile-doctype.ts
```diff
@@ -18,7 +18,7 @@ import {
     UnsignedRecord,
 } from "@ceramicnetwork/common"
 
-const DOCTYPE = 'tile'
+export const DOCTYPE = 'tile'
```
stbrody, line 21: ditto re: renaming this to DOCTYPE_NAME

stbrody (commented):
Looks great!  A few small comments sprinkled throughout, but they're all pretty small nitpicks.  Overall looking quite good

packages/cli/package.json
ukstv: This reminds me of all the recommendations and punishments from 10 years ago when people tried to enforce code style by human labor. There were no prettier or gofmt. State of the art IDE was effectively vim. Seems like a wasted effort here, honestly.

packages/core/src/__tests__/document.test.ts
ukstv: I'll just get it back.

packages/doctype-caip10-link-handler/src/__tests__/caip10-link-doctype-handler.test.ts
ukstv: Same.

packages/cli/package.json
oed: Hey, vim is great! 😅

ukstv: vim is awesome indeed.

stbrody (approved):
LGTM!  Thanks!

oed (approved):

PR #764, opened 2021-01-12
Record is now named Commit

ukstv:

PR #773, opened 2021-01-12
Record is now named Commit

ukstv:
Somehow failed to force push the branch in #764. So, here it is @oed @stbrody.


stbrody (approved):
I didn't actually re-read this whole thing, but approving under the assumption that it's the same as the earlier version that I already approved

oed (approved):

PR #775, opened 2021-01-12
fix(core): Encode network name into pin store directory name

stbrody:
This also changes the interpretation of the `stateStorePath` ceramic config option slightly.  Now if it is explicitly provided, rather than putting the leveldb data directly into the provided path, we will create always create a `.pinning.store.<network name>` directory within the provided `stateStorePath` directory, and put the leveldb state there.  This allows us to ensure that the data for different networks stays separate, even in the case of user error where the user restarts their ceramic on a different network but forgets to update the `stateStorePath`


packages/cli/src/bin/ceramic.ts
```diff
@@ -9,7 +9,7 @@ program
     .option('--anchor-service-api <url>', 'The anchor service URL to use')
     .option('--validate-docs', 'Validate documents according to their schemas. It is enabled by default')
     .option('--pinning <url...>', 'Pinning endpoints')
-    .option('--pinning-store-path <url>', `The directory path used for pinning service. Defaults to WORKING_DIR/${DEFAULT_PINNING_STORE_PATH}`)
+    .option('--pinning-store-path <url>', `The directory path used for pinning service. Defaults to WORKING_DIR`)
```
oed, line 12: Don't think this should say `<url>`. A path is not a uri afaik. Would also prefer to rename this to something like `pinset-directory`.

oed (changes requested):
I'd prefer if we do this differently. Right now it's a bit confusing that the config is stored in `~/.ceramic/config.json` and pinned documents in CWD.

Let's do something like this instead by default:

`~/.ceramic/pinset/<network-name>`

Now if the user configures a different pinset-directory there will always be specific network sub directories, `<user-path>/<network-name>`.

packages/core/src/ceramic.ts
```diff
@@ -178,15 +178,15 @@ class Ceramic implements CeramicApi {
   }
 
   private static async _generateNetworkOptions(config: CeramicConfig, anchorService: AnchorService): Promise<CeramicNetworkOptions> {
-    const networkName = config.networkName || DEFAULT_NETWORK
```
ukstv, line 181: Why mutation like this? It is effect could be negligible here, but it certainly introduces new risk down the evolution line.

Aha, if I understand correctly, this modifies `config` here to only pass proper network name to `PinStoreFactory`.

May I suggest instead of this to explicitly create a new `effectiveConfig` (or something like it) variable that sets all the defaults for the internal consumers like PinStoreFactory and Ceramic class itself? Being explicit here makes it clear what configuration is effectively used, where it is set and based on what.

For pin-store it makes sense to extend internal/effective config with `pinStorePath` property that is set to appropriate stateStorePath + suffix + network.

Data flow becomes more clear cut `configuration set by caller -> effective configuration -> consumers`.

Also, mea culpa, `process.cwd` should have never been used inside PinStoreFactory. It should silently accept whatever path it is constructed with.

stbrody:
Okay, new version is up and ready for review!

packages/cli/src/bin/ceramic.ts
```diff
@@ -1,15 +1,15 @@
...
 
 program
     .command('daemon')
     .option('--ipfs-api <url>', 'The ipfs http api to use')
     .option('--ethereum-rpc <url>', 'The Ethereum RPC URL used for communicating with Ethereum blockchain')
     .option('--anchor-service-api <url>', 'The anchor service URL to use')
     .option('--validate-docs', 'Validate documents according to their schemas. It is enabled by default')
-    .option('--pinning <url...>', 'Pinning endpoints')
-    .option('--pinning-store-path <url>', `The directory path used for pinning service. Defaults to WORKING_DIR/${DEFAULT_PINNING_STORE_PATH}`)
+    .option('--pinning-endpoints <url...>', 'Pinning endpoints')
```
oed, line 11: I think this should be a singular since you have to pass the flag multiple times to specify multiple endpoints.

packages/cli/src/bin/ceramic.ts
```diff
@@ -1,15 +1,15 @@
...
+    .option('--pinset-directory <string>', `The directory path used for pinning service. Defaults to HOME_DIR/.ceramic/pinning`)
```
oed, line 12:
```suggestion
    .option('--pinset-directory <string>', `The directory path used for pinning service. Defaults to HOME_DIR/.ceramic/pinset`)
```

oed (approved):
Looks good, just one typo

oed (approved):

PR #779, opened 2021-01-13
fix: add cache to 3idv0 resolver

oed:
Not urgent to review/merge.

### DIFF
--- a/packages/3id-did-resolver/src/legacyResolver.ts
+++ b/packages/3id-did-resolver/src/legacyResolver.ts
@@ -1,5 +1,6 @@
 import fetch from 'cross-fetch'
 import * as u8a from 'uint8arrays'
+import { LRUMap } from 'lru_map'
 
 interface DAG {
   get(cid: string): any;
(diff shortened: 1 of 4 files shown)


packages/3id-did-resolver/src/legacyResolver.ts
```diff
@@ -11,11 +12,17 @@ interface IPFS {
 
 // Legacy 3ids available from 3box, other v1 will always be resolved through ipfs and other services 
 const THREEBOX_API_URL = 'https://ipfs.3box.io'
+const LIMIT = 100
+const fetchCache = new LRUMap<string, any>(LIMIT)
 
 const fetchJson = async (url: string): Promise<any> =>  {
+  const cached = fetchCache.get(url)
+  if (cached) return cached
```
stbrody, line 20: can v0 3ids ever be updated?  Do we need to worry about cache invalidation at all?

ukstv: They can not. They are static, so no cache invalidation.

ukstv (approved):
🏇

packages/3id-did-resolver/src/legacyResolver.ts
oed: Nope! 3IDv0 are just static ipfs objects.

PR #791, opened 2021-01-14
fix(blockchain-utils-validation): add missing dep

oed:
Having trouble using the new creamic version. Looks like this dependency is missing.


ukstv:
OHSHI. True finding.

ukstv (approved):

Issue #778, opened 2021-01-13
Add nodes on our infra as bootstrap nodes when running ceramic cli with internal ipfs.


oed:
We likely want this only to happen only if you start the cli with testnet-clay (and mainnet in the future once enabled)

For Clay we want to bootstrap ipfs with:
* Ceramic Node IPFS
* Ceramic Gateway IPFS
* Ceramic Anchor Service IPFS

@val to update us with the actual libp2p multiaddrs!

PR #801, opened 2021-01-15
Establish IPFS connectivity for PubSub

ukstv:
Why manual connection? Apparently, it takes a lot of time for IPFS to connect to bootstrap nodes.

Related: #778

### DIFF
--- a/packages/cli/src/build-ipfs-connection.util.ts
+++ b/packages/cli/src/build-ipfs-connection.util.ts
@@ -1,0 +1,66 @@
+import IPFS from "ipfs-core"
+
+import dagJose from 'dag-jose'
+// @ts-ignore
+import multiformats from 'multiformats/basics'
+// @ts-ignore
+import legacy from 'multiformats/legacy'
+import ipfsClient from "ipfs-http-client"
+import { IpfsApi } from "@ceramicnetwork/common"
+
+multiformats.multicodec.add(dagJose)
+const dagJoseFormat = legacy(multiformats, dagJose.name)
+
+const IPFS_DHT_SERVER_MODE = process.env.IPFS_DHT_SERVER_MODE === 'true'
+
+const BOOTSTRAP_LIST = [
+    '/dns4/ipfs-dev.3boxlabs.com/tcp/4012/wss/p2p/QmeRWpCaEpQLY1yJFFHf7pFbZZapXoj5E4pKvUtyExMUKk',
+    '/dns4/ipfs-dev.ceramic.network/tcp/4012/wss/p2p/QmadvKqYyLMHYjnwswd2bHM3UCFp4F4oSFdobYUKSPjBaZ',
+    '/dns4/ipfs-dev-cas.3boxlabs.com/tcp/4012/wss/p2p/QmPHLQoWhK4CMPPgxGQxjNYEp1fMB8NPpoLaaR2VDMNbcr',
+    '/dns4/ipfs-clay-cas.3boxlabs.com/tcp/4012/wss/p2p/QmQotCKxiMWt935TyCBFTN23jaivxwrZ3uD58wNxeg5npi',
+    '/dns4/ipfs-clay.ceramic.network/tcp/4012/wss/p2p/QmQqq1oVkwpJdKNdiWvrftgWE2nov4AFGPBauZ2AakxmoR',
+    '/dns4/ipfs-clay.3boxlabs.com/tcp/4012/wss/p2p/QmQotCKxiMWt935TyCBFTN23jaivxwrZ3uD58wNxeg5npi'
+]
+
+export async function buildIpfsConnection(network: string, ipfsEndpoint?: string): Promise<IpfsApi>{
+    if (ipfsEndpoint) {
+        return ipfsClient({ url: ipfsEndpoint, ipld: { formats: [dagJoseFormat] } })
+    } else {
+        const ipfs = await IPFS.create({
+            ipld: {
+                formats: [dagJoseFormat]
+            },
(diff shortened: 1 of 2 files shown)


oed:
> Why manual connection? Apparently, it takes a lot of time for IPFS to connect to bootstrap nodes.

Ok that's still the case -.- We saw the same thing happen with 3box way back.

packages/cli/src/build-ipfs-connection.util.ts
```diff
@@ -0,0 +1,66 @@
...
+            config: {
+                Routing: {
+                    Type: IPFS_DHT_SERVER_MODE ? 'dhtserver' : 'dhtclient',
+                },
+                Bootstrap: BOOTSTRAP_LIST
+            }
+        })
+        const shouldConnect = network === "testnet-clay" || network === "mainnet"
+        if (shouldConnect) {
+            await Promise.all(BOOTSTRAP_LIST.map(async node => {
```
oed, line 54: Do we really need to await this? I don't think this is blocking for the rest of the code?

packages/cli/src/build-ipfs-connection.util.ts
```diff
@@ -0,0 +1,66 @@
...
+
+const IPFS_DHT_SERVER_MODE = process.env.IPFS_DHT_SERVER_MODE === 'true'
+
+const BOOTSTRAP_LIST = [
+    '/dns4/ipfs-dev.3boxlabs.com/tcp/4012/wss/p2p/QmeRWpCaEpQLY1yJFFHf7pFbZZapXoj5E4pKvUtyExMUKk',
+    '/dns4/ipfs-dev.ceramic.network/tcp/4012/wss/p2p/QmadvKqYyLMHYjnwswd2bHM3UCFp4F4oSFdobYUKSPjBaZ',
+    '/dns4/ipfs-dev-cas.3boxlabs.com/tcp/4012/wss/p2p/QmPHLQoWhK4CMPPgxGQxjNYEp1fMB8NPpoLaaR2VDMNbcr',
+    '/dns4/ipfs-clay-cas.3boxlabs.com/tcp/4012/wss/p2p/QmQotCKxiMWt935TyCBFTN23jaivxwrZ3uD58wNxeg5npi',
+    '/dns4/ipfs-clay.ceramic.network/tcp/4012/wss/p2p/QmQqq1oVkwpJdKNdiWvrftgWE2nov4AFGPBauZ2AakxmoR',
+    '/dns4/ipfs-clay.3boxlabs.com/tcp/4012/wss/p2p/QmQotCKxiMWt935TyCBFTN23jaivxwrZ3uD58wNxeg5npi'
```
oed, line 22: Can we make sure to connect to the clay nodes only when `network` is configured to `testnet-clay`?

packages/cli/src/build-ipfs-connection.util.ts
```diff
@@ -0,0 +1,66 @@
...
+
+multiformats.multicodec.add(dagJose)
+const dagJoseFormat = legacy(multiformats, dagJose.name)
+
+const IPFS_DHT_SERVER_MODE = process.env.IPFS_DHT_SERVER_MODE === 'true'
+
+const BOOTSTRAP_LIST = [
+    '/dns4/ipfs-dev.3boxlabs.com/tcp/4012/wss/p2p/QmeRWpCaEpQLY1yJFFHf7pFbZZapXoj5E4pKvUtyExMUKk',
+    '/dns4/ipfs-dev.ceramic.network/tcp/4012/wss/p2p/QmadvKqYyLMHYjnwswd2bHM3UCFp4F4oSFdobYUKSPjBaZ',
+    '/dns4/ipfs-dev-cas.3boxlabs.com/tcp/4012/wss/p2p/QmPHLQoWhK4CMPPgxGQxjNYEp1fMB8NPpoLaaR2VDMNbcr',
```
oed, line 19: Not sure we want to connect to these. Are they also configured with Clay @valmack ?

packages/cli/src/build-ipfs-connection.util.ts
oed: Ah, see that you did something like this 👍 I imagine when we have the mainnet nodes up we will connect to a different set of nodes of that configuration :)

packages/cli/src/build-ipfs-connection.util.ts
ukstv: It's controversial. If you do not wait, the daemon _certainly_ proceeds with an empty set of peers. If you wait, it leaves you with the "best effort" connectivity, which most probably is all right for application to proceed.

packages/cli/src/build-ipfs-connection.util.ts
ukstv: So, as for now, all the nodes are still here @oed @valmack ?

ukstv: Clarification regarding network names: if the "bootstrap" nodes do not follow some other pubsub topic like "/testing/bugging", the nodes connected to the bootstrap ones would not hear from the topic, unless finally connected to each other.

oed: These three nodes we should connect to only if we are on `testnet-clay`.

packages/cli/src/build-ipfs-connection.util.ts
oed: These three nodes we should connect to only if we are on `dev-unstable`.

ukstv: aha, clear

oed (approved):
lgtm

PR #805, opened 2021-01-16
fix(cli): Properly connect to bootstrap nodes

oed:
This PR contains two changes:
* Properly make connections to the bootstrap nodes


stbrody:
Can we split out the DHT disabling from the rest of this?  I think that deserves its own github story and own PR, for anyone outside the company who may be following our git activity

oed:
Removed the commit from this branch 👍

stbrody (approved):

PR #806, opened 2021-01-18
fix(cli): Disable DHT

oed:
**This PR disables the use of the DHT in the js-ipfs node used by the cli.**

Having the DHT enabled in it's current configuration is causing quite severe degradation issues when making method calls to ceramic. It doesn't only slow things down, but makes request times very variable (sometimes up to minutes). With the DHT disabled the same calls work much more consistently.


ukstv (approved):
🐎

stbrody (approved):

PR #817, opened 2021-01-18
fix(core): Schema validation not enforced during update

stbrody:


packages/core/src/document.ts
```diff
@@ -528,7 +528,7 @@ class Document extends EventEmitter implements DocStateHolder {
         // it's a signed commit
         const tmpState = await this._doctypeHandler.applyCommit(commit, cid, this._context, state)
         if (this._validate) {
-          const schemaId = payload.header?.schema
+          const schemaId = tmpState.next.metadata.schema
```
zachferland, line 531: does line above https://github.com/ceramicnetwork/js-ceramic/pull/817/files#diff-0cbb693f9af1c04e8581f15abbab4c56b056e71312d6cbbef0ab9eea11c4b83fR518 also not have to change similarly? payload.metadata.schema

zachferland (approved):
im less familiar with this part of the code, but looks good, and makes sense with the tests, just added one comment

packages/core/src/document.ts
zachferland: oh that link no good, line 518 i meant

stbrody: Well, it would be fine because that's a genesis record so the current payload has all the relevant state needed.  But I just put up a small cleanup to merge these code paths so it should be clearer now

zachferland: 👍

zachferland (approved):

PR #822, opened 2021-01-19
feat(core): Allow specifying infura project id

stbrody:


michaelsena:
@stbrody I think we should make this generic as there are other alternative providers besides Infura that offer NaaS API access. And I imagine there might be an off chance that someone might want to run their own ETH node for an API. Thoughts on these?

michaelsena:
[Alchemy](https://docs.alchemyapi.io/alchemy/guides/getting-started#1-create-an-alchemy-key) is another example. Also I'm wondering how the Ceramic node network configuration (clay: rinkeby/ropsten; mainnet: mainnet) interacts with the API URLs of the different providers. Infura likely has one URL configuration, Alchemy another, etc...

valmack:
Agree with @michaelsena We should think this through a bit more. For now I think it's fine having the functionality to pass a json rpc endpoint if the developer is experiencing rate limiting with ethers.

The alternative should be, the developer specifies which type of provider they would like to use and the relevant configuration. See https://docs.ethers.io/ethers.js/v5-beta/api-providers.html#provider-specific-properties-and-methods

oed (changes requested):
What's the reason for adding this? Are we running into limits somewhere? Ideally here we wouldn't need to rely on etherjs to make our requests. Not sure though what the state of the art is in terms of communicating with ethereum in an agnostic way. I agree with above that it's weird to assume infura here.

stbrody:
Closing in favor of https://github.com/ceramicnetwork/js-ceramic/issues/826

PR #830, opened 2021-01-20
fix(core): Honor ethereumRpcUrl config option

stbrody:


packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -235,10 +253,17 @@ export default class EthereumAnchorService extends AnchorService {
             throw new Error('Invalid chain ID according to CAIP-2')
```
valmack, line 253: Can we update this error message (it's not exactly accurate for example if I were to pass `bip122:000000000019d6689c085ae165831e93`)


Maybe just `Unsupported chain ID. Must be eip155 namespace.`

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -75,6 +78,23 @@ export default class EthereumAnchorService extends AnchorService {
...
     }
 
+    async init(): Promise<void> {
+        // Get the chainIds supported by our anchor service
+        const response = await fetch(this.chainIdApiEndpoint)
+        const json = await response.json()
+        this._supportedChains = json.supportedChains
+
+        // Confirm that we have an eth provider that works for each of the chains that the anchor service supports
+        for (const chain of this._supportedChains) {
```
valmack, line 88: Hmm trying to understand what is happening here. We have supported chains from the perspective of the anchor service, but also only 2 supported chains listed here eip155:1 and eip155:3.

Should we just be checking that the anchor service supports __*at least* one of these two__ chains?
Seems like this loop should not be needed. Whichever chain is in our supported list is the one we can use to get the provider.

And then, in the case that they pass an rpc endpoint, we also ensure it is one of the two chains.

Maybe I'm missing something but not sure entirely what we are trying to achieve here yet.

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
stbrody, line 253:
```suggestion
            throw new Error('Unsupported chainId - must be eip155 namespace')
```

stbrody: Done

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
stbrody: So we do support chainids other than `eip155:1` and `eip155:3` (for instance ganache uses chainid `eip155:1337`).  But to use the default ethers provider we need the network name not the chainId, so these are the two hardcoded mappings from chainid to network name.  We should probably add Rinkeby to this list.

The loop probably isn't necessary at the moment, but is future proofing in case the CAS supports multiple chains in the future.  In that world, we'd need to make sure that the ceramic node had an eth provider for every chain that the CAS supports.  Right now the CAS will never return more than one supported chainId at a time, so it doesn't really matter either way.

oed (approved):
lgtm

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
valmack: So idea is to allow anchoring to multiple chains at once?

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -235,10 +253,17 @@ export default class EthereumAnchorService extends AnchorService {
             throw new Error('Invalid chain ID according to CAIP-2')
         }
 
+        if (!this._supportedChains.includes(chain)) {
```
valmack, line 256: This line seems redundant? this._supportedChains is the CAS chains and chain is one of the CAS chains from the loop.

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
valmack: Also not sure it’s fully future proof it seems because we only allow you to pass 1 rpc endpoint which I think can only have 1 network at a time. I could be wrong..

oed: Interesting point @valmack! Maybe we need to allow passing multiple endpoints for specific chainIds?

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -75,6 +78,23 @@ export default class EthereumAnchorService extends AnchorService {
...
+        const response = await fetch(this.chainIdApiEndpoint)
+        const json = await response.json()
+        this._supportedChains = json.supportedChains
+
+        // Confirm that we have an eth provider that works for each of the chains that the anchor service supports
+        for (const chain of this._supportedChains) {
+            const provider = this._getEthProvider(chain)
+            const provider_chain_idnum = (await provider.getNetwork()).chainId
+            const provider_chain = BASE_CHAIN_ID + ':' + provider_chain_idnum
+            if (chain != provider_chain) {
```
ukstv, line 92: If one ethereum rpc is passed (chainId 1), and CAS returns multiple supported chains (chainId 1, 2, 3), this would always fail on [2, 3].

ukstv: Guess, this thing with supported chains and checking of support from providers is totally unnecessary. Have ethereumRpc => use this provider (in `_getEthereumProvider`), check if it supports the chainId of validated anchor. If supports, use it. If it does not, use fallback provider.

oed: Note for fixing this: You can call `eth_chainId` to get the chainId of the provider: https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/blockchain-utils-linking/src/ethereum.ts#L31

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
stbrody: Yeah, we don't actually support multiple chains yet, there would be more work to do to make that work well.  We just made the API from the CAS for fetching the supported chain(s) use an array so that we could add this support in the future without breaking the API.  I added an assert that there is only one supported chain returned from the CAS for now, which makes this more explicit.

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
stbrody: `chain` can also come from an anchorProof that needs validating, which was discovered from an anchorCommit and could theoretically be any chain if the user anchored manually

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
stbrody: I made it explicit that we only support one chain at a time right now

ukstv: In the next two weeks, the comment in this pull request is forgotten by everyone, but code remains. It would be fun to watch a person doing digital archeology to figure the reasoning here.

ukstv: If one chain, the loop does not make sense. If multiple chains, it conveys a misleading semantics.

ukstv: the relation implied here is `some` or `includes`, not `every` which is currently encoded.

stbrody: > the relation implied here is `some` or `includes`, not `every` which is currently encoded.

It actually should be `every`.  If the CAS supported multiple chains, then ceramic would need to be prepared to validate any anchorProofs from any of the supported chains that might get used.

stbrody: I'll make a change to get the first element of the array returned from the CAS and from that point on structure the code to assume only one supported chain, so the code more explicitly matches the current state of the world, rather than trying to code for a hypothetical future we don't yet have

ukstv: > If the CAS supported multiple chains, then ceramic would need to be prepared to validate any anchorProofs from any of the supported chains that might get used.

Mea culpa. That's right. I was stuck on current "return provider just based on single ethereum rpc param" logic in `getEthereumProvider`.

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
ukstv: Maybe leave the check here, and remove it from `init` then?

stbrody: Not sure I follow.  The checks in `init` are different.  Which part are you suggesting is unnecessary?

ukstv: OHSHI. Sorry. They really are different. Please, do not mind.

ukstv (approved):
🏇 LGTM

PR #833, opened 2021-01-21
fix: update package-lock

oed:
Seems like there was an issue with eos tests due to some package mismatch. Updated the package-lock.


ukstv (approved):
🏇

ukstv:
npm is no good :/

PR #839, opened 2021-01-21
fix(core): Don't erroneously set anchorStatus to FAILED when a later CID is anchored instead of an earlier one

stbrody:


packages/core/src/document.ts
```diff
@@ -580,39 +580,73 @@ class Document extends EventEmitter implements DocStateHolder {
...
    * Request anchor for the latest document state
    */
   async anchor (): Promise<void> {
-    this._context.anchorService.on(this.id.toString(), async (asr: AnchorServiceResponse): Promise<void> => {
+    // eslint-disable-next-line @typescript-eslint/no-this-alias
+    const doc = this
+    const tip: CID = this.tip
+
+    this._context.anchorService.on(this.id.toString(), async function listener (asr: AnchorServiceResponse) {
+      if (asr.cid.toString() != tip.toString()) { // todo don't use toString
```
stbrody, line 610: I have no idea why this is necessary, but I tried comparing the cids directly as well as comparing their `.bytes` and in both cases the CIDs were comparing unequal even when they were in fact the same.  Using `toString()` solved the issue, but it really shouldn't be necessary... Open to ideas on how to debug this further...

packages/core/src/document.ts
```diff
@@ -580,39 +581,73 @@ class Document extends EventEmitter implements DocStateHolder {
...
-    this._context.anchorService.on(this.id.toString(), async (asr: AnchorServiceResponse): Promise<void> => {
+    // eslint-disable-next-line @typescript-eslint/no-this-alias
+    const doc = this
+    const tip: CID = this.tip
+
+    this._context.anchorService.on(this.id.toString(), async function listener (asr: AnchorServiceResponse) {
+      if (!asr.cid.equals(tip)) {
+        // This message is about a different tip for the same document
+        return
+      }
```
oed, line 614: Why do we need the condition if `FAILED` if we just return here when the cid is not equal to the tip?

oed (approved):
Nice catch

packages/core/src/document.ts
stbrody: I don't understand the question.  What condition are you referring to?

oed: Sorry that was way to unclear!

packages/core/src/document.ts
```diff
@@ -580,39 +581,73 @@ class Document extends EventEmitter implements DocStateHolder {
...
-          const state = this._doctype.state
+          doc._context.anchorService.removeListener(doc.id.toString(), listener)
+
+          if (await doc._isCidAnchored(tip)) {
+            // Even though the anchor request for this specific CID came back as FAILED, the cid
+            // still appears to be anchored.  This means that a later CID built on top of this
+            // one was anchored instead. In that case we should do nothing and simply return since
+            // the later CID that was successfully anchored should have already updated the state appropriately
+            return
+          }
```
oed, line 647: This is the condition I was referring to. Seems like if we check that `asr.cid` equals `tip` then this will never happen?

stbrody: This could happen if the notifications from the anchor service come in out of order. Basically if you learn about the later CID being anchored, and *then* get notified about the earlier CID failing.  You don't want to transition to FAILED in that case

oed (approved):

packages/core/src/document.ts
oed: makes sense, thanks

PR #844, opened 2021-01-25
fix(core): Properly ignore old FAILED anchor responses

oed:
We set the `tip` to a constant which was problematic if there is another update on the old listener. This PR changes this to use `doc.tip` which is dynamic.

### DIFF
--- a/packages/core/src/document.ts
+++ b/packages/core/src/document.ts
@@ -582,33 +582,15 @@ class Document extends EventEmitter implements DocStateHolder {
   }
 
   /**
-   * Returns whether the given CID is in the log, and there is a later log entry that is an anchor commit
-   * @param cid
-   */
-  async _isCidAnchored(cid: CID): Promise<boolean> {
-    let foundCid = false
-    for (const entry of this.state.log) {
-      if (entry.cid.equals(cid)) {
-        foundCid = true
-      }
-
-      if (foundCid && entry.type == CommitType.ANCHOR) {
-        return true
-      }
-    }
-    return false
-  }
-
-  /**
    * Request anchor for the latest document state
    */
   async anchor (): Promise<void> {
     // eslint-disable-next-line @typescript-eslint/no-this-alias
     const doc = this
-    const tip: CID = this.tip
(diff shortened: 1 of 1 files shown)


ukstv (approved):
LGTM

packages/core/src/document.ts
```diff
@@ -605,10 +605,9 @@ class Document extends EventEmitter implements DocStateHolder {
   async anchor (): Promise<void> {
     // eslint-disable-next-line @typescript-eslint/no-this-alias
     const doc = this
-    const tip: CID = this.tip
 
     this._context.anchorService.on(this.id.toString(), async function listener (asr: AnchorServiceResponse) {
-      if (!asr.cid.equals(tip)) {
+      if (!asr.cid.equals(doc.tip)) {
```
stbrody, line 610: I don't understand this change at all.  The point of this if-statement with the captured `tip` value was to make sure that there is one `listener` per anchor request and that each listener instance only processes the response for the CID it was created for.  That's how we were able to switch from `anchorService.removeAllListeners` to `anchorService.removeListener(listener)` below and know that we're removing the right listener instance.

Can you explain further the problem you were seeing that lead to this change?

stbrody (changes requested):

packages/core/src/document.ts
oed: You make two updates to the doc: *u1*, *u2* in that order. *u2* returns anchor `COMPLETE`, doc is updated. *u2* returns anchor `FAILED`, now since the `tip` was saved the document will change it's status to failed. This in various variations made me do this change. 

I see your point about the listeners though. And now that I look more closely it would seem that the issue would get solved by `_isCidAnchored`. But for some reason it isn't.

stbrody: Right, that's exactly the purpose of the `_isCidAnchored` check.  If that's not working, then that might be the real bug

stbrody (approved):

stbrody:
LGTM if tests pass

stbrody (approved):

PR #845, opened 2021-01-25
Harmonise versions of widely-used packages

ukstv:
No more complaints from lerna about mis-hoisted dependencies.

- Same uint8arrays in all the packages (no change function-wise),
- Same babel-jest in all the packages (dev, so do not care much if it is functionally different as soon as tests pass)
- Removed unused deps
- Reorganised zondax deps: with link/validation and handler/tile it makes sense to include the one in the caip10 handler only

Really, I just wanted to add one more package, and then it turned into this.


packages/3id-did-resolver/package.json
```diff
@@ -32,23 +32,20 @@
...
     "@babel/core": "^7.9.0",
     "@babel/preset-env": "^7.8.4",
     "@babel/preset-typescript": "^7.9.0",
-    "@types/bs58": "^4.0.1",
-    "@types/events": "^3.0.0",
     "@types/node": "^13.13.15",
     "@typescript-eslint/eslint-plugin": "^4.6.0",
     "@typescript-eslint/parser": "^4.6.0",
-    "babel-jest": "^25.1.0",
+    "babel-jest": "^26.6.3",
```
oed, line 44: At some point it might make more sense to switch to `ts-jest` so that we don't need all of these babel deps.

packages/core/src/anchor/memory/in-memory-anchor-service.ts
```diff
@@ -258,7 +257,8 @@ class InMemoryAnchorService extends AnchorService {
     const { payload, signatures } = commit
     const { signature, protected: _protected } = signatures[0]
 
-    const decodedHeader = JSON.parse(base64url.decode(_protected))
+    const jsonAsBase64url = uint8arrays.toString(uint8arrays.fromString(_protected, 'base64url'))
+    const decodedHeader = JSON.parse(jsonAsBase64url)
```
oed, line 261:
```suggestion
    const decodedJsonString = uint8arrays.toString(uint8arrays.fromString(_protected, 'base64url'))
    const decodedHeader = JSON.parse(decodedJsonString)
```
More clear?

oed (approved):
Nice! 

uint8arrays <3

oed:
Let's wait with merging this though!

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -75,7 +69,7 @@ export default class EthereumAnchorService extends AnchorService {
      * @param ceramic - Ceramic API used for various purposes
      */
     set ceramic(ceramic: CeramicApi) {
-        this._ceramic = ceramic
+        // Do Nothing
```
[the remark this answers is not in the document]
ukstv: Right.

packages/3id-did-resolver/package.json
ukstv: Interesting point. To date I thought that was an explicit decision to jump onto Babel bandwagon. The most obvious reason I saw is speed - Babel avoids type checking.

Personally, prefer ts-jest. Less clutter in the dependency list.

packages/core/src/anchor/memory/in-memory-anchor-service.ts
ukstv: Yup.

packages/3id-did-resolver/package.json
oed: I think when I initially set up this repo I was not aware of ts-jest. How big is the speed difference?

ukstv: Have seen anecdotal evidence only like https://iamturns.com/typescript-babel/

Fanboy-ish "it's fast".

In my experience, it's like 1.2-1.5 times slower. Not good, but not critical, and it really complains on broken tests wrt types.

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -131,19 +125,15 @@ export default class EthereumAnchorService extends AnchorService {
             }
         })
 
-        if (!response.ok) {
+        if (response.ok) {
+            const json = await response.json()
+            this.processRemoteResponse(cidDocPair, json)
+        } else {
             this.emit(cidDocPair.docId, {
-                cid: cidDocPair.cid,
```
stbrody, line 136: why is this removed?

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -271,4 +235,32 @@ export default class EthereumAnchorService extends AnchorService {
         return providers.getDefaultProvider(ethNetwork.network)
     }
 
+    private processRemoteResponse(cidDoc: CidDoc, json: any): boolean {
+        switch (json.status) {
+            case "PENDING": {
```
stbrody, line 240: While we're at it, can we explicitly combine the PENDING and PROCESSING cases since the logic is the same?

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -271,4 +235,32 @@ export default class EthereumAnchorService extends AnchorService {
         return providers.getDefaultProvider(ethNetwork.network)
     }
 
+    private processRemoteResponse(cidDoc: CidDoc, json: any): boolean {
```
stbrody, line 238: Can you add a comment for this method?  At the very least it should clarify what the return value represents

stbrody (commented):
Wow, what a great set of cleanups.  Thanks for doing this!  A couple of small comments but mostly looks good

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
ukstv: A mistake certainly.

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
ukstv: This is intentional in anticipation of proper enum-based handling as in https://github.com/ceramicnetwork/js-ceramic/pull/860/files#diff-7920408ff457031025a51c6c64241f297dc91febb1a810e724475548c154c781R251

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
ukstv: The replacement for this is coming in the next batch: https://github.com/ceramicnetwork/js-ceramic/pull/860/files#diff-7920408ff457031025a51c6c64241f297dc91febb1a810e724475548c154c781R251

packages/3id-did-resolver/package.json
oed: So let's keep babel unless we have an explicit reason to change?
cc @zachferland if you have any thoughts here.

ukstv:
Dropping this, as relevant changes have been merged into upstream.

PR #856, opened 2021-01-26
fix: Use node 14

oed:
Switch from node 12 to 14 in:
* Dockerfile
* CircleCI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ orbs:
 jobs:
   build-and-test:
     docker:
-      - image: circleci/node:12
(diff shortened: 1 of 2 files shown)


valmack (approved):
I guess if tests are passing we should be ok? (A bit worried about compatibility here but rather be fixing bugs on 14 anyway). We should likely bump js-IPFS-ceramic also

oed:
I don't think we need to bump js-ipfs-ceramic right now, but we should in the future. I've been using node 14 locally since forever btw, not seen any issues.

stbrody (approved):

PR #857, opened 2021-01-26
fix(core): Improve pubsub resubscribe logic

oed:
Only tries to resubscribe if we are not currently subscribed to the pubsub topic. Will unsubscribe before trying to subscribe. 

Have tested this locally by running an external ipfs node with ceramic. Closed the ipfs node, then restarted it and confirmed that I still receive pubsub messages.

### DIFF
--- a/packages/core/src/dispatcher.ts
+++ b/packages/core/src/dispatcher.ts
@@ -61,7 +61,7 @@ export default class Dispatcher extends EventEmitter {
    */
   async init(): Promise<void> {
     this._peerId = this._peerId || (await this._ipfs.id()).id
-    await this._subscribe()
+    await this._subscribe(true)
     // If ipfs.libp2p is defined we have an internal ipfs node, this means that
     // we don't want to resubscribe since it will add multiple handlers.
     if (!TESTING && !this._ipfs.libp2p) {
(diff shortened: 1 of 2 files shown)


packages/core/src/dispatcher.ts
```diff
@@ -74,18 +74,24 @@ export default class Dispatcher extends EventEmitter {
...
+          this.handleMessage,
+          // {timeout: IPFS_GET_TIMEOUT} // ipfs-core bug causes timeout option to throw https://github.com/ipfs/js-ipfs/issues/3472
+        )
+        this._log({peer: this._peerId, event: 'subscribed', topic: this.topic })
+      }
     } catch (error) {
       // TODO: use logger
       if (error.message.includes('Already subscribed')) {
         this.logger.debug(error.message)
+      } else if (error.message.includes('The user aborted a request')) {
```
valmack, line 92: Never seen this. Is it on node 14?

Also should we be using latest IPFS http client?

packages/core/src/dispatcher.ts
```diff
@@ -74,18 +74,24 @@ export default class Dispatcher extends EventEmitter {
...
-  async _subscribe(): Promise<void> {
+  async _subscribe(force = false): Promise<void> {
     try {
-      await this._ipfs.pubsub.subscribe(
-        this.topic,
-        this.handleMessage,
-        // {timeout: IPFS_GET_TIMEOUT} // ipfs-core bug causes timeout option to throw https://github.com/ipfs/js-ipfs/issues/3472
-      )
-      this._log({peer: this._peerId, event: 'subscribed', topic: this.topic })
+      if (force || !(await this._ipfs.pubsub.ls()).includes(this.topic)) {
```
valmack, line 79: Assumes 1 client for 1 node (otherwise this breaks). I think that’s a safe assumption as ipfs team says they weren’t designing for multi-client.

valmack (approved):

packages/core/src/dispatcher.ts
oed: I've used node 14 when testing yes. 

Not updating to latest ipfs-http-client because it might break other stuff (CIDs, multiformats).

packages/core/src/dispatcher.ts
oed: Makes sense, thanks.

packages/core/src/dispatcher.ts
```diff
@@ -74,18 +74,24 @@ export default class Dispatcher extends EventEmitter {
...
+          // {timeout: IPFS_GET_TIMEOUT} // ipfs-core bug causes timeout option to throw https://github.com/ipfs/js-ipfs/issues/3472
+        )
+        this._log({peer: this._peerId, event: 'subscribed', topic: this.topic })
+      }
     } catch (error) {
       // TODO: use logger
       if (error.message.includes('Already subscribed')) {
         this.logger.debug(error.message)
+      } else if (error.message.includes('The user aborted a request')) {
+        // for some reason the first call to pubsub.subscribe throws this error
```
stbrody, line 93: 🤨

stbrody (approved):

stbrody:
How often are we getting disconnected?  Do we need to decrease the IPFS_RESUBSCRIBE_INTERVAL_DELAY to less than a minute?

packages/core/src/dispatcher.ts
oed: Yeah, I couldn't make any sense of it.

PR #860, opened 2021-01-26
Drop EventEmitter in favour of rxjs

ukstv:
Don't look at this. Cleaner PR is in the works.

<s>Pro: cleaner code.
Contra: might be non-obvious.

Fact: Put myself in a tricky branch-situation, with a lot of (hopefully) goodness stuck in-between reviews, as this thing is based on #845.

Going to stop pushing code for now, till this thing moves to some upstream-land.

@PaulLeCam @oed have a look at `packages/core/src/anchor/memory/in-memory-anchor-service.ts` and `packages/core/src/anchor/ethereum/ethereum-anchor-service.ts` RxJS goodness is here.</s>


stbrody:
Can you set the base branch for this PR to `su/pending-anchors-2` so that the changes from https://github.com/ceramicnetwork/js-ceramic/pull/845 don't show up as well, and we can focus the review on the new changes?

ukstv:
@stbrody I managed to merge develop branch here, so either way there's going to be a great deal of garbage here. 

Better consider a preview. Marked as draft because of this.

packages/core/src/anchor/memory/in-memory-anchor-service.ts
```diff
@@ -36,134 +35,139 @@ interface InMemoryAnchorConfig {
...
-    const groupedCandidates = await this._groupCandidatesByDocId(this._queue)
-    return await this._selectValidCandidates(groupedCandidates)
+    const groupedCandidates = await this._groupCandidatesByDocId(this.#queue);
+    return this._selectValidCandidates(groupedCandidates);
   }
 
   async _groupCandidatesByDocId(candidates: Candidate[]): Promise<Record<string, Candidate[]>> {
     const result: Record<string, Candidate[]> = {};
+    await Promise.all(
+      candidates.map(async (req) => {
```
stbrody, line 100: Nice

packages/core/src/anchor/memory/in-memory-anchor-service.ts
```diff
@@ -36,134 +35,139 @@ interface InMemoryAnchorConfig {
...
+    );
+    return result;
   }
 
-  async _selectValidCandidates(groupedCandidates: Record<string, Candidate[]>): Promise<Candidate[]> {
-    const result: Candidate[] = []
+  _selectValidCandidates(groupedCandidates: Record<string, Candidate[]>): Candidate[] {
+    const result: Candidate[] = [];
     for (const compositeKey of Object.keys(groupedCandidates)) {
-      const candidates: Candidate[] = groupedCandidates[compositeKey]
```
stbrody, line 127: why remove the type annotation?  It may not be necessary, but it doesn't hurt does it?  And it makes the code more readable IMO

packages/core/src/anchor/memory/in-memory-anchor-service.ts
```diff
@@ -36,134 +35,139 @@ interface InMemoryAnchorConfig {
...
-      }
-    }
-
-    return result
+      })
+    );
+    return result;
   }
 
-  async _selectValidCandidates(groupedCandidates: Record<string, Candidate[]>): Promise<Candidate[]> {
```
stbrody, line 124: when I first set out to learn javascript a few months ago I read something that cautioned against mixing async and non-async functions in the same code.  I guess that isn't actually that big of a problem after all?  I don't recall what specifically the thing I read was worried could go wrong...

packages/core/src/anchor/memory/in-memory-anchor-service.ts
```diff
@@ -205,23 +209,35 @@ class InMemoryAnchorService extends AnchorService {
...
-    const candidate: Candidate = new Candidate(cid, docId)
-
-    if (this._anchorOnRequest) {
-      await this._process(candidate)
+  requestAnchor(docId: DocID, tip: CID): Observable<AnchorServiceResponse> {
+    const candidate = new Candidate(tip, docId);
+
+    if (this.#anchorOnRequest) {
+      queue
+        .add(() => {
```
stbrody, line 226: why do we add this to the queue if it's already been processed?

stbrody (commented):
> @stbrody I managed to merge develop branch here, so either way there's going to be a great deal of garbage here.
> 
> Better consider a preview. Marked as draft because of this.
> 
> Actually, it is better to adjust the branch after the previous one is merged. Whenever that might be.

okay, then I'm just going to hold off on doing too much of a review until this is cleaned up and in a more readable state

stbrody:
Oops, looks like my draft comments got published accidentally.  oh well, that's fine.  feel free to either respond now or wait until the proper full review once this is cleaned up

stbrody:
I guess before getting too into the weeds on this review, I'd like to better understand why we're doing it?  It seems like a pretty sweeping change for unclear gain?  Do we think rxjs is more performant than EventEmitter?  Does it allow us to do something that EventEmitter doesn't?  The PR description just says "cleaner code" as the pro.  From what I skimmed thus far it doesn't seem immediately obvious that rxjs leads to cleaner code than EventEmitter, and even if it does, it seems the benefit is marginal at best.

If we do make this change, we should probably make the change to every place we use EventEmitter, so we're not mixed with two different models that need to be known to understand our code.  If we're going to make such a wide-reaching refactor, it'd be nice to understand better what we're looking to gain from it.

packages/core/src/ceramic.ts
```diff
@@ -585,6 +586,20 @@ class Ceramic implements CeramicApi {
     return this._networkOptions.supportedChains
   }
 
+  /**
+   * Load all the pinned documents, re-request PENDING or PROCESSING anchors.
+   */
+  async recoverDocuments() {
```
oed, line 592: Shouldn't this method be called on startup?

oed (commented):
What's the benefit of using rxjs here over an AsyncIterator? Seems like there will never be more than once place we are listening for updates so seems like an AI should be sufficient here?

packages/core/src/ceramic.ts
ukstv: It is called, in cli/daemon. Current tests apply a lot of mocks, that (a) are too daunting to change for this little thing, (b) will be heavily redone wrt mocking while state refactoring.

oed: But what if I start a ceramic core instance programmatically? Seems like it should be called when you create a ceramic instance?

oed:
My main worry about using `rxjs` as done here is that it adds 47kB to the bundle size without any benefits over AsyncIterators as far as I can tell.

ukstv:
I am curious to now what matters more here: bundle size or a limited ability to shoot yourself in the foot.

Let's look at the current upstream. Scenario: start Ceramic instance, make it poll for anchors by creating a document, then stop the instance via `Ceramic#close()`.
Expectation: it stops everything. Reality: polling continues.
I wonder what other interesting behaviour is lurking in the code, eating resources for nothing.

Theoretically, AI is better due to back-pressure and native support. Practically, it would take a lot of plumbing  to _really_ control AI-based control flow like you can do now with piping and explicit `Subscription` of rxjs.

oed:
> Yes, you could add a signalling mechanism to stop all of the polling. Then you have a prospective change when documents are instantiated on demand, so this kind of signalling has to be per document, and for me this sounds like a big enough opportunity for drowning in concurrency issues.

Not sure i follow why this would be? Seems like we could just use one [AbortSignal](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) within the `ethereum-anchor-service.ts` class that is passed to all calls to fetch from there. When `.close` is called we would simply tell the ethereum anchor service instance to abort, which would abort all current polling. 

Am I missing something here?

ukstv:
> that is passed to all calls to fetch from there

This is what I call an invitation for drowning into all sorts of concurrency issues.

It's one anchor service, not a big deal. Consider amount of plumbing for the other cases. One signal for "document is done", another for "app is done" just here. I guess there will be few more for particular operations.

ukstv:
One more thing: timing to perform an operation. Right now a call to Ceramic node results in a non-interrupted sequence of operations, like changing state, saving state to store, push tip to pubsub, anchor probably. While this all makes sense, I see no reason while all of this should be executed sequentially. rx shaves nicely precious milliseconds of request processing.

PR #862, opened 2021-01-27
fix(core): Fully process incoming tips on first document load

oed:
This is a no-brainer.

### DIFF
--- a/packages/core/src/document.ts
+++ b/packages/core/src/document.ts
@@ -728,9 +728,10 @@ class Document extends EventEmitter implements DocStateHolder {
     // add response timeout for network change
     return new Promise(resolve => {
       let tid: any // eslint-disable-line prefer-const
-      const clear = (): void => {
+      const clear = async (): Promise<void> => {
         clearTimeout(tid)
         this._doctype.off('change', clear)
+        await this._applyQueue.onEmpty()
         resolve()
       }
       tid = setTimeout(clear, 3000)


ukstv (approved):
🏇

oed:
Tested this on my local node. Works well 👍

packages/core/src/document.ts
```diff
@@ -728,9 +728,10 @@ class Document extends EventEmitter implements DocStateHolder {
     // add response timeout for network change
     return new Promise(resolve => {
       let tid: any // eslint-disable-line prefer-const
-      const clear = (): void => {
+      const clear = async (): Promise<void> => {
         clearTimeout(tid)
         this._doctype.off('change', clear)
+        await this._applyQueue.onEmpty()
```
stbrody, line 734: This is interesting because it doesn't wait for the first tip heard about to be processed, but for *all* incoming tips to be processed.  So if the node keeps learning about new tips, even new invalid tips that don't wind up being successfully applied, it will continue to delay returning from the `create` or `load` call that triggered this `wait`.

Spamming messages about invalid tips continues to be an attack vector even without this change, one that we'll have to deal with at some point, but probably not here.  And if you're learning about actually valid tips, then it probably makes sense to continue to delay as long as it takes to get the most recent version of the document.  It means we'll break down if there's ever a constant stream of valid updates to a documents, but we'd break down in other ways anyway if that was the case, if only from the document log getting unusably long.

So all that is to say that I think this change is fine.  It amplifies some already existing rough edges of the system, but doesn't introduce any new ones, and in the majority of cases provides a real benefit to user experience.  LGTM

stbrody (approved):

packages/core/src/document.ts
oed: That's right. This is a problem regardless and needs to be addressed in the tip handling and potentially peering layer.

PR #865, opened 2021-01-28
Organised dependencies

ukstv:
Instead of #845. No controversy, easy stuff here to silence lerna.


package.json
```diff
@@ -10,24 +10,27 @@
     "publish:latest": "lerna publish --allow-branch master --conventional-graduate --create-release github",
     "lint": "lerna run lint",
     "clear-deps": "lerna exec 'rm -rf package-lock.json node_modules' ; rm -rf package-lock.json node_modules",
```
oed, line 12: I guess this line can be removed now that you added `clean:deps`.

packages/blockchain-utils-linking/src/cosmos.ts
```diff
@@ -1,7 +1,7 @@
 import { AccountID } from 'caip';
 import { AuthProvider } from './auth-provider';
 import { getConsentMessage, LinkProof } from './util';
-import { Tx, SignMeta } from '@tendermint/sig';
+import type { Tx, SignMeta } from '@tendermint/sig';
```
oed, line 4: Nice catch!

types/uint8arrays/index.d.ts
```diff
@@ -3,3 +3,11 @@ declare module 'uint8arrays' {
...
     export function concat(bs: Array<Uint8Array>): Uint8Array
 }
+
+declare module 'uint8arrays/concat' {
+    export default function concat(bs: Array<Uint8Array>): Uint8Array
+}
+
+declare module 'uint8arrays/to-string' {
+    export default function toString(b: Uint8Array, enc?: string): string
+}
```
oed, line 13: Doesn't v 2.0.x include types?

oed (approved):

types/uint8arrays/index.d.ts
ukstv: It does, sort of. https://github.com/achingbrain/uint8arrays/issues/13

oed: grr, ok thanks.

PR #866, opened 2021-01-28
chore: Even cleaner dependencies

ukstv:
- Drop ancillary packages like js-sha256, @ethersproject/random, etc.
- Replace typestub-multihashes with multihashes.

Part 2 of #845. The final one.


stbrody (commented):
looks good to me, but I'll leave it to Joel for final approval since he's more familiar with these libraries than I am.

packages/doctype-tile/package.json
```diff
@@ -43,6 +40,7 @@
     "@typescript-eslint/eslint-plugin": "^4.6.0",
     "@typescript-eslint/parser": "^4.6.0",
     "babel-jest": "^26.6.3",
+    "dids": "^1.1.0",
```
oed, line 43: Can we please use 1.1.1?

packages/doctype-tile/src/tile-doctype.ts
```diff
@@ -67,7 +67,7 @@ export class TileDoctype extends Doctype {
      */
     static async makeGenesis<T extends CeramicCommit>(params: DocParams, context: Context): Promise<T> {
         // If 'deterministic' is undefined, default to creating document uniquely
-        const unique = params.deterministic ? '0' : base64Encode(randomBytes(12))
+        const unique = params.deterministic ? '0' : uint8arrays.toString(randomBytes(12), 'base16')
```
oed, line 70: Why was this changed from base64 to base16?

oed (approved):
Lgtm mod comments.

packages/doctype-tile/src/tile-doctype.ts
ukstv: so true. mea culpa

packages/doctype-tile/package.json
ukstv: yup

PR #868, opened 2021-01-30
Clean up public-facing ceramic API

stbrody:


stbrody:
CC @zachferland @valmack @michaelsena

oed:
> Add an `updateDocument` function...

This makes sense to me.

> Remove `applyCommit` from CeramicApi...

I don't get this. Why would the method signatures diverge? We definitely need the `applyCommit` method in the http-client. In general I'd prefer  if core and http-client keep the same api so that you can use them interchangeably. 

> Remove `loadDocumentCommits` from CeramicApi...

I believe this functionality is very important for some of the tools that @PaulLeCam is building. I'll let him fill in here. I don't see any drawback to having this here though. 

> it's functionality could easily be recreated in application code by inspecting the document state after loading it.

This is not true since the DocState object only contains the commitIds not the actual commits.

ukstv:
Do not understand motivation. What is to be achieved or solved here? Please, @stbrody could you elaborate a bit on that?

And +1 to @oed questions.

PaulLeCam:
> Add an `updateDocument` function to `CeramicApi`.  This is important just to make our primary API clear and predictable.  Having create and load in one place but update in another is weird and confusing.  I also mark `Doctype.change()` as deprecated as part of this, but I could convinced to leave that part out and continue to support `Doctype.change()` in perpetuity if others think that is important syntactic sugar to maintain.

I'm all up for deprecating `Doctype.change()` and removing it as soon as possible, I think it's better to have a canonical way to make changes rather than two that would lead to people asking what's the difference. Let's just make sure we update all the docs and examples ASAP.

> Remove `applyCommit` from `CeramicApi` (but leave its implementations in `Ceramic` and `CeramicClient`).  This being in CeramicApi constrains it to having the same signature in both the core and http-client implementations, which ties our hands a little in how we can implement/optimize it for the different client models.  I noticed this while designing the state refactor.  Allowing these two implementations to diverge frees us up a bit.

I already discussed this with Joel and Janko in the past, I think it's problematic to treat the core and HTTP client as "the same" because it limits possibilities on both. My take is that we should have a "frontend" interface used by the apps/clients and a "backend" one that takes care of networking and similar lower-level APIs (ideally with a transport interface in-between but that's another subject).
That said, we're currently widely using the `CeramicApi` type to represent either Ceramic client, so I think making the change you're suggesting would break typechecking in IDX and likely most of other apps and libraries. We still need a type representing this abstract Ceramic "frontend" interface.

> Remove `loadDocumentCommits` from `CeramicApi`.  It's really hard for me to imagine how any end user would actually use this function in a meaningful way.  Even if we were to remove it entirely, it's functionality could easily be recreated in application code by inspecting the document state after loading it.  I almost completely deleted its implementation entirely, but I wound up leaving it in to support the `commits` http endpoint, since I didn't want to change the http API lightly, but the same argument could be made to support deleting it from there as well.  If we really think it's important to keep this syntactic sugar around, I would argue to move it onto `Doctype` and out of `CeramicApi` to keep `CeramicApi` simple and targeted (although the `Doctype` interface is user-facing also, so doing that would only help so much).

I don't think it's something end-users would interact with, but it's useful for tooling. I'm using it to "export" documents to a serialized state so they can be copied over to another Ceramic node for example (still WIP but the use case I'm thinking is equivalent to `mongodump` + `mongorestore` for Ceramic/IDX).
If it's more convenient to have it on the `Doctype` interface no problem making the change, as long as there is a way to keep the functionality it's all good for me.

ukstv:
Actually, this `updateDocument` worries me a lot, as it touches a question of "what is document accessor?" (i.e. how it is modified), and I believe this should be a responsibility of Doctype.

oed:
> Actually, this updateDocument worries me a lot, as it touches a question of "what is document accessor?" (i.e. how it is modified), and I believe this should be a responsibility of Doctype.

Actually I agree with this. The applyCommit method should be enough!

stbrody:
> I don't think it's something end-users would interact with, but it's useful for tooling. I'm using it to "export" documents to a serialized state so they can be copied over to another Ceramic node for example (still WIP but the use case I'm thinking is equivalent to mongodump + mongorestore for Ceramic/IDX).
Very interesting!  I'm fine with putting this back.  I had thought maybe it would make sense to move it to the doctype interface to keep the CeramicApi interface clean and minimal, but the Doctype interface is actually pretty clean and minimal right now, so it's not clear to me that it's better off in one place or the other.  Lacking any particular preference between leaving it in CeramicApi or moving it to Doctype, I'll default to leaving it where it is.  I'll just revert the commit that made any changes to `loadDocumentCommits`

oed:
Thinking a bit more about this. The reason we have the `applyCommit` method on `CeramicApi` is that is can be agnostic of the Doctype. Different document types will have different interfaces for updating the document. The `.change` method is mostly intended for `tile`, I think we can make a better interface for `caip10-link` given the difference in functionalty. Therefore I think that the `updateDocument` function doesn't make that much sense because it would require the CeramicApi to know about implementation details of each doctype.

stbrody:
> I already discussed this with Joel and Janko in the past, I think it's problematic to treat the core and HTTP client as "the same" because it limits possibilities on both. My take is that we should have a "frontend" interface used by the apps/clients and a "backend" one that takes care of networking and similar lower-level APIs (ideally with a transport interface in-between but that's another subject).

I completely agree with this

> I don't get this. Why would the method signatures diverge? We definitely need the applyCommit method in the http-client. In general I'd prefer if core and http-client keep the same api so that you can use them interchangeably.

You need an `applyCommit` http endpoint.  You don't need an `applyCommit` function in the developer-facing code API.  `applyCommit` is logically an implementation detail of `CeramicApi.updateDocument()`/`Doctype.change()`.  If we had a separate public API from internal API as Paul suggests, then this could go there, although I'm still not totally sure it makes sense to require the internal and http implementations to have the same signature when this functionality could easily be abstracted away under `CeramicApi.updateDocument()`



Let me provide more context on what was motivating this change:

`CeramicApi.applyCommit` takes a `DocId` and a `CeramicCommit` (and `DocOpts`, but that's not relevant to the discussion).  It notably does not take the current `DocState`.  This is weird, given that the passed in commit was just created off a specific view of the document state, and will not apply on top of any other state.  This means that the core Ceramic class' applyCommit implementation has to [load the document](https://github.com/ceramicnetwork/js-ceramic/blob/bef3320943823d98a20411023e2516559fd7de83/packages/core/src/ceramic.ts#L363) before applying the commit.  This is wasted and unnecessary work - if you just created a commit to apply you must have already had the `DocState`, why are we throwing it out just to load it again?  So as far as the core API is concerned it would make sense to remove the `DocId` arg and replace it with a `DocState` arg.  If we want to add the `DocState` to the core `Ceramic.applyCommit()` function, however, then we need to also add it to the http client's `CeramicClient.applyCommit()` function as well, since they implement the same interface.  That would require serializing the current DocState over the wire from the http client to the daemon.  Given that that same DocState was most likely just loaded from the daemon, that's redundant information moving across the wire unnecessarily.  So allowing the signatures of `applyCommit()` to diverge between the core and http interface would allow the core implementation to always take the DocState as an argument, while letting the http implementation keep the same behavior as today and send just the docId and rely on the daemon to load the base doc state.

If we don't want this interface to diverge between the http and core implementations, we could just go ahead and make the change that `applyCommit` takes the entire `DocState` as an argument, and accept that we're going to be re-serializing the `DocState` back from the http client to the daemon whenever we go to apply a commit.  The same reasoning behind why we want that in the core implementation does apply to the http implementation as well.  It's a bit more network traffic, but maybe that's not actually huge deal for point-to-point http traffic.  It's certainly less costly than additional pub/sub traffic like we have from loading the document unnecessarily now.  That's probably the direction I'd lean towards going in now.

stbrody:
> Actually, this updateDocument worries me a lot, as it touches a question of "what is document accessor?" (i.e. how it is modified), and I believe this should be a responsibility of Doctype.

Can you elaborate on your concern here?  Please note that the existing `Doctype.change()` approach is a function on the top-level `Doctype` interface, which means that all doctype implementations are beholden to the same function signature.  So if your concern is to allow different doctypes to have different function signatures for performing updates, that is already not possible currently, and would be pretty difficult to enable without a larger change to the design of the system.  The *implementation* of update/change can certainly be different for different doctypes, so long as they use the same signature and take the same arguments, and adding `CeramicApi.updateDocument()` doesn't make that any harder.

stbrody:
Okay, I think I responded to all major comments here.  PTAL @oed @ukstv @PaulLeCam

stbrody:
> Thinking a bit more about this. The reason we have the `applyCommit` method on `CeramicApi` is that is can be agnostic of the Doctype. Different document types will have different interfaces for updating the document. The `.change` method is mostly intended for `tile`, I think we can make a better interface for `caip10-link` given the difference in functionalty. Therefore I think that the `updateDocument` function doesn't make that much sense because it would require the CeramicApi to know about implementation details of each doctype.

Okay, so if the goal is to make all the update functionality doctype-specific, that requires a pretty significant rethink of our HTTP API (we'd likely need different endpoints for each doctype), which will in turn require a major rework of our developer APIs.  If that's the direction we want to be going in, then maybe the thing to do is to go the other direction, and take `createDocument` and `loadDocument` out of `CeramicApi`, and `change()` out of `Doctype`, and put the entire document CRUD API on the specific doctype implementations.

stbrody:
If we want to be moving in the direction of the CRUD API is entirely doctype-specific, I don't think we need to actually add the new HTTP endpoints right now, but I do think we should have our core developer API somewhat stable before mainnet launch, which is why I'm pushing us to think about this now.

ukstv:
Oh, I would love to focus on circular logic here 😈, but it too would miss the point of the conversation.

stbrody:
> Oh, I would love to focus on circular logic here , but it miss the point of the conversation too.

@ukstv, which logic here is circular?  If I'm thinking about this in the wrong way or if you have an idea for a better way to frame the conversation, I'd certainly welcome input!

ukstv:
Yeah-yeah, it just takes time to write down a comprehensive narrative.

ukstv:
Note on terminology:
_Change_ - part of a commit that is meaningful from application/developer/consumer standpoint. Introduced to separate semantics from technicality of _commit_.

Let’s start with document lifecycle. First, as a pure abstraction:
```
Null State --change--> State1 --change--> State2 -->...
```

And here is more real document lifecycle, with some technicalities omitted:
1. Create initial change (as a change from null state),
2. Prepare initial commit (and apply to null state, but we omit application here for illustrative purposes; or create initial state, it is debatable)
3. Prepare a meaningful change (from developer/consumer perspective)
4. Prepare an update commit
5. Apply commit to the document. This involves validation, pubsub, anchoring, changing log of commit-ids, everything to maintain state.

I emphasise here changes, as they are only things that matter to external humans. Also, this emphasis shows a point that there is _something_ that serves as an interface between a human and the commits (or technical details in general).

What actually performs these steps:
1. developer,
2. `**Doctype.makeGenesis` based on parameters from (1),
3. developer (with `**Doctype#change`),
4. `**Doctype#_makeCommit` based on (3),
5. `**DoctypeHandler#applyCommit` after stripped from all the ancillary bells and whistles.

So, here we have `Doctype` as a something that turns human will into commit, and `DoctypeHandler` that  prepares new state of the document. A handler is the arrow on a diagram above.

Of these, `DoctypeHandler` is the only necessary component. `Doctype` is a mere human interface. You note this when you find that `.makeGenesis` is called from obviously human `Ceramic#createDocument`. And `_makeCommit` only happens in http client, which is human too.

I would declare the current situation as human component intertwined with machine one, while they better lay one on top of the other.

This was a prelude. Now the real thing.

Constraints:
* Commits are necessary, they are part of the system. One can not get away with it. One can only abstract over it. S.U.: And the abstraction better not leak.
* DoctypeHandler of some sort is necessary as a state-transition mechanism.

Based on this, one could model API based on commits only. Yet, part of the system interacts with humans, so

Desideratum:
* Some form of human API abstracts over commits and ever-changing state.

--

The whole discussion up here for me is an attempt to figure what kind of abstraction seems more appropriate. One way is to focus on machine part while laying human part on top (`applyCommit` and richer doctype direction, with richer doctype means domain-specific `Doctype#change` or no `Doctype#change` at all). Another way is to bring human part even closer to the core with `updateDocument`, which intertwines the parts even more.

I believe the two approaches are interchangeable if done properly. The only difference is how complex they will be.  `updateDocument` way of changes introduces more complexity by tangling human and machine components together.

For human-friendliness one better properly enrich Doctype methods, and eliminate or move somewhere from CeramicAPI human-related functions like `#createDocument` . With proper Doctype `Ceramic#createDocumentFromGenesis` would work just all right.

—

Now to the fancy points:
> This means that the core Ceramic class’ applyCommit implementation has to  [load the document](https://github.com/ceramicnetwork/js-ceramic/blob/bef3320943823d98a20411023e2516559fd7de83/packages/core/src/ceramic.ts#L363)  before applying the commit

Unless it is a virgin node receiving its first commit out of the blue, it has already state in memory. 

> …your concern is to allow different doctypes to have different function signatures for performing updates, that is already not possible currently, and would be pretty difficult to enable without a larger change to the design of the system.

If Doctype handles ever-changing state rather than a snapshot of uncertain freshness, this looks rather possible to me. It is state refactor time anyway.

> …if the goal is to make all the update functionality doctype-specific, that requires a pretty significant rethink of our HTTP API…

The system is a per-document state machine, not a CRUD app, `applyCommit`  is enough.

ukstv:
Would like here to have input from @PaulLeCam: What do you dislike in `Doctype#change`?

stbrody:
> Unless it is a virgin node receiving its first commit out of the blue, it has already state in memory.

Unless the state has aged out of the cache.  But yeah, your point is still valid.  99.9% of the time it will be in memory, so we could add some code to always check the state store first and only if it isn't there do we go to the network to load the document.  We would still have to handle the case though where the state no longer matches the state for which the commit was created.  I guess at that point it should be left to the doctype-specific conflict resolution rules.

> The system is a per-document state machine, not a CRUD app, applyCommit is enough.

Yep, you're totally right, not sure why I got it in my head we needed to change the HTTP api.


So I think we're basically in agreement as to the core problem(s).  The main one being that CeramicApi is a conflation of an internal and user-facing API.  I'm fine with scrapping the `updateDocument` function in favor of moving all CRUD operations onto the specific Doctype implementations.  My main concern was to make sure the developer-facing API was clear and in one place.

stbrody:
This PR was opened with 3 separate commits/concerns to discuss.  I've been convinced that we don't need to do anything for the latter 2.  The first was always the most important in my mind, and that's to clarify our developer-focused API.  I propose we close this PR without merging anything in it, and use the new [Finalize developer API](https://www.notion.so/threebox/Finalize-developer-API-59c989317bae4a81b6f6312b5312552e) I filed in Notion to plan a project to move the user-facing CRUD API off of both `CeramicApi` and `Doctype` and into the specific doctype implementations.

PaulLeCam:
> Would like here to have input from @PaulLeCam: What do you dislike in `Doctype#change`?

I don't mind `Doctype#change`, I just wanted to avoid having both this and `Ceramic#updateDocument` as 2 APIs doing the same thing.

ukstv:
@PaulLeCam Got it, thanks.

PR #875, opened 2021-02-02
Recover anchors on startup

ukstv:
BTW, shaves 5-20 ms from `POST /documents` request as tested locally.

Replaces #860

### DIFF
--- a/packages/common/src/anchor-service.ts
+++ b/packages/common/src/anchor-service.ts
@@ -1,43 +1,80 @@
 import CID from "cids";
+import type { Observable } from "rxjs";
+import { AnchorProof, AnchorStatus } from "./doctype";
+import { CeramicApi } from "./ceramic-api";
+import DocID from "@ceramicnetwork/docid";
 
-import { EventEmitter } from "events";
-import { AnchorProof } from "./doctype"
-import { CeramicApi } from "./ceramic-api"
+export interface AnchorServicePending {
+  readonly status: AnchorStatus.PENDING;
+  readonly docId: DocID;
+  readonly cid: CID;
+  readonly message: string;
+  readonly anchorScheduledFor: number;
+}
+
+export interface AnchorServiceProcessing {
+  readonly status: AnchorStatus.PROCESSING;
+  readonly docId: DocID;
+  readonly cid: CID;
+  readonly message: string;
+}
+
+export interface AnchorServiceAnchored {
+  readonly status: AnchorStatus.ANCHORED;
+  readonly docId: DocID;
+  readonly cid: CID;
+  readonly message: string;
+  readonly anchorRecord: CID;
+}
+
+export interface AnchorServiceFailed {
+  readonly status: AnchorStatus.FAILED;
+  readonly docId: DocID;
+  readonly cid: CID;
+  readonly message: string;
+}
+
+/**
+ * Describes anchor service response
+ */
+export type AnchorServiceResponse =
+  | AnchorServicePending
+  | AnchorServiceProcessing
+  | AnchorServiceAnchored
+  | AnchorServiceFailed;
 
 /**
  * Describes anchoring service behavior
  */
-export abstract class AnchorService extends EventEmitter {
+export abstract class AnchorService {
+  /**
+   * Performs whatever initialization work is required by the specific anchor service implementation
+   */
+  abstract init(): Promise<void>;
 
-    /**
-     * Performs whatever initialization work is required by the specific anchor service implementation
-     */
-    abstract init(): Promise<void>;
+  /**
+   * Set Ceramic API instance
+   *
+   * @param ceramic - Ceramic API used for various purposes
+   */
+  abstract set ceramic(ceramic: CeramicApi);
 
-    /**
-     * Set Ceramic API instance
-     *
-     * @param ceramic - Ceramic API used for various purposes
-     */
-    abstract set ceramic(ceramic: CeramicApi);
+  /**
+   * Request anchor commit on blockchain
+   * @param docId - Document ID
+   * @param tip - CID tip
+   */
+  abstract requestAnchor(docId: DocID, tip: CID): Observable<AnchorServiceResponse>;
 
-    /**
-     * Request anchor commit on blockchain
-     * @param docId - Document ID
-     * @param tip - CID tip
-     */
-    abstract requestAnchor(docId: string, tip: CID): Promise<void>;
+  /**
+   * Validate anchor proof commit
+   * @param anchorProof - Proof of blockchain inclusion
+   */
+  abstract validateChainInclusion(anchorProof: AnchorProof): Promise<void>;
 
-    /**
-     * Validate anchor proof commit
-     * @param anchorProof - Proof of blockchain inclusion
-     */
-    abstract validateChainInclusion(anchorProof: AnchorProof): Promise<void>;
-
-    /**
-     * @returns An array of the CAIP-2 chain IDs of the blockchains that are supported by this
-     * anchor service.
-     */
-    abstract getSupportedChains(): Promise<Array<string>>;
-
+  /**
+   * @returns An array of the CAIP-2 chain IDs of the blockchains that are supported by this
+   * anchor service.
+   */
+  abstract getSupportedChains(): Promise<Array<string>>;
 }
--- a/packages/common/src/anchor-service.ts
+++ b/packages/common/src/anchor-service.ts
@@ -46,35 +46,35 @@ export type AnchorServiceResponse =
 /**
  * Describes anchoring service behavior
  */
-export abstract class AnchorService {
+export interface AnchorService {
   /**
    * Performs whatever initialization work is required by the specific anchor service implementation
    */
-  abstract init(): Promise<void>;
+  init(): Promise<void>;
 
   /**
    * Set Ceramic API instance
    *
    * @param ceramic - Ceramic API used for various purposes
    */
-  abstract set ceramic(ceramic: CeramicApi);
+  ceramic: CeramicApi;
 
   /**
    * Request anchor commit on blockchain
    * @param docId - Document ID
    * @param tip - CID tip
    */
-  abstract requestAnchor(docId: DocID, tip: CID): Observable<AnchorServiceResponse>;
+  requestAnchor(docId: DocID, tip: CID): Observable<AnchorServiceResponse>;
 
   /**
    * Validate anchor proof commit
    * @param anchorProof - Proof of blockchain inclusion
    */
-  abstract validateChainInclusion(anchorProof: AnchorProof): Promise<void>;
+  validateChainInclusion(anchorProof: AnchorProof): Promise<void>;
 
   /**
    * @returns An array of the CAIP-2 chain IDs of the blockchains that are supported by this
    * anchor service.
    */
-  abstract getSupportedChains(): Promise<Array<string>>;
+  getSupportedChains(): Promise<Array<string>>;
 }
--- a/packages/core/src/__tests__/ceramic-recover-documents.test.ts
+++ b/packages/core/src/__tests__/ceramic-recover-documents.test.ts
@@ -1,0 +1,154 @@
+import IPFS from "ipfs-core";
+import Ceramic from "../ceramic";
+import { Ed25519Provider } from "key-did-provider-ed25519";
+import tmp from "tmp-promise";
+import getPort from "get-port";
+import { DoctypeUtils, DocState, Doctype, IpfsApi, AnchorStatus } from "@ceramicnetwork/common";
+
+import dagJose from "dag-jose";
+import basicsImport from "multiformats/cjs/src/basics-import.js";
+import legacy from "multiformats/cjs/src/legacy.js";
+import * as u8a from "uint8arrays";
+
+const PUBSUB_TOPIC = "/ceramic/inmemory/test";
+const SEED = u8a.fromString("6e34b2e1a9624113d81ece8a8a22e6e97f0e145c25c1d4d2d0e62753b4060c83", "base16");
+
+/**
+ * Create an IPFS instance
+ * @param overrideConfig - IPFS config for override
+ */
+const createIPFS = (overrideConfig: Record<string, unknown> = {}): Promise<any> => {
+    basicsImport.multicodec.add(dagJose);
+    const format = legacy(basicsImport, dagJose.name);
+
+    const config = {
+        ipld: { formats: [format] },
+    };
+
+    Object.assign(config, overrideConfig);
+    return IPFS.create(config);
+};
+
+const expectEqualStates = (state1: DocState, state2: DocState): void => {
+    expect(DoctypeUtils.serializeState(state1)).toEqual(DoctypeUtils.serializeState(state2));
+};
+
+const anchor = async (ceramic: Ceramic): Promise<void> => {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    await ceramic.context.anchorService.anchor();
+};
+
+const syncDoc = async (doctype: Doctype): Promise<void> => {
+    await new Promise<void>((resolve) => {
+        doctype.on("change", () => {
+            resolve();
+        });
+    });
+};
+
+async function createCeramic(ipfs: IpfsApi, pinsetDirectory: string) {
+    const ceramic = await Ceramic.create(ipfs, {
+        pinsetDirectory: pinsetDirectory,
+        anchorOnRequest: false,
+        pubsubTopic: PUBSUB_TOPIC, // necessary so Ceramic instances can talk to each other
+    });
+    const provider = new Ed25519Provider(SEED);
+    await ceramic.setDIDProvider(provider);
+    return ceramic;
+}
+
+jest.setTimeout(60000);
+let ipfs1: IpfsApi;
+let ipfs2: IpfsApi;
+let ipfs3: IpfsApi;
+let multaddr1: string;
+let multaddr2: string;
+let multaddr3: string;
+let tmpFolder: any;
+
+const DOCTYPE_TILE = "tile";
+
+let p1Start = 4000;
+let p2Start = 4100;
+let p3Start = 4200;
+
+const pOffset = 100;
+
+let port1: number;
+let port2: number;
+let port3: number;
+
+beforeEach(async () => {
+    tmpFolder = await tmp.dir({ unsafeCleanup: true });
+
+    const buildConfig = (path: string, port: number): Record<string, unknown> => {
+        return {
+            repo: `${path}/ipfs${port}/`,
+            config: {
+                Addresses: { Swarm: [`/ip4/127.0.0.1/tcp/${port}`] },
+                Bootstrap: [],
+            },
+        };
+    };
+
+    const findPort = async (start: number, offset: number): Promise<number> => {
+        return await getPort({ port: getPort.makeRange(start + 1, start + offset) });
+    };
+
+    [port1, port2, port3] = await Promise.all([p1Start, p2Start, p3Start].map((start) => findPort(start, pOffset)));
+    [ipfs1, ipfs2, ipfs3] = await Promise.all(
+        [port1, port2, port3].map((port) => createIPFS(buildConfig(tmpFolder.path, port)))
+    );
+    [p1Start, p2Start, p3Start] = [p1Start, p2Start, p3Start].map((start) => start + pOffset);
+
+    multaddr1 = (await ipfs1.id()).addresses[0].toString();
+    multaddr2 = (await ipfs2.id()).addresses[0].toString();
+    multaddr3 = (await ipfs3.id()).addresses[0].toString();
+
+    const id1 = await ipfs1.id();
+    const id2 = await ipfs2.id();
+    const id3 = await ipfs3.id();
+    multaddr1 = id1.addresses[0].toString();
+    multaddr2 = id2.addresses[0].toString();
+    multaddr3 = id3.addresses[0].toString();
+});
+
+afterEach(async () => {
+    await ipfs1.stop(() => console.log("IPFS1 stopped"));
+    await ipfs2.stop(() => console.log("IPFS2 stopped"));
+    await ipfs3.stop(() => console.log("IPFS3 stopped"));
+
+    await tmpFolder.cleanup();
+});
+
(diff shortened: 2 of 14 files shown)


packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -97,179 +95,216 @@ export default class EthereumAnchorService extends AnchorService {
...
+  }
+
+  /**
+   * Start polling for CidDocPair mapping
+   * @param cidDoc - CID to Doc mapping
+   * @param pollTime - Single poll timeout
+   * @param maxPollingTime - Global timeout for max polling in milliseconds
+   * @private
+   */
+  private poll(
```
stbrody, line 177: we've been using the convention of a leading underscore for private method names, can you put that back?

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -97,179 +95,216 @@ export default class EthereumAnchorService extends AnchorService {
...
+    const requestUrl = [this.requestsApiEndpoint, cidDoc.cid.toString()].join(
+      "/"
+    );
+
+    return interval(DEFAULT_POLL_TIME).pipe(
+      concatMap(async () => {
+        const now = new Date().getTime();
+        if (now > maxTime) {
+          throw new Error("Exceeded max timeout");
+        } else {
```
stbrody, line 193: Can remove the `else` statement and remove a layer of indentation

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -97,179 +95,216 @@ export default class EthereumAnchorService extends AnchorService {
...
+          const response = await fetch(requestUrl);
+          if (response.status === HTTP_STATUS_NOT_FOUND) {
```
stbrody, line 195: hmm, I wonder if we should be handling all failed statuses here, not just `HTTP_STATUS_NOT_FOUND`...

Probably out of scope for this change though

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -97,179 +95,216 @@ export default class EthereumAnchorService extends AnchorService {
...
         }
+      })
+    );
+  }
+
+  /**
+   * Validate anchor proof on the chain
+   * @param anchorProof - Anchor proof instance
+   */
+  async validateChainInclusion(anchorProof: AnchorProof): Promise<void> {
```
stbrody, line 210: This is already a large and complex review.  Can we leave formatting changes out of it and do them in a dedicated review if needed?  As-is there's lots of noise in this review from tabbing changes that aren't relevant to the core change.  I don't really have strong feelings about number of spaces to indent with anyway, so I would rather just leave it as-is until we adapt an autoformatter to handle formatting for us.

packages/core/src/anchor/memory/in-memory-anchor-service.ts
```diff
@@ -1,32 +1,36 @@
...
-  }
+  constructor(
+    readonly cid: CID,
+    readonly docId?: DocID,
+    readonly log?: CID[]
+  ) {}
 
   get key(): string {
-    return this.docId + this.did
+    return this.docId.toString();
```
stbrody, line 32: whooops, I thought I'd done that already.  Nice catch

packages/core/src/anchor/memory/in-memory-anchor-service.ts
```diff
@@ -35,134 +39,143 @@ interface InMemoryAnchorConfig {
...
-  private readonly _verifySignatures: boolean
+class InMemoryAnchorService implements AnchorService {
+  #ceramic: Ceramic;
+  #dispatcher: Dispatcher;
 
-  private _queue: Candidate[] = []
+  readonly #anchorDelay: number;
+  readonly #anchorOnRequest: boolean;
+  readonly #verifySignatures: boolean;
+  readonly #feed: Subject<AnchorServiceResponse> = new Subject();
```
stbrody, line 58: So as best I can tell `Subject` is the same as `Observable` but allows multiple subscribers instead of just one.  Why do we need to support multiple subscribers here?

packages/core/src/anchor/memory/in-memory-anchor-service.ts
```diff
@@ -35,134 +39,143 @@ interface InMemoryAnchorConfig {
...
+    const groupedCandidates = await this._groupCandidatesByDocId(this.#queue);
+    return this._selectValidCandidates(groupedCandidates);
   }
 
-  async _groupCandidatesByDocId(candidates: Candidate[]): Promise<Record<string, Candidate[]>> {
+  async _groupCandidatesByDocId(
+    candidates: Candidate[]
+  ): Promise<Record<string, Candidate[]>> {
     const result: Record<string, Candidate[]> = {};
+    await Promise.all(
```
stbrody, line 105: Nice improvement!

packages/core/src/anchor/memory/in-memory-anchor-service.ts
```diff
@@ -35,134 +39,143 @@ interface InMemoryAnchorConfig {
...
-      result.push(selected)
+      result.push(selected);
     }
 
-    return result
+    return result;
   }
 
-  async _failCandidate(candidate: Candidate, message?: string): Promise<void> {
+  _failCandidate(candidate: Candidate, message?: string): void {
```
stbrody, line 169: So I had read somewhere when I was first teaching myself about javascript/typescript a few months ago that it's a bad idea to mix sync and async calls.  I don't totally remember why it was a problem though, something about how the node event loops works.  Do you know what I'm referring to?  Is this something I shouldn't actually be worried about?

packages/core/src/anchor/memory/in-memory-anchor-service.ts
```diff
@@ -204,22 +221,46 @@ class InMemoryAnchorService extends AnchorService {
...
-
-    if (this._anchorOnRequest) {
-      await this._process(candidate)
+  requestAnchor(docId: DocID, tip: CID): Observable<AnchorServiceResponse> {
+    const candidate = new Candidate(tip, docId);
+    const feed$ = this.#feed.pipe(
+      filter((asr) => asr.docId.equals(docId) && asr.cid.equals(tip))
+    );
+    if (this.#anchorOnRequest) {
+      requestAnchorQueue
```
stbrody, line 239: why do the anchor requests need to get serialized here?  What's wrong with letting them happen in parallel like they are currently?

packages/core/src/anchor/memory/in-memory-anchor-service.ts
stbrody: Okay, I see now, it's because we send all the anchor requests through this single feed, then filter the global feed to individual per-request observables that we return from `requestAnchor`.  So then why have this single unified `#feed` at all?  The ethereum-anchor-service doesn't do this - why can't the in-memory anchor service do the same and just have a bunch of totally separate per-request observables without ever serializing them into a single feed?

packages/core/src/ceramic.ts
```diff
@@ -72,6 +72,7 @@ export interface CeramicConfig {
   cacheDocCommits?: boolean; // adds 'docCacheLimit' additional cache entries if commits can be cached as well
 
   useCentralizedPeerDiscovery?: boolean;
+  restoreDocuments?: boolean;
```
stbrody, line 75: I'm a little sad to be adding testing-only configuration options here.  Not the end of the world certainly, but would be nice to avoid if we can do so without causing too much pain in our tests.  I'm actually kind of surprised that this causes issues in the tests - I would have expected the `restoreDocuments()` call to be a no-op because it never actually finds any pinned documents that need restoring.

stbrody (commented):
Overall this looks very nice.  A bunch of little comments, mostly style nitpicks and questions to firm up my understanding of rxjs, but nothing major.  I'm definitely starting to see the value of rxjs more as I understand this review better - the improved error handling story alone is pretty motivating.

My biggest concern remains whether this is the right time to roll out rxjs.  I'm a little hesitant to put this in without pretty quickly thereafter following it up with removing EventEmitter and replacing it with rxjs throughout the entire codebase.  Otherwise there's two mental models for handling events and synchronization that developers need to get familiar with to operate in our code.  So I wonder if we decide to go in this direction if we should put together a project soon to fully switch over.  If not, then do we want to hold off on putting rxjs *anywhere* until we can roll it out throughout the codebase?  Would be interested in your thoughts here @oed

packages/common/src/anchor-service.ts
```diff
@@ -1,43 +1,80 @@
...
+  readonly status: AnchorStatus.FAILED;
+  readonly docId: DocID;
+  readonly cid: CID;
+  readonly message: string;
+}
+
+/**
+ * Describes anchor service response
+ */
+export type AnchorServiceResponse =
```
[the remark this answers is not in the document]
ukstv: Type-safety. Less opportunity to shoot yourself in the foot. Here it allows one to... Whoops, it does not, yet.

Now it does with one more commit.

From just one single ASR you could not figure what fields are expected to be populated for what status. Source of confusion, source of errors, source of delays. It too quickly deteriorates into stringly-typed programming. With this change we have total clarity regarding fields, and compiler [helps to remind](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#union-exhaustiveness-checking) if some of the variants one forgets to handle.

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
oed: Just fyi, the autoformatter (pretty) we use in other repos enforce 2 spaces indention.

packages/core/src/anchor/memory/in-memory-anchor-service.ts
oed: Why is that relevant here? This function is completely sync so it should not be async.

oed (commented):
This looks mostly good to me :) 

> My biggest concern remains whether this is the right time to roll out rxjs. I'm a little hesitant to put this in without pretty quickly thereafter following it up with removing EventEmitter and replacing it with rxjs throughout the entire codebase. Otherwise there's two mental models for handling events and synchronization that developers need to get familiar with to operate in our code.

@stbrody I don't feel super strongly that this needs to happen fast. However we might make a large chunk of it part of the state refactor?

packages/core/src/__tests__/ceramic-recover-documents.test.ts
```diff
@@ -0,0 +1,154 @@
...
+    // Retrieve after being closed
+    const ceramic2 = await createCeramic(ipfs2, pinsetDirectory);
+
+    const doc2 = await ceramic2.loadDocument(doc1.id);
+    expect(doc2.state.anchorStatus).toEqual(AnchorStatus.PENDING);
+    // doc2 is exact replica of doc1
+    expectEqualStates(doc1.state, doc2.state);
+    // Now CAS anchors
+    await anchor(ceramic2);
+    await syncDoc(doc2);
```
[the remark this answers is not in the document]
ukstv: Good catch. I guess, for this purpose, InMemoryAnchor service adds a timeout when anchoring. Anyway, the whole `ceramic.test.ts` is to be fixed at least.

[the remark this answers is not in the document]
stbrody: Yep that's right.  `ceramic.test.ts` hasn't been updated yet since we've started switching to the new paradigm to prevent races

packages/core/src/document.ts
```diff
@@ -584,56 +585,55 @@ class Document extends EventEmitter implements DocStateHolder {
...
-      }
-    })
-    await this._context.anchorService.requestAnchor(this.id.toString(), this.tip)
-    const state = this._doctype.state
-    state.anchorStatus = AnchorStatus.PENDING
-    this._doctype.state = state
+    const anchorStatus$ = this._context.anchorService.requestAnchor(this.id.baseID, this.tip);
+    const subscription = anchorStatus$
+        .pipe(
+            concatMap(async (asr) => {
```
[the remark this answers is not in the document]
ukstv: Unlike `mergeMap`, concatMap stops the feed propagation further until the function inside `concatMap` is done.
`interval` inside EthereumAnchorService pushes events every N seconds. I would like to be sure, that processing of those happens sequentially, instead of having multiple request-processing instances at a time.

packages/core/src/subscription-set.ts
```diff
@@ -0,0 +1,27 @@
...
+ * If a subscription gets unsubscribed, it auto-removes from the set.
+ */
+export class SubscriptionSet {
+  constructor(readonly subscriptions: Set<Subscription> = new Set()) {}
+
+  /**
+   * Start tracking a subscription, so that it is closed with others on +close+.
+   */
+  add(subscription: Subscription) {
+    subscription.add(() => {
```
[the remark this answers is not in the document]
ukstv: I was as baffled as you when I read it first.

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
ukstv: Okay, but this convention makes me sad, like Hungarian notation.

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
ukstv: #879

packages/core/src/anchor/memory/in-memory-anchor-service.ts
ukstv: The logic here is different. Unlike independent requests for ethereuem, all of these requests are triggered by  single call to `#anchor`. That's why.

packages/core/src/anchor/memory/in-memory-anchor-service.ts
ukstv: https://nodejs.org/en/docs/guides/blocking-vs-non-blocking/ - irrelevant here. Nothing here relies on event-loop.

packages/core/src/ceramic.ts
ukstv: You would be surprised to know, that the whole Inmemory anchor service is just for test purpose only.

ukstv: This is the least intrusive option to avoid re-mocking tests that apparently heavily rely on state store being fully mocked.

ukstv: And, honestly, if the initialisation of Ceramic is done differently, this should go away.

packages/core/src/anchor/memory/in-memory-anchor-service.ts
stbrody: Ah right, that makes sense.  The in-memory anchor service supports `anchorOnRequest:false` which the eth anchor service does not.

packages/core/src/document.ts
```diff
@@ -584,56 +586,57 @@ class Document extends EventEmitter implements DocStateHolder {
...
+              if (!asr.cid.equals(requestTip)) {
```
stbrody, line 598: Can we get rid of this now that we're getting an observable per anchor request instead of a single event feed from the entire anchor service?  The Observable returned from `AnchorService.requestAnchor` should be able to only emit events that are responses to the specific anchor request right?

packages/core/src/anchor/memory/in-memory-anchor-service.ts
stbrody: Yeah, that must have been what I was remembering.  Guess I was overthinking it.  Thanks!

packages/core/src/ceramic.ts
stbrody: > You would be surprised to know, that the whole Inmemory anchor service is for test purpose only.

This is the first InMemoryAnchorService-specific config option to show up in this `CeramicConfig` interface, that's why I'm calling it out.

stbrody: > This is the first InMemoryAnchorService-specific config option to show up in this CeramicConfig interface, that's why I'm calling it out.

I guess there's also `anchorOnCommit`, which doesn't actually show up explicitly here in CeramicConfig, but is implicitly a member as it is passed to the InMemoryAnchorService via this object because `CeramicConfig` allows arbitrary extra options.  I don't really like that either, but maybe doing something similar would be better than polluting this highly user-visible configuration interface with testing-only parameters

stbrody: > This is the least intrusive option to avoid re-mocking tests that apparently heavily rely on state store being fully mocked.

So the issue is that many of the unit tests don't have a real state store, so in order to get the tests to pass without this option, you'd need to update the mocked state store in those tests to expect the initial call to `stateStore.list()` and return an empty list?

Assuming that is indeed the issue, then I think it's fair to want to avoid doing that in a large number of our tests.  I think in that case the thing to do would be to remove this argument from `CeramicConfig` and handle it similarly to how `anchorOnCommit` is handled.  Would be nice to also set it to default to false if we're running in testing mode, similar to how the `useCentralizedPeerDiscovery` default is set

stbrody: > And, honestly, if the initialisation of Ceramic is done differently, this should go away.

How would that work?

packages/core/src/document.ts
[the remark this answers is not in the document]
stbrody: interesting.  So `pipe` doesn't process the events sequentially already?  It just seems weird to me since the docs for `concatMap` all talk about merging multiple input observables into one output observable, but here we only have 1 input observable.  Is there no more specific and semantically meaningful operator to take a single observable and just tell it to finish processing one event before starting to process the next?

stbrody (commented):
Okay, I think I responded to all the comment threads that you responded to, though there are still at least a few comments from my original pass that didn't get a response yet

packages/core/src/document.ts
ukstv: True.

ukstv:
"Is there no more specific and semantically meaningful operator to take a single observable and just tell it to finish processing one event before starting to process the next" Please, go find and propose one, if you feel like there is a better way.

ukstv:
> S.U.: And, honestly, if the initialisation of Ceramic is done differently, this should go away
> S.B.: How would that work?

As I told you before, separate parameter configuration based on input params from actual initialisation. You have external parameters, internal configuration, and components initialisation as three separate entities.

> but maybe doing something similar would be better than polluting

Spencer, please! Gorilla InMemoryAnchorService exists just for testing purposes, yet it is a full-fledged part of the production code. InMemoryAnchorService's totally opaque parameters, that are not declared anywhere, are casted from `config as any`. Tests call `.anchor` on it after telling compiler, "please, close your eyes, here is a dirty thing going". _This_ is pollution.

Some day someone does proper initialisation. For now, we have both `restoreDocuments` and `useCentralizedPeerDiscovery` as part of config. Oops, mind doing `git blame` for the latter?

oed:
Seems like we need to do some initialization cleanup at some point for sure. I'd rather not get stuck on if we should or shouldn't use a param only for testing rn though, unless there is a strong reason for either :)

oed:
> For now, we have both restoreDocuments and useCentralizedPeerDiscovery as part of config. Oops, mind doing git blame for the latter?

Hey! Let's try to not point fingers here. Seems like Spencer is trying to shoot for the ideal, and maybe that's not the right thing here now, but please try to be more friendly when interacting other people :)

stbrody:
> "Is there no more specific and semantically meaningful operator to take a single observable and just tell it to finish processing one event before starting to process the next" Please, go find and propose one, if you feel like there is a better way.

I'm asking because I'm not familiar at all with rxjs, so I was wondering if you knew more about the available APIs.  So I guess you're saying that the standard way in rxjs to make an event stream fully process one event before beginning to process the next is to always use `mergeMap`, even if there's actually only one stream and therefore nothing to "merge"?

stbrody:
> Gorilla InMemoryAnchorService exists just for testing purposes, yet it is a full-fledged part of the production code. InMemoryAnchorService's totally opaque parameters, that are not declared anywhere, are casted from `config as any`. Tests call `.anchor` on it after telling compiler, "please, close your eyes, here is a dirty thing going". _This_ is pollution.

It's true that a bunch of things about the `InMemoryAnchorService` are ugly, especially the part that it ships as part of our production code when it is only intended for testing, but the difference here is that none of the config options for `InMemoryAnchorService` are part of our publicly documented APIs.  `restoreDocuments` would be the first.  I'm fine with a certain amount of ugliness under the covers, but generally try to be more careful with what shows up in our user-visible APIs.

> Some day someone does proper initialisation. For now, we have both `restoreDocuments` and `useCentralizedPeerDiscovery` as part of config.

`useCentralizedPeerDiscovery` controls public behavior, it makes sense for it to be part of the public configuration API.  We change the default based on whether or not we're in testing mode, but the actual behavior being configured is relevant outside of tests.

oed:
Is there any blockers here besides bike shedding around constructor arguments?

ukstv:
> I'm asking because I'm not familiar at all with rxjs, so I was wondering if you knew more about the available APIs. So I guess you're saying that the standard way in rxjs to make an event stream fully process one event before beginning to process the next is to always use mergeMap, even if there's actually only one stream and therefore nothing to "merge"?

A `Promise` is an `Observable`, semantically. The difference is Observable is a feed of generally unknown length,  while Promise is guaranteed to emit one value. So, `concatMap`/`mergeMap` actually merges two observables, even if Observable nature of Promise is hidden behind technicalities.

packages/core/src/anchor/memory/in-memory-anchor-service.ts
stbrody: I'm still curious about this for my own understanding in case I'm missing something, but it's not a blocker, especially since this is test-only code

stbrody (approved):

ukstv:
> I'm still curious about this for my own understanding in case I'm missing something, but it's not a blocker, especially since this is test-only code

Actually, this is a totally legit concern. Fixed. Now anchors go in parallel.

packages/core/src/anchor/memory/in-memory-anchor-service.ts
ukstv: Aargh, should be posted here. Valid concern. Fixed.

ukstv:
@oed approval to merge this?

oed (approved):

ukstv:
Yahoo!

PR #878, opened 2021-02-02
feat(core): Add new logger package

stbrody:
Based on the one currently in the CAS.

Just putting it in place and demonstrating how it will get wired in, configured, and initialized.  Still lots of work to go to actually set up the necessary event-specific loggers and replace the existing logger package


oed (approved):
lgtm.
I guess you need to add the package to [tsconfig.json](https://github.com/ceramicnetwork/js-ceramic/blob/develop/tsconfig.json) to make the tests pass.

packages/common/src/logger-provider.ts
```diff
@@ -1,279 +1,38 @@
...
-    level: 'info',
-    format: 'text',
-    stacktrace: {
-        levels: ['trace', 'warn', 'error'],
-        depth: 3,
-        excess: 0,
-    },
-}
+// Invalid to access before calling `initializeLogger()`
+export let logger: DiagnosticsLogger
```
valmack, line 6: Why not just `export const` after it is initialized?

packages/common/src/logger-provider.ts
```diff
@@ -1,279 +1,38 @@
...
-        prefix.apply(log, {
-            format(level, name, timestamp) {
-                return LoggerProvider._toText(timestamp, level, name)
-            },
-            timestampFormatter(date) {
-                return date.toISOString()
-            }
-        })
+        const logLevel: LogLevel = typeof config.logLevel == "string" ? logLevelMapping[config.logLevel] : config.logLevel
+        logger = new DiagnosticsLogger(config.logPath, logLevel, config.logToFiles);
```
valmack, line 36: Why does logger need to be assigned here?

Why not something like
```
export const logger = new LoggerProvider()
```

valmack: Oh I see because the config options get passed in when we create the Ceramic instance. So would this be a problem if I import a package from the monorepo without actually starting a Ceramic instance? Then the logger would break...?

valmack: Ok so probably better to use different loggers so that we don't require that. For example if I want to import @ceramicnetwork/ipfs I shouldn't be required to call `LoggerProvider.init()` to use it, or @ceramicnetwork/docid, etc.

stbrody: Hmm... what exactly are you proposing?  There has to be **somewhere** that handles the configuration of what file path to put the logs.  Either we have one central place, but then all packages that use the logger are dependent on that central place to configure the logger appropriately, or every package that uses the logger needs to parse the config file and configure the logging path themselves, making each package dependent on our config file format.

stbrody:
weirdly all the tests pass for me locally, not sure why they keep failing in CI.  Are these new errors because the package hasn't been published to NPM? @oed

packages/common/src/logger-provider.ts
valmack: So it seems like each package needs to configure its own logger. And yes need to be able to read config. I would propose env vars, unless it's the cli where we get config from commander cli flags.

oed:
>  Are these new errors because the package hasn't been published to NPM? 
@stbrody they shouldn't be. Are you sure you have the package in package.json everywhere which it's used?

packages/common/src/logger-provider.ts
stbrody: Okay, I think the core problem was trying to export a global logger instance that could be imported directly.  I changed `LoggerProvider` to be a more traditional factory so everything that needs a logger instance will need to explicitly construct what it needs with whatever config it wants.  For now only the diagnostic logger is created, but I imagine `Ceramic.create()` eventually creating specific ServiceLoggers and giving them to the services that need them.  For instance I imagine there will be an `httpLogger` that is given to the `Dispatcher` in its constructor.  My hope is that only the `DiagnosticLogger` is global-enough to be on the `Context` and the other loggers will all be service-specific.

valmack (approved):
LGTM

PR #882, opened 2021-02-04
feat: Export auth providers

ukstv:


oed (approved):
sweet

packages/blockchain-utils-linking/src/index.ts
```diff
@@ -5,3 +5,9 @@ export * as filecoin from "./filecoin";
 export * as polkadot from "./polkadot";
 export * as eosio from "./eosio";
 export * as cosmos from "./cosmos";
```
oed, line 7: Any reason to still do these exports?

ukstv: One has to export helpers, like "prepare canonical message for signing". These are per-blockchain, and I would like to avoid name clash.

PR #885, opened 2021-02-04
Add explicit dependency on 'events' package

stbrody:


oed (approved):
This the only package it's needed?

stbrody:
> This the only package it's needed?

I *think* so?  I didn't exhaustively search, but I expect any of our packages using EventEmitters probably already depend on `common`

stbrody:
Looks like "core" was the only other package using EventEmitter, so I gave it an explicit dependency on events as well

PR #886, opened 2021-02-04
fix(core): Add ipfs timeout everywhere we get from the dag

stbrody:


packages/cli/src/build-ipfs-connection.util.ts
```diff
@@ -33,8 +34,9 @@ export async function buildIpfsConnection(network: string, ipfsEndpoint?: string
     let ipfs: IpfsApi
 
     if (ipfsEndpoint) {
-        ipfs = ipfsClient({ url: ipfsEndpoint, ipld: { formats: [dagJoseFormat] } })
+        ipfs = ipfsClient({ url: ipfsEndpoint, ipld: { formats: [dagJoseFormat] }, timeout: IPFS_GET_TIMEOUT })
     } else {
+        // TODO: Add global timeout, remove timeout from every call to ipfs.dag.get()
```
stbrody, line 39: I tried my hardest to find a way to do this in the docs and even from reading the source, but I failed.  I opened this question in the IPFS support forum to find out if this is possible: https://discuss.ipfs.io/t/global-timeout-option-for-ipfs-create-from-ipfs-core/10119.

oed (approved):
lgtm

PR #893, opened 2021-02-05
Move IPFS daemon inside

ukstv:
Scenarios in mind:
- start ipfs daemon with no extra params: starts api, gateway, no peer discovery
- start ipfs daemon with CERAMIC_NETWORK in env: ^^ + set up peer discovery
- start ceramic daemon with ceramic network specified (default `testnet-clay` covered too): similar to ^^
  - added benefit: it also starts IPFS API, IPFS gateway
  - if in memory or local => no peer discovery
- Ceramic.create with external IPFS node: initiates peer discovery on the node
  - even if IPFS node is peer-discovery-enabled, it does no harm,
  - reasoning here is one can not be sure that IPFS node ever does peer discovery, and it is Ceramic responsibility to set up peer discovery

How it is achieved:
- moved js-ipfs-ceramic into `ipfs-daemon`,
- extracted peer discovery logic into `ipfs-topology` (because it sets up IPFS topology)
- replaced js-ipfs initialisation in cli with ipfs-daemon.


ukstv:
I just hope that this shit will soon go away.

Screenshot 2021-02-05 at 18 57 03

packages/core/src/ceramic.ts
```diff
@@ -312,7 +313,7 @@ class Ceramic implements CeramicApi {
 
     const doPeerDiscovery = config.useCentralizedPeerDiscovery ?? !TESTING
     if (doPeerDiscovery) {
-      await periodicallyConnectToPeers(networkOptions.name, ipfs)
+      await topology.start()
```
stbrody, line 316: Instead of doing this here, since this functionality also exists in IpfsDaemon, can we just make sure to pass the proper value for `useCentralizedPeerDiscovery` when we construct the IpfsDaemon in `build-ipfs-connection.ts`?

packages/cli/src/build-ipfs-connection.util.ts
```diff
@@ -14,64 +14,16 @@ const dagJoseFormat = legacy(multiformats, dagJose.name)
...
-            },
-            preload: {
-                enabled: false
-            }
-        })
+        const ipfsDaemon = await IpfsDaemon.create({
+            ipfsDhtServerMode: IPFS_DHT_SERVER_MODE,
+            ipfsEnableApi: true,
+            ipfsEnableGateway: true,
+            useCentralizedPeerDiscovery: false
```
stbrody, line 25: If we pass the proper value from the config file for this here, then we don't need the code to check for this flag and call `topology.start()` in `Ceramic.create()`, right?

stbrody (commented):
Basically looks good, made one comment to de-dupe some logic.  The package names are different than what we had previously discussed, but I don't actually feel strongly about what they should be, so I'll defer to Joel there.

stbrody:
> I just hope that this shit will soon go away.
> 
> Screenshot 2021-02-05 at 18 57 03

Yeah, those test failures are getting pretty annoying.  I filed https://github.com/ceramicnetwork/js-ceramic/issues/894 to try to address them

oed:
With this is it possible to install `@ceramicnetwork/ipfs-daemon` globally as a cli command? I think that would be quite useful!

packages/core/src/ceramic.ts
ukstv: I am worried about one scenario: dummy (=no knowledge of Ceramic network) IPFS node and Ceramic core instance running atop. In the scenario, if `topology.start` is removed here, the connection will never be established to the network.

IPFS Daemon can be run as an independent node with CERAMIC_NETWORK passed in, so that the daemon establishes connection to other IPFS nodes. Also, it can be run without CERAMIC_NETWORK passed in. Then the daemon has no idea what to connect to. Ceramic core instance though always knows a network, so it can freely command a dummy IPFS daemon to connect to proper peers.

Ceramic core instance commands peering topology for IPFS node - that's the main scenario, even more valid for go/rust IPFS nodes. IPFS daemon making own topology based on network is a nice benefit, which is valuable from deployment perspective.

packages/cli/src/build-ipfs-connection.util.ts
ukstv: Replied to a previous comment. One scenario that makes `topology.start` stay: Ceramic instance running atop dummy IPFS node.

Resolving this conversation in favour of the one above.

ukstv:
> With this is it possible to install @ceramicnetwork/ipfs-daemon globally as a cli command? I think that would be quite useful!

I think too this is useful: `ipfs-daemon` bin is declared in `package.json`. Installing `@ceramicnetwork/ipfs-daemon` makes `ipfs-daemon` available.

oed:
Thanks @ukstv 🙏

packages/core/src/ceramic.ts
stbrody: Yeah, that makes sense, good call.  I'd probably prefer to put this logic into `buildIpfsConnection` so its contained within the rest of the ipfs-specific logic rather than up here in the Ceramic initializer, but not a huge deal either way.

packages/cli/src/build-ipfs-connection.util.ts
stbrody, line 25: Can we maybe just add a comment explaining why we're setting `useCentralizedPeerDiscovery` to false instead of using the value from the config?

stbrody (approved):
Made 2 small optional suggestions, but otherwise LGTM

packages/cli/src/build-ipfs-connection.util.ts
ukstv: Okay. Makes sense.

packages/core/src/ceramic.ts
ukstv: `buildIpfsConnection` is called _only_ when Ceramic daemon is started from cli. One _can_ (and it is useful) run ceramic instance from a Nodejs script against an unknown IPFS node. In this case, if peer-discovery is out of Ceramic instance codebase, no peer discovery is set up.

ukstv: If one considers a case of Ceramic node running against dummy IPFS node to be valid, peer-discovery has to reside in Ceramic instance.

I acknowledge, there is alternative way of doing things: expose network name from CeramicAPI, and let a consumer apply peer-discovery by herself. Not sure, though if it makes sense, as peer-discovery anyway has to be setup out of the box, be it for IPFS or Ceramic-Ceramic connection.

stbrody: oh whoops, sorry my mistake.  I thought we were using `buildIpfsConnection` within Ceramic.create() to get the ipfs connection handle, but I see it's actually passed in and expected to be already set up. So yeah, the way you're doing it here makes perfect sense.  Thanks for bearing with me.

ukstv: Resolving the conversation then :)

ukstv:
Mind approving the review @stbrody please?

stbrody:
> Mind approving the review @stbrody please?

Already did!

ukstv:
Huh, my fault. Thought it was Joel's one approval.

PR #934, opened 2021-02-09
fix: load flickering document synchronously

ukstv:
Maybe works.

### DIFF
--- a/packages/core/src/__tests__/ceramic-anchor.test.ts
+++ b/packages/core/src/__tests__/ceramic-anchor.test.ts
@@ -311,7 +311,7 @@ describe('Ceramic anchoring', () => {
     expect(doctype1.content).toEqual({ x: 6 })
     expect(doctype1.state.log.length).toEqual(8)
 
-    const doctype2 = await ceramic2.loadDocument(doctype1.id)
+    const doctype2 = await ceramic2.loadDocument(doctype1.id, {sync: true})
     expect(doctype1.content).toEqual(doctype2.content)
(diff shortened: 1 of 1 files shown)


packages/core/src/__tests__/ceramic-anchor.test.ts
```diff
@@ -311,7 +311,7 @@ describe('Ceramic anchoring', () => {
     expect(doctype1.content).toEqual({ x: 6 })
     expect(doctype1.state.log.length).toEqual(8)
 
-    const doctype2 = await ceramic2.loadDocument(doctype1.id)
+    const doctype2 = await ceramic2.loadDocument(doctype1.id, {sync: true})
```
oed, line 314: Isn't this the default value for this option?

ukstv: It is indeed.

oed: So does this really solve the problem?

ukstv: Not at all.

stbrody:
I think what we need to do here is something similar to [what we do in the integration tests](https://github.com/3box/ceramic-integration-tests/blob/main/src/utils.ts#L11) for waiting for an anchor.  Basically a loop that waits on change events and then checks some predicate - in this case whether the update has been loaded - repeatedly until a timeout is hit.

ukstv:
Dropped in favour of #960

PR #935, opened 2021-02-09
fix: make Docker build again

ukstv:
Logger package is to blame this time.


valmack (approved):
👍

oed (approved):

ukstv:
I wonder if there could be a `npm docker:prepare` command that builds dockerfile based on the actual file structure, or it is an over-complication 😈

PR #948, opened 2021-02-09
chore(cli): Add new express logger to ceramic daemon

stbrody:
As implemented, this doesn't actually replace or remove the existing logging behavior from the ceramic daemon, it simply adds the new logger in place in addition to the old.  So now we'll have two log files containing the http logs, with the following output formats.  Note in the below log samples I added a line `throw new Error("FAILED TO LOAD DOC!!")` right before [this line](https://github.com/ceramicnetwork/js-ceramic/blob/01562c92ea1b837530f751d5737ec6ade0305608/packages/cli/src/ceramic-daemon.ts#L272), to simulate a failed response.

*ceramicdaemon.log*:
```
request.timestamp=1612906056137 request.headers.content-type=application/json request.headers.accept=*/* request.headers.content-length=706 request.headers.user-agent="node-fetch/1.0 (+https://github.com/bitinn/
node-fetch)" request.headers.accept-encoding=gzip,deflate request.headers.connection=close request.headers.host=localhost:7007 request.httpVersion=1.1 request.method=POST request.remoteAddress=::ffff:127.0.0.1 r
equest.remoteFamily=IPv6 request.url=/api/v0/documents response.timestamp=1612906056137 response.processingTime=1 response.body="" response.statusCode=200 response.statusMessage=OK response.requestError=
Error: FAILED TO LOAD DOC!!
request.timestamp=1612906061976 request.headers.accept=*/* request.headers.user-agent="node-fetch/1.0 (+https://github.com/bitinn/node-fetch)" request.headers.accept-encoding=gzip,deflate request.headers.connect
ion=close request.headers.host=localhost:7007 request.httpVersion=1.1 request.method=GET request.remoteAddress=::ffff:127.0.0.1 request.remoteFamily=IPv6 request.url=/api/v0/documents/kjzl6cwe1jw14atbi2xajrecsyr
02ub3oazhrxbtafga1wu5kseqrzg8qr5vdgu response.timestamp=1612906061976 response.processingTime=0 response.body= response.statusCode=500 response.statusMessage="Internal Server Error" response.requestError=
```

*http.log*:
```
[Tue, 09 Feb 2021 21:27:36 GMT] service=http ip=::ffff:127.0.0.1 ts=2021-02-09T21:27:36.136Z method=POST path=/api/v0/documents http_version=1.1 req_header- status=200 content_length=361 content_type="applicatio
n/json; charset=utf-8" ref=- user_agent=node-fetch/1.0 (+https://github.com/bitinn/node-fetch) elapsed_ms=53.983
 
[Tue, 09 Feb 2021 21:27:41 GMT] service=http ip=::ffff:127.0.0.1 ts=2021-02-09T21:27:41.977Z method=GET path=/api/v0/documents/kjzl6cwe1jw14atbi2xajrecsyr02ub3oazhrxbtafga1wu5kseqrzg8qr5vdgu http_version=1.1 req
_header- status=500 content_length=32 content_type="application/json; charset=utf-8" ref=- user_agent=node-fetch/1.0 (+https://github.com/bitinn/node-fetch) elapsed_ms=3.375
 ```

Note that the new `http.log` file seems to contain less information than the existing `ceramicdaemon.log` file does, namely the `statusMessage` and the actually exception contents seem to be lost.  I'm not sure if this is something that could be easily added right now, or if it requires the larger project to rethink how error handling in the daemon happens @valmack.  This is why I decided to leave the old log handling in place for now.  I'm wondering if you think this is good enough for now to get in and let me move on to other stuff, and then we can revisit this as part of the daemon refactor project?

### DIFF
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -60,7 +60,7 @@ interface MultiQueries {
 const ACCESS_LOG_FMT = 'ip=:remote-addr ts=:date[iso] method=:method path=:url http_version=:http-version req_header:req[header] status=:status content_length=:res[content-length] content_type=":res[content-type]" ref=:referrer user_agent=:user-agent elapsed_ms=:total-time[3]';
 
 const makeExpressMiddleware = function (config: LoggerConfig) {
-  const logger = LoggerProvider.makeServiceLogger("http", config)
+  const logger = LoggerProvider.makeServiceLogger("http-access", config)
 
   return [morgan(ACCESS_LOG_FMT, { stream: logger })]
 };
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -57,9 +57,19 @@ interface MultiQueries {
   queries: Array<MultiQuery>
 }
 
-const ACCESS_LOG_FMT = 'ip=:remote-addr ts=:date[iso] method=:method path=:url http_version=:http-version req_header:req[header] status=:status content_length=:res[content-length] content_type=":res[content-type]" ref=:referrer user_agent=:user-agent elapsed_ms=:total-time[3]';
+const ACCESS_LOG_FMT = 'ip=:remote-addr ts=:date[iso] method=:method original_url=:original-url base_url=:base-url path=:path http_version=:http-version req_header:req[header] status=:status content_length=:res[content-length] content_type=":res[content-type]" ref=:referrer user_agent=:user-agent elapsed_ms=:total-time[3]';
 
 const makeExpressMiddleware = function (config: LoggerConfig) {
+  morgan.token<Request, Response>('original-url', function (req, res): any {
+    return req.originalUrl;
+  });
+  morgan.token<Request, Response>('base-url', function (req, res): any {
+    return req.baseUrl;
+  });
+  morgan.token<Request, Response>('path', function (req, res): any {
+    return req.path;
+  });
+
   const logger = LoggerProvider.makeServiceLogger("http-access", config)
 
   return [morgan(ACCESS_LOG_FMT, { stream: logger })]
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -1,11 +1,10 @@
 import express, { Request, Response, NextFunction } from 'express'
 import Ceramic from '@ceramicnetwork/core'
 import type { CeramicConfig} from "@ceramicnetwork/core"
-import { ServiceLogger } from "@ceramicnetwork/logger"
+import { DiagnosticsLogger } from "@ceramicnetwork/logger"
 import {
   DoctypeUtils,
   RootLogger,
-  Logger,
   IpfsApi,
   MultiQuery,
   LoggerConfig,
@@ -110,16 +109,16 @@ const makeCeramicConfig = function (opts: CreateOpts): CeramicConfig {
  */
 class CeramicDaemon {
   private server: any
-  private logger: Logger
   private maxHealthyCpu: number
   private maxHealthyMemory: number
   private readonly debug: boolean
+  private readonly logger: DiagnosticsLogger
 
   constructor (public ceramic: Ceramic, opts: CreateOpts) {
     this.debug = opts.debug
-    this.logger = RootLogger.getLogger(CeramicDaemon.name)
     this.maxHealthyCpu = opts.maxHealthyCpu
     this.maxHealthyMemory = opts.maxHealthyMemory
+    this.logger = ceramic.context.logger
 
     const app: core.Express = express()
     app.use(express.json())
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by stbrody** [comment]
Also please let me know if you have any preference for the service name besides just "http".  "http-access"?  "http-request"?

File: packages/logger/src/index.ts [hunk 1]
```diff
@@ -108,21 +113,25 @@ export class ServiceLogger {
   /**
    * Converts the service log to logfmt and writes it to `this.filePath`
    * @param serviceLog Service log object
-   * @param logToConsole True to log to console in addition to file
    */
-  public log(serviceLog: ServiceLog, logToConsole?: boolean): void {
```
stbrody, line 113: I removed the `logToConsole` arg since it's not being used anywhere and was complicating some of the logic.  I figure if we have code that wants to make sure we always log to console it can always use the diagnostic logger directly at a higher log level.

valmack: I think in debug mode the service loggers should always also log to console

valmack: Otherwise it will be harder for us to see what is going on

valmack:
> Note that the new `http.log` file seems to contain less information than the existing `ceramicdaemon.log` file does, namely the `statusMessage` and the actually exception contents seem to be lost. I'm not sure if this is something that could be easily added right now, or if it requires the larger project to rethink how error handling in the daemon happens @valmack. 

I'm not sure either. We definitely want that exception but I'm guessing it is getting messed up. For now I think this is ok as long as we are getting the 500 status.

We can come back to the headers as well--not using that data atm.

packages/cli/src/ceramic-daemon.ts
```diff
@@ -48,6 +58,53 @@ interface MultiQueries {
   queries: Array<MultiQuery>
 }
 
+const ACCESS_LOG_FMT = 'ip=:remote-addr ts=:date[iso] method=:method path=:url http_version=:http-version req_header:req[header] status=:status content_length=:res[content-length] content_type=":res[content-type]" ref=:referrer user_agent=:user-agent elapsed_ms=:total-time[3]';
```
valmack, line 60: Can we keep this as a constant in the LoggerProvider? Will be nice to easily reuse.

packages/logger/src/index.ts
stbrody: Yep, they still do that

packages/cli/src/ceramic-daemon.ts
```diff
@@ -48,6 +58,53 @@ interface MultiQueries {
...
+
+const makeExpressMiddleware = function (config: LoggerConfig) {
+  const logger = LoggerProvider.makeServiceLogger("http", config)
```
valmack, line 63: `("access", config)` ? To stay consistent with CAS. Also don't want to confuse this with http-client log file.

stbrody:
@valmack, just added a small change to give the daemon access to a diagnostic logger and remove it's use of `console.log`.  Sorry for the post-PR addition

packages/logger/src/index.ts
valmack: Oh ok I see you kept that anyway!

packages/cli/src/ceramic-daemon.ts
valmack: I guess what I just merged into CAS should go in here too if you don't mind! https://github.com/ceramicnetwork/ceramic-anchor-service/commit/5a604c7e6bceef4a5c49c7b27e8e981631901bbe

The format and tokens

valmack (approved):
Great. Just small requests!

valmack:
> @valmack, just added a small change to give the daemon access to a diagnostic logger and remove it's use of `console.log`. Sorry for the post-PR addition

Gotcha. Yeah this works for now.

packages/cli/src/ceramic-daemon.ts
stbrody: I originally tried to put the entire `makeExpressMiddleware` function in the LoggerProvider, but then I realized that was introducing a dependency on `morgan` into the `common` package.  So I decided to move into the daemon directly as the `cli` package is likely to be the only one that starts an HTTP server.  If/when we have another HTTP server then we can move this somewhere more common then

packages/cli/src/ceramic-daemon.ts
stbrody: What's the http-client log file?  Also really feels like the fact that these are all http requests getting logged is pretty relevant.  How about "http-access" as the service name?

valmack: Oh if we were to add logging to the http-client, but I guess that naming wouldn't be relevant.

"http-access" sounds good to me

packages/cli/src/ceramic-daemon.ts
valmack: @stbrody Yeah that's fair

valmack (approved):
🙏

PR #949, opened 2021-02-09
chore(core): Log ceramic config at startup

stbrody:
Example output:

```
> ceramic daemon --ipfs-api=http://localhost:5011
Starting Ceramic node at version 0.19.7 with config: 
{
  "logLevel": "silent",
  "gateway": false,
  "networkName": "testnet-clay",
  "anchorServiceUrl": "https://cas-clay.3boxlabs.com"
}
Connecting to ceramic network 'testnet-clay' using pubsub topic '/ceramic/testnet-clay' with supported anchor chains ['eip155:3']
Connecting to peers found in 'https://raw.githubusercontent.com/ceramicnetwork/peerlist/main/testnet-clay.json'
Connected to peers: /dns4/ipfs-dev.3boxlabs.com/tcp/4012/wss/p2p/Qmc4BVsZbVkuvax6SKgwq5BrcKjzBdwx5dW45cWfLVHabx, /dns4/ipfs-clay-cas.3boxlabs.com/tcp/4012/wss/p2p/QmbeBTzSccH8xYottaYeyVX8QsKyox1ExfRx7T1iBqRyCd, 
/dns4/ipfs-dev-internal.3boxlabs.com/tcp/4012/wss/p2p/QmYkpxusRem2iup8ZAfVGYv7iq1ks1yyq2XxQh3z2a8xXq, /dns4/ipfs-clay.3boxlabs.com/tcp/4012/wss/p2p/QmWiY3CbNawZjWnHXx3p3DXsg21pZYTj4CRY1iwMkhP8r3, /dns4/ipfs-clay
.ceramic.network/tcp/4012/wss/p2p/QmSqeKpCYW89XrHHxtEQEWXmznp6o336jzwvdodbrGeLTk, /dns4/ipfs-clay-internal.3boxlabs.com/tcp/4012/wss/p2p/QmQotCKxiMWt935TyCBFTN23jaivxwrZ3uD58wNxeg5npi, /dns4/ipfs-dev.ceramic.net
work/tcp/4012/wss/p2p/QmStNqcAjwh6s2sxUWr2ZXT3MhRZmqpJ9Dj6fp3gPdHr6E, /dns4/ipfs-dev-cas.3boxlabs.com/tcp/4012/wss/p2p/QmPHLQoWhK4CMPPgxGQxjNYEp1fMB8NPpoLaaR2VDMNbcr
Ceramic API running on port 7007
```

### DIFF
--- a/packages/core/src/ceramic.ts
+++ b/packages/core/src/ceramic.ts
@@ -31,7 +31,6 @@ import { Resolver } from "did-resolver"
 import { DID } from 'dids'
 import { TileDoctypeHandler } from "@ceramicnetwork/doctype-tile-handler"
 import { Caip10LinkDoctypeHandler } from "@ceramicnetwork/doctype-caip10-link-handler"
-import { DiagnosticsLogger } from "@ceramicnetwork/logger";
 import { PinStoreFactory } from "./store/pin-store-factory";
 import { PinStore } from "./store/pin-store";
 import { PathTrie, TrieNode, promiseTimeout } from './utils'
@@ -50,15 +49,19 @@ const IPFS_GET_TIMEOUT = 60000 // 1 minute
 const TESTING = process.env.NODE_ENV == 'test'
 
 /**
- * Ceramic configuration parameters
+ * Ceramic configuration
  */
 export interface CeramicConfig {
   ethereumRpcUrl?: string;
   anchorServiceUrl?: string;
   pinsetDirectory?: string;
 
+  didResolver?: Resolver;
+  didProvider?: DIDProvider;
+
   validateDocs?: boolean;
   pinningEndpoints?: string[];
+  pinningBackends?: PinningBackendStatic[];
 
   logLevel?: string;
   logToFiles?: boolean;
@@ -80,16 +83,6 @@ export interface CeramicConfig {
   restoreDocuments?: boolean;
 
   [index: string]: any; // allow arbitrary properties
-}
-
-/**
- * Ceramic inner components provided during creation.
- */
-export interface CeramicProvidedComponents {
-  didResolver?: Resolver;
-  didProvider?: DIDProvider;
-  pinningBackends?: PinningBackendStatic[];
-  logger?: DiagnosticsLogger;
 }
 
 /**
@@ -239,12 +232,24 @@ class Ceramic implements CeramicApi {
   }
 
   /**
+   * Returns a copy of the given CeramicConfig object but with any potentially sensitive fields
+   * removed so that it is safe to log the whole thing.
+   * @param config
+   * @returns Copy of `config` with potentially sensitive information removed
+   * @private
+   */
+  private static _redactConfigForLogging(config: CeramicConfig): CeramicConfig {
+    const redactedConfig = {...config}
+    delete redactedConfig.didProvider
+    return redactedConfig
+  }
+
+  /**
    * Create Ceramic instance
    * @param ipfs - IPFS instance
    * @param config - Ceramic configuration
-   * @param components - Ceramic internal components provided to `create` from higher level
    */
-  static async create(ipfs: IpfsApi, config: CeramicConfig = {}, components?: CeramicProvidedComponents): Promise<Ceramic> {
+  static async create(ipfs: IpfsApi, config: CeramicConfig = {}): Promise<Ceramic> {
     // todo remove
     LoggerProviderOld.init({
       level: config.logLevel? config.logLevel : 'silent',
@@ -262,10 +267,10 @@ class Ceramic implements CeramicApi {
 
     // Initialize ceramic loggers
     const loggerConfig = {logLevel: config.logLevel, logToFiles: config.logToFiles, logPath: config.logPath}
-    const logger = components?.logger ?? LoggerProvider.makeDiagnosticLogger(loggerConfig)
+    const logger = LoggerProvider.makeDiagnosticLogger(loggerConfig)
     const pubsubLogger = LoggerProvider.makeServiceLogger("pubsub", loggerConfig)
 
-    logger.imp(`Starting Ceramic node at version ${packageJson.version} with config: \n${JSON.stringify(config, null, 2)}`)
+    logger.imp(`Starting Ceramic node at version ${packageJson.version} with config: \n${JSON.stringify(Ceramic._redactConfigForLogging(config), null, 2)}`)
 
     const anchorService = config.anchorServiceUrl ? new EthereumAnchorService(config) : new InMemoryAnchorService(config as any)
     await anchorService.init()
@@ -285,7 +290,7 @@ class Ceramic implements CeramicApi {
       networkName: networkOptions.name,
       pinsetDirectory: config.pinsetDirectory,
       pinningEndpoints: config.pinningEndpoints,
-      pinningBackends: components?.pinningBackends
+      pinningBackends: config.pinningBackends
     }
     const pinStoreFactory = new PinStoreFactory(context, pinStoreProperties)
     const pinStore = await pinStoreFactory.open()
@@ -296,11 +301,11 @@ class Ceramic implements CeramicApi {
     const keyDidResolver = KeyDidResolver.getResolver()
     const threeIdResolver = ThreeIdResolver.getResolver(ceramic)
     ceramic.context.resolver = new Resolver({
-      ...components?.didResolver, ...threeIdResolver, ...keyDidResolver,
+      ...config.didResolver, ...threeIdResolver, ...keyDidResolver,
     })
 
-    if (components?.didProvider) {
-      await ceramic.setDIDProvider(components.didProvider)
+    if (config.didProvider) {
+      await ceramic.setDIDProvider(config.didProvider)
     }
 
     const doPeerDiscovery = config.useCentralizedPeerDiscovery ?? !TESTING
--- a/packages/core/src/ceramic.ts
+++ b/packages/core/src/ceramic.ts
@@ -88,6 +88,7 @@ export interface CeramicConfig {
 export interface CeramicProvidedComponents {
   didResolver?: Resolver;
   didProvider?: DIDProvider;
+  ipfs: IpfsApi;
   pinningBackends?: PinningBackendStatic[];
   logger?: DiagnosticsLogger;
 }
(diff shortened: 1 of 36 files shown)


valmack:
That version number makes me so happy :)

oed:
Should we really be logging this if `--debug` is not enabled?

stbrody:
> Should we really be logging this if `--debug` is not enabled?
> I'd rather keep the amount of noise here minimal.

I strongly disagree.  If we put all the useful logging into log level "debug" then there becomes no point in even having log levels.  It'll either be no logs without --debug or all the logs with --debug.  The debug log level is *already* quite verbose and not something I think most people are going to want to run with, which makes it even more important to have useful logging at the default level, or else our ability to diagnose problems in the field becomes much harder.  I for one never want to have to diagnose a user's issue by reading their log files without at least knowing what version they are running and what config options they are running with.  Having to *ask* the user for that information separately every time we want to diagnose an issue creates unnecessary friction and extra back-and-forths in the already lengthy and stressful support process.

packages/core/src/ceramic.ts
```diff
@@ -287,6 +290,8 @@ class Ceramic implements CeramicApi {
     const logger = LoggerProvider.makeDiagnosticLogger(loggerConfig)
     const pubsubLogger = LoggerProvider.makeServiceLogger("pubsub", loggerConfig)
 
+    logger.imp(`Starting Ceramic node at version ${packageJson.version} with config: \n${JSON.stringify(config, null, 2)}`)
```
oed, line 293: Wouldn't this output the users private key which is usually contained in the `~/.ceramic/config.json`?

oed:
I'm totally fine with outputting the version. Printing the config is strange though. Wouldn't it be easier to just ask the user for their config? I've just never seen any cli daemons similar to ceramic that outputs the complete config file on startup.

stbrody:
> I'm totally fine with outputting the version. Printing the config is strange though. Wouldn't it be easier to just ask the user for their config? I've just never seen any cli daemons similar to ceramic that outputs the complete config file on startup.

Interesting.  The point about printing the user's private key is definitely an important one.  We do need to be careful to redact any potentially sensitive information from the logs.  But printing the majority of the process config seems like no-brainer table stakes to me.  MongoDB binaries always log their config at startup, and I really can't imagine debugging issues with user's deployments without that.

packages/core/src/ceramic.ts
stbrody: Can you point me to where this private key information is loaded?  I'm not seeing it obviously within the [CeramicConfig](https://github.com/ceramicnetwork/js-ceramic/blob/be2b472640d35165efeb61437b83a81f29e878be/packages/core/src/ceramic.ts#L50).  Maybe it's within the `didProvider` object?  If we know which fields in the config may contain sensitive information, we could add a `redactSensitiveConfigInfo()` method that we call before logging the config to filter such fields out

oed:
Ok I guess that makes sense @stbrody.
Here is the code that loads the seed from the config: https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/cli/src/ceramic-cli-utils.ts#L307-L311

stbrody:
New version up that redacts the didProvider, but adds logging when a did gets authenticated

oed:
@stbrody not sure exactly what the `didProvider` does. In my config I see this:
```json
{
  "seed": "..."
}
```
So shouldn't it be the `seed` property we redact?

stbrody:
> @stbrody not sure exactly what the `didProvider` does. In my config I see this:
> 
> ```json
> {
>   "seed": "..."
> }
> ```
> 
> So shouldn't it be the `seed` property we redact?

The didProvider is the object representing the authenticated did and performs the signing when creating SignedCommits.  It has the seed stored within it in a variable somewhere deep within several layers of object abstractions.  Since it's an object not actually a config it probably doesn't make sense to try and stringify it and print the whole thing out anyway.  I'm actually thinking now that we should remove the actual objects that aren't really string or number config options out of the `CeramicConfig` argument to `Ceramic.create()` and have them come in a separate argument.  That can also help with my issue of getting the logger to the IpfsDaemon, and I think lets us get rid of the `_redactConfigForLogging` function entirely.  I'll give that a shot later today and put up a new review here.

oed:
Oh ok @stbrody. I thought we were just logging the contents of `config.json`.

stbrody:
> Oh ok @stbrody. I thought we were just logging the contents of `config.json`.

Nope, the `config` object being logged here is the `CeramicConfig` passed to `Ceramic.create()`.  Much of its contents are *derived* from the `config.json`, but that happens higher up within the `CeramicDaemon`

packages/core/src/ceramic.ts
```diff
@@ -82,6 +82,16 @@ export interface CeramicConfig {
   [index: string]: any; // allow arbitrary properties
 }
 
+/**
+ * Ceramic inner components provided during creation.
+ */
+export interface CeramicProvidedComponents {
```
stbrody, line 88: Definitely open to ideas for better names for this

stbrody:
New version up that separates out the more basic config options from the parts of the config that are actually providing fully-instantiated objects with runnable logic

packages/core/src/ceramic.ts
```diff
@@ -230,10 +241,10 @@ class Ceramic implements CeramicApi {
 
   /**
    * Create Ceramic instance
-   * @param ipfs - IPFS instance
+   * @param components - Ceramic internal components provided to `create` from higher level
    * @param config - Ceramic configuration
    */
-  static async create(ipfs: IpfsApi, config: CeramicConfig = {}): Promise<Ceramic> {
+  static async create(components: CeramicProvidedComponents, config: CeramicConfig = {}): Promise<Ceramic> {
```
oed, line 247: I guess technically this is a breaking API change. Is the main reason we're breaking them apart that it will be easier to log the config? Wouldn't it be better to just filter the config object in that case?

stbrody: Yeah, there's no technical reason why this *has* to be broken out.  We certainly could continue to just add more things into the `CeramicConfig`.  It's less about the filtering in my mind and more about logically separating where we're providing configuration options to Ceramic, vs where we plug in implementations of entire subsystems.  I already need to add a way to provide a logger instance (while continuing to create the logger internally if no logger instance is provided, but config options about how to construct one are), and I'm worried that it'll be confusing to have a full logger instance and the config options that describe how to *construct* a logger instance in the same object.  We also have other places where we allow providing an entire instance of a subcomponent rather than just config options that control what subcomponents get used, with more to come.  The different `StateStore` backends are a good example.  Is an entire implementation of the `StateStore` interface really a config option?  Or should we have a separate mechanism for pluggable components that extend the core functionality?

I could go either way on this one.  Would be curious if @ukstv has an opinion here

oed: Another option could be to have two properties in the config type. One for modules, and one for simple configs?
Idk what's best here tbh, but it feels a bit weird to have two separate types.

stbrody: > Another option could be to have two properties in the config type.

You mean two properties at the top-level, with everything else moving into sub-objects?  So basically changing `CeramicConfig` to look something like:

```
export interface CeramicConfig {
  flags: {
    ethereumRpcUrl?: string;
    anchorServiceUrl?: string;
    pinsetDirectory?: string;
    // ... <the rest of the current CeramicConfig properties>
  };
  components: {
    ipfs: IpfsApi;
    logger?: DiagnosticLogger;
    stateStore?: StateStore;
    // ...
  };
}
```

oed: Yeah something like that. Maybe call them `modules` and `params`?

stbrody:
Okay, new version up that splits `CeramicConfig` into separate `params` and `modules` subobjects

oed (approved):
I like this more 👍 In theory I guess we could include ipfs in `modules` since this is a breaking change anyway. I'm fine either way though!

valmack:
> MongoDB binaries always log their config at startup, and I really can't imagine debugging issues with user's deployments without that.

I also have only ever seen full config behind a debug flag or some other flag along the lines of `--show-config`. I'm not sure what the thinking behind it is/tradeoffs--I think if the config becomes very large (20+ lines?) it will just be noisy. In terms of debugging we still have the `config show` command.

Not making any real suggestion here just my 2 cents

ukstv:
Suggestion: let's leave `Ceramic.create` as it is for a usual consumer. Maybe slim down the parameters list a bit.

Let `Ceramic.create` call well-known initialisation logic that returns a bunch of components. `new Ceramic` would consume the components and nothing more. 

For a usual consumer, everything is super friendly.

For a tech savvy, Ceramic promises that `Ceramic.create` will ever contain two lines:
```
const components = await Ceramic.components(<currently-second-argument-of-ceramic-create>)
return new Ceramic(components)
```

The tech-savvy person then passes own components by inserting them in the middle:
```
const components = await Ceramic.components(<currently-second-argument-of-ceramic-create>)
return new Ceramic({...components, logger: fancyLogger })
```
No more nice `Ceramic.create`, but anyway if someone passes own logger, she initialises it somehow.

The whole choice is between bad and ugly, and this way seems to provide
- nice interface for a novice (or "do-not-care") person through `Ceramic.create`,
- powerful interface for a tech-savvy person through "create your own instance with our little help" proposal.

Brain dump behind this: https://gist.github.com/ukstv/5e7fb68fc69eff2be2052a022c3585a2

stbrody:
> > MongoDB binaries always log their config at startup, and I really can't imagine debugging issues with user's deployments without that.
> 
> I also have only ever seen full config behind a debug flag or some other flag along the lines of `--show-config`. I'm not sure what the thinking behind it is/tradeoffs--I think if the config becomes very large (20+ lines?) it will just be noisy. In terms of debugging we still have the `config show` command.
> 
> Not making any real suggestion here just my 2 cents

Interesting. My personal philosophy on logging is to be careful about being too verbose in code paths that get hit a lot of times, but for things that only happen once for the lifetime of a process it's better to err on the side of including more in the logs rather than less, so you don't wind up without information that you wish you had.

stbrody:
@ukstv has some pretty good ideas that I need to think about further.  This PR has gone pretty far afield from its original focus on adding some log messages.  I'm going to revert this to the version that doesn't change the config at all and just redacts the didProvider manually, and then work on a follow-up patch to clean up the initialization in advance of providing a logger instance to the IpfsDaemon.

PR #951, opened 2021-02-10
fix(common): Show timezone with anchorScheduledFor property

### LINKED ISSUE #897: Show timezone with anchorScheduledFor property
https://github.com/ceramicnetwork/js-ceramic/blob/8efe3daf4c8702a45a3d6489b587f5ce8045d908/packages/common/src/doctype.ts#L116

For example getting the doc state shows something like this
```
"anchorScheduledFor": "2/8/2021, 3:09:00 AM"
```

We should format this to something that shows the timezone

carlbarrdahl:
Fixes #897


stbrody (approved):
Looks good, thank you for the PR!

Issue #829, opened 2021-01-20
Change console.log calls to use logger


ukstv:
I doubt it is a good idea to post it to console right away. I believe this kind of information should be properly logged.

stbrody:
Agreed, we just don't have a good logging framework in place in js-ceramic yet.  We have [a project](https://github.com/ceramicnetwork/js-ceramic/issues/609) to add one.  As part of that project I plan to grep for every place we `console.log()` and change them to use the logger instead.

stbrody:
I'm looking through the current places we access the `console` directly.  One of those places is [this line](https://github.com/ceramicnetwork/js-ceramic/blob/be2b472640d35165efeb61437b83a81f29e878be/packages/blockchain-utils-validation/src/blockchains/eosio.ts#L23) within the `validateLink` function for eosio blockchain validation, where it catches all exceptions and swallows them, logging a warning in the process.  I'm wondering - why are we swallowing all exceptions here?  The [ethereum](https://github.com/ceramicnetwork/js-ceramic/blob/be2b472640d35165efeb61437b83a81f29e878be/packages/blockchain-utils-validation/src/blockchains/ethereum.ts#L79-L81) and [filecoin](https://github.com/ceramicnetwork/js-ceramic/blob/be2b472640d35165efeb61437b83a81f29e878be/packages/blockchain-utils-validation/src/blockchains/filecoin.ts#L24-L26) integrations also swallow all exceptions, but without logging a warning.  The [cosmos](https://github.com/ceramicnetwork/js-ceramic/blob/be2b472640d35165efeb61437b83a81f29e878be/packages/blockchain-utils-validation/src/blockchains/cosmos.ts#L12), [near](https://github.com/ceramicnetwork/js-ceramic/blob/be2b472640d35165efeb61437b83a81f29e878be/packages/blockchain-utils-validation/src/blockchains/near.ts#L11), and [polkadot](https://github.com/ceramicnetwork/js-ceramic/blob/be2b472640d35165efeb61437b83a81f29e878be/packages/blockchain-utils-validation/src/blockchains/polkadot.ts#L8) integrations, however, do not swallow exceptions and allow them to bubble up to the calling code.

[...]

PR #958, opened 2021-02-10
chore(core): Allow exceptions to escape from blockchain validateLink functions

stbrody:
See [this comment](https://github.com/ceramicnetwork/js-ceramic/issues/829#issuecomment-777056611) for more context

### DIFF
--- a/packages/blockchain-utils-validation/src/blockchains/__tests__/filecoin.test.ts
+++ b/packages/blockchain-utils-validation/src/blockchains/__tests__/filecoin.test.ts
@@ -41,6 +41,6 @@ describe('validateLink', () => {
     const testAddr = await testnetProvider.getAccounts()
     const testAcc = new AccountID(`${testAddr[0]}@fil:t`)
     proof.account = testAcc.toString()
-    await expect(validateLink(proof)).resolves.toEqual(null)
+    await expect(validateLink(proof)).rejects.toEqual('Error verifying signature: BLS error | Size mismatch')
   })
 })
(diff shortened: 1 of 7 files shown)


oed:
These functions verify signatures. They are expected to either return the proof if the singature is valid, or null if it's not. Changing them here would probably break the `caip10-link` doctype handler. 
I guess it could make sense to bubble up the error to there where we can log it?

stbrody:
> These functions verify signatures. They are expected to either return the proof if the singature is valid, or null if it's not. Changing them here would probably break the `caip10-link` doctype handler.
> I guess it could make sense to bubble up the error to there where we can log it?

Did you see my comment in the linked ticket?  Some of the other blockchains already don't suppress exceptions (I guess it's possible that they just never throw exceptions, but that seems unlikely?) and the callsite in caip10-link-doctype-handler is already prepared to throw exceptions.  So yeah, it seems like we should be able to let the error bubble up and get handled and logged at a higher level

ukstv:
TBH, It seems like a legit proposal, if the handler code handles the error properly.

oed:
Ok, thanks for the additional context @stbrody. This seems good to me then 👍

packages/blockchain-utils-validation/src/blockchains/__tests__/filecoin.test.ts
```diff
@@ -41,6 +41,6 @@ describe('validateLink', () => {
     const testAddr = await testnetProvider.getAccounts()
     const testAcc = new AccountID(`${testAddr[0]}@fil:t`)
     proof.account = testAcc.toString()
-    await expect(validateLink(proof)).resolves.toEqual(null)
+    await expect(validateLink(proof)).rejects.toEqual('Error verifying signature: BLS error | Size mismatch')
```
stbrody, line 44: It seems the `@zondax/filecoin-signing-tools` is throwing a raw string rather than throwing an `Error` on signature verification failure :(

oed (approved):
Lgtm :)

ukstv (approved):
🏇

PR #959, opened 2021-02-10
chore(core): Use the logger in InMemoryAnchorService

stbrody:
One part of https://github.com/ceramicnetwork/js-ceramic/issues/829

### DIFF
--- a/packages/core/src/anchor/memory/in-memory-anchor-service.ts
+++ b/packages/core/src/anchor/memory/in-memory-anchor-service.ts
@@ -3,6 +3,7 @@ import * as uint8arrays from "uint8arrays";
 import { Observable, Subject, concat, of } from "rxjs";
 import { filter } from "rxjs/operators";
 import * as didJwt from "did-jwt";
+import { Resolver } from "did-resolver"
 import {
   AnchorProof,
   AnchorService,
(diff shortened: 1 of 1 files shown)


oed (approved):

packages/core/src/anchor/memory/in-memory-anchor-service.ts
```diff
@@ -114,7 +115,7 @@ class InMemoryAnchorService implements AnchorService {
           }
           result[candidate.key].push(candidate);
         } catch (e) {
-          console.error(e.message);
+          this.#logger.err(e.message)
```
ukstv, line 118: @stbrody Is it possible to pass full error, with stack trace to the log? I think, a stack trace passes valuable information.

packages/core/src/anchor/memory/in-memory-anchor-service.ts
```diff
@@ -216,9 +217,10 @@ class InMemoryAnchorService implements AnchorService {
    *
    * @param ceramic - Ceramic API used for various purposes
    */
-  set ceramic(ceramic: CeramicApi) {
-    this.#ceramic = (ceramic as unknown) as Ceramic;
+  set ceramic(ceramic: Ceramic) {
```
ukstv, line 220: +1

PR #960, opened 2021-02-10
chore(core): Stabilize ceramic-anchor.test.ts

stbrody:
I'm curious for thoughts on this approach.  If we like this we can roll out the `waitForState` helper in more of our tests over time

### DIFF
--- a/packages/common/src/utils/test-utils.ts
+++ b/packages/common/src/utils/test-utils.ts
@@ -1,0 +1,42 @@
+import {DocState, Doctype} from "../doctype";
+
+export class TestUtils {
+
+    /**
+     * Returns a Promise that resolves when there is an update to the given document's state.
+     * @param doc
+     */
+    static registerChangeListener(doc: Doctype): Promise<void> {
+        return new Promise(resolve => {
+            doc.on('change', () => {
+                resolve()
+            })
+        })
+    }
+
+    /**
+     * Given a document and a predicate that operates on the document state, continuously waits for
+     * changes to the document until the predicate returns true.
+     * @param doc
+     * @param timeout - how long to wait for
+     * @param predicate - function that takes the document's DocState as input and returns true when this function can stop waiting
+     * @param onFailure - function called if we time out before the predicate becomes true
+     */
+    static async waitForState(doc: Doctype,
+                              timeout: number,
+                              predicate: (state: DocState) => boolean,
+                              onFailure: () => void): Promise<void> {
+        const timeoutPromise = new Promise(resolve => setTimeout(resolve, timeout))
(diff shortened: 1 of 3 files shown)


oed (approved):

packages/common/src/utils/test-utils.ts
```diff
@@ -0,0 +1,42 @@
...
+     * @param timeout - how long to wait for
+     * @param predicate - function that takes the document's DocState as input and returns true when this function can stop waiting
+     * @param onFailure - function called if we time out before the predicate becomes true
+     */
+    static async waitForState(doc: Doctype,
+                              timeout: number,
+                              predicate: (state: DocState) => boolean,
+                              onFailure: () => void): Promise<void> {
+        const timeoutPromise = new Promise(resolve => setTimeout(resolve, timeout))
+        const completionPromise = async function () {
```
ukstv, line 30: Guess, could be made simpler:
```
const completionPromise = new Promise<void>(resolve => {
  // first, check if current state satisfies the predicate
  // it could happen that doc has already been changed
  if (predicate(doc.state)) return resolve();
  // if not:
  const handler = () => {
    if (predicate(doc.state)) {
      document.off('change', handler)
      resolve()
    }
  }
  document.on('change', handler)
})

ukstv:
+1 for `waitForState`. A bit concerned about putting the test utility to `common`. I think though, it could be a useful general-purpose function. Maybe just raw export of the function, rather than putting it into `TestUtil` container?

ukstv (approved):

packages/common/src/utils/test-utils.ts
stbrody: Thanks.  I still really find the api for interacting with Promises in javascript to be really unintuitive

stbrody:
> +1 for `waitForState`. A bit concerned about putting the test utility to `common`. I think though, it could be a useful general-purpose function. Maybe just raw export of the function, rather than putting it into `TestUtil` container?

What's your concern exactly?  I just followed the same convention as we have for `DoctypeUtils`.  In my mind having the test-specific functions within a `TestUtil` scope seems nice

ukstv:
I am concerned with valuable function being treated as test-only. IMO, it is more than test-only. Don't mind though. We'll see shortly if there is an interest in exposing it.

ukstv:
Lgtm

stbrody:
> I am concerned with valuable function being treated as test-only. IMO, it is more than test-only. Don't mind though. We'll see shortly if there is an interest in exposing it.

ah I see.  yeah, let's leave it here for now and we can always move it somewhere more general-purpose if we find a reason to use it in non-test code

Issue #879, opened 2021-02-03
Bring proper prettier and .editorconfig into repo


ukstv:
Benefits are obvious.

PR #972, opened 2021-02-11
chore: Add editorconfig and prettier

ukstv:


oed:
I see that we have the story here: https://github.com/ceramicnetwork/js-ceramic/issues/879
When you open PRs agains stories please connect it to the story using the zenhub plugin:
image

package.json
```diff
@@ -25,11 +25,17 @@
...
     "ts-node": "^9.0.0",
     "typedoc": "^0.19.2",
     "typedoc-plugin-lerna-packages": "^0.3.1",
     "typedoc-plugin-pages": "^1.0.1",
     "typescript": "^4.0.5"
+  },
+  "prettier": {
+    "printWidth": 120,
+    "trailingComma": "all",
+    "singleQuote": true
```
oed, line 39: Would be nice if we could start using the https://github.com/3box/eslint-config-3box package for this entire repo. But that is likely a much larger change.

oed (approved):

ukstv:
Okay.

ZenHub lacks Safari plugin. So sad.

oed:
That's unfortunate. It works for Firefox and chromium. Does safari actually have a plugin system?

oed:
You should be able to connect it within the zenhub app as well fyi :)

package.json
ukstv: It definitely is bigger. 3box eslint rules are too strict for this code 😉

Issue #754, opened 2021-01-08
Ceramic.close() doesn't close all open handles and background work


stbrody:
Steps to reproduce: add this test to `ceramic.test.ts`
```
  it.only('can close ceramic right after creating document', async () => {
    const ceramic = await createCeramic(ipfs1)
    await delay(1000)
    await ceramic.close()
  })
```

Have jest run just that test, and you get the warning:
```
Jest did not exit one second after the test run has completed.

This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue
```

ukstv:
Note to self: look for FIXME 754

ukstv:
As demonstrated by https://github.com/ceramicnetwork/js-ceramic/pull/992, not reproducible anymore. Closing.

stbrody:
Still reproduces for me

ukstv:
@stbrody You mean https://github.com/ceramicnetwork/js-ceramic/pull/992/files#diff-3dc086303a234c0aa9cb5b7b4b675827819e66906135448b777b221da01e0d60R50 hangs?

 Could you please, provide an example of the minimal test suite to reproduce, like I did with https://github.com/ceramicnetwork/js-ceramic/pull/992 ?

stbrody:
@ukstv, finally got around to taking another look at this.  PR that repros the issue here: https://github.com/ceramicnetwork/js-ceramic/pull/1071

ukstv:
@stbrody I feel like I solved the case. IPFS is to blame here. Apparently, when `ipfs.stop` promise fulfils, IPFS has still not freed all the resources. Adding `delay(1000)` after all the `ipfs.stop` statements fixes the issue.

stbrody:
> @stbrody I feel like I solved the case. IPFS is to blame here. Apparently, when `ipfs.stop` promise fulfils, IPFS has still not freed all the resources. Adding `delay(1000)` after all the `ipfs.stop` statements fixes the issue.

Boom, I think that's exactly right.  Confirmed that that fixes it for me as well.  I also now seem to remember a problem in a different test at some point a while ago where starting a new ipfs instance too quickly after closing an old one would conflict in some way.  So yeah, definitely seems like js-ipfs has an issue with `.stop()` not actually waiting for all resources to be fully freed

stbrody:
I guess we should just close this issue then if it's not actually a bug in ceramic?

PR #979, opened 2021-02-11
chore: Make IPFS instances DRY way

ukstv:
This started as "I want to create an IPFS instance in my test for LocalPinApi".

- Extract `createIPFS`: creates an IPFS instance with random repository folder, and random port,
- Extract `swarmConnect`: connects two IPFS instances together via `swarm.connect`, eliminates need for `multiaddr` variables in tests.

FYI Found problematic test wrt #754, commented it for now, made a note in #754.


packages/core/src/__tests__/ceramic-anchor.test.ts
```diff
@@ -2,35 +2,14 @@ import Ceramic from '../ceramic'
...
 import tmp from 'tmp-promise'
-import IPFS from 'ipfs-core'
 import * as u8a from 'uint8arrays'
-
-import getPort from 'get-port'
-
-import dagJose from 'dag-jose'
-import basicsImport from 'multiformats/cjs/src/basics-import.js'
-import legacy from 'multiformats/cjs/src/legacy.js'
+import * as _ from 'lodash'
```
stbrody, line 6: this seems potentially confusing.  Is this a common convention in javascript code to import a package as `_`?  Wouldn't it be clearer just to call it `lodash`?

packages/core/src/__tests__/ceramic-api.test.ts
```diff
@@ -584,20 +557,21 @@ describe('Ceramic API', () => {
...
-        },
-        {
-          docId: notExistDocId,
-          paths: ['/e/f' , '/d']
-        }
-      ]
-      const docs = await ceramic.multiQuery(queries, 1000)
-      expect(Object.keys(docs).length).toEqual(3)
-    })
+    // FIXME 754 Problematic test for #754
```
stbrody, line 560: How did the change from this PR make this test worse?  It looks totally unrelated to me?

packages/core/src/__tests__/ceramic-api.test.ts
```diff
@@ -467,6 +439,7 @@ describe('Ceramic API', () => {
     const notExistDocId = DocID.fromString('kjzl6cwe1jw1495fyn7770ujykvl1f8sskbzsevlux062ajragz9hp3akdqbmdg')
 
     beforeAll(async () => {
+      ceramic = await createCeramic()
```
stbrody, line 442: ugh.  So I'm guessing these tests worked if they were run after the previous tests, but failed if run on their own because the `ceramic` object hadn't been set?  In any case if we're adding this to `beforeAll` we probably need a corresponding `afterAll` to close it right?

packages/core/src/__tests__/ceramic-api.test.ts
```diff
@@ -96,20 +76,12 @@ describe('Ceramic API', () => {
...
-        Bootstrap: []
-      }
-    })
+    ipfs = await createIPFS()
   })
 
   afterAll(async () => {
     await ipfs.stop(() => console.log('IPFS stopped'))
-    await tmpFolder.cleanup()
+    await ceramic?.close()
```
stbrody, line 84: Should we change this from an `afterAll` to an `afterEach`?  Otherwise if a test reassigns the `ceramic` variable without closing the existing one then the previous ceramic gets leaked

packages/core/src/__tests__/ipfs-util.ts
```diff
@@ -0,0 +1,54 @@
...
+  };
+
+  const config = _.merge({}, defaultConfig, overrideConfig);
+  const instance = await IPFS.create(config);
+
+  // IPFS does not notify you when it stops.
+  // Here we intercept a call to `ipfs.stop` to clean up IPFS repository folder.
+  // Poor man's hook.
+  return new Proxy(instance, {
+    get(target: any, p: PropertyKey): any {
```
stbrody, line 36: woah.  javascript really is a wild language

stbrody (commented):
Great cleanup, thanks for doing it!  A couple of small questions inline but nothing big

packages/core/src/__tests__/ceramic-anchor.test.ts
ukstv: It is a convention. Comes from times of a package called [underscore](https://underscorejs.org). Lodash is effectively a replacement for underscore, that's why "_".

packages/core/src/__tests__/ceramic-api.test.ts
ukstv: Totally unrelated feature-wise. At one point I anticipated dropping "--runInBand", and this test still consumer resources after "ceramic.close".

For future reference: multi-query timeout is a shallow one. The timeout does not propagate to `Document.load`, so it hangs till IPFS timeout expires.

ukstv: Going to put it back in a few.

ukstv: CC @zachferland about multi-query timeout.

packages/core/src/__tests__/ceramic-api.test.ts
ukstv: Yes! Sad thing about it I has just wiped the changes that make it concur with your logic. Restoring.

packages/core/src/__tests__/ceramic-api.test.ts
ukstv: Changed.

packages/core/src/__tests__/ceramic-api.test.ts
ukstv: And, restored.

packages/core/src/__tests__/ceramic-api.test.ts
ukstv: Uncommented back.

packages/core/src/__tests__/ceramic-anchor.test.ts
```diff
@@ -72,55 +51,21 @@ describe('Ceramic anchoring', () => {
...
-      return await getPort({port: getPort.makeRange(start + 1, start + offset)})
-    }
-
-    ([port1, port2, port3] = await Promise.all([p1Start, p2Start, p3Start].map(start => findPort(start, pOffset))));
-    ([ipfs1, ipfs2, ipfs3] = await Promise.all([port1, port2, port3].map(port => createIPFS(buildConfig(tmpFolder.path, port)))));
-    ([p1Start, p2Start, p3Start] = [p1Start, p2Start, p3Start].map(start => start + pOffset))
-
-    multaddr1 = (await ipfs1.id()).addresses[0].toString()
-    multaddr2 = (await ipfs2.id()).addresses[0].toString()
+    [ipfs1, ipfs2, ipfs3] = await Promise.all(_.times(3).map(() => createIPFS()));
```
carlbarrdahl, line 58: You can use this instead of the lodash dep: `Array.from({length: 3}).map()`

packages/core/src/__tests__/ceramic-anchor.test.ts
stbrody: Ah okay.  Yeah I guess this package provides such fundamental utilities it's reasonable to give it someone special treatment.  Sounds good

stbrody (approved):
LGTM

packages/core/src/__tests__/ceramic-anchor.test.ts
ukstv: At the time it was groundbreaking indeed. Removing it away.

ukstv: Was that sarcasm in "fundamental"?

stbrody: No, it wasn't sarcastic, I was being genuine.  It seems like this library is designed to give some pretty basic core primitives that are generally useful and can almost be thought of as extensions to the language's standard library.

PR #992, opened 2021-02-12
Ceramic.close() <s>doesn't</s> do close all open handles and background work

ukstv:
DO NOT MERGE!!!!

Demonstrates that `Ceramic#close` works all right. Jest does not notice any hanging handles here.

@stbrody if you are satisfied, feel free to close the pull request.


stbrody:
Hmm... I can still repro https://github.com/ceramicnetwork/js-ceramic/issues/754 quite easily

PR #1009, opened 2021-02-16
feat: Unbundle DocID into DocID and CommitID

ukstv:
# Rationale

Current DocID represents both document identifier and a reference to document state at commit. This leads to confusion: what is actually requested here - document or its snapshot.

# Changes

`DocID` represents a reference to a document. It does not have any commit information. `CommitID` references a particular commit. `DocRef` provides (1) interface shared by both `CommitID` and `DocID`, (2)  `from` method that parses input into `CommitID` or `DocID`.

Removed `fromOther` method in favour of `DocID`, `CommitID` constructors, and `travel` method of both classes.

`DocID#travel` and `CommitID#travel` create a new CommitID that references a commit of the document.

`DocID.fromString` and `DocID.fromBytes` do not accept an additional commit parameter. Previous behaviour could be achieved by an additional `#travel` call: `DocID.fromString(input).travel(commit)`. Same is true for CommitID.

`CommitID.fromString` and `CommitID.fromBytes` throw an error if there is no commit identifier in the input. One can not parse `DocID` representation with `CommitID.fromString`/`CommitID.fromBytes` and get a reference to a genesis commit.

Re: `DocRef` namespace: I find it more concise to write `DocRef.from` than `docRefFrom`.


ukstv:
Just in case, I am not entirely happy with `travel` name, so suggestions are highly welcome.

packages/docid/src/commit-id.ts
```diff
@@ -0,0 +1,225 @@
...
+   * Commit CID
+   */
+  get commit(): CID {
+    return this.#commit || this.#cid;
+  }
+
+  /**
+   * Bytes representation
+   */
+  @Memoize()
```
oed, line 165: What does this do?

packages/docid/src/commit-id.ts
```diff
@@ -0,0 +1,225 @@
...
+  toUrl(): string {
+    return `ceramic://${this.toString()}`;
+  }
+
+  /**
+   * If genesis commit:
+   * CommitID(k3y52l7mkcvtg023bt9txegccxe1bah8os3naw5asin3baf3l3t54atn0cuy98yws, 0)
+   *
+   * If commit:
+   * CommitID(k3y52l7mkcvtg023bt9txegccxe1bah8os3naw5asin3baf3l3t54atn0cuy98yws, bagjqcgzaday6dzalvmy5ady2m5a5legq5zrbsnlxfc2bfxej532ds7htpova)
```
oed, line 207: Can we please output this as an encoded CommitID? 
I.e. this.toString().

packages/docid/src/doc-id.ts
```diff
@@ -0,0 +1,157 @@
...
+   * @readonly
+   */
+  get baseID(): DocID {
+    return new DocID(this.#doctype, this.#cid);
+  }
+
+  /**
+   * Construct new CommitID for the same document, but a new `commit` CID.
+   */
+  travel(commit: CID | string | number): CommitID {
```
oed, line 115: `toCommitID`?

oed (changes requested):
This seems mostly good. Left some feedback.

packages/docid/src/commit-id.ts
ukstv: As the name suggests, it memoizes the getter. No longer need to manually track last-calculated value in a private variable like _bytes.

If called for the first time, the getter calculates the result. If called later, the previously calculated value is returned.

ukstv: This is the smallest memoization library I know. Has no external dependencies.

packages/docid/src/commit-id.ts
ukstv: The point of this inspection is to provide an engineer with a hint about document identifier and commit. IMO two-part output is more comprehensible while triage. And if not triage, I do not see why the inspection is ever needed. Do you think that one-part makes it easier to read?

packages/docid/src/commit-id.ts
oed: neat!

packages/docid/src/doc-id.ts
ukstv: For DocID it is perfect. For CommitID, I feel like `commitId.toCommitID` is weird, and seems like something surprising.

Something like this goes across my mind: "Here is commit id, we transform it into another commit id. Commit id - to commit id. Hmm, it must be something fishy then"

If it doesn't cause a similar effect on you, I will change the name.

oed: Do they need to have the same name in both classes? Could be `toOtherCommitID` in the CommitID class?

ukstv: atCommit maybe??

oed: Makes sense to me 👍

PaulLeCam (approved):
Looks great!
I like the separation between the 2 types while still having the `DocRef` interface share between the 2!

I agree with Joel about the `travel` method name though, I don't think it's an obvious/common name for such a method so would be good to have a more explicit one.

ukstv:
Just in case, kudos to @stbrody for suggesting `DocID | CommitID`.

ukstv:
`travel` is now `atCommit` !

packages/docid/src/commit-id.ts
ukstv: Summoning @stbrody and @valmack as consumers of the API here I have a suspicion, two-part display here makes sense, but as I do not consume the API (yet), I have little to say.

packages/docid/src/__tests__/doc-id.test.ts
```diff
@@ -0,0 +1,210 @@
...
+  test('create from string', () => {
+    const docid = DocID.fromString(DOC_ID_STRING);
+
+    expect(docid.type).toEqual(0);
+    expect(docid.cid.toString()).toEqual(CID_STRING);
+    expect(docid.toString()).toMatchSnapshot();
+  });
+
+  test('create from string including commit', () => {
+    const docid = DocID.fromString(DOC_ID_WITH_COMMIT);
```
stbrody, line 81: It feels like maybe we should throw in this case?

packages/cli/src/ceramic-cli-utils.ts
```diff
@@ -137,13 +137,6 @@ export class CeramicCliUtils {
      */
     static async change(docId: string, content: string, controllers: string, schemaDocId?: string): Promise<void> {
         const id = DocID.fromString(docId)
-
-        const commit = id.commit
-        if (commit) {
-            console.error(`No commits allowed. Invalid docId: ${id.toString()}`)
```
stbrody, line 143: This is a slight behavior change since `DocID.fromString()` does not currently error if you give it a string that represents a CommitID.  So where currently we'd error if you passed a CommitId string to `ceramic change` on the CLI, now we'll allow it and operate on the base DocId.  I'm not sure that's a change we actually want?  I kind of feel like `DocID.fromString()` should error if given a CommitID

packages/core/src/document.ts
```diff
@@ -652,11 +645,8 @@ class Document extends EventEmitter implements DocStateHolder {
    */
   static async loadSchemaById<T extends Doctype>(context: Context, schemaDocId: string): Promise<T> {
     if (schemaDocId) {
-      const schemaDocIdParsed = DocID.fromString(schemaDocId)
-      if (!schemaDocIdParsed.commit) {
-        throw new Error("Commit missing when loading schema document")
```
stbrody, line 657: Can we keep this error around, which is a bit more descriptive than the `No commit information provided` message you'll get without it?  It's useful to know that we're in the context of loading a schema and that schemas must always be set at a specific commit

packages/core/src/document.ts
```diff
@@ -206,7 +199,7 @@ class Document extends EventEmitter implements DocStateHolder {
 
     const commit = await dispatcher.retrieveCommit(doc._genesisCid)
     if (commit == null) {
-      throw new Error(`No commit found for CID ${id.commit.toString()}`)
+      throw new Error(`No commit found for CID ${id.toString()}`)
```
stbrody, line 202:
```suggestion
      throw new Error(`No genesis commit found with CID ${doc._genesisCid.toString()}`)
```

stbrody (commented):
This looks really great!  The APIs all seem much clearer now.  A few small questions/comments but nothing major, the biggest one is probably around whether or not DocID.fromString should throw when given a CommitID string representation.

packages/docid/src/doctypes.ts
```diff
@@ -0,0 +1,22 @@
+const registry: Record<string, number | undefined> = {
+  tile: 0,
+  'caip10-link': 1,
```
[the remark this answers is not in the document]
ukstv: `caip10-link` is not a valid literal for JS key, so must be quoted. tile can be quoted as well, but not required.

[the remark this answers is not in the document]
stbrody: okay makes sense.  The asymmetry just threw me

zachferland (approved):
nice, +1 on DocId/CommitID and atCommit, and some other nice code improvements along the way, looks good

packages/cli/src/ceramic-cli-utils.ts
ukstv: Makes sense.

packages/docid/src/__tests__/doc-id.test.ts
ukstv: should.

packages/core/src/document.ts
ukstv: OK

packages/docid/src/commit-id.ts
ukstv: > if a debugger was attached

yes.

ukstv:
Addressed all the comments:
- CommitID inspection now returns its string representation
- DocID.fromString and DocID.fromBytes are strict now

PR #1011, opened 2021-02-16
feat(core): Remove 'exists' method from StateStore API

stbrody:


ukstv:
I oppose this change, sorry. In general `exists` query for any database has same or better performance than loading a value ([Postgres for example](https://blog.jooq.org/2016/09/14/avoid-using-count-in-sql-when-you-could-use-exists/)). Here, we have also an overhead in `load` for deserialising a found record. Even for LevelDB, this change definitely gets worse performance, and I see no definitive gain out of the move. 

It was my mistake to call `load` inside `exists`, while `level-ts` supports `exists` directly. If `exists` is in focus, it is better to base it on `level#exists`, rather than on the current `LevelStateStore#load`.

ukstv:
Nah, would like to retract my comment. Sorry.

Makes sense.

ukstv (approved):
🏇

oed (approved):

PR #1021, opened 2021-02-18
feat(core): Clean up Ceramic.create() and accept LoggerProvider in CeramicConfig

stbrody:


packages/core/src/ceramic.ts
```diff
@@ -237,93 +290,109 @@ class Ceramic implements CeramicApi {
...
+  /**
+   * Create Ceramic instance
+   * @param ipfs - IPFS instance
+   * @param config - Ceramic configuration
+   */
+  static async create(ipfs: IpfsApi, config: CeramicConfig = {}): Promise<Ceramic> {
+    // todo: Remove ipfs argument, pull logic to start in-process IPFS here and out of CeramicDaemon
+    const [params, modules] = await Ceramic._processConfig(config)
+
+    modules.ipfs = ipfs
```
stbrody, line 361: I wasn't able to totally get this review to the place I wanted it today, which is why this PR is a draft, but there's enough moving pieces here I thought it makes sense to get some other eyes on it sooner than later.  The main things I still want to do are:
* add a `setIpfs()` method to the `Ceramic` object that is called after the constructor but before `_init()` and can take care of initializing these ipfs-dependent modules
* Remove the `ipfs` argument from Ceramic.create(), let ceramic internally handle creating the IpfsDaemon or the ipfs http client based on the config, rather than having the CeramicDaemon do it.
* Finally provide the logger instance to the IpfsDaemon and IpfsTopology so they no longer use `console.log` directly.

The above is my plan for tomorrow (today for you already), but if you have any feedback about anything I'm doing here I wanted to give you a chance to give it before I start work in the morning and go further down this path.

stbrody:
Ready for real review @ukstv @oed

packages/core/src/ceramic.ts
```diff
@@ -120,32 +147,52 @@ const tryDocId = (id: string): DocID | null => {
...
+  private readonly _doctypeHandlers: Record<string, DoctypeHandler<Doctype>>
   private readonly _docCache: DocCache
+  private readonly _ipfsTopology: IpfsTopology
+  private readonly _logger: DiagnosticsLogger
+  private readonly _networkOptions: CeramicNetworkOptions
+  private readonly _pinStoreFactory: PinStoreFactory
+  private readonly _validateDocs: boolean
+
+  constructor (params: CeramicParameters,
+               modules: CeramicModules) {
```
oed, line 166: I would flip the order of these params, but it really doesn't matter 😄

oed (approved):
Lgtm 👍

packages/core/src/ceramic.ts
```diff
@@ -237,93 +284,107 @@ class Ceramic implements CeramicApi {
...
-   * Create Ceramic instance
-   * @param ipfs - IPFS instance
-   * @param config - Ceramic configuration
+   * Parses the given `CeramicConfig` and generates the appropriate `CeramicParameters` and
+   * `CeramicModules` from it. This usually should not be called directly - most users will prefer
+   * to call `Ceramic.create()` instead which calls this internally.
    */
-  static async create(ipfs: IpfsApi, config: CeramicConfig = {}): Promise<Ceramic> {
-    // todo remove
+  static async _processConfig(ipfs: IpfsApi, config: CeramicConfig): Promise<[CeramicParameters, CeramicModules]> {
```
ukstv, line 291: @stbrody As it can be called by a consumer, this better not have a leading underscore.

packages/core/src/ceramic.ts
```diff
@@ -237,93 +284,107 @@ class Ceramic implements CeramicApi {
...
+    return ceramic
+  }
+
+  /**
+   * Finishes initialization and startup of a Ceramic instance. This usually should not be called
+   * directly - most users will prefer to call `Ceramic.create()` instead which calls this internally.
+   * @param doPeerDiscovery - Controls whether we connect to the "peerlist" to manually perform IPFS peer discovery
+   * @param restoreDocuments - Controls whether we attempt to load pinned document state into memory at startup
+   */
+  async _init(doPeerDiscovery: boolean, restoreDocuments: boolean): Promise<void> {
```
ukstv, line 375: I dislike having it here, but I see no other way :/

packages/core/src/ceramic.ts
ukstv: `_init` will go away, but this will remain, so better rename it now.

ukstv (approved):
🏇 Except one leading underscore.

packages/core/src/ceramic.ts
stbrody: So I did this on purpose, but it is admittedly an unusual/controversial choice.  Basically I was thinking of `_processConfig` and `_init` as kind of "semi-private" functions.  They're not technically private so they *can* be called by a consumer, but most of the time they *shouldn't*.  So the leading underscore is a signal to consumers that these aren't really part of the standard API

stbrody: I'm going to go ahead and merge this since it's ready to go, but I have another PR coming up that touches this code so I can change it there if you'd like me to drop the underscore, I don't feel super strongly either way

ukstv: Well, let it be then along with _init, and let's wait for complaints, if they ever appear.

PR #1041, opened 2021-02-19
feat(cli): Add S3StateStore

stbrody:
No automated testing yet as I can't see a good way to test this outside of an integration test that actually speaks to S3, so that will come in https://github.com/ceramicnetwork/js-ceramic/issues/990.  But I have tested manually and everything looked good.

To use this you need to provide the S3 configuration in a way that the AWS library knows to look for it.  I did all my testing by setting the AWS_REGION, AWS_ACCESS_KEY_ID, and AWS_SECRET_ACCESS_KEY environment variables, although I believe there may also be a standard config file location that the AWS library knows to look in, but I didn't test that


packages/core/src/store/level-state-store.ts
```diff
@@ -22,8 +24,11 @@ export class LevelStateStore implements StateStore {
     /**
      * Open pinning service
      */
-    async open(): Promise<void> {
-        this.#store = new Level(this.storePath);
+    async open(networkName: string): Promise<void> {
+        // Always store the pinning state in a network-specific directory
+        const storePath = path.join(this.storeRoot, networkName)
+        await fs.mkdir(storePath, { recursive: true }) // create dir if it doesn't exist
```
valmack, line 30: Can we add some error handling here?

For example if the process doesn't have write access to the given path, it'd be helpful to return a descriptive error and im guessing we want to exit in that case.

packages/core/src/store/pin-store-factory.ts
```diff
@@ -2,38 +2,45 @@ import { IpfsApi, PinningBackendStatic } from "@ceramicnetwork/common";
...
     }
 
-    async open(): Promise<PinStore> {
-        await fs.mkdir(this.stateStorePath, { recursive: true }) // create dir if it doesn't exist
-        const stateStore = new LevelStateStore(this.stateStorePath)
+    public setStateStore(stateStore: StateStore): void {
+      this._stateStore = stateStore
+    }
+
+    public async open(): Promise<PinStore> {
```
valmack, line 39: Seems like this method name should be `create`? Bc it creates and returns a new StateStore

packages/core/src/index.ts
```diff
@@ -0,0 +1,5 @@
+import Ceramic from "./ceramic";
+
+export * from './store/state-store'
```
valmack, line 3: What do we gain from this export?

packages/core/src/store/pin-store.ts
```diff
@@ -14,8 +14,8 @@ export class PinStore {
       readonly resolve: (path: string) => Promise<CID>
     ) {}
 
-    async open(): Promise<void> {
-        await this.stateStore.open()
+    async open(networkName: string): Promise<void> {
+        await this.stateStore.open(networkName)
```
valmack, line 18: So the "StateStore" holds both the state of any document commits/tips loaded and the tips of any documents created. Does that all happen through the PinStore?

Here it seems there could be other entities besides the PinStore that could use the same stateStore, however the `open` method presumably should be called only once so it seems out of place here.

If my understanding is correct my suggestion would be to open the stateStore in the Ceramic initialization instead.

packages/core/src/store/level-state-store.ts
stbrody: If this fails for any reason it throws an error which bubbles up to the top level.  As an experiment I tried deleting the `statestore` directory and taking away write access from the `~/.ceramic` directory and ran the daemon.  I got an error: 
`Error: EACCES: permission denied, mkdir '/home/spencer/.ceramic/statestore/testnet-clay'`.  That kinda seems good enough for now?

packages/cli/src/bin/ceramic.ts
```diff
@@ -9,7 +9,8 @@ program
     .option('--anchor-service-api <url>', 'The anchor service URL to use')
     .option('--validate-docs', 'Validate documents according to their schemas. It is enabled by default')
     .option('--ipfs-pinning-endpoint <url...>', 'Ipfs pinning endpoints')
-    .option('--state-store-directory <string>', `The directory path used for storing pinned document state. Defaults to HOME_DIR/.ceramic/pinset`)
+    .option('--state-store-directory <string>', `The directory path used for storing pinned document state. Defaults to HOME_DIR/.ceramic/statestore`)
```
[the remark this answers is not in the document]
stbrody: Renamed to log-directory

packages/core/src/index.ts
stbrody: The ability to import the `StateStore` interface in the `cli` package, where the `S3StateStore` is defined

packages/core/src/store/pin-store-factory.ts
stbrody: renamed to `createPinStore()`

packages/core/src/store/pin-store.ts
stbrody: hmm, the `StateStore` *is* sometimes used directly, but always via a handle from the pinStore.  I can't imagine there ever being a time when we would have a `StateStore` but not have a PinStore.  In fact, with the stuff Sergey is working on now, I expect all these internal pieces to become even more hidden away behind the new `DocStateRepository` interface.  So I think we want to be moving in the direction of pushing more of these pieces down behind interfaces rather than pulling more of them up to be used directly in higher layers

oed:
> To use this you need to provide the S3 configuration in a way that the AWS library knows to look for it. I did all my testing by setting the AWS_REGION, AWS_ACCESS_KEY_ID, and AWS_SECRET_ACCESS_KEY environment variables, although I believe there may also be a standard config file location that the AWS library knows to look in, but I didn't test that

Is there no way to pass this to the library on construction time? Would be nice if we could set this configuration in `~/.ceramic/config.json`.

stbrody:
> > To use this you need to provide the S3 configuration in a way that the AWS library knows to look for it. I did all my testing by setting the AWS_REGION, AWS_ACCESS_KEY_ID, and AWS_SECRET_ACCESS_KEY environment variables, although I believe there may also be a standard config file location that the AWS library knows to look in, but I didn't test that
> 
> Is there no way to pass this to the library on construction time? Would be nice if we could set this configuration in `~/.ceramic/config.json`.

Yes we could.  In fact, my first version of this PR did it that way.  Take a look at [this commit](https://github.com/ceramicnetwork/js-ceramic/pull/1041/commits/8f52adb4436bdba84da39718a04ef3df44ae0a9a) from this PR's commit history.  I could revert that commit and it would be possible to explicitly configure these settings.  The reason I ripped that out is that I came to understand that the AWS package is already configured to look for aws configuration options in certain well known places, so if we don't provide the options explicitly, it will load them via the standard AWS ways, which includes its [own config file|https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/setting-region.html#setting-region-config-file].  So I opted to leave things to work in the "standard" way for AWS configs.

I have no idea though how other libraries that utilize the AWS api tend to expose these configs - whether its actually common to let the inner AWS library handle the configuration in the way it wants to, or if it's most common to repackage those config options as configs to the higher-level library.

stbrody:
> Would be nice if we could set this configuration in ~/.ceramic/config.json

Just FYI, the `config.json` file currently can only be used to set options for the http-client started by the CLI when interacting *with* a daemon for creating documents, etc.  It is not used when *starting* the daemon.  If we want a way for a config file to control the actual full ceramic node that is started when running `ceramic daemon` we'll need to do more work to enable that.

packages/core/src/store/pin-store.ts
valmack: Ok so state store is only used if you explicitly choose to pin something? If that is the case then it makes sense how it is

valmack:
> If we want a way for a config file to control the actual full ceramic node that is started when running `ceramic daemon` we'll need to do more work to enable that.

it’s definitely helpful to have be forced to set config from two different places

As long as we keep ability to use env vars I’m happy.

packages/cli/src/bin/ceramic.ts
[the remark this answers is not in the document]
valmack: Oof I should have said, path makes more sense because we need full path, not just directory name.

oed:
> Yes we could. In fact, my first version of this PR did it that way. Take a look at this commit from this PR's commit history. I could revert that commit and it would be possible to explicitly configure these settings. The reason I ripped that out is that I came to understand that the AWS package is already configured to look for aws configuration options in certain well known places, so if we don't provide the options explicitly, it will load them via the standard AWS ways, which includes its own config file. So I opted to leave things to work in the "standard" way for AWS configs.

> I have no idea though how other libraries that utilize the AWS api tend to expose these configs - whether its actually common to let the inner AWS library handle the configuration in the way it wants to, or if it's most common to repackage those config options as configs to the higher-level library.

Ehm, do we document this anywhere in that case? Seems kind of bad that you need to know how an underlaying library works in order to use this feature in ceramic?

stbrody:
> > Yes we could. In fact, my first version of this PR did it that way. Take a look at this commit from this PR's commit history. I could revert that commit and it would be possible to explicitly configure these settings. The reason I ripped that out is that I came to understand that the AWS package is already configured to look for aws configuration options in certain well known places, so if we don't provide the options explicitly, it will load them via the standard AWS ways, which includes its own config file. So I opted to leave things to work in the "standard" way for AWS configs.
> 
> > I have no idea though how other libraries that utilize the AWS api tend to expose these configs - whether its actually common to let the inner AWS library handle the configuration in the way it wants to, or if it's most common to repackage those config options as configs to the higher-level library.
> 
> Ehm, do we document this anywhere in that case? Seems kind of bad that you need to know how an underlaying library works in order to use this feature in ceramic?

Yep, that's a totally fair concern.  I'm happy to revert that commit and put these options into explicit config settings for the CeramicDaemon if you think that's best.  I just wanted to raise the issue that there does seem to be some sort of standard way to provide this configuration.  Also if there are other client-side AWS options that users might want to control, we'd have to expose each one ourselves in that route, vs if we let the AWS library handle it then users have access to all the underlying config options they might want.  We will certainly have to make sure to document this, but we'd have to have some docs of these options in either case anyway.

packages/cli/src/bin/ceramic.ts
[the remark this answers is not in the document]
stbrody: I guess I went with directory over path b/c in my mind "path" sounds like it's the full path to the specific file where the logs or state store will be, when in fact it's just the path to the directory where we will create files and possibly sub-directories according to a layout we decide.

packages/core/src/store/pin-store.ts
stbrody: Yeah, there will only be state in the state store for docs that are pinned

valmack: 🙏 thanks for clarity

packages/cli/src/bin/ceramic.ts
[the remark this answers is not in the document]
valmack: Either way consistency is most important thing so this is good :)

valmack (approved):
👍

zachferland:
yeah its very common and i think expected really, for most all libraries to leave that to underlying aws sdk, and you may configure other options there as well. As spencer mentioned it already pull these vars from some default places and this allows aws sdk to make assumptions about where its going to fetch them during runtime depending on the aws service it is running in (you wont necessarily be able to easily fetch these values ahead of runtime in some service configurations).

Also for testing, may make sense to test locally with something like https://github.com/minio/minio, object store with same s3 interface.

stbrody:
> Also for testing, may make sense to test locally with something like https://github.com/minio/minio, object store with same s3 interface.

Interesting.  You're suggesting starting an in-process minio instance in the jest test, and pointing the ceramic node at that for its S3 storage?  When I looked briefly at minio it looked much more heavyweight than that, like it can actually be used in production deployments rather than just within tests.  But having some kind of lightweight S3-api compatible service we could start within the tests would certainly be really great.  But so far I can't even find a way to start a minio server from within javascript?  Do you know if this is possible @zachferland?

zachferland:
yeah agree on second look again, likely too heavy for tests here still, although i thought that project/community also had some test/mock tooling. Just in 3box before we ran a docker image with it in local configurations, local development and for some more integration style tests.

stbrody:
@oed, okay to merge as-is with config management delegated to AWS library?

oed (approved):

nahsi:
> > Also for testing, may make sense to test locally with something like https://github.com/minio/minio, object store with same s3 interface.
> 
> Interesting. You're suggesting starting an in-process minio instance in the jest test, and pointing the ceramic node at that for its S3 storage? When I looked briefly at minio it looked much more heavyweight than that, like it can actually be used in production deployments rather than just within tests. But having some kind of lightweight S3-api compatible service we could start within the tests would certainly be really great. But so far I can't even find a way to start a minio server from within javascript? Do you know if this is possible @zachferland?

Usually you start a development instance of minio in docker and then use it to run tests against.

There is a library for that for golang https://github.com/testcontainers/testcontainers-go . For javascripts something similar should exist.

PR #1043, opened 2021-02-21
chore(core): Add additional ceramic logging

stbrody:


packages/cli/src/bin/ceramic.ts
```diff
@@ -13,6 +13,7 @@ program
     .option('--gateway', 'Makes read only endpoints available. It is disabled by default')
     .option('--port <int>', 'Port daemon is available. Default is 7007')
     .option('--debug', 'Enable debug logging level. Default is false')
+    .option('--verbose', 'Enable verbose logging level. Default is false')
```
oed, line 16: What happens if you pass both of these?

oed (approved):

packages/cli/src/bin/ceramic.ts
stbrody: --verbose wins

PR #1044, opened 2021-02-22
feat: Introduce Repository

ukstv:
Dummy one. No cache eviction, no persistence, even Dispatcher is the same.

### DIFF
--- a/packages/core/src/ceramic.ts
+++ b/packages/core/src/ceramic.ts
@@ -459,15 +459,6 @@ class Ceramic implements CeramicApi {
   }
 
   /**
-   * Get document from cache by DocID
-   * @param docId - Document ID
-   * @private
-   */
-  private _getDocFromCache(docId: DocID): Document {
-    return this._docCache.get(docId) as Document
-  }
-
-  /**
    * Create doctype instance
    * @param doctype - Document type
    * @param params - Create parameters
--- a/packages/core/src/ceramic.ts
+++ b/packages/core/src/ceramic.ts
@@ -22,10 +22,8 @@ import {
   PinApi,
   MultiQuery,
   PinningBackendStatic,
-  DocCache,
   AnchorStatus,
   LoggerProvider,
-  LoggerConfig,
 } from "@ceramicnetwork/common"
 import { Resolver } from "did-resolver"
 
@@ -157,7 +155,7 @@ class Ceramic implements CeramicApi {
   public pinStore: PinStore // Set during init()
 
   private readonly _doctypeHandlers: Record<string, DoctypeHandler<Doctype>>
-  private readonly _docCache: DocCache
+  private readonly _repository: Repository
   private readonly _ipfsTopology: IpfsTopology
   private readonly _logger: DiagnosticsLogger
   private readonly _networkOptions: CeramicNetworkOptions
@@ -193,7 +191,7 @@ class Ceramic implements CeramicApi {
       'caip10-link': new Caip10LinkDoctypeHandler()
     }
 
-    this._docCache = new DocCache(params.docCacheLimit, params.cacheDocumentCommits)
+    this._repository = modules.repository
   }
 
   /**
@@ -401,7 +399,7 @@ class Ceramic implements CeramicApi {
    */
   async _init(doPeerDiscovery: boolean, restoreDocuments: boolean): Promise<void> {
     this.pinStore = await this._pinStoreFactory.createPinStore()
-    this.pin = new LocalPinApi(this.pinStore, this._docCache, this._loadDoc.bind(this), this._logger)
+    this.pin = new LocalPinApi(this.pinStore, this._loadDoc.bind(this), this._logger)
 
     if (doPeerDiscovery) {
       await this._ipfsTopology.start()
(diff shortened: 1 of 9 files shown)


packages/core/src/dispatcher.ts
```diff
@@ -284,18 +278,15 @@ export default class Dispatcher extends EventEmitter {
    * @private
    */
   async _handleUpdateMessage(message: any): Promise<void> {
-    // TODO Add validation the message adheres to the proper format.
-
     const { doc, tip } = message
-    if (!this._documents[doc]) {
-      return
+    const docId = DocID.fromString(doc)
+    if (await this.repository.has(docId)) {
```
stbrody, line 283: Can we invert this if-statement and do an early-return if we don't care about the message, as was done previously?  I'm a big fan of early-returns rather than introducing additional layers of indentation and scope nesting.  I find the code tends to be easier to read and less error-prone.

Here and in the other handle*Message functions below

packages/core/src/dispatcher.ts
```diff
@@ -354,7 +349,7 @@ export default class Dispatcher extends EventEmitter {
 
     clearInterval(this._resubscribeInterval)
 
-    await Promise.all(Object.values(this._documents).map(async (doc) => await doc.close()))
+    await this.repository.close()
```
stbrody, line 352: I'm not convinced it makes sense for the Dispatcher to logically "own" the Repository and be responsible for its lifecycle like this.  Kinda feels like this should happen up in `Ceramic.close()`?

packages/core/src/repository.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  /**
+   * Stub for async loading of the document.
+   */
+  async get(docId: DocID): Promise<Document> {
+    const found = this.#map.get(docId.toString());
+    if (found) {
+      return found;
+    } else {
+      throw new Error(`No document found for id ${docId}`);
```
stbrody, line 15: I'm concerned about the paradigm of always calling `Repository.has()` before `Repository.get()` with the assumption that if `.has()` returns true then the doc will still be there when calling `.get()`.  If the document is at the very end of the LRU cache about to age out, we call `.has()`, it returns true, then the schedule switches contexts and a new doc is inserted into the cache, then when `.get()` is schedule the document can be missing, despite having previously returned true for `.has()`.

stbrody: I guess if the call to `.has()` queries the cache and marks the doc as accessed it will reset its order in the LRU, so maybe this is unlikely to ever be an actual problem in practice.  Nevertheless I wonder is there actually value to having this separation between `has` and `get`?  Can't `get()` just return `null` if it doesn't have it?

packages/core/src/repository.ts
```diff
@@ -0,0 +1,46 @@
+import Document from './document';
+import DocID from '@ceramicnetwork/docid';
+
+export class Repository {
+  readonly #map: Map<string, Document> = new Map();
+
+  /**
+   * Stub for async loading of the document.
```
stbrody, line 8: Are you expecting that the document loading will actually happen within the Repository?  I had imagined the Repository just being focused on managing the caching and state persistence layer, without any of the actual business logic for loading a document.  I had hoped when a document isn't in the Repository and needs to be loaded, it would be loaded outside of the Repository and then provided to the Repository for caching/persistence once the load was complete.

stbrody (commented):
Looks good, my primary question is around how documents get loaded and added to the Repository

oed (approved):
So this is mostly to introduce the repo interface? Looks good.

ukstv:
It just consolidates caching behavior. Nothing more. Does not do fancy stuff. Damn simple.

Re: has-get. „Has“=true means „I track the document“. „Get“ means „load the document no matter where it is stored“. Two primitive actions that together give you proper semantics. „I want the document  if you track it“ or „I just want this document”.

stbrody:
> Get“ means „load the document no matter where it is stored“.

Yeah, I'm not actually that concerned about the `has` method.  It's unlikely to ever really be a problem in practice.  It's the part where `Repository.get()` loads the document that I'm actually more concerned about.  I'd prefer to keep a greater separation of concerns.  `Repository` handles caching and persistence.  `StateManager` handles document loading and transformation.  `Ceramic` handles tying them together.

Mixing the networking and business logic related to document loading with the comparatively simple logic around state caching and persistence changes the Repository from what could a pretty simple and focused subsystem into something more complicated.  It seems better to me to keep the logic for document loading alongside the logic for applying document transformations, since all of that has to do networking i/o and deal with doctype-specific logic, and keep all of that out of such a low-level component as the Repository.

stbrody:
That said, I'm not going to block this PR over this.  If you want to move forward with the approach you have here we can do so and see how it looks when it all comes together.

packages/core/src/repository.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  /**
+   * The document is in memory or on disk to load.
+   */
+  async has(docId: DocID): Promise<boolean> {
+    return this.#map.has(docId.toString());
+  }
+
+  /**
+   * Remove from memory. Stub.
```
valmack, line 34: What does “stub” mean?

ukstv: Only that it does not do much.

packages/core/src/ceramic.ts
```diff
@@ -502,14 +494,13 @@ class Ceramic implements CeramicApi {
...
     const docId = new DocID(doctype, genesisCid)
 
-    let doc = this._getDocFromCache(docId)
-    if (doc) {
-      return doc
+    if (await this._repository.has(docId)) {
+      return this._repository.get(docId)
+    } else {
+      const document = await Document.create(docId, doctypeHandler, this.dispatcher, this.pinStore, this.context, opts, this._validateDocs);
+      // this.repository.add(document) TODO See Document#register, it adds to the repository too
```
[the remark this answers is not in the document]
ukstv: Uhums, just not all at once. This mark is a reminder not to forget to to have a replacement for this `this.repository.add` statement.

packages/core/src/ceramic.ts
```diff
@@ -465,14 +458,13 @@ class Ceramic implements CeramicApi {
     const genesisCid = await this.dispatcher.storeCommit(genesis)
     const docId = new DocID(doctype, genesisCid)
 
-    let doc = this._getDocFromCache(docId)
-    if (doc) {
-      return doc
+    if (await this._repository.has(docId)) {
```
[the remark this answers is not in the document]
ukstv: Maybe. In the next incarnation.

ukstv:
I need this element of the puzzle. Going to merge.

PR #1058, opened 2021-02-24
feat: Feed of pubsub messages

ukstv:
State refactor plan needs a proper pubsub message feed.


packages/core/src/pubsub/incoming-channel.ts
```diff
@@ -0,0 +1,107 @@
...
+      this.tasks.add(() => resubscribe(ipfs, topic, pubsubLogger, handler));
+
+      const ensureSubscribed = interval(resubscribeEvery).subscribe(() => {
+        this.tasks.add(() => resubscribe(ipfs, topic, pubsubLogger, handler));
+      });
+
+      return () => {
+        ensureSubscribed.unsubscribe();
+        // Remove pending subscription attempts.
+        this.tasks.clear();
```
stbrody, line 65: Could more tasks get added in between this `clear()` and when the unsubscribe task added below actually runs?

packages/core/src/pubsub/incoming-channel.ts
```diff
@@ -0,0 +1,107 @@
...
+        });
+      };
+    });
+  }
+
+  // We want all the tasks added to execute in FIFO order.
+  // Subscription attempts must be sequential.
+  // Last call to unsubscribe must execute after all the attempts are done,
+  // and all the attempts yet inactive are cleared.
+  @Memoize()
```
stbrody, line 78: This is a little more magical/indirect than I prefer.  Can we just make a private `#tasks` member that we initialize once in the constructor, rather than memoizing the getter like we're doing here?

packages/core/src/pubsub/incoming-channel.ts
```diff
@@ -0,0 +1,107 @@
...
+      }
+    });
+  }
+}
+
+/**
+ * Pass only messages from other IPFS nodes.
+ * @param peerId$ - Own peer id.
+ */
+export function filterOuter(peerId$: Observable<string>): MonoTypeOperatorFunction<IPFSPubsubMessage> {
```
stbrody, line 97: How about `filterOwnMessages`?  `filterOuter` isn't really clear to me from the name what it's doing

packages/core/src/pubsub/pubsub-message.ts
```diff
@@ -0,0 +1,119 @@
...
+  doc: DocID;
+};
+
+export type ResponseMessage = {
+  typ: MsgType.RESPONSE;
+  id: string;
+  tips: Map<string, CID>;
+};
+
+export type PubsubMessage = UpdateMessage | QueryMessage | ResponseMessage;
```
stbrody, line 37: 😍

packages/core/src/pubsub/pubsub.ts
```diff
@@ -0,0 +1,86 @@
...
+      const incoming$ = new IncomingChannel(ipfs, topic, resubscribeEvery, pubsubLogger, logger);
+
+      incoming$
+        .pipe(filterOuter(this.peerId$), ipfsToPubsub(this.peerId$, pubsubLogger, topic))
+        .subscribe(subscriber);
+    });
+  }
+
+  @Memoize()
+  private get peerId$() {
```
stbrody, line 61: ditto my earlier comment.  Can this just be a const member rather than a memoized accessor function?

packages/core/src/pubsub/pubsub.ts
```diff
@@ -0,0 +1,86 @@
...
+    return from<Promise<string>>(this.ipfs.id().then((_) => _.id));
+  }
+
+  /**
+   * Publish message to IPFS Pubsub as a "fire and forget" command.
+   * Returned Subscription is free to ignore.
```
stbrody, line 67: is there ever a case in which you would use it?  If so, how?  If not, why return it at all?

packages/core/src/pubsub/task-queue.ts
```diff
@@ -0,0 +1,43 @@
...
+   */
+  constructor(private readonly onError: (error: Error, retry?: () => void) => void = noop) {}
+
+  /**
+   * Add task to queue.
+   */
+  add(f: Task<void>): void {
+    this.#pq.add(f).catch((error) => {
+      const retry = () => this.add(f);
+      this.onError(error, retry);
```
stbrody, line 26: does this mean that all tasks in the queue that error will be retried indefinitely?  Should we have a cap of how many retry attempts it gets so we don't get stuck in an infinitely failing loop?

EDIT: Nevermind, I see now it's the `onError` callback that decides if/when to retry.  I think we need some comments to make that clearer.  Maybe give the onError function signature a type and have a comment describing how it's used?

stbrody (commented):
This looks really awesome.  I love how simple the Dispatcher becomes here.

I'm still a little bit concerned about how opaque some of the RXJS code is to me (and thus presumably will be to others trying to operate in our codebase who aren't already familiar with rxjs).  I kind of have to just trust that a lot of this code is actually doing what I think it is since I don't know the rxjs apis and many of them seem not super obvious at first glance, which is a bit disappointing.  I'm also a bit worried about what it will be like trying to debug issues in the pipelines.  But maybe all of that is just an unavoidable cost of moving to this model.  There certainly are lots of advantages that likely outweigh the steeper learning curve for our codebase.  Just seemed worth flagging that there is a cost here @oed 

But anyway, that's all more of a statement on rxjs itself than this PR.  I still think this is the right way to go, and I'm excited about the foundation this PR lays for us.  I left a few comments throughout, but nothing major.  Nice work!

packages/core/src/pubsub/incoming-channel.ts
ukstv: The only thing can do this is controlled by `ensureSubscribed`. After it is effectively stopped (`ensureSubscribed.unsubscribe`), nothing can interfere.

packages/core/src/pubsub/incoming-channel.ts
ukstv: Agree on magical/indirect. All the other ways I can think of to structure this are even more indirect an definitely less clear: https://gist.github.com/ukstv/acebcddbf77a330f8f2d0d340a2a7910 

Maybe https://gist.github.com/ukstv/acebcddbf77a330f8f2d0d340a2a7910#file-4-ts is okay to you?

packages/core/src/pubsub/incoming-channel.ts
ukstv: _Fillter_ is a tough word. In JS (and other languages I know), `filter(predicate)` passes an item, if `predicate` returns `true`. If it returns `false`, an item is omitted.

Here we would like to pass _outer_ messages, and omit _own_, so `filterOuter`.

packages/core/src/pubsub/pubsub.ts
ukstv: Do you find https://gist.github.com/ukstv/1d5943744a31dbe7151f763c3bf5ee71 clearer?

Memoize and getter explicitly state "this is lazy evaluation, executed once, it is just a function". 
In gist, I have to recall ordering of JS object instantiation, and it is a bit frightening. But maybe this is only my quirk.

packages/core/src/pubsub/pubsub.ts
ukstv: Maybe you would like to do something exactly after the message is published.

ukstv: Like in https://github.com/ceramicnetwork/js-ceramic/pull/1058/files/2729068e8f050414983bb508add23123680246b4#diff-0e6de57da2fc7ba36929ce8a926162b43593faa16fe1750c83a5e5c622560dfcR106

packages/core/src/pubsub/incoming-channel.ts
stbrody: Ah I see.  And this is because `tasks` isn't actually used for incoming pub/sub messages, only to serialize resubscribe attempts.  Okay makes sense

packages/core/src/pubsub/incoming-channel.ts
stbrody: I kind of like a mix of versions 3 and 4.  Constructing `tasks` within the body of the constructor like in version 3, but using a helper function like exists in version 4 (although I'd probably name it `makeTaskQueue` or `makeResubscribeQueue`).

Not a huge deal either way though, so go with whatever you think is cleanest

stbrody: I also like how version 3 pulls out `onSubscribe` into a helper function too, to make the constructor body much smaller and easier to digest

packages/core/src/pubsub/incoming-channel.ts
stbrody: Ah yeah, good point.  I still don't love the word "outer" here though cause it's not clear to me what it's referring to.  How about `filterExternalMessages`?

packages/core/src/pubsub/pubsub.ts
stbrody: Ah okay, then the issue is that `super()` has to be the first line of the constructor, so then we can't set `this.peerId$` until after the code that references it, which looks weird unless you really understand how the subscription is lazy.  Making it a memoized function would make it work even if the callback passed to `super()` was run immediately.

I'd actually probably still vote for the form from the Gist and just add a comment above where `this.peerId$` is initialized saying that this is safe because the above code from `super` is run lazily.  But I do understand better now why you did it this way so if you prefer it this way that's fine by me

packages/core/src/pubsub/pubsub.ts
stbrody: Makes sense, but can we just add a message to the comment clarifying that?

packages/core/src/pubsub/task-queue.ts
stbrody: I'd still love some more comments or something here as I think the relationship between the `onError` callback given to the constructor and the `retry` argument it receives isn't super obvious at first glance

stbrody (approved):
LGTM mod a few requests for some additional comments

packages/core/src/pubsub/pubsub.ts
ukstv: Okay.

packages/core/src/pubsub/task-queue.ts
ukstv: Ack. Hope now it is more clear.

packages/core/src/pubsub/pubsub.ts
ukstv: Yup.

packages/core/src/pubsub/incoming-channel.ts
ukstv: Let's settle on `filterExternal` :) I guess, my broken English strikes here and romantic memories of [Plan 9 from Outer Space](https://en.wikipedia.org/wiki/Plan_9_from_Outer_Space).

oed (approved):
Had a glance over this. Seems reasonable.

packages/core/src/__tests__/dispatcher.test.ts
```diff
@@ -100,20 +101,23 @@ describe('Dispatcher', () => {
   })
 
   it('publishes tip correctly', async () => {
-    const docId = DocID.fromString(FAKE_DOC_ID)
     const tip = new CID('QmSnuWmxptJZdLJpKRarxBMS2Ju2oANVrgbr2xWbie9b2D')
-    dispatcher.publishTip(docId, tip)
-    expect(ipfs.pubsub.publish).toHaveBeenCalledWith(TOPIC, JSON.stringify({ typ: MsgType.UPDATE, doc: FAKE_DOC_ID, tip: tip.toString() }))
+    const subscription = dispatcher.publishTip(FAKE_DOC_ID, tip)
+    subscription.add(() => { // Could be delay, but this is faster
+      expect(ipfs.pubsub.publish).toHaveBeenCalledWith(TOPIC, serialize({ typ: MsgType.UPDATE, doc: FAKE_DOC_ID, tip: tip }))
```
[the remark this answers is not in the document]
ukstv: Converted this to promise. If by any reason (should never happen though) the subscription does not end, jest would complain about timeouts here.

PR #1069, opened 2021-02-26
chore(core): Set controller to the authenticated DID even if no content provided

stbrody:


packages/doctype-tile/src/tile-doctype.ts
```diff
@@ -71,7 +71,7 @@ export class TileDoctype extends Doctype {
 
         const metadata: GenesisHeader = params.metadata || { controllers: [] }
         if (!metadata.controllers || metadata.controllers.length === 0) {
-            if (params.content && context.did) {
```
stbrody, line 74: Was there some reason for this requirement I'm not seeing @oed?

oed: I guess no. As long as the genesis commit is not signed when content is undefined we should be all good.

oed (approved):

PR #1080, opened 2021-03-04
chore(core): Update state store path in S3 to include ceramic prefix

stbrody:
It was mentioned the other day that we wanted the ceramic-related stuff in S3 under a `ceramic/` prefix to parallel the `ipfs/` prefix for the ipfs data


oed (approved):
Thanks ❤️

PR #1081, opened 2021-03-04
Bring cache eviction

ukstv:
Finally.

Not with problems:
- initialisation is a mess,
- concurrency-control is a bit too multilayered,
- Document is present still.


packages/core/src/dispatcher.ts
```diff
@@ -171,8 +162,8 @@ export class Dispatcher {
     // TODO Add validation the message adheres to the proper format.
 
     const { doc: docId, id } = message
-    if (await this.repository.has(docId)) {
-      const document = await this.repository.get(docId)
+    const document = await this.repository.get(docId)
+    if (document) {
```
oed, line 166: Does this actually load the document into memory if we reply to a query?
Seems like it should be possible to just retrieve the tip without loading the document into memory?

oed (commented):
Had a quick look through the changes. One main concern about queries, other than that I think this looks good.

packages/core/src/dispatcher.ts
ukstv: It is possible. And, I see the point in doing that.

ukstv:
Note on pinstore changes:
- tests need in-memory anchor service,
- it needs fully-specified context with dispatcher set,
- dispatcher needs repository,
- repository needs `pinStore.stateStore`

So, this whole bundle has to be set at once. One could do all of it in _init to preserve async stateFactory. Really, there was just a single async operation that makes pinStore initialisation async: it was `fs.promises.mkdir`. I see really no _need_ to leave it async since it is just a directory creation. So, here we are with sync `pinStoreFactory.createPinStore`.

packages/core/src/state-management/named-p-queue.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  /**
+   * Remove a queue. Called after a task is done.
+   *
+   * There can only be one `queue` or `remove` operation running at a time.
+   */
+  private remove(name: string): Promise<void> {
+    return this.#operations.use(async () => {
+      const found = this.#lanes.get(name);
+      if (found && found.size === 0) {
```
stbrody, line 42: this `.size` check looks racy to me.  Is there anything that prevents an item from being added or removed from the queue after this check has occurred?

stbrody: I think this would work if the `#operations` mutex were held not just when create/fetching the queue but also while the task is being added to it.  That way if you see the size is 0 here you know nothing can get added until the mutex is released.

packages/core/src/state-management/repository.ts
```diff
@@ -0,0 +1,85 @@
+import { Document } from '../document';
+import DocID from '@ceramicnetwork/docid';
+import { AsyncLruMap } from './async-lru-map';
+import { StateStore } from '../store/state-store';
+import { DocumentFactory } from './document-factory';
+import { DocState } from '@ceramicnetwork/common';
+
+export class Repository {
```
stbrody, line 8: class doc-comment

packages/core/src/state-management/repository.ts
```diff
@@ -0,0 +1,85 @@
...
+      await entry.value.close();
+    });
+  }
+
+  // This should not be here, but here we are with the initialization process.
+  setDocumentFactory(documentFactory: DocumentFactory): void {
+    this.#documentFactory = documentFactory;
+  }
+
+  // This should not be here, but here we are with the initialization process.
```
stbrody, line 24:
```suggestion
  // Ideally this would be provided in the constructor, but circular dependencies in our initialization process make this necessary for now
```

packages/core/src/state-management/repository.ts
```diff
@@ -0,0 +1,85 @@
...
+  #documentFactory?: DocumentFactory;
+  #stateStore?: StateStore
+
+  constructor(limit: number) {
+    this.#map = new AsyncLruMap(limit, async (entry) => {
+      await entry.value.close();
+    });
+  }
+
+  // This should not be here, but here we are with the initialization process.
```
stbrody, line 19:
```suggestion
  // Ideally this would be provided in the constructor, but circular dependencies in our initialization process make this necessary for now
```

packages/core/src/state-management/repository.ts
```diff
@@ -0,0 +1,85 @@
...
+      return fromMemory.state;
+    } else {
+      if (this.#stateStore && this.#documentFactory) {
+        return this.#stateStore.load(docId);
+      }
+    }
+  }
+
+  /**
+   * Stub for adding the document.
```
stbrody, line 66:
```suggestion
   * Adds the document to the cache
```

packages/core/src/state-management/repository.ts
```diff
@@ -0,0 +1,85 @@
...
+   */
+  async add(document: Document): Promise<void> {
+    await this.#map.set(document.id.toString(), document);
+  }
+
+  /**
+   * Remove from memory. Stub.
```
stbrody, line 73:
```suggestion
   * Removes the document from the cache
```

packages/core/src/state-management/repository.ts
```diff
@@ -0,0 +1,85 @@
...
+      if (this.#stateStore && this.#documentFactory) {
+        return this.#stateStore.load(docId);
+      }
+    }
+  }
+
+  /**
+   * Stub for adding the document.
+   */
+  async add(document: Document): Promise<void> {
```
stbrody, line 68: Since this and `delete` below don't remove the document from the state store, but just from the in-memory cache, should they be called `addToCache` and `removeFromCache`?

packages/core/src/state-management/repository.ts
stbrody, line 8: This Repository is doing a lot less than I had imagined it would.  It seems like you've mostly turned the old `DocCache` into this new `Repository`, but they still do more or less the same thing.  I had expected the Repository to be an abstraction layer around the DocCache, StateStore, PinStore, and PinningBackends, but it seems like the latter 3 of those all still exist outside of the Repository, and the Repository has no knowledge of pinning.  Is the consolidation planned for a follow-up commit?  I'm worried about how many different but related objects we have that all have to do with managing document state in some form or fashion, and that there's a lot of cognitive overhead to developers to keep track of them all.

packages/core/src/state-management/repository.ts
```diff
@@ -0,0 +1,85 @@
...
+import { DocumentFactory } from './document-factory';
+import { DocState } from '@ceramicnetwork/common';
+
+export class Repository {
+  readonly #map: AsyncLruMap<Document>;
+  #documentFactory?: DocumentFactory;
+  #stateStore?: StateStore
+
+  constructor(limit: number) {
+    this.#map = new AsyncLruMap(limit, async (entry) => {
```
stbrody, line 14: I know you haven't gotten to this yet, but can you help me understand how it will eventually work that if a user has subscribed to updates on a Doctype instance that the corresponding Document/StateManager doesn't get evicted from the cache?

packages/core/src/store/__tests__/state-store-integration.test.ts
```diff
@@ -223,14 +223,14 @@ describe('Level data store', () => {
...
-    store = await storeFactory.createPinStore()
+    store = storeFactory.createPinStore()
   })
 
   it('pins document correctly without IPFS pinning', async () => {
     const genesis = await TileDoctype.makeGenesis({ content: initialContent, metadata: { controllers, tags: ['3id'] } }, context)
     const genesisCid = await dispatcher.storeCommit(genesis)
     const docId = new DocID('tile', genesisCid)
-    const doc = await Document.create(docId, doctypeHandler, dispatcher, store, context)
+    const doc = await Document.load(docId, doctypeHandler, dispatcher, store, context, DEFAULT_WRITE_DOCOPTS)
```
stbrody, line 233: maybe we should put back `Document.create()` but as just a super-simple one-line wrapper around `Document.load()` with a different default DocOpts?

packages/core/src/state-management/async-lru-map.ts
```diff
@@ -0,0 +1,197 @@
...
+
+  constructor(readonly key: string, public value: A) {}
+}
+
+const noop = async () => {
+  // Do Nothing
+};
+
+/**
+ * LRU Map with add/set/delete operations linearized per key.
```
stbrody, line 18: I'm finding this class complicated and quite difficult to decipher.

I'm also wondering if it's trying to do a lot more than it actually needs to.  For one, why do we need the linearization per key?  Isn't there only ever going to be one element per key anyway?  Can we remove all concurrency control from this class and do it at a higher level?

Also, the code here seems to be heavily based on [lru-map](https://github.com/rsms/js-lru/blob/master/lru.js).  Is there any reason we couldn't just use that package directly?

packages/core/src/state-management/named-p-queue.ts
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Set of named PQueues.
+ * When a task is done, it checks for pending tasks. No pending tasks means PQueue for the name is cleared.
+ */
+export class NamedPQueue {
+  #lanes: Map<string, PQueue>;
+  #operations: Mutex = new Mutex();
+
+  constructor(lanes: Map<string, PQueue> = new Map()) {
```
stbrody, line 12: is there ever a reason you wouldn't want to start with an empty map?  Can we remove this argument from the constructor?

stbrody (commented):
Lots of little comments throughout.  The biggest set of comments is related to the abstraction layers between Repository, DocCache, StateStore, PinStore, and PinningBackend, although it may make sense to get this in then do another commit to address that.

packages/core/src/__tests__/ceramic-anchor.test.ts
```diff
@@ -287,15 +279,23 @@ describe('Ceramic anchoring', () => {
...
-
-    await handle1
-    await handle2
+    await anchorService.anchor()
+
+    if (update1ShouldWin) {
+      // doctype1 anchors successfully, doctype2 fails to anchor
+      // doctype1 sends a tip, doctype2 updates state based on the tip
+      // so we wait till doctype2 handles the tip and changes anchorStatus to anchored
+      await TestUtils.waitForState(doctype2, 2000, state => state.anchorStatus === AnchorStatus.ANCHORED, () => {
```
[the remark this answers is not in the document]
ukstv: Fog of war.

packages/core/src/ceramic.ts
```diff
@@ -165,24 +166,25 @@ const tryDocId = (id: string): DocID | null => {
 class Ceramic implements CeramicApi {
 
   public readonly context: Context
-  public readonly dispatcher: Dispatcher
 
-  public pin: PinApi // Set during init()
-  public pinStore: PinStore // Set during init()
+  public readonly dispatcher: Dispatcher; // Set during init()
```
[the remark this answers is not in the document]
ukstv: changed

packages/core/src/ceramic.ts
```diff
@@ -165,24 +166,25 @@ const tryDocId = (id: string): DocID | null => {
...
+  public readonly pin: PinApi; // Set during init()
+  public readonly pinStore: PinStore; // Set during init()
```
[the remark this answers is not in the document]
ukstv: You want a bigger pull request and/or even longer time till one can run Ceramic in production? :)

Right from the PR description: "Initialisation is a mess".

packages/core/src/ceramic.ts
```diff
@@ -165,24 +166,25 @@ const tryDocId = (id: string): DocID | null => {
...
   private readonly _supportedChains: Array<string>
   private readonly _validateDocs: boolean
+  private readonly loadingQueue: LoadingQueue
 
   constructor (modules: CeramicModules, params: CeramicParameters) {
     this._ipfsTopology = modules.ipfsTopology
     this._logger = modules.loggerProvider.getDiagnosticsLogger()
-    this._pinStoreFactory = modules.pinStoreFactory
+    this.pinStore = modules.pinStoreFactory.createPinStore()
+    this.pin = new LocalPinApi(this.pinStore, this._loadDoc.bind(this), this._logger)
```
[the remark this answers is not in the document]
ukstv: Ditto.

packages/core/src/ceramic.ts
```diff
@@ -356,14 +359,16 @@ class Ceramic implements CeramicApi {
...
     const ipfsTopology = new IpfsTopology(ipfs, networkOptions.name, logger)
     const pinStoreFactory = new PinStoreFactory(ipfs, pinStoreOptions)
-    const repository = new Repository()
+    const repository = new Repository(docCacheLimit)
     const dispatcher = new Dispatcher(ipfs, networkOptions.pubsubTopic, repository, logger, pubsubLogger)
 
     const params = {
       cacheDocumentCommits: config.cacheDocCommits ?? true,
-      docCacheLimit: config.docCacheLimit ?? DEFAULT_DOC_CACHE_LIMIT,
+      docCacheLimit: docCacheLimit,
```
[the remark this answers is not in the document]
ukstv: It can, but it would be a breaking change, that I avoid till the very last moment, when it is inevitable.

packages/core/src/handlers-map.ts
```diff
@@ -0,0 +1,28 @@
+import { TileDoctypeHandler } from '@ceramicnetwork/doctype-tile-handler';
+import { Caip10LinkDoctypeHandler } from '@ceramicnetwork/doctype-caip10-link-handler';
+import { Doctype, DoctypeHandler } from '@ceramicnetwork/common';
+import { DiagnosticsLogger } from '@ceramicnetwork/logger';
+
+export class HandlersMap {
```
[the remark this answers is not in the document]
ukstv: ok

packages/core/src/handlers-map.ts
```diff
@@ -0,0 +1,28 @@
...
+  private readonly handlers: Map<string, DoctypeHandler<Doctype>>;
+
+  constructor(private readonly logger: DiagnosticsLogger, handlers?: Map<string, DoctypeHandler<Doctype>>) {
+    this.handlers =
+      handlers || new Map().set('tile', new TileDoctypeHandler()).set('caip10-link', new Caip10LinkDoctypeHandler());
+  }
+
+  get<A extends Doctype>(doctypeName: string): DoctypeHandler<A> {
```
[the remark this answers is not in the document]
ukstv: makes sense.

packages/core/src/state-management/named-p-queue.ts
ukstv: Testing is the reason. Either this or exposing lanes to the outside.

packages/core/src/state-management/loading-queue.ts
```diff
@@ -0,0 +1,54 @@
...
+      const found = await this.repository.get(docId);
+      if (found) {
+        this.logger.verbose(`Document ${docId.toString()} loaded from cache`);
+        return found;
+      } else {
+        // Load the current version of the document
+        const handler = this.handlers.get(docId.typeName);
+        opts = { ...DEFAULT_LOAD_DOCOPTS, ...opts };
+        const genesisCid = docId.cid;
+        const commit = await this.dispatcher.retrieveCommit(genesisCid);
```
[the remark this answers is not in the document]
ukstv: Indeed there is. `Document.load` gets ditched. Too much time pressure to remove it here, since it is used in a bunch of tests. And, really, it is used in tests only.

packages/core/src/state-management/document-factory.ts
```diff
@@ -0,0 +1,26 @@
...
+export class DocumentFactory {
+  constructor(
+    private readonly dispatcher: Dispatcher,
+    private readonly pinStore: PinStore,
+    private readonly context: Context,
+    private readonly validateDocs: boolean,
+    private readonly handlers: HandlersMap,
+  ) {}
+
+  async build(initialState: DocState) {
```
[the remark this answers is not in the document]
ukstv: The purpose of Context is to provide minimal necessary execution context (huh) to a doctype handler function and a doctype instance, regardless of what environment those run on: full Ceramic node or thin HTTP client. There is no Dispatcher or PinStore on HTTP client.

I would like to not use Context as a dependency injection mechanism internally.

packages/core/src/state-management/named-p-queue.ts
ukstv: This is puzzling. JS execution model is a single-threaded event-loop with events [processed until completion](https://developer.mozilla.org/en-US/docs/Web/JavaScript/EventLoop#run-to-completion). Nothing can stop the execution here after the check, and start executing something else.

I wrapped addition, but what the case you are describing here is a total impossibility.

packages/core/src/dispatcher.ts
ukstv: done, btw.

packages/core/src/state-management/async-lru-map.ts
stbrody: Sounds like this will go away in a future version, so not gonna worry about it right now

packages/core/src/state-management/named-p-queue.ts
stbrody: Yep, you are right of course.  Still getting used to the Node concurrency model.

stbrody (approved):
LGTM, thanks!

PR #1094, opened 2021-03-08
fix(blockchain-utils-linking): Pin @ethersproject/providers to previous version

stbrody:
Looks like our test depends on internal state from the `providers` package: https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/blockchain-utils-linking/src/__tests__/ethereum.test.ts#L89-L95.  That state seems to have been updated in a recent release from `@ethersproject`.  Ideally we would fix the way this test works, but in the interest of time, this is a quick fix that just pins us to the second-to-most recent version of the `providers` package

### DIFF
--- a/packages/blockchain-utils-linking/package.json
+++ b/packages/blockchain-utils-linking/package.json
@@ -27,6 +27,7 @@
   "dependencies": {
     "@stablelib/sha256": "^1.0.0",
     "caip": "^0.9.2",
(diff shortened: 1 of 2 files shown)


stbrody:
@oed @ukstv, feel free to merge this immediately if it looks acceptable to you

ukstv (approved):

PR #1117, opened 2021-03-12
feat(core): enable the use of timestamps

oed:
This PR does two things mainly:
### Puts timestamps in log entries for AnchorCommits
This is useful when you want to be able to know when certain updates happened and will be critical to determine if a signature done by a DID was done by a key that was valid at that point in time.

### Enables loading of documents using an `atTime` parameter
Both `loadDocument` and `multiQuery` now supports the `atTime` parameter. When specified Ceramic will load the AnchorCommit closest to that timestamp. If *atTime* is before any AnchorCommit the genesis commit will be loaded. Otherwise it will load the AnchorCommit which is has a timestamp smaller than, or equal to *atTime*


packages/common/src/doctype.ts
```diff
@@ -123,14 +126,25 @@ export interface DocState {
  * behaviors that are performed as part of the operation.
  */
 export interface DocOpts {
-    // Whether or not to request an anchor after performing the operation.
+    /**
+     * Whether or not to request an anchor after performing the operation.
+     */
```
oed, line 131: This should put the comment into the generated docs!

packages/core/src/ceramic.ts
```diff
@@ -630,11 +631,14 @@ class Ceramic implements CeramicApi {
...
 
     // If DocID is requested, return the document
-    if (docRef instanceof DocID) {
-      return doc
-    } else {
+    if (docRef instanceof CommitID) {
       // Here CommitID is requested, let's return document at specific commit
       return doc.rewind(docRef)
+    } else if (opts.atTime) {
+      const commitId = doc.findCommitAt(opts.atTime)
```
ukstv, line 638: This, with `findCommitAt` do not belong here IMO, but let it live for now. Going to move it out of the document after the merge.

oed: Ok, let me know where I should move it? Or do you want to later?

ukstv (approved):

packages/core/src/ceramic.ts
ukstv: We'll do it later. It's all about having snapshots, which seems to be pretty targeted change. For now let it live here.

packages/core/src/__tests__/ceramic-api.test.ts
```diff
@@ -431,14 +433,26 @@ describe('Ceramic API', () => {
                    notDoc: '123' },
         metadata: { controllers: [controller] }
       })
+
+      // test data for the atTime feature
+      docFStates.push(docF.state)
+      docFTimestamps.push(Date.now())
```
stbrody, line 439: is this adding a timestamp to an update that was never anchored?

packages/core/src/document.ts
```diff
@@ -271,6 +272,27 @@ export class Document implements DocStateHolder {
     this.subscriptionSet.add(subscription);
   }
 
+  /**
+   * Find the relevant AnchorCommit given a particular timestamp.
+   * Will return an AnchorCommit whos timestamp is earlier to or
```
stbrody, line 277:
```suggestion
   * Will return an AnchorCommit whose timestamp is earlier to or
```

stbrody (approved):

packages/core/src/__tests__/ceramic-api.test.ts
oed: No this is just a timestamp for the points of times before and in between the actual anchor commits.

stbrody: Now I'm more confused.  I thought only AnchorCommits should have timestamps

oed: They do, but here I'm just capturing the time before and after each anchor commit timestamp.

PR #1118, opened 2021-03-12
feat: Introduce Running State

ukstv:
Actually, does not do much. Just a minuscule necessary step for turning Document inside out.


stbrody (approved):
LGTM!  I love having a nice small easy to understand review like this.  The fact that there's no behavior change is actually even better since it allows me as a reviewer to just focus on the refactoring piece without having to think about the user-experience/behavior piece

PR #1130, opened 2021-03-15
Preliminary bunch of changes before concurrent per-doc task execution

ukstv:
- Rename NamedPQueue to NamedTaskQueue
- Introduce ContextfulHandler - handler that does not require context on calls
- Document no longer uses context


oed (commented):
Had a quick glance over this. Makes sense overall. Will leave final confirmation to Spencer.

packages/core/src/__tests__/ceramic.test.ts
```diff
@@ -104,7 +100,7 @@ describe('Ceramic integration', () => {
     const doctype1 = await ceramic1.createDocument(DOCTYPE_TILE, { content: { test: 123 } }, { anchor: false, publish: false })
     const doctype2 = await ceramic2.loadDocument(doctype1.id)
     expect(doctype1.content).toEqual(doctype2.content)
-    expectEqualStates(doctype1.state, doctype2.state)
+    expect(DoctypeUtils.statesEqual(doctype1.state, doctype2.state));
```
stbrody, line 103: This does basically the same thing as what was there previously, but loses the ability to see the diff between expected and received if the check fails.  Would prefer to leave it as it was

packages/core/src/__tests__/create-ceramic.ts
```diff
@@ -0,0 +1,23 @@
+import Ceramic, { CeramicConfig } from '../ceramic';
+import { IpfsApi } from '@ceramicnetwork/common';
+import * as uint8arrays from 'uint8arrays';
+import * as sha256 from '@stablelib/sha256';
+import tmp from 'tmp-promise';
+import { Ed25519Provider } from 'key-did-provider-ed25519';
+
+export async function createCeramic(ipfs: IpfsApi, config?: CeramicConfig & { seed?: string }): Promise<Ceramic> {
+  const appliedConfig = {
+    stateStoreDirectory: await tmp.tmpName(),
```
stbrody, line 10: I think this should be `tmp.dir`?

packages/core/src/__tests__/dispatcher.test.ts
```diff
@@ -1,107 +1,127 @@
...
+const FAKE_CID2 = new CID('bafybeig6xv5nwphfmvcnektpnojts44jqcuam7bmye2pb54adnrtccjlsu');
+const FAKE_DOC_ID = DocID.fromString('kjzl6cwe1jw147dvq16zluojmraqvwdmbh61dx9e0c59i344lcrsgqfohexp60s');
 
 const ipfs = {
   pubsub: {
     ls: jest.fn(async () => Promise.resolve([])),
     subscribe: jest.fn(),
     unsubscribe: jest.fn(async () => Promise.resolve()),
-    publish: jest.fn()
+    publish: jest.fn(),
```
stbrody, line 36: I assume you have an editor that is auto-performing these small formatting changes?  Don't worry about it for this review, but for the future I would appreciate it if you could turn that off for now, as all these little changes create a lot of noise in the code reviews that I then have to sift through.  At some point we should enable an auto-formatter and do a single pass over the entire codebase to enforce a single style.  But until then, having these little style changes mixed in with actual functional changes is very distracting and slows down the review process

packages/core/src/pubsub/task-queue.ts
```diff
@@ -33,6 +40,13 @@ export class TaskQueue {
     });
   }
 
+  /**
+   * Add task and wait till it is completed.
+   */
+  run<T>(task: Task<T>): Promise<T> {
+    return this.#pq.add(task)
```
stbrody, line 47: Do we need any retry logic slash error handling here like `add()` above has?

stbrody (approved):
LGTM mod 2 small comments.

I do wonder though if using `Context` for dependency injection is so bad that it's worth all this effort to remove it from as many places as possible.

packages/core/src/__tests__/ceramic.test.ts
ukstv: okay. changed back in su/state-refactor-tasks-execution

packages/core/src/__tests__/dispatcher.test.ts
ukstv: makes sense. thanks.

packages/core/src/pubsub/task-queue.ts
ukstv: The point of `run` (as opposed to `add`) is to pass an error to the caller if it is throw inside a task. Note "fire-and-forget" comment for the `doc` method.

So, no special error handling here.

Added that to the doc-comment.

packages/core/src/__tests__/create-ceramic.ts
ukstv: `tmpName` only generates a temporary name. Ceramic then in `LevelStateStore` makes a directory. IMO, `tmpName` is more aligned with real-world usage, when a user passes some name (may be non-existent yet), and the code creates whatever is needed at that path.

This is driven by my big disappointment with some CLI utilities that require you to create a folder/file with _correct_ (god knows what it means for every different command) permissions and structure before the first run. One tries to use the command for the first time, has no knowledge yet on how the command works, and then it spits out to you a (usually) cryptic message like "Incorrect path". Go figure what it meant, and what the command expects as an input.

packages/core/src/__tests__/ceramic.test.ts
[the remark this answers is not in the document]
ukstv: changed back :)

ukstv:
@stbrody "if using Context for dependency injection is so bad"

Come on, single responsibility principle have some merit.

Context is an important piece of architecture as it is passed to doctypes and handlers, and _exposed to the consumer_. Right now it is a bucket of nullable values. Every nullable value with an unclear lifecycle means possible surprise down the road. Can I expect anchorService to be available on http client side? How about DID? When exactly is it available? When can I use DID resolver exposed from Context? I can not reason about it without looking at _a lot_ of adjacent code.

Every time I see Context now, I can not deduce what is expected here, what is actually the minimal required functionality here. It is then too hard for me to reason about code, about its properties. Changes are hard then, and unreliable.

packages/core/src/__tests__/create-ceramic.ts
stbrody: Hmm... so if `stateStoreDirectory` is just a string like "foo" and not a path, then where will the directory be created [here](https://github.com/ceramicnetwork/js-ceramic/blob/56912f56b74f0b97fdb95df972d6182389b2a7e4/packages/core/src/store/level-state-store.ts#L29-L30) (`storeRoot` will be the `stateStoreDirectory`)?  Won't it be a relative path to the CWD?  Or will it be from the root dir (ie `/foo`)?  Either way feels wrong to me?  I assume `tmp.dir` generates something in `/tmp`, which seems like the right place

packages/core/src/pubsub/task-queue.ts
stbrody: Makes sense, thanks for clarifying!

PR #1136, opened 2021-03-16
fix(core): Fix test by waiting long enough for new anchor timestamp

stbrody:


packages/core/src/__tests__/document.test.ts
```diff
@@ -433,9 +433,8 @@ describe('Document', () => {
       // create invalid change that happened after main change
       const doc2 = await Document.load(docId, doctypeHandler, dispatcher, pinStore, context, { sync: false })
       await doc2._handleTip(tipPreUpdate)
-      // add short wait to get different anchor time
-      // sometime the tests are very fast
-      await new Promise(resolve => setTimeout(resolve, 1))
+      // add one second wait to get different anchor time
+      await new Promise(resolve => setTimeout(resolve, 1000))
```
stbrody, line 437: now that timestamps from the InMemoryAnchorService have second-granularity, this test needs to wait at least a second to ensure that this anchor gets a different timestamp than the earlier anchor in the test

oed: Ah, makes sense!

oed (approved):

packages/core/src/__tests__/document.test.ts
oed: Thanks for addressing my blunder :)

PR #1138, opened 2021-03-16
fix(http-client): Don't use the logger in the http-client

stbrody:


oed (approved):
Ok, so we just completely remove logging from the http-client for now?

stbrody:
> Ok, so we just completely remove logging from the http-client for now?

Yep, that's right.  I made a note in https://www.notion.so/threebox/Enable-core-package-to-run-in-browser-e4d779ccf34b4086bf953de6df10526f to put this logging back

PR #1140, opened 2021-03-17
fix: use npm package for ipfs-daemon in docker

valmack:
### Motivation

We want to use @ceramicnetwork/ipfs-daemon instead of js-ipfs-ceramic on our infra nodes so this updates the Dockerfile that will be used to run the former.

### Changes

- Updates Dockerfile to pull ipfs-daemon from npm and run the binary


ukstv (approved):
🏇

PR #1141, opened 2021-03-18
State Refactor Complete

ukstv:
Proper concurrency model in place.


packages/core/src/state-management/execution-queue.ts
```diff
@@ -0,0 +1,60 @@
...
+import { Task, TaskQueueLike } from '../pubsub/task-queue';
+import { DiagnosticsLogger } from '@ceramicnetwork/logger';
+import { RunningState } from './running-state';
+
+export interface ExecLike extends TaskQueueLike {
+  addE: (task: (state$: RunningState) => Promise<void>) => void;
+  runE<T>(task: (state$: RunningState) => Promise<T>): Promise<T>;
+}
+
+export class ExecutionQueue {
```
stbrody, line 12: Can you add some comments on the relationship and differences between TaskQueue and ExecutionQueue?

packages/core/src/state-management/execution-queue.ts
```diff
@@ -0,0 +1,60 @@
+import { NamedTaskQueue } from './named-task-queue';
+import { DocID } from '@ceramicnetwork/docid';
+import { Task, TaskQueueLike } from '../pubsub/task-queue';
+import { DiagnosticsLogger } from '@ceramicnetwork/logger';
+import { RunningState } from './running-state';
+
+export interface ExecLike extends TaskQueueLike {
+  addE: (task: (state$: RunningState) => Promise<void>) => void;
```
stbrody, line 8: If you have an ExecutionQueue, when would you call `.addE` or `runE` vs `.add` or `.run`?  Some comments here would help

packages/core/src/store/level-state-store.ts
```diff
@@ -36,7 +36,7 @@ export class LevelStateStore implements StateStore {
      * @param docStateHolder - Document instance
      */
     async save(docStateHolder: DocStateHolder): Promise<void> {
-        await this.#store.put(docStateHolder.id.baseID.toString(), DoctypeUtils.serializeState(docStateHolder.state))
+        await this.#store.put(docStateHolder.id.toString(), DoctypeUtils.serializeState(docStateHolder.state))
```
stbrody, line 39: why this change?  Don't we always want to store by baseID?  I'm not sure it makes sense to cache state at specific commitIds in the state store.

Maybe we should just throw if the doc is pinned to a specific CommitID when trying to store it into the state store?

packages/core/src/state-management/repository.ts
```diff
@@ -1,70 +1,110 @@
...
+  anchorService: AnchorService,
+  conflictResolution: ConflictResolution,
+  stateValidation: StateValidation
+}
 
 export class Repository {
-  readonly loadingQ: NamedTaskQueue = new NamedTaskQueue();
-  readonly #map: AsyncLruMap<Document>;
-  #documentFactory?: DocumentFactory;
+  readonly loadingQ: NamedTaskQueue;
```
stbrody, line 27: I think it'd be helpful if all of these members of Repository had brief comments on what they represent

packages/core/src/state-management/repository.ts
```diff
@@ -98,62 +138,44 @@ export class Repository {
...
     } else {
-      if (this.pinStore) {
-        return this.pinStore.stateStore.load(docId);
-      }
+      return this.#deps.pinStore.stateStore.load(docId);
     }
   }
 
   /**
    * Stub for adding the document.
```
stbrody, line 150:
```suggestion
   * Adds the document to the in-memory cache
```

packages/core/src/state-management/snapshot-state.ts
```diff
@@ -0,0 +1,17 @@
+import { Observable, of } from 'rxjs';
+import { RunningStateLike } from './running-state';
+import { DocState } from '@ceramicnetwork/common';
+import { DocID } from '@ceramicnetwork/docid';
+
+export class SnapshotState extends Observable<DocState> implements RunningStateLike {
```
stbrody, line 6: class comment please

packages/core/src/state-management/state-manager.ts
```diff
@@ -0,0 +1,191 @@
...
+   * the most recent Tip to it.
+   * @param state$
+   * @param opts
+   */
+  syncGenesis(state$: RunningState, opts: DocOpts): Promise<void> {
+    return this.applyOpts(state$, opts);
+  }
+
+  /**
+   * Takes the most recent known-about version of a document and a specific commit and returns a new
```
stbrody, line 38: Comment is out of date now

stbrody (changes requested):
This looks really great!
Most of my suggestions are around adding more comments and clarifying the names of various components and their relationships to other components.  But the core of the change looks great, and I think overall the code looks much cleaner now.  Thanks for all your hard work here!

stbrody:
One other question: As best as I can tell this hasn't changed the behavior around how and when new states become visible to doctype instances.  I assume there will be another PR coming that implements the changes that we discussed in https://www.notion.so/threebox/User-visible-behavior-on-document-updates-382c1c8e24b040f2a54b78b06a2466f9?

packages/core/src/ceramic.ts
```diff
@@ -629,22 +645,25 @@ class Ceramic implements CeramicApi {
...
-      const commitId = doc.findCommitAt(opts.atTime)
-      return doc.rewind(commitId)
+      return this.repository.stateManager.atTime(state$, opts.atTime)
     } else {
-      return doc
+      return state$
     }
   }
 
+  async __loadDoc(docId: DocID, opts: DocOpts = {}): Promise<RunningState> {
```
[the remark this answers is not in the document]
oed: Also don't get why both are needed 🤔

packages/core/src/conflict-resolution.ts
```diff
@@ -205,18 +207,33 @@ export async function fetchLog(
   return fetchLog(dispatcher, prevCid, stateLog, log);
 }
 
+export function commitAtTime(state$: RunningStateLike, timestamp: number): CommitID {
```
oed, line 210: 👍

packages/core/src/store/level-state-store.ts
oed: Storing states for specific CommitIDs in the state store might make sense when you are loading a commit of a document multiple times. This happens in 3ID and for schemas for example.

oed (commented):
Looks good :)

packages/core/src/store/level-state-store.ts
stbrody: But you could just store the current state of the doc in the state store, load it, then reset to the old version

oed: That's going to be quite slow if you are loading it multiple times.

oed: Ideally that's the case yes. I would guess that's not how it's implemented right now however.

packages/core/src/__tests__/document.test.ts
```diff
@@ -1,29 +1,21 @@
-import { DocID } from '@ceramicnetwork/docid';
-import { AnchorStatus, CommitType, DocState, DoctypeUtils, IpfsApi, SignatureStatus } from '@ceramicnetwork/common';
+import {
```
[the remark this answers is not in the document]
ukstv: True.

packages/core/src/ceramic.ts
```diff
@@ -209,12 +208,19 @@ class Ceramic implements CeramicApi {
 
     this._doctypeHandlers = new HandlersMap(this._logger)
 
+    // This initialization block below has to be redone.
```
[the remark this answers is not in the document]
ukstv: Changed.

packages/core/src/ceramic.ts
```diff
@@ -209,12 +208,19 @@ class Ceramic implements CeramicApi {
...
     this.stateValidation = this._validateDocs ? new RealStateValidation(this.loadDocument.bind(this)) : new FauxStateValidation()
-    const documentFactory = new DocumentFactory(this.dispatcher, pinStore, this.context, this._doctypeHandlers, this.stateValidation)
-    const networkLoad = new NetworkLoad(this.dispatcher, this._doctypeHandlers, this.context, this._logger, documentFactory)
-    this.repository.setPinStore(pinStore)
-    this.repository.setDocumentFactory(documentFactory);
-    this.repository.setNetworkLoad(networkLoad);
+    const conflictResolution = new ConflictResolution(this.context.anchorService, this.stateValidation, this.dispatcher, this.context, this._doctypeHandlers)
+    const pinStore = modules.pinStoreFactory.createPinStore()
+    this.repository.setDeps({
```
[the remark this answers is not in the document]
ukstv: My position is Context is not for DI. For example, AnchorService should not be a part of Context IMO.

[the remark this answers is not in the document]
ukstv: Discussion about Context: worth it.

packages/core/src/ceramic.ts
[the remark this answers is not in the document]
ukstv: Forgotten remnant of dev time. Only one is required.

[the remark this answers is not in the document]
ukstv: Changed.

packages/core/src/ceramic.ts
```diff
@@ -629,22 +645,25 @@ class Ceramic implements CeramicApi {
    * @param docId - Document ID
    * @param opts - Initialization options
    */
-  async _loadDoc(docId: DocID | CommitID | string, opts: DocOpts = {}): Promise<Document> {
+  async _loadDoc(docId: DocID | CommitID | string, opts: DocOpts = {}): Promise<RunningStateLike> {
```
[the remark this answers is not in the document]
ukstv: Renamed `__loadDoc` as `_loadDoc` with old `_loadDoc` gone. Hope now it is clearer.

packages/core/src/ceramic.ts
```diff
@@ -656,16 +675,17 @@ class Ceramic implements CeramicApi {
...
-      const toRecover = document.state?.anchorStatus === AnchorStatus.PENDING || document.state?.anchorStatus === AnchorStatus.PROCESSING
-      if (toRecover) {
-        document.anchor()
-      }
+  restoreDocuments() {
+    this.repository.listPinned().then(async list => {
+      let n = 0
+      await Promise.all(list.map(async docId => {
+        await this.__loadDoc(DocID.fromString(docId))
+        n++;
```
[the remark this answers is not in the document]
ukstv: ¯\_(ツ)_/¯

packages/core/src/conflict-resolution.ts
```diff
@@ -205,18 +207,33 @@ export async function fetchLog(
...
-    private readonly handler: ContextfulHandler<Doctype>,
+    private readonly context: Context,
+    private readonly handlers: HandlersMap
   ) {}
 
   /**
    * Applies the log to the document and updates the state.
    */
-  async applyLogToState(log: Array<CID>, state?: DocState, breakOnAnchor?: boolean): Promise<DocState> {
+  private async applyLogToState(handler: ContextfulHandler, log: Array<CID>, state?: DocState, breakOnAnchor?: boolean): Promise<DocState> {
```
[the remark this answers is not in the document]
ukstv: Uhums.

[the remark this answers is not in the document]
ukstv: Too lazy to add ContextfulHandlersMap instead :)

packages/core/src/state-management/execution-queue.ts
ukstv: Replaced `addE`/`runE` with just `add`/`run`. Sans-E functions are not used anyway.

packages/core/src/store/level-state-store.ts
ukstv: Surprised about this. `DocStateHolder#id` is defined as `DocID` type. `DocID#baseID` is just redundant here.

packages/core/src/state-management/execution-queue.ts
ukstv: Added comments to the ExecutionQueue.

packages/core/src/state-management/repository.ts
ukstv: Ok.

packages/core/src/state-management/snapshot-state.ts
ukstv: Okay.

packages/core/src/state-management/state-manager.ts
ukstv: Changed.

ukstv:
"One other question: As best as I can tell this hasn't changed the behavior around how and when new states become visible to doctype instances"

Even more so, this does not change how doctype relates to state. I extracted the functionality to https://github.com/ceramicnetwork/js-ceramic/pull/1141/files#diff-d32156236fb949a52abdd6b3238467d9ee30f43a96eb746de316dae81304c995R5 to track this easier. It is the next step to change Doctype.

packages/core/src/store/level-state-store.ts
stbrody: > Ideally that's the case yes. I would guess that's not how it's implemented right now however.

Shouldn't the plan then be to move in the direction towards optimizing the implementation rather than attempting to optimize by storing more data? @oed

stbrody: > Surprised about this. `DocStateHolder#id` is defined as `DocID` type. `DocID#baseID` is just redundant here.

Ah right, sorry, my mistake.

How do we ensure that we aren't putting the state from a "readonly" document at was loaded at a specific CommitID into the state store as if it was the most recent version of that document?  Can we add an assert somewhere into this process that would catch that type of programming or user error?  It would be a pretty nasty bug if we ever accidentally overrode the state store with less up-to-date info for a given DocID than what was already there

packages/core/src/ceramic.ts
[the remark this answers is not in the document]
stbrody: Yep.  Your comment on the other PR about how using Context reduces visibility into which parts of the Context are actually needed by any given piece of code is a good point.  But then if Context isn't for dependency injection it raises the question: what *is* Context for?  Should we be moving towards eliminating Context entirely?  If so, should we be replacing it with some other type of DI system?  Or should we just get used to passing around a lot of references everywhere to what are basically process-wide singleton objects?

In any case, this seems like the topic for a separate more dedicated conversation about our approach to dependency injection in this codebase.  Which I think can safely wait until after mainnet is out the door

packages/core/src/state-management/execution-queue.ts
```diff
@@ -1,14 +1,31 @@
 import { NamedTaskQueue } from './named-task-queue';
```
stbrody, line 1: Nice cleanup of this file over the last version, I find it much easier to make sense of now!

stbrody (commented):
> It is the next step to change Doctype.

Perfect!

stbrody (approved):
Looks great!  Thanks for all the nice cleanups!

packages/core/src/state-management/execution-queue.ts
ukstv: ^_^

packages/core/src/store/level-state-store.ts
ukstv: @stbrody I believe, this check could be made on type level. Right now all of RunningState, RunningStateLike, SnapshotState along with Doctype implement DocStateHolder interface. Maybe at the next iteration, we think on proper typing to avoid this kind of accident.

PR #1147, opened 2021-03-22
fix(js-ceramic): Fix deploy-docs github workflow

stbrody:
Is there any way to test this?


oed (approved):
You can test it by changing 
```
on:
  push:
    branches: [ master ]
 ```
 To the name of this branch, then change back before you merge :)

PR #1148, opened 2021-03-22
chore(core): Move 'unique' field into GenesisHeader

stbrody:
With this change now if a deterministic document is created then the `unique` field simply doesn't show up in the doc state at all (unlike today where it's there with the value `"0"`).  When a non-deterministic document is made, the unique field and its random string value are now a part of the document metadata (and thus will now show up on the cli when running `ceramic state`, which isn't currently the case)

### DIFF
--- a/packages/3id-did-resolver/src/__tests__/index.test.ts
+++ b/packages/3id-did-resolver/src/__tests__/index.test.ts
@@ -6,11 +6,11 @@ jest.mock('cross-fetch', () => {
       json: async () => {
         const call = mockedCalls[url]
         if (opts.method === 'post') {
-          const { response } = call.find(({ expectedBody }) => opts.body === JSON.stringify(expectedBody))
-          if (response) {
-            return response
+          const result = call.find(({ expectedBody }) => opts.body === JSON.stringify(expectedBody))
+          if (result?.response) {
+            return result.response
           } else {
-            throw new Error('invalid http request for mock')
+            throw new Error('Could not find response for http body: ' + opts.body)
           }
         }
         return call.response
--- a/packages/3id-did-resolver/src/__tests__/vectors.json
+++ b/packages/3id-did-resolver/src/__tests__/vectors.json
@@ -159,13 +159,13 @@
       }
     },
     "http://localhost:7007/api/v0/documents": [{
-      "expectedBody": {"doctype":"tile","genesis":{"header":{"controllers":["did:key:zQ3shcMwX6vEZkbCWrZqL5W47yditTzauxC5x4WNbgo5r2PbW"],"family":"3id"},"unique":"0"},"docOpts":{"anchor":false,"publish":false}},
+      "expectedBody": {"doctype":"tile","genesis":{"header":{"controllers":["did:key:zQ3shcMwX6vEZkbCWrZqL5W47yditTzauxC5x4WNbgo5r2PbW"],"family":"3id"}},"docOpts":{"anchor":false,"publish":false}},
       "response": {
         "docId":"k2t6wyfsu4pfzxkvkqs4sxhgk2vy60icvko3jngl56qzmdewud4lscf5p93wna",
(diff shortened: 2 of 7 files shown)


stbrody:
@oed do you know what's going on with the 3id test?  I assume the issue is that the vectors need to be regenerated since the JWS's will be different, although the way its failing looks different from what I would have expected.

If the issue is indeed just that the vectors need to be regenerated, is there an easy way to do that off of my working branch without needing to do a release first?

oed:
> I assume the issue is that the vectors need to be regenerated since the JWS's will be different, although the way its failing looks different from what I would have expected.

Ehm, this is **not** ok. We can't break the http api. Looks like it's an issue with how the deterministic flag is passed to the body of http requests. Can you make this change such that the http api works with the old way of passing the deterministic flag?

oed:
I fixed the issue by modifying the http request response mock.

Question: What will happen if someone now makes a http request with the `unique` property? Will it result in the same Ceramic document or a different one?

oed (changes requested):
Ok, this is a problem. These two requests need to generate the same docid, otherwise things will break.

**Request 1, no unique**
```sh
curl https://ceramic-clay.3boxlabs.com/api/v0/documents -X POST -d '{
   "doctype": "tile",
   "genesis": {
     "header": {
       "family": "test",
       "controllers": ["did:key:z6MkfZ6S4NVVTEuts8o5xFzRMR8eC6Y1bngoBQNnXiCvhH8H"]
     }
   },
   "docOpts": { "publish": false, "anchor": false }
 }' -H "Content-Type: application/json"

{"docId":"k2t6wyfsu4pg2qvoorchoj23e8hf3eiis4w7bucllxkmlk91sjgluuag5syphl","state":{"doctype":"tile","content":{},"metadata":{"family":"test","controllers":["did:key:z6MkfZ6S4NVVTEuts8o5xFzRMR8eC6Y1bngoBQNnXiCvhH8H"]},"signature":0,"anchorStatus":"NOT_REQUESTED","log":[{"cid":"bafyreihtdxfb6cpcvomm2c2elm3re2onqaix6frq4nbg45eaqszh5mifre","type":0}]}}
```

**Request 1, unique set to 0**
```sh
curl https://ceramic-clay.3boxlabs.com/api/v0/documents -X POST -d '{
   "doctype": "tile",
   "genesis": {
     "header": {
       "family": "test",
       "controllers": ["did:key:z6MkfZ6S4NVVTEuts8o5xFzRMR8eC6Y1bngoBQNnXiCvhH8H"]
     }, "unique": "0"
   },
   "docOpts": { "publish": false, "anchor": false }
 }' -H "Content-Type: application/json"
 
{"docId":"k2t6wyfsu4pg1788blcx8r48j01eaafwrq349ikskig7qlaq1ql26wxxnnivme","state":{"doctype":"tile","content":{},"metadata":{"family":"test","controllers":["did:key:z6MkfZ6S4NVVTEuts8o5xFzRMR8eC6Y1bngoBQNnXiCvhH8H"]},"signature":0,"anchorStatus":"NOT_REQUESTED","log":[{"cid":"bafyreifvc62lw3c455aycxg6tw3uqepzxsbpw6sqxxuhfu7n7jduhxxtsy","type":0}]}}
```

stbrody:
I don't understand why those two documents would be expected to have the same DocID, they're different?  A `unique` field with the value of `"0"` is still a unique field, same as if the string was a random sequence of characters

oed:
> I don't understand why those two documents would be expected to have the same DocID, they're different? A unique field with the value of "0" is still a unique field, same as if the string was a random sequence of characters

That's the problem. Previous versions of ceramic generating a deterministic document would add this field when using the http client. Now if a dev upgrades to a version that uses this PR they would get a different docid. This is not acceptable because it breaks the expectation, not just the structure of the api.

stbrody:
I guess this goes back to your same comment from https://github.com/ceramicnetwork/js-ceramic/issues/1149.  Like that issue, this is also a breaking change to document state.  So it comes down to whether we're willing to break existing documents on Clay.  To me, one of the main advantages of having a period of time where we're on testnet but not on mainnet is so that we can make breaking changes as we discover them without having to be burdened by backwards compatibility concerns.  The moment mainnet launches every protocol-level change becomes **much** harder, I don't know why we would choose to tie our own hands any earlier than we have to.  Now's our last chance to make final breaking protocol changes before mainnet goes out and we have to do the elaborate backward compatibility dance for every little change and thus accumulate complexity and tech debt until the end of time.  No one should have data they can't afford to lose on the testnet, that's literally the point of the testnet.

oed (approved):

PR #1150, opened 2021-03-23
feat(core, http-client, common): Doctype accepts Running State

ukstv:
WARNING: BREAKING CHANGE

- Doctype is no longer an instance of EventEmitter.
- Updates to state are only available after subscription or `sync`.


packages/core/src/dispatcher.ts
```diff
@@ -126,13 +126,10 @@ export class Dispatcher {
     // TODO Add validation the message adheres to the proper format.
 
     const { doc: docId, tip } = message
-    const document = await this.repository.get(docId)
-    if (document) {
```
stbrody, line 130: It took me a while of digging to figure out why removing this doesn't cause us to now do the work of processing incoming tips for every update message we receive even if the doc isn't in the repository.  I believe the answer is because of how we use the ExecutionQueue within the StateManager and how the ExecutionQueue abandons its task if its `get()` callback returns null, but that's pretty subtle.  I'm honestly not sure the best way to make this more clear, maybe a combination of comments within the ExecutionQueue and the StateManager?

packages/doctype-caip10-link/src/caip10-link-doctype.ts
```diff
@@ -67,7 +67,7 @@ export class Caip10LinkDoctype extends Doctype {
             throw new Error('Exactly one controller must be specified')
         }
 
-        const [address, linkedChainId] = metadata.controllers[0].split('@') // eslint-disable-line @typescript-eslint/no-unused-vars
+        const [, linkedChainId] = metadata.controllers[0].split('@') // eslint-disable-line @typescript-eslint/no-unused-vars
```
stbrody, line 70: Can we remove the eslint suppression comment now?

packages/http-client/src/ceramic-http-client.ts
```diff
@@ -73,11 +66,12 @@ export default class CeramicClient implements CeramicApi {
   private readonly _config: CeramicClientConfig
   public readonly _doctypeConstructors: Record<string, DoctypeConstructor<Doctype>>
 
-  constructor (apiHost: string = CERAMIC_HOST, config: CeramicClientConfig = {}) {
+  constructor (apiHost: string = CERAMIC_HOST, config: Partial<CeramicClientConfig> = {}) {
     this._config = { ...DEFAULT_CLIENT_CONFIG, ...config }
 
     this._apiUrl = combineURLs(apiHost, API_PATH)
-    this._docCache = new DocCache(config.docCacheLimit, this._config.cacheDocCommits)
+    // this._docCache = new LRUMap(config.docCacheLimit) Not now. We do not know what to do when document is evicted on HTTP client.
```
stbrody, line 73: Do we even still need this `_docCache` in the http client?  What is it gaining us?

EDIT: Nevermind, feel free to ignore this for now, we can revisit this in the cleanup http client project

stbrody (approved):
Looks really good! Biggest comment is around more exhaustive testing to ensure that the user-visible behavior around when updates are made visible works as expected, and isn't affected based on whether or not the document happens to be in cache

stbrody (changes requested):
Whoops, clicked approve by accident

packages/core/src/dispatcher.ts
```diff
@@ -126,13 +126,10 @@ export class Dispatcher {
...
-      // TODO: add cache of cids here so that we don't emit event
-      // multiple times if we get the message more than once.
-      this.repository.stateManager.update(document, tip)
-      // TODO: Handle 'anchorService' if present in message
-    }
+    // TODO: add cache of cids here so that we don't emit event
+    // multiple times if we get the message more than once.
```
oed, line 130:
```suggestion
    // TODO: add cache of cids here so that we don't emit event
    // multiple times if we get the message from more than one peer.
```

packages/http-client/src/document.ts
```diff
@@ -1,63 +1,52 @@
...
+import { CeramicCommit, DocOpts, DocState, DoctypeUtils, RunningStateLike } from '@ceramicnetwork/common';
+import { DocID, CommitID } from '@ceramicnetwork/docid';
+import { fetchJson } from './utils'
+
+export class Document extends Observable<DocState> implements RunningStateLike {
+  private readonly state$: BehaviorSubject<DocState>;
+
+  constructor (initial: DocState, private _apiUrl: string, docSyncInterval: number) {
+    super(subscriber => {
+      const periodicUpdates = timer(0, docSyncInterval).pipe(throttle(() => this._syncState())).subscribe();
```
oed, line 12: Is it still possible to disable syncing here like before?

oed: Maybe `0` means disabled?

packages/http-client/src/ceramic-http-client.ts
```diff
@@ -46,9 +40,8 @@ export const DEFAULT_CLIENT_CONFIG: CeramicClientConfig = {
  */
 export interface CeramicClientConfig {
```
oed, line 41: Can you add tsdoc to each of the config params here?

oed (commented):
A few small comments.

packages/common/src/doctype.ts
```diff
@@ -200,34 +203,31 @@ export abstract class Doctype extends EventEmitter implements DocStateHolder {
...
+        return cloneDeep(this.state$.value)
     }
 
     get context(): Context {
         return this._context
     }
 
+    async sync(): Promise<void> {
+      const document = await this._context.api.loadDocument(this.id)
+      this.state$.next(document.state)
```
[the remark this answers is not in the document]
ukstv: It will not.

packages/common/src/utils/test-utils.ts
```diff
@@ -27,22 +38,14 @@ export class TestUtils {
...
-        const completionPromise = new Promise<void>(resolve => {
-            const handler = () => {
-                if (predicate(doc.state)) {
-                    doc.off('change', handler)
-                    resolve()
-                }
-            }
-            doc.on('change', handler)
-
-            // Call handler() manually once to see if the predicate was already true
```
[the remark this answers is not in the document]
ukstv: OK.

packages/core/src/__tests__/state-manager.test.ts
```diff
@@ -318,19 +319,25 @@ test('enforce previously assigned schema during future update', async () => {
...
 });
 
 test('should announce change to network', async () => {
   const publishTip = jest.spyOn(ceramic.dispatcher, 'publishTip');
   const doctype1 = await ceramic.createDocument('tile', { content: INITIAL_CONTENT, metadata: { controllers } });
+  doctype1.subscribe();
   const doc1 = await ceramic.repository.load(doctype1.id);
   expect(publishTip).toHaveBeenCalledTimes(1);
-  expect(publishTip).toHaveBeenCalledWith(doctype1.id, doctype1.tip);
+  // Can not use `toHaveBeenCalledWith` as below due to CID incompatibilities (hello symbol equality)
```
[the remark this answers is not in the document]
ukstv: Probably was a glitch. Reset back to usual thing.

packages/core/src/__tests__/state-manager.test.ts
```diff
@@ -318,19 +319,25 @@ test('enforce previously assigned schema during future update', async () => {
...
+  // expect(publishTip).toHaveBeenCalledWith(doctype1.id, doctype1.tip);
+  // So fall back to manual check
+  expect(publishTip.mock.calls[0][0].toString()).toEqual(doctype1.id.toString())
+  expect(publishTip.mock.calls[0][1].toString()).toEqual(doctype1.tip.toString())
   await publishTip.mockClear();
 
   const updateRec = await TileDoctype._makeCommit(doctype1, ceramic.did, { foo: 34 }, doctype1.controllers);
   await ceramic.repository.stateManager.applyCommit(doc1, updateRec);
-  expect(publishTip).toHaveBeenCalledTimes(1);
   expect(publishTip).toHaveBeenCalledWith(doctype1.id, doctype1.tip);
```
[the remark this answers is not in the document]
ukstv: Ditto.

packages/core/src/state-management/__tests__/cache-eviction.test.ts
```diff
@@ -0,0 +1,88 @@
...
+  await ceramic.createDocument('tile', { content: { purpose: 'evict-one' } }); // Now doc1 is evicted
+  expect(ceramic.repository.inmemory.size).toEqual(1);
+  expect(runningState1.isStopped).toBeTruthy(); // RunningState is stopped after eviction
+
+  const document2 = await createDocument();
+  expect(document2.id).toEqual(document1.id); // Same ID
+
+  await document2.change({ content: { stage: 'changed-concurrently' } });
+  expect(document2.content).toEqual({ stage: 'changed-concurrently' });
+  expect(runningState1.state.next.content).toEqual({ stage: 'changed-1' }); // Running state did not get update
```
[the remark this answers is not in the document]
ukstv: Actually, pinning does not matter. If a document is not pinned and is not present in the cache, and `applyCommit` is requested, the state is loaded as RunningState and updates are still pushed to Repository#feed$. By the nature of `applyCommit`, the state is computed right at (before the commit) time, so pinning does not matter really.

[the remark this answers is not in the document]
ukstv: Anyway, 4 cases added.

packages/core/src/state-management/doctype-from-state.ts
```diff
@@ -1,19 +1,18 @@
-import { Context, Doctype, DoctypeUtils } from '@ceramicnetwork/common';
+import { Context, DocState, Doctype, DoctypeUtils } from '@ceramicnetwork/common';
+import { Observable } from 'rxjs';
+import { filter } from 'rxjs/operators';
 import { HandlersMap } from '../handlers-map';
-import { RunningStateLike } from './running-state';
+import { StateLink } from './state-link';
 
-// Should be removed when Doctype accepts RunningStateLike
 export function doctypeFromState<T extends Doctype>(
```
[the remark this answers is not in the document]
ukstv: k.

packages/core/src/state-management/repository.ts
```diff
@@ -163,8 +155,11 @@ export class Repository {
   /**
    * Adds the document to the in-memory cache
    */
-  add(state: RunningState): void {
-    this.#map.set(state.id.toString(), state);
+  add(state$: RunningState): void {
+    state$.subscribe({
+      next: this.feed$.next.bind(this.feed$),
```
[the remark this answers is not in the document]
ukstv: This is the standard semantics of Rxjs `subscribe`  ¯\_(ツ)_/¯ Would be even more confusing to add a clarification, like it is something non-standard.

packages/core/src/state-management/repository.ts
```diff
@@ -38,7 +38,9 @@ export class Repository {
   /**
    * In-memory cache of the currently running documents.
    */
-  readonly #map: LRUMap<string, RunningState>;
+  readonly inmemory: LRUMap<string, RunningState>;
+
+  readonly feed$: Subject<DocState> = new Subject<DocState>();
```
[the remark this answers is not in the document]
ukstv: k

packages/core/src/dispatcher.ts
ukstv: That's why ARD and guide are in the pipeline.

packages/doctype-caip10-link/src/caip10-link-doctype.ts
ukstv: k

packages/http-client/src/document.ts
ukstv: Syncing now has explicit opt-in semantics (API semantics by @stbrody). There is no global `sync all` or `does not sync anything` flag. It does not make sense anymore. Adding "no-sync" flag to opt-in semantics seems confusing.

packages/http-client/src/ceramic-http-client.ts
ukstv: k

packages/core/src/state-management/repository.ts
```diff
@@ -163,8 +158,11 @@ export class Repository {
   /**
    * Adds the document to the in-memory cache
```
stbrody, line 159:
```suggestion
   * Adds the document's RunningState to the in-memory cache and subscribes the Repository's global feed$ to receive changes emitted by that RunningState
```

packages/core/src/state-management/repository.ts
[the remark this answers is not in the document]
stbrody: Yeah wasn't suggesting documenting the rxjs semantics but rather the semantics of this Repository operation.  Anyway I added a suggested comment change that I think covers it.

stbrody (commented):
This looks great!  Thanks for all the extra testing work.
The one thing I want to confirm is that if an application has a doctype handle and has explicitly called `subscribe()` on it, do they continue to receive updates when they come in via pub-sub even after the RunningState is evicted from the cache?  It looks like if the `RunningState` isn't in the `Repository` then the `ExecutionQueue` won't process the new tip, and I don't see any code to keep subscribed RunningStates in the Repository, but maybe I missed it?  That was really the behavior I was hoping to test with the "4 cases" from my comment on the previous round, but I think the `cache-eviction.test.ts` might be testing at too low a level to prove that unfortunately.

In any case, assuming that works then consider this PR approved, but I do want to make sure that that case works.

stbrody:
NOTE: This PR should not be merged until after the next release of js-ceramic that contains the moving of `unique` into the genesis header and the fix for the http-client, since this change introduces user-visible API changes and we want to wait and bundle that in with the rest of the developer API changes into the following release happening in 2 weeks

oed:
> The one thing I want to confirm is that if an application has a doctype handle and has explicitly called subscribe() on it, do they continue to receive updates when they come in via pub-sub even after the RunningState is evicted from the cache?

This seems very important. However I would assume that if the user has called `subscribe()` we should not evict the `RunningState` from cache at all?

packages/http-client/src/document.ts
oed: Aha, yes ofc! Thanks for clarifying

stbrody:
> > The one thing I want to confirm is that if an application has a doctype handle and has explicitly called subscribe() on it, do they continue to receive updates when they come in via pub-sub even after the RunningState is evicted from the cache?
> 
> This seems very important. However I would assume that if the user has called `subscribe()` we should not evict the `RunningState` from cache at all?

Right, sorry what I should have said is that if the user called subscribe() then the doctype instance continues to receive pubsub updates regardless of how large the cache is and how many other documents are being accessed.  Whether that is achieved by preventing the doc from being evicted from the cache, keeping the doc around in the Repository outside the cache, or something else that relies on knowing about open subscriptions - that's an implementation detail that doesn't really matter so long as the user-visible behavior is right.

ukstv:
Done. The case of a doctype hanging around and later subscribed is solved.

packages/core/src/state-management/repository.ts
```diff
@@ -159,9 +151,6 @@ export class Repository {
    * Adds the document's RunningState to the in-memory cache and subscribes the Repository's global feed$ to receive changes emitted by that RunningState
```
stbrody, line 151:
```suggestion
   * Adds the document's RunningState to the in-memory cache
```

packages/core/src/state-management/state-cache.ts
```diff
@@ -0,0 +1,75 @@
...
+    } else {
+      return this.volatile.get(key);
+    }
+  }
+
+  set(key: string, value: T): void {
+    if (this.durable.has(key)) {
+      this.durable.set(key, value);
+    }
+    this.volatile.set(key, value);
```
stbrody, line 30: Right now the behavior of `endure` and `set` are different with respect to whether or not a doc should stay in `volatile` if it's also in `durable` (`endure` deletes from `volatile` when adding to `durable`, but `set` will add it back to `volatile` even if it's already in `durable`).

I could see a case for either behavior.  If we keep docs only in one place, so that documents in `durable` do not show up in `volatile`, then that would mean that if an app subscribed to a large number of documents, they'd wind up with a number of documents in-memory in the StateCache  that is equal to the number of subscribed docs PLUS the configured cache size.  I feel like it probably makes sense for docs you've actively subscribed to to count towards your "budget" of docs to keep in the cache.  So in that case I think it probably makes sense to always have docs that are in `durable` also exist in `volatile` as well.  This would also have the nice behavior that unsubscribing from a doc (and thus calling free() on it here in the StateCache) doesn't put the doc to the back of the LRU queue but would allow it to actually be immediately deleted if it hadn't had any updates in a long while.

So that's the way I'd probably go, though that said I could see an argument for going the other way as well.  In either case, we should make the behavior of `endure` and `set` consistent.

CC @oed in case you have an opinion here.

stbrody (commented):
Looks good!  Thanks for all your patience and hard work on this.  I know it was a long review process.  Everything looks good except one comment about whether subscribed docs should eat into the cache or not

oed:
> Done. The case of a doctype hanging around and later subscribed is solved.

Thanks both @ukstv and @stbrody. This was one of the main things I wanted to make sure wasn't missed with the refactor so glad it didn't get looked over! Glad you both spent the time to make sure it was included!

ukstv:
@stbrody How about these two cases?
Common disposition: The budged is 100 documents. One has subscribed to 100 documents.
Case 1: Request to one more document comes. Rejecting the request is not an option (hope, this is obvious). Effective budget equals 1 now. The budget semantics breaks: it is 100, but really 101.
Case 2: One subscribes to one more document, so that one is subscribed to 101 documents. You could add a request to one more document. The budget semantics breaks again.

That is why two buckets here. A subscription to a document is a clear signal "you will consume additional resources". No surprises.

`set` / `endure` symmetry: ok. Does not hurt.

stbrody:
> @stbrody How about these two cases?
> Common disposition: The budged is 100 documents. One has subscribed to 100 documents.
> Case 1: Request to one more document comes. Rejecting the request is not an option (hope, this is obvious). Effective budget equals 1 now. The budget semantics breaks: it is 100, but really 101.
> Case 2: One subscribes to one more document, so that one is subscribed to 101 documents. You could add a request to one more document. The budget semantics breaks again: it is 100, but really max[any-number-of-subscribed + 1, 100].
> 
> That is why two buckets here. A subscription to a document is a clear signal "you will consume additional resources". No surprises.
> 
> Consider the environment where this subscription case unfolds: a developer runs Ceramic node inside her own code. If someone does this, I consider the person smart enough to adjust cache limit according to the anticipated load profile, i.e. figuring how many spare slots to have for running documents apart the subscriptions.
> 
> `set` / `endure` symmetry: ok. Does not hurt.


Yep, I agree with most of what you said there, just not with the conclusion that those concerns mean we shouldn't have subscribed documents in the `volatile` set.  It basically boils down to, as written in the PR currently, the number of docs in memory is `num_subscribed + cache_size` vs in the approach I'm proposing it's `max(num_subscribed, cache_size)`.  I don't think either approach "breaks" the semantics, they just have different semantics.  In the version we have here, we keep slightly more docs in memory, and when the last subscriber to a doc unsubscribes, the doc gets put into the `volatile` LRU cache, possibly evicting some other recently accessed document. In the version I'm proposing, we keep slightly fewer docs in memory, and when the last subscriber to a doc unsubscribes it either is already in `volatile` and just keeps its current place in the eviction order, or it's already been evicted from `volatile`, implying it hasn't been accessed in a while, and so it immediately gets deleted from the `StateCache`.  In my mind the latter seems like a bit nicer behavior, but I don't think it's a critical decision either way, so if you feel strongly that it is better this way then that sounds fine to me.

As for the set/endure symmetry change, sorry, I had been thinking about the case where `set` gets called on a doc that is already in `durable`, but now I see that that case shouldn't ever happen.  `set` is only called if the doc isn't found anywhere in the `StateCache`.  Added one comment about making that clear with an exception.


Anyway, to summarize: LGTM!

stbrody (approved):

packages/core/src/state-management/state-cache.ts
```diff
@@ -0,0 +1,100 @@
...
+  get(key: string): T | undefined {
+    return this.durable.get(key) || this.volatile.get(key);
+  }
+
+  /**
+   * Set value to volatile bucket. Remove from durable.
+   */
+  set(key: string, value: T): void {
+    this.volatile.set(key, value);
+    this.durable.delete(key);
```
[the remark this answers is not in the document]
ukstv: This would break a symmetry, wouldn't it?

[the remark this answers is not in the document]
stbrody: Aren't they actually asymmetrical in this regard though?  It's safe to call `endure` after `set` (will happen anytime a doc is loaded and then subscribed to), it's not safe to call `set` after `endure`.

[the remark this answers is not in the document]
ukstv: And if we consider making calls _safe_, we get back to the original logic with update if `durable` is found :P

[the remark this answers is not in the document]
stbrody: Yep, you're right, that would do it too.  I had misunderstood how `set` is used.  Sorry for the confusion.

PR #1157, opened 2021-03-25
Addition of Avalanche C-Chain Support

kylecbetts:
Avalanche is an open-source platform for launching decentralized applications and enterprise blockchain deployments in one interoperable, highly scalable ecosystem.

Avalanche C-Chain uses the CAIP-3 eip155 namespace with two chain IDs:
43113 for Avalanche Fuji Testnet
43114 for Avalanche Mainnet


oed:
Hey @kylecbetts the `eip155` namespace is already supported by the "ethereum" AuthProvider. There should be no need to add any extra logic here!

kylecbetts:
So should we still have an AvalancheAuthProvider class that is a subclass of EthereumAuthProvider with no logic, or are you suggesting adding nothing?

oed:
@kylecbetts yeah, unless there is something that you do differently than the *EthereumAuthProvider* then there is no reason to add duplicate code :)

kylecbetts:
@oed thanks for your patience, I have changed the implementation of AvalancheAuthProvider to simply inherit everything from EthereumAuthProvider.

oed:
@kylecbetts I think you might have misunderstood. Seems like this PR is not needed at all if you can just use the *EthereumAuthProvider* directly? There is nothing added here in terms of functionality afaict.

kylecbetts:
I guess I was just thinking for users of js-ceramic it would eliminate any confusion of wanting to use the Avalanche platform, as they might not know we use the same functionality as the EthereumAuthProvider. I agree nothing here in terms of functionality, just ease of use.

packages/blockchain-utils-linking/src/index.ts
```diff
@@ -1,12 +1,14 @@
...
 export * as ethereum from "./ethereum";
+export * as avalanche from "./avalanche";
 export * from "./util";
 export * as filecoin from "./filecoin";
 export * as polkadot from "./polkadot";
 export * as eosio from "./eosio";
 export * as cosmos from "./cosmos";
 
 export { EthereumAuthProvider } from "./ethereum";
+export { AvalancheAuthProvider } from "./avalanche";
```
oed, line 11:
```suggestion
export { EthereumAuthProvider as AvalancheAuthProvider } from "./ethereum";
```

oed (changes requested):
Ok if that's what we need, then I would just do the suggested change and remove everything else.

kylecbetts:
@oed Sorry for all the confusion, I do appreciate the help.

oed (approved):
Thanks @kylecbetts

PR #1158, opened 2021-03-25
Remove dependency on RotatingFileStream from common and core packages

stbrody:
This moves the primary logging type definitions `DiagnosticsLogger` and `ServiceLogger` into the `common` package (from the `logger` package) and makes the `LoggerProvider` receive a factory function for constructing the `RotatingFileStream` needed for file logging.  The `RotatingFileStream` remains in the `logger` package (and that is now the only thing contained in the `logger` package).  The logger package is now only used directly by the CLI to inject a `RotatingFileStream` constructor into the `LoggerProvider`

Note that in order to get this to build I had to do an RC release so that the ipfs packages could depend on a version of `common` that had the `DiagnosticsLogger` type exported.


packages/core/src/state-management/execution-queue.ts
```diff
@@ -31,7 +31,7 @@ export class ExecutionQueue {
 
   constructor(logger: DiagnosticsLogger, readonly get: (docId: DocID) => Promise<RunningState>) {
     this.tasks = new NamedTaskQueue((error) => {
-      logger.err(error);
+      logger.err(error.toString());
```
valmack, line 34: Same comment as above

packages/core/src/pubsub/incoming-channel.ts
```diff
@@ -24,7 +24,7 @@ function buildResubscribeQueue(logger: DiagnosticsLogger) {
       // For some reason the first call to pubsub.subscribe throws this error.
       retry();
     } else {
-      logger.err(error);
+      logger.err(error.toString());
```
valmack, line 27: This doesn't seem right.. `logger.err` should take in an error and handle it correctly.

See: https://github.com/seanpmaxwell/overnight/blob/master/src/logger/lib/Logger.ts#L213-L215

packages/core/src/state-management/repository.ts
```diff
@@ -52,7 +52,7 @@ export class Repository {
 
   constructor(limit: number, private readonly logger: DiagnosticsLogger) {
     this.loadingQ = new NamedTaskQueue((error) => {
-      logger.err(error);
+      logger.err(error.toString());
```
valmack, line 55: Same comment as above

valmack (commented):
So ideally we would rename the @ceramicnetwork/logger package to `@ceramicnetwork/file-logger`. I'm guessing that is too big of an ask right now?

Also this needs a change in CAS to use the logger from `@ceramicnetwork/common` correct?

Lastly, I think we should pass error objects to `logger.err` not strings so we can get a clean stack trace in the logs. From what I understand, overnight is make this call when it gets an error: https://nodejs.org/api/util.html#util_util_inspect_object_showhidden_depth_colors

stbrody:
> So ideally we would rename the @ceramicnetwork/logger package to `@ceramicnetwork/file-logger`. I'm guessing that is too big of an ask right now?

Yeah that would be nice, but I'll defer to @oed on that one since I know we can't actually *delete* the logger package, just mark it deprecated

> Also this needs a change in CAS to use the logger from `@ceramicnetwork/common` correct?

Yes correct, I'll file a ticket

> Lastly, I think we should pass error objects to `logger.err` not strings so we can get a clean stack trace in the logs. From what I understand, overnight is make this call when it gets an error: [nodejs.org/api/util.html#util_util_inspect_object_showhidden_depth_colors](https://nodejs.org/api/util.html#util_util_inspect_object_showhidden_depth_colors)

Done

valmack (approved):
👍

packages/key-did-resolver/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "key-did-resolver",
-  "version": "1.1.0",
+  "version": "1.1.1-rc.0",
```
oed, line 3: I don't get why this package is getting a version bump?

oed (approved):
lgtm, thanks

valmack:
> err, y that would be nice. Not sure we don't have a use for the logger pkg in the future. I would just punt on it for now until we have a need for multiple different loggers.

Well as it stands the names are very confusing. We do have multiple loggers now. We have the primary logging functionality in @ceramicnetwork/common and the only thing left in @ceramicnetwork/logger is file logging.

So people coming to our repo are going to be confused. I think we should rename it.

oed:
@valmack aha, so the logger in common is just logging to console or similar? Then yeah a rename might make sense if it's not too much effort!

Issue #1163, opened 2021-03-26
Enable building http-client with webpack


stbrody:
Simple test of building http-client with webpack found the following errors:

```

> ceramic-browser@ start /home/spencer/3box/webpack-test
> webpack --mode=production .

assets by status 448 KiB [cached] 1 asset
runtime modules 1.02 KiB 6 modules
orphan modules 3.97 KiB [orphan] 2 modules
modules by path ./node_modules/ 1.02 MiB
  javascript modules 1.01 MiB 160 modules
  json modules 3.9 KiB
    ./node_modules/@ceramicnetwork/http-client/package.json 2.1 KiB [built] [code generated]
    ./node_modules/elliptic/package.json 1.8 KiB [built] [code generated]
optional modules 30 bytes [optional]
  crypto (ignored) 15 bytes [optional] [built] [code generated]
  buffer (ignored) 15 bytes [optional] [built] [code generated]
./index.js 39 bytes [built] [code generated]

ERROR in ./node_modules/@ceramicnetwork/common/lib/logger-provider.js 8:31-46
Module not found: Error: Can't resolve 'path' in '/home/spencer/3box/webpack-test/node_modules/@ceramicnetwork/common/lib'

BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
	- add a fallback 'resolve.fallback: { "path": require.resolve("path-browserify") }'
	- install 'path-browserify'
If you don't want to include a polyfill, you can use an empty module like this:
	resolve.fallback: { "path": false }
 @ ./node_modules/@ceramicnetwork/common/lib/index.js 19:13-41
 @ ./node_modules/@ceramicnetwork/http-client/lib/ceramic-http-client.js 10:17-50
 @ ./index.js 1:0-38

ERROR in ./node_modules/@ceramicnetwork/common/lib/logger-provider.js 9:29-42
Module not found: Error: Can't resolve 'os' in '/home/spencer/3box/webpack-test/node_modules/@ceramicnetwork/common/lib'

BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
	- add a fallback 'resolve.fallback: { "os": require.resolve("os-browserify/browser") }'
	- install 'os-browserify'
[...]

PR #1165, opened 2021-03-26
fix(common): Don't use node packages at runtime in the logger if we might be running in-browser

stbrody:
Doesn't actually remove the dependency from the package, so bundlers will still get mad, but now at least if you can get the code into the browser (by providing polyfills, suppressing bundler errors, or not using a bundler at all), then things should work

### DIFF
--- a/packages/common/src/logger-provider.ts
+++ b/packages/common/src/logger-provider.ts
@@ -17,9 +17,11 @@ export interface LoggerConfig {
     logToFiles?: boolean,
 }
 
+// Can't have a default log directory as that would require using the `path` module before we know
+// for sure if it's safe to do so. If `logToFiles` is false we must avoid using any node-specific
+// functionality as we might be running in-browser
 const DEFAULT_LOG_CONFIG = {
     logLevel: LogLevel.important,
-    logDirectory: path.join(os.homedir(), ".ceramic", "logs/"),
     logToFiles: false
 }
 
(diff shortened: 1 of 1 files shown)


stbrody:
@zachferland, is there any way we can test the assumption that this works as long as you can make your bundler happy or don't use a bundler?  Maybe by deploying in 3id-connect?

packages/common/src/logger-provider.ts
```diff
@@ -46,9 +48,19 @@ export class LoggerProvider {
       this._diagnosticLogger = this._makeDiagnosticLogger()
     }
 
+    private _getLogPath(filename: string) {
+        if (!this.config.logToFiles) {
+            throw new Error("Tried to generate log path when logToFiles is false") // This indicates a programming error
+        }
+        const logDirectory = this.config.logDirectory || path.join(os.homedir(), ".ceramic", "logs/")
```
oed, line 55: Why couldn't the logger itself take care of setting the path?

stbrody: It could, but I don't think that helps in any way?

oed: Doesn't it? The `logDirectory` should only be relevant to the filesystem logger. Am I missing something?

stbrody: So instead of having just the `RotatingFileStream` provided into the logging system from outside (in our case from the `cli` package when setting up the daemon), you want a implementation of the `DiagnosticLogger` and `ServiceLogger` to be provided, and have those handle the config file parsing to set the log directory?  That would work, but is a larger refactoring.

oed: Ok, I don't know exactly how the different components fit together. I just intuitively felt that it would make sense for directory to only be set for the loggers that use it. I'm fine with not implementing that now though.

zachferland:
yeah would say merge anyways for now - better after would be split out into to two logger-providers - ref https://github.com/ceramicnetwork/js-ceramic/issues/1163

and this alone doesnt seem to unblock anything, ie http-client seems fine without this change anyways, and then core still has other issues to resolve (but this does seem to resolve logger issue towards that, unless overnight logger as some issues without node libs)

PR #1180, opened 2021-03-30
feat(tile-doctype): Update Tile API

stbrody:
Builds on https://github.com/ceramicnetwork/js-ceramic/pull/1150


packages/3id-did-resolver/CHANGELOG.md
```diff
@@ -3,6 +3,22 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# [1.2.0-rc.0](/compare/@ceramicnetwork/3id-did-resolver@1.1.0...@ceramicnetwork/3id-did-resolver@1.2.0-rc.0) (2021-03-29)
```
stbrody, line 6: Sorry, release candidate version bump snuck in while trying to debug some test failures due to what appeared to be circular dependency issues

packages/doctype-tile/src/tile-doctype.ts
```diff
@@ -27,45 +30,130 @@ export interface TileParams extends DocParams {
...
     /**
-     * Change existing Tile doctype
-     * @param params - Change parameters
-     * @param opts - Initialization options
+     * Creates a Tile document.
+     * @param ceramic - Instance of CeramicAPI used to communicate with the Ceramic network
+     * @param content - Genesis contents. If 'null', then no signature is required to make the genesis commit
+     * @param metadata - Genesis metadata
+     * @param opts - Additional options
+     * TODO: Better way to specify "no content"? Should 'null' get converted to 'undefined'? What about the empty document {}?
```
stbrody, line 81: In the old API "no content" was expressed by just leaving the `content` field out of the `DocParams` entirely.  Now that `content` is a top-level argument, we need to decide how to represent "no content" especially as it pertains to the rule that genesis commits with no content don't need to be signed.  The version I went with here accepts both `undefined` and `null` as meaning "no content" and will ensure that the `data` field doesn't show up in the `GenesisCommit` in either case.  We could consider adding `{}` as an option for "no content", making `{}` the *only* way to specify "no content" (so that the `content` arg is always an object), whether we should include the `data` field in the GenesisCommit as an empty object, and any combination of the above.

But I think what I did here is likely fine.

packages/http-client/src/ceramic-http-client.ts
```diff
@@ -127,6 +127,9 @@ export default class CeramicClient implements CeramicApi {
     }
   }
 
+  /**
+   * @deprecated
+   */
   async createDocument<T extends Doctype>(doctype: string, params: DocParams, opts?: DocOpts): Promise<T> {
```
stbrody, line 133: I tried to remove this method entirely, but when I do I get the following compilation error:
```
src/ceramic-http-client.ts(83,57): error TS2345: Argument of type 'this' is not assignable to parameter of type 'Ceramic'.
  Property 'createDocument' is missing in type 'CeramicClient' but required in type 'Ceramic'.
```

I'm not actually sure what is happening here TBH.  I feel like it might be due to the circular dependency between the http-client and the 3id-did-resolver, but then shouldn't the error message say that `createDocument` is required in type `CeramicApi`?  Not in type `Ceramic`?  I'm not sure how the `Ceramic` type is in play here at all...

stbrody:
I'm not sure what happened with the docs build here?  This PR doesn't even include any meaningful changes to any of the pinning packages... @oed any ideas?

packages/doctype-tile/src/tile-doctype.ts
```diff
@@ -27,45 +30,137 @@ export interface TileParams extends DocParams {
...
 
+/**
+ * Arguments used to generate the metadata for Tile documents
+ */
+export interface TileMetadataArgs {
+  controllers?: Array<string>
+  family?: string
+  schema?: CommitID | string
+  tags?: Array<string>
+  deterministic?: boolean
```
stbrody, line 41: We still need to make a final decision on how developers should specify `deterministic`. I'm a little less opposed to putting it into `DocOpts` than I was initially.  It would just mean splitting `DocOpts` up to be doctype-specific, but that's likely something we'll want to do eventually anyway.  I personally still like it here in metadata, even though it's not quite 1-1 to what comes back when metadata is loaded.  But if others think this would be better in `DocOpts` I could possibly be swayed.

stbrody:
> I'm not sure what happened with the docs build here? This PR doesn't even include any meaningful changes to any of the pinning packages... @oed any ideas?

Actually, looks like it's failing on `develop` branch now also.  Maybe some upstream package got updated and broke us?  https://app.circleci.com/pipelines/github/ceramicnetwork/js-ceramic/2702/workflows/d79fa8b3-bfd6-4b36-a611-76d098d0236d/jobs/2880

packages/3id-did-resolver/generateVectorDocuments.js
```diff
@@ -123,8 +123,9 @@ const legacyDid = async (threeId, threeIdGenesisCopy, ceramic) => {
   // uses the first public key of the seed for the v1 threeId as the keydid.
   // sorry for the magic here -.-
   await ceramic.setDIDProvider(threeIdGenesisCopy.getDidProvider())
-  const docParams =  { deterministic: true, metadata: { controllers: [firstKeyDid], family: '3id' } }
-  const doc = await ceramic.createDocument('tile', docParams, { anchor: false, publish: false })
+  const metadata = { controllers: [firstKeyDid], family: '3id', deterministic: true }
+  const doc = await TileDoctype.create(
+      ceramic, {}, metadata, { anchor:false, publish: false })
```
oed, line 128: Is there a difference here between passing undefined (what happened before) and an empty object?

packages/3id-did-resolver/src/index.ts
```diff
@@ -131,8 +132,8 @@ const legacyResolve = async (ceramic: CeramicApi, didId: string, verNfo: Version
   // This would remove one request to ceramic.
   //const genesisCommit = { header: { family: '3id', controllers: [legacyPublicKeys.keyDid] }, unique: '0' }
   //const docid = new DocID('tile', await dagCBOR.util.cid(new Uint8Array(dagCBOR.util.serialize(genesisCommit))))
-  const docParams =  { deterministic: true, metadata: { controllers: [legacyPublicKeys.keyDid], family: '3id' } }
-  const docid = (await ceramic.createDocument('tile', docParams, { anchor: false, publish: false })).id
+  const metadata =  { controllers: [legacyPublicKeys.keyDid], family: '3id', deterministic: true }
+  const docid = (await TileDoctype.create(ceramic, null, metadata, { anchor: false, publish: false })).id
```
oed, line 136: hm, here you are passing `null`.

packages/common/src/doctype.ts
```diff
@@ -22,6 +22,7 @@ export enum AnchorStatus {
 
 export interface CommitHeader {
     controllers: Array<string>
+    family?: string
```
oed, line 25: 👍

packages/core/src/ceramic.ts
```diff
@@ -506,6 +506,7 @@ class Ceramic implements CeramicApi {
    * @param doctype - Document type
    * @param params - Create parameters
    * @param opts - Initialization options
+   * @deprecated
    */
   async createDocument<T extends Doctype>(doctype: string, params: DocParams, opts?: DocOpts): Promise<T> {
```
oed, line 511: Mind also adding a `console.warn` here telling devs that it's deprecated?

packages/doctype-tile/src/tile-doctype.ts
oed: Thanks for clarifying. This seems good.

packages/http-client/src/ceramic-http-client.ts
oed: Hm, yeah that's odd.

oed (approved):
lgtm. Don't have a strong view on the *deterministic* flag.

ukstv (changes requested):
All good to me, apart code style. Yes, I know, it is annoying. `TileDoctype` has a lot of static methods. Few of them are really part of the doctype API. Others (`_makeGenesis`, `_signDagJWS`),  better be present as functions belonging to module, rather than TileDoctype class.

ukstv:
On similar topic: @stbrody Why `headerFromMetadata` and `throwReadOnlyError` are declared as `const` rather than `function headerFromMetadata(){}`?

packages/doctype-tile/src/tile-doctype.ts
```diff
@@ -27,45 +30,137 @@ export interface TileParams extends DocParams {
...
+     * @param opts - Additional options
+     */
+    static async load(ceramic: CeramicApi, docId: DocID | CommitID | string, opts: DocOpts = {}): Promise<TileDoctype> {
+        const docRef = DocRef.from(docId)
+        if (docRef.type != TileDoctype.DOCTYPE_ID) {
+            throw new Error(`DocID ${docRef.toString()} does not refer to a 'tile' doctype, but to a ${docRef.typeName}`)
         }
 
-        const updateCommit = await TileDoctype._makeCommit(this, this.context.did, params.content, params.metadata?.controllers, params.metadata?.schema)
+        return await ceramic.loadDocument<TileDoctype>(docRef, opts)
```
ukstv, line 118: await is redundant here

packages/doctype-tile/src/tile-doctype.ts
```diff
@@ -27,45 +30,137 @@ export interface TileParams extends DocParams {
...
+     * @param ceramic - Instance of CeramicAPI used to communicate with the Ceramic network
+     * @param content - Genesis contents. If 'null', then no signature is required to make the genesis commit
+     * @param metadata - Genesis metadata
+     * @param opts - Additional options
+     * TODO: Better way to specify "no content"? Should 'null' get converted to 'undefined'? What about the empty document {}?
      */
-    async change(params: TileParams, opts: DocOpts = {}): Promise<void> {
-        if (this.context.did == null) {
-            throw new Error('No DID authenticated')
+    static async create<T>(ceramic: CeramicApi, content: T | null, metadata?: TileMetadataArgs, opts: DocOpts = {}): Promise<TileDoctype> {
```
ukstv, line 90: `Promise<TileDoctype<T>>` I guess?

Also, `content: T | null | undefined`

packages/doctype-tile/src/tile-doctype.ts
```diff
@@ -27,45 +30,137 @@ export interface TileParams extends DocParams {
...
-     * @param context - Ceramic context
-     * @param opts - Initialization options
+     * Update the contents of an existing Tile document based on a JSON-patch diff from the existing
+     * contents to the desired new contents
+     * @param jsonPatch - JSON patch diff of document contents
+     * @param opts - Additional options
+     */
+    async patch(jsonPatch: Operation[], opts: DocOpts = {}): Promise<void> {
+        const header = headerFromMetadata(this.metadata)
+        const unsignedCommit: UnsignedCommit = { header, data: jsonPatch, prev: this.tip, id: this.state$.value.log[0].cid }
```
ukstv, line 141: `id: this.state$.id.cid`

packages/core/src/ceramic.ts
stbrody: I hope to delete this method entirely ASAP and before the next release.  The only reason it isn't done in this PR is because of this compiler error I mentioned: https://github.com/ceramicnetwork/js-ceramic/pull/1180#discussion_r604348910

packages/3id-did-resolver/generateVectorDocuments.js
stbrody: Yeah, this should probably have been `null`. Updated

packages/doctype-tile/src/tile-doctype.ts
stbrody: Good call, updated

stbrody (commented):
New version up addressing review comments

stbrody:
> On similar topic: @stbrody Why `headerFromMetadata` and `throwReadOnlyError` are declared as `const` rather than `function headerFromMetadata(){}`?

No reason.  Is there a meaningful difference?  Should I always use the `function foo()` form?

stbrody:
> All good to me, apart code style. Yes, I know, it is annoying. `TileDoctype` has a lot of static methods. Few of them are really part of the doctype API. Others (`_makeGenesis`, `_signDagJWS`), better be present as functions belonging to module, rather than TileDoctype class.

Done, though I'm not totally clear on why this way is better?  Since they were private methods anyway it's not like they're polluting the public TileDoctype API

ukstv (approved):
> > On similar topic: @stbrody Why `headerFromMetadata` and `throwReadOnlyError` are declared as `const` rather than `function headerFromMetadata(){}`?
> 
> No reason. Is there a meaningful difference? Should I always use the `function foo()` form?

There is a difference, but it is irrelevant here. Thought, you have some good reasoning behind this, actually :)

ukstv:
> > All good to me, apart code style. Yes, I know, it is annoying. `TileDoctype` has a lot of static methods. Few of them are really part of the doctype API. Others (`_makeGenesis`, `_signDagJWS`), better be present as functions belonging to module, rather than TileDoctype class.
> 
> Done, though I'm not totally clear on why this way is better? Since they were private methods anyway it's not like they're polluting the public TileDoctype API. Just asking to learn if there's something about javascript that makes this approach better in some way that I don't understand.

Irrelevant FYI, just to be sure we are on the same page: This kind of `private` exists only on TS level. JS have no idea on restrictions here.

It might be just my quirk, but I have a hard time reading classes/files with a lot of static methods. I can not understand responsibilities of the code, what is important for the purpose, and what is not. After the change, there is a clear separation between relevant functions (1.5 screens of code, _clear relationships_ between the methods), and ancillary stuff. One can clearly see if the ancillary stuff can be DRYed/shared.

Again, functionally there is no difference. For code comprehension it is a boon (IMO).

packages/doctype-tile/src/tile-doctype.ts
```diff
@@ -27,45 +30,137 @@ export interface TileParams extends DocParams {
...
+        const commit = await TileDoctype._signDagJWS(unsignedCommit, this.context.did, this.controllers[0])
+        const updated = await this.context.api.applyCommit(this.id.toString(), commit, opts)
```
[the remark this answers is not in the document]
stbrody: Nice catch

packages/doctype-tile/src/tile-doctype.ts
```diff
@@ -85,35 +180,65 @@ export class TileDoctype extends Doctype {
...
+      return await TileDoctype._signDagJWS(commit, did, metadata.controllers[0])
+    }
+
+    /**
+     * Make a commit to update the document
+     * @param did - DID making (and signing) the commit
+     * @param newContent
+     * @param newMetadata
+     */
+    async _makeCommit(did: DID, newContent: T | undefined, newMetadata?: TileMetadataArgs): Promise<CeramicCommit> {
```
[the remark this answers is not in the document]
stbrody: That's right

packages/http-client/src/ceramic-http-client.ts
stbrody: There really isn't much else interesting besides what I posted.  Here's the error message with more of the context on either side:

```
lerna ERR! npm run build exited 2 in '@ceramicnetwork/http-client'
lerna ERR! npm run build stdout:

> @ceramicnetwork/http-client@0.11.0-rc.1 prebuild /home/spencer/js-ceramic2/packages/http-client
> npm run clean


> @ceramicnetwork/http-client@0.11.0-rc.1 clean /home/spencer/js-ceramic2/packages/http-client
> rm -rf ./lib


> @ceramicnetwork/http-client@0.11.0-rc.1 build /home/spencer/js-ceramic2/packages/http-client
> ../../node_modules/.bin/tsc -p tsconfig.json

src/ceramic-http-client.ts(83,57): error TS2345: Argument of type 'this' is not assignable to parameter of type 'Ceramic'.
  Property 'createDocument' is missing in type 'CeramicClient' but required in type 'Ceramic'.

lerna ERR! npm run build stderr:
npm ERR! code ELIFECYCLE
npm ERR! errno 2
npm ERR! @ceramicnetwork/http-client@0.11.0-rc.1 build: `../../node_modules/.bin/tsc -p tsconfig.json`
npm ERR! Exit status 2
npm ERR!
npm ERR! Failed at the @ceramicnetwork/http-client@0.11.0-rc.1 build script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/spencer/.npm/_logs/2021-04-02T21_51_08_181Z-debug.log

```

stbrody (commented):
> > > All good to me, apart code style. Yes, I know, it is annoying. `TileDoctype` has a lot of static methods. Few of them are really part of the doctype API. Others (`_makeGenesis`, `_signDagJWS`), better be present as functions belonging to module, rather than TileDoctype class.
> > 
> > 
> > Done, though I'm not totally clear on why this way is better? Since they were private methods anyway it's not like they're polluting the public TileDoctype API. Just asking to learn if there's something about javascript that makes this approach better in some way that I don't understand.
> 
> Irrelevant FYI, just to be sure we are on the same page: This kind of `private` exists only on TS level. JS have no idea on restrictions here.
> 
> It might be just my quirk, but I have a hard time reading classes/files with a lot of static methods. I can not understand responsibilities of the code, what is important for the purpose, and what is not. After the change, there is a clear separation between relevant functions (1.5 screens of code, _clear relationships_ between the methods), and ancillary stuff. One can clearly see if the ancillary stuff can be DRYed/shared.
> 
> Again, functionally there is no difference. For code comprehension it is a boon (IMO).

Thanks for explaining, that's helpful for me to understand!

PR #1182, opened 2021-03-31
fix: remove cpu and memory from healthcheck

valmack:
The readings are off on 3Box Ceramic infra--maybe it doesn't work properly when running the process in a container. CPU usage said it was over .95 but we have been around 25% consistently per cloudwatch/grafana dashboards. Also potentially caused 502 gateway errors.

*Anyone can merge!*


ukstv (approved):
🏇

PR #1183, opened 2021-04-01
fix(blockchain-utils-linking): Pin version of @glif/local-managed-provider to unbreak test

stbrody:


stbrody:
Merged without review to unbreak tests, let me know if you have any concerns

PR #1186, opened 2021-04-02
feat: seperate node and browser loggers

zachferland:
~~will revisit tmrrw, some cleanup/types, and had some trouble testing locally and getting dependencies to resolve properly~~

likely better long term solutions after, but can be revisited at same time as we work on core working in browser again, there may be other options for libraries as well, but wanted to minimize any changes there since we have a lot built on current log formats.

### DIFF
--- a/packages/common/src/logger-base.ts
+++ b/packages/common/src/logger-base.ts
@@ -1,0 +1,186 @@
+export enum LogStyle {
+  verbose = 'info',
+  info = 'info',
+  imp = 'imp',
+  warn = 'warn',
+  err = 'err'
+}
+
+export enum LogLevel {
+  verbose,
+  debug,
+  important,
+  warn,
+}
+
+export interface WriteableStream {
+  write: (message: string) => void
+}
+
+/**
+ * Logs to the console based on log level
+ */
+export class DiagnosticsLoggerBase {
+  public readonly logLevel: LogLevel;
+  protected logger: any;
+  protected readonly fileLogger: WriteableStream;
+  protected readonly includeStackTrace: boolean;
+  protected readonly logToFiles
+
+  constructor(logLevel: LogLevel, logToFiles: boolean, fileLogger?: WriteableStream) {
+    this.logLevel = logLevel;
+    this.includeStackTrace = this.logLevel <= LogLevel.debug ? true : false;
+    this.logToFiles = logToFiles
+    this.fileLogger = fileLogger
+  }
+
+  /**
+   * Calls `this.debug`. Used for stream interfaces.
+   * @param content Content to log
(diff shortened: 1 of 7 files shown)


zachferland:
This and https://github.com/StableLib/stablelib/pull/30 will allow http-client to be bundled by default (tested webpack5, probably helps with other as well, or at least optimizes those by removing polyfills not needed) 

This is bit of a bandaid that minimizes making changes anywhere else, but we should trend towards organizing more like described here https://github.com/ceramicnetwork/js-ceramic/issues/1163 when we revisit/choose to prioritize getting ceramic core to work/bundle in browser

packages/common/src/logger-base.ts
```diff
@@ -0,0 +1,186 @@
...
+  write: (message: string) => void
+}
+
+/**
+ * Logs to the console based on log level
+ */
+export class DiagnosticsLoggerBase {
+  public readonly logLevel: LogLevel;
+  protected logger: any;
+  protected readonly fileLogger: WriteableStream;
```
stbrody, line 26: Seems like this should be moved out of the base class and into the `DiagnosticsLogger` that extends this and actually uses it?

packages/common/src/logger-base.ts
```diff
@@ -0,0 +1,186 @@
...
+}
+
+/**
+//  * Global Logger factory
+//  */
+export class LoggerProviderBase {
+  public readonly config: LoggerConfig
+
+  protected readonly _diagnosticLogger: DiagnosticsLoggerBase
+  protected readonly _fileLoggerFactory: FileLoggerFactory
```
stbrody, line 160: Ditto, seems like this doesn't belong in the base class if it's not used here anywhere

stbrody (approved):
LGTM mod question about moving file loggers out of base classes and into derived classes

zachferland:
tried changing, but then remembered why i did that, otherwise the extended class types would not match the base class types which are then used everywhere else. (although not really sure why it complains when protected, public interfaces the same). The better implementation again would be complete separation and likely shared logger interfaces that they both implement instead.

stbrody:
> tried changing, but then remembered why i did that, otherwise the extended class types would not match the base class types which are then used everywhere else. (although not really sure why it complains when protected, public interfaces the same). The better implementation again would be complete separation and likely shared logger interfaces that they both implement instead.

Ah I see. `LoggerProviderBase` would need to be called `LoggerProvider`, but then you'd need a factory for getting LoggerProvider instances, which defeats the whole purpose of the provider.  Okay, this is fine for now then until we can do the better fix

stbrody (approved):

PR #1187, opened 2021-04-02
feat: add networks enum and elp

valmack:
### Motivation

We have some infra up for ELP but now need a way to run the Ceramic nodes!

### Changes

- Added "elp" network
- Created enum for different networks (will punt this change if there's hesitancy because just want to get this running asap ideally)

### Notes

Do we need a test for this? If so will sync with @stbrody about how because I didn't really understand if/how we are testing these networks.

### DIFF
--- a/packages/core/src/ceramic.ts
+++ b/packages/core/src/ceramic.ts
@@ -25,6 +25,7 @@ import {
   MultiQuery,
   PinningBackendStatic,
   LoggerProvider,
+  Networks,
 } from "@ceramicnetwork/common"
 import { Resolver } from "did-resolver"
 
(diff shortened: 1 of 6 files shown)


oed (approved):
Why do we have both ELP and MAINNET?

valmack:
> Why do we have both ELP and MAINNET?

@oed what we discussed briefly the other day is that we don't want people outside of ELP to have the misconceptions about mainnet and ideally we can prevent them from using it. Obviously we can't literally do that but having "elp" be a somewhat hidden network can help to some degree.

If this feels frivolous (valid opinion) we can just use mainnet instead of adding a new elp network.

packages/core/src/ceramic.ts
```diff
@@ -52,19 +53,22 @@ const IPFS_GET_TIMEOUT = 60000 // 1 minute
 const TESTING = process.env.NODE_ENV == 'test'
 
 const DEFAULT_ANCHOR_SERVICE_URLS = {
-  "testnet-clay": "https://cas-clay.3boxlabs.com",
-  "dev-unstable": "https://cas-dev.3boxlabs.com",
-  "local": "http://localhost:8081",
+  [Networks.MAINNET]: "https://cas.3boxlabs.com",
```
stbrody, line 56: Nice

stbrody (approved):
Looks great.  And FWIW I like having the separate "elp" hidden network flag

PR #1190, opened 2021-04-02
chore(js-ceramic): Only allow creating release candidates on the 'release-candidate' branch

stbrody:


valmack (approved):
Why `--no-push`?

stbrody:
> Why `--no-push`?

I just left what was there.  You're right, now if we're limiting to just this one branch it probably makes sense to push automatically as part of the release process.  Updated

PR #1193, opened 2021-04-02
chore(blockchain-utils-validation): Increase test timeout for eosio.test.ts

stbrody:
Looks like this might be enough to stabilize the test

### DIFF
--- a/packages/3id-did-resolver/generateVectorDocuments.js
+++ b/packages/3id-did-resolver/generateVectorDocuments.js
@@ -11,6 +11,10 @@
 
 const Ceramic = require('@ceramicnetwork/http-client').default
 const ThreeIdProvider = require('3id-did-provider').default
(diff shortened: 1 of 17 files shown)


ukstv:
Whoops, seems like this self-merged PR contains also did setter.

ukstv (commented):
🏇

stbrody:
> Whoops, seems like this self-merged PR contains also did setter.

Yeah, I built the DID cleanup PR on top of this one so that the tests wouldn't keep failing in the other PR, but then when I went to merge I forgot to set the base branch back to develop so that PR got merged into this one.  Github UI makes this way too easy to do IMO.  Anyway that's why I went ahead and merged this PR pre-review, since I wanted to get the DID cleanup work in.  I supposed I probably could have done some git surgery to separate them again but it didn't seem worth it.  Anyway thanks for taking a look

PR #1194, opened 2021-04-03
js-ipfs dependency updates, now with full types (WIP)

rvagg:
This is against an RC of the ipfs suite of packages which have full typing compiled and exported. Using this to test the RC for now but hopefully this will be a helpful addition when complete!

@achingbrain the things that stand out here are:

* The `JSIPFS` type in ipfs-http-client, defined as `IPFS & { ipld: IPLD, libp2p: libp2p }`, is maybe too broad? I'm not sure what to do with this for `new HttpApi()` and I see we even still `@ts-ignore` those calls in tests too.
* Probably need some integration work between dag-jose and the latest multiformats update _and_ probably needs https://github.com/multiformats/js-multiformats/issues/67 to be done
* `IPFS.create()` config `{ pubsub: { enabled: boolean } }` did that go away or do we have typing wrong for `enabled`?

### DIFF
--- a/packages/core/src/dispatcher.ts
+++ b/packages/core/src/dispatcher.ts
@@ -61,7 +61,8 @@ export class Dispatcher {
    * @param cid - Commit CID
    */
   async retrieveCommit (cid: CID | string): Promise<any> {
-    const record = await this._ipfs.dag.get(cid, { timeout: IPFS_GET_TIMEOUT })
+    const asCid = typeof cid === 'string' ? new CID(cid) : cid
+    const record = await this._ipfs.dag.get(asCid, { timeout: IPFS_GET_TIMEOUT })
     await this._restrictRecordSize(cid)
     return cloneDeep(record.value)
   }
@@ -72,7 +73,8 @@ export class Dispatcher {
    * @param path - optional IPLD path to load, starting from the object represented by `cid`
    */
   async retrieveFromIPFS (cid: CID | string, path?: string): Promise<any> {
-    const record = await this._ipfs.dag.get(cid, { timeout: IPFS_GET_TIMEOUT, path })
+    const asCid = typeof cid === 'string' ? new CID(cid) : cid
+    const record = await this._ipfs.dag.get(asCid, { timeout: IPFS_GET_TIMEOUT, path })
     return cloneDeep(record.value)
   }
 
--- a/packages/core/src/pubsub/pubsub.ts
+++ b/packages/core/src/pubsub/pubsub.ts
@@ -4,7 +4,9 @@ import { IpfsApi } from '@ceramicnetwork/common';
 import { map, catchError, mergeMap, withLatestFrom } from 'rxjs/operators';
 import { IncomingChannel, filterExternal, IPFSPubsubMessage } from './incoming-channel';
 import { DiagnosticsLogger, ServiceLogger } from '@ceramicnetwork/common';
-import uint8ArrayFromString from 'uint8arrays/from-string';
+import { TextDecoder } from 'util';
+
+const textDecoder = new TextDecoder('utf-8')
 
 /**
  * Deserialize incoming message in an internal observable that does not emit if error happens.
@@ -26,7 +28,7 @@ function ipfsToPubsub(
         map((incoming) => {
           const message = deserialize(incoming);
           const serializedMessage = serialize(message);
-          const logMessage = { ...incoming, ...JSON.parse(serializedMessage) };
+          const logMessage = { ...incoming, ...JSON.parse(textDecoder.decode(serializedMessage)) };
           delete logMessage.key;
           delete logMessage.signature;
           pubsubLogger.log({ peer: peerId, event: 'received', topic: topic, message: logMessage });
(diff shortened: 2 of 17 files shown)


packages/ipfs-daemon/src/ipfs-daemon.ts
```diff
@@ -91,19 +89,20 @@ export class IpfsDaemon {
...
-                    },
+                dht: {
+                    enabled: false,
+                    clientMode: !configuration.ipfsDhtServerMode,
+                    randomWalk: false,
+                },
+                pubsub: {
+                    // TODO: @rvagg is this gone?
+                    // @ts-ignore
+                    enabled: configuration.ipfsEnablePubsub,
```
oed, line 105: Were we doing this wrong or did it change?

packages/ipfs-topology/src/ipfs-topology.ts
```diff
@@ -52,7 +51,7 @@ export class IpfsTopology {
   ) {}
 
   async forceConnection(): Promise<void> {
-    const base: string[] = BASE_BOOTSTRAP_LIST[this.ceramicNetwork] || [];
+    const base: Multiaddr[] = BASE_BOOTSTRAP_LIST[this.ceramicNetwork] || [];
```
oed, line 54: Is there a change needed to `BASE_BOOTSTRAP_LIST` as well?

oed (commented):
Thanks @rvagg! Glad to see the ipfs types working better.

> Probably need some integration work between dag-jose and the latest multiformats update and probably needs multiformats/js-multiformats#67 to be done

Right, the dag-jose codec has not been updated in quite a while. It could definitely use an update to more recent packages (multiformats, cids, cbor).

rvagg:
@oed I'd like to have a go at updating dag-jose, it'll be good validation of work we've been doing around multiformats and codecs recently, and we want to get the types right (tho the `multiformats/legacy` connection to js-ipfs is currently not properly done). The rest of the issues here will have to use more expert help from @achingbrain who's been deep in the js-ipfs types and knows much more about the API changes over time than I.

oed:
@rvagg sounds great! Let us know if you run into any problems along the way.

packages/core/src/pubsub/pubsub.ts
```diff
@@ -71,7 +72,7 @@ export class Pubsub extends Observable<PubsubMessage> {
     return this.peerId$
       .pipe(
         mergeMap(async (peerId) => {
-          await this.ipfs.pubsub.publish(this.topic, serialize(message));
+          await this.ipfs.pubsub.publish(this.topic, uint8ArrayFromString(serialize(message)));
```
ukstv, line 75: Guess, this to uint8array convertation should be put into `serialise` function @rvagg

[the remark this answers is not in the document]
ukstv: @oed It's already present https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/core/src/pubsub/pubsub-message.ts#L89

ukstv:
Concerned about dag-jose codec (does it work with types enabled?) and typed `Multiaddr` in bootstrap list. For the latter assumption was that strings are passed.

packages/core/src/dispatcher.ts
```diff
@@ -61,7 +61,8 @@ export class Dispatcher {
    * @param cid - Commit CID
    */
   async retrieveCommit (cid: CID | string): Promise<any> {
-    const record = await this._ipfs.dag.get(cid, { timeout: IPFS_GET_TIMEOUT })
+    const asCid = typeof cid === 'string' ? new CID(cid) : cid
+    const record = await this._ipfs.dag.get(asCid, { timeout: IPFS_GET_TIMEOUT })
```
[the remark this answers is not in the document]
rvagg: yes, but actually probably shouldn't! https://github.com/ipfs/js-ipfs/issues/3637

packages/ipfs-daemon/src/ipfs-daemon.ts
rvagg: @vasco-santos can you answer this one? I can't get to the bottom of the change here. `dht` and `pubsub` have been lifted out of `config` to the top level in the `libp2p` config options. Was that always wrong here, or a recent change, or a problem with types?

packages/ipfs-topology/src/ipfs-topology.ts
rvagg: Probably shouldn't be so restrictive, opened an issue @ https://github.com/ipfs/js-ipfs/issues/3638

packages/ipfs-daemon/src/ipfs-daemon.ts
vasco-santos: This is still inside the config, so this should not really be changed. Were there type errors configuring inside the config property?

vasco-santos: The in progress branch of `js-ipfs` with the types improvements also has it inside the config, so the types should be correct upstream: https://github.com/ipfs/js-ipfs/blob/chore/update-buffer-deps/packages/ipfs-core/src/runtime/libp2p-nodejs.js#L42

rvagg:
rebased to master and added some minor changes, still a few open questions about the js-ipfs core types

packages/ipfs-daemon/src/ipfs-daemon.ts
rvagg: See ipfs-core/src/types.d.ts and how it pulls in `Libp2pConfig` rather than `Libp2pOptions`. I _think_ this is the source of the problem but you'll have to confirm. https://github.com/ipfs/js-ipfs/issues/3640

rvagg:
continued in #1337, because that's a better number, obviously! thanks @achingbrain

PR #1195, opened 2021-04-03
feat(streamid): Rename DocID to StreamID

oed:
This is a WIP. For now only renames the package and internals. This needs to be used by all other packages.


oed:
Just updated this to use the `@ceramicnetwork/streamid` thoughout all packages.

Tried to make it fully backwards compatible, but in particular I'd like @ukstv and @valmack to have a look at the changes around state management and logging!

packages/cli/src/ceramic-daemon.ts
```diff
@@ -218,52 +226,76 @@ class CeramicDaemon {
...
   }
 
   /**
    * Apply one commit to the existing document
    */
   async applyCommit (req: Request, res: Response): Promise<void> {
-    const { docId, commit, docOpts } = req.body
-    if (!(docId && commit)) {
-      throw new Error('docId and commit are required in order to apply commit')
+    const { streamId, commit, docOpts } = req.body
```
ukstv, line 274:
```suggestion
    const { streamId, docId, commit, docOpts } = req.body
```

ukstv (approved):
🏇 One suggestion for backwards compatibility of `applyCommit` endpoint.

oed:
@valmack will the renaming of the `docid.log` file be a problem?

valmack:
> @valmack will the renaming of the `docid.log` file be a problem?

Not a problem. Requires some small updates in `ceramic-stats` repo

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -150,7 +150,7 @@ export default class EthereumAnchorService implements AnchorService {
       fetch(this.requestsApiEndpoint, {
         method: "POST",
         body: JSON.stringify({
-          docId: cidDocPair.docId.toString(),
+          streamId: cidDocPair.streamId.toString(),
```
stbrody, line 153: Can we hold off on updating the request/response to/from the CAS right now?  Otherwise we need to update and deploy the CAS this sprint as well and I'm worried about how quickly the tail end of this project is growing.  We should be able to safely update the CAS apis after ELP launch right?

oed: For sure, and good catch. I didn't think about this!

stbrody (approved):
LGTM, thanks!

PR #1196, opened 2021-04-04
feat(core, http-client): Remove default DID Resolver and make apps provide via CeramicAPI.setDID()

stbrody:
This breaks a circular dependency between the http-client and the 3id-did-resolver

### DIFF
--- a/packages/3id-did-resolver/generateVectorDocuments.js
+++ b/packages/3id-did-resolver/generateVectorDocuments.js
@@ -11,6 +11,10 @@
 
 const Ceramic = require('@ceramicnetwork/http-client').default
 const ThreeIdProvider = require('3id-did-provider').default
+const ThreeIdResolver = require('@ceramicnetwork/3id-did-resolver').default
+const KeyDidResolver = require('key-did-resolver').default
+const { Resolver } = require('did-resolver')
+const { DID } = require('dids')
 const TileDoctype = require('@ceramicnetwork/doctype-tile').default
 const u8a = require('uint8arrays')
 const { randomBytes } = require('@stablelib/random')
@@ -77,14 +81,24 @@ const waitForAnchor = doc => new Promise(resolve => {
   })
 })
 
+const makeDID = function(provider, ceramic) {
+  const keyDidResolver = KeyDidResolver.getResolver()
+  const threeIdResolver = ThreeIdResolver.getResolver(ceramic)
+  const resolver = new Resolver({
+    ...threeIdResolver, ...keyDidResolver,
+  })
+  return new DID({ provider, resolver })
+}
+
 const setLegacyDoc = async (ceramic, doc, keyset) => {
   const signing = encodeKey(keyset.signingPub, 'secp256k1')
   const encryption = encodeKey(keyset.encPub, 'x25519')
   const oldProvider = ceramic.did._client.connection
   const provider = new Ed25519Provider(keyset.seed)
-  await ceramic.setDIDProvider(provider)
-  const didstr = ceramic.did.id
-  await ceramic.setDIDProvider(oldProvider)
+  const didWithProvider = makeDID(provider, ceramic)
+  const didWithOldProvider = makeDID(oldProvider, ceramic)
+  const didstr = didWithProvider.id
+  await ceramic.setDID(didWithOldProvider)
 
   await doc.update(
       { publicKeys: {
@@ -93,7 +107,7 @@ const setLegacyDoc = async (ceramic, doc, keyset) => {
         }},
       { controllers: [didstr] })
 
-  await ceramic.setDIDProvider(provider)
+  await ceramic.setDID(didWithProvider)
   await waitForAnchor(doc)
 }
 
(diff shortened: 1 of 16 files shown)


packages/3id-did-resolver/generateVectorDocuments.js
```diff
@@ -77,14 +81,24 @@ const waitForAnchor = doc => new Promise(resolve => {
...
+
 const setLegacyDoc = async (ceramic, doc, keyset) => {
   const signing = encodeKey(keyset.signingPub, 'secp256k1')
   const encryption = encodeKey(keyset.encPub, 'x25519')
   const oldProvider = ceramic.did._client.connection
   const provider = new Ed25519Provider(keyset.seed)
-  await ceramic.setDIDProvider(provider)
-  const didstr = ceramic.did.id
-  await ceramic.setDIDProvider(oldProvider)
+  const didWithProvider = makeDID(provider, ceramic)
```
stbrody, line 98: The changes in this file were done manually and were not tested.  May be worth testing @oed to make sure I didn't break anything.

I want to state for the record that I don't like the way this `generateVectorDocuments.js` is set up right now.  I feel like if there is any sort of auto-generated files as part of your build or test process, the auto-generation should happen every time as part of the standard build and test process so that it remains current.  A "run once" system where you're supposed to re-run manually at certain special moments (that might not even be clear to everyone when is supposed to happen) seems destined to bitrot and cause headaches down the road.

oed: I just included this script for completeness in case something goes very wrong. We should never need to regenerate them, if we need to it means that we broke the http api in a way we are not supposed to break it.

packages/core/src/__tests__/ceramic-anchor.test.ts
```diff
@@ -11,20 +11,34 @@ import { createIPFS, swarmConnect } from './ipfs-util';
...
+const makeDID = function(seed: Uint8Array, ceramic: Ceramic): DID {
+  const provider = new Ed25519Provider(seed)
+
+  const keyDidResolver = KeyDidResolver.getResolver()
+  const threeIdResolver = ThreeIdResolver.getResolver(ceramic)
+  const resolver = new Resolver({
+    ...threeIdResolver, ...keyDidResolver,
+  })
+  return new DID({ provider, resolver })
+}
```
oed, line 32: I guess this could be made into a util somewhere.

packages/core/src/ceramic.ts
```diff
@@ -465,17 +448,16 @@ class Ceramic implements CeramicApi {
...
-  async setDIDProvider (provider: DIDProvider): Promise<void> {
-    this.context.provider = provider;
-    this.context.did = new DID({ provider, resolver: this.context.resolver })
-
-    if (!this.context.did.authenticated) {
-      await this.context.did.authenticate()
+  async setDID(did: DID): Promise<void> {
+    this.context.did = did
+    if (!did.authenticated) {
+      await this.did.authenticate()
```
oed, line 458: I think we should avoid doing this here. Instead we can do it inside of the *TileDocument* class on demand.

packages/http-client/src/ceramic-http-client.ts
```diff
@@ -230,12 +216,10 @@ export default class CeramicClient implements CeramicApi {
...
-  async setDIDProvider(provider: DIDProvider): Promise<void> {
-    this.context.provider = provider;
-    this.context.did = new DID( { provider, resolver: this.context.resolver })
-
-    if (!this.context.did.authenticated) {
-      await this.context.did.authenticate()
+  async setDID(did: DID): Promise<void> {
+    this.context.did = did
+    if (!did.authenticated) {
+      await this.did.authenticate()
```
oed, line 222: same thing here.

oed (changes requested):
Looks good. Main thing is that we should change where we do the authentication step.

packages/3id-did-resolver/generateVectorDocuments.js
stbrody: What about when we update to V1 of the http api?  It all just seems very brittle to me.  But definitely not high priority to change right now

packages/core/src/__tests__/ceramic-anchor.test.ts
stbrody: Yeah agreed.  I wanted to do that here but just wasn't sure the best place to put it to avoid re-introducing dependency cycles.  See this discord thread I just started: https://discord.com/channels/484729862368526356/638777352402239538/828334185658843166

packages/core/src/ceramic.ts
```diff
@@ -454,10 +454,7 @@ class Ceramic implements CeramicApi {
    */
   async setDID(did: DID): Promise<void> {
     this.context.did = did
-    if (!did.authenticated) {
-      await this.did.authenticate()
-    }
-    this._logger.imp(`Now authenticated as DID ${did}`)
+    this._logger.imp(`Now using DID ${did}`)
```
oed, line 457: Not sure if this is really needed, but we should log when authenticating in the other places (then log `did.id` instead of `did`)

oed (approved):
Looks good, mod one comment

packages/core/src/ceramic.ts
stbrody: okay will do, but it's kind of an a big change so I'm going to do it as a follow-up on top of this and https://github.com/ceramicnetwork/js-ceramic/pull/1197 to minimize rebase conflicts

PR #1197, opened 2021-04-04
chore(common,doctype-tile): Remove dead code related to old createDocument API

stbrody:
Builds on https://github.com/ceramicnetwork/js-ceramic/pull/1196. Breaking the circular dependency between the http-client and the 3id-did-resolver eliminated the weird compile error that was preventing this from happening previously.

### DIFF
--- a/packages/doctype-caip10-link/src/caip10-link-doctype.ts
+++ b/packages/doctype-caip10-link/src/caip10-link-doctype.ts
@@ -3,9 +3,10 @@ import {
   DoctypeConstructor,
   DoctypeStatic,
   DocOpts,
-  DocParams,
   Context,
-  CeramicCommit, UnsignedCommit,
+  CeramicCommit,
+  UnsignedCommit,
+  DocMetadata,
 } from '@ceramicnetwork/common';
 
 export const DOCTYPE_NAME = 'caip10-link'
@@ -13,8 +14,16 @@ export const DOCTYPE_NAME = 'caip10-link'
 /**
  * Caip10Link parameters
  */
-export interface Caip10LinkParams extends DocParams {
+export interface Caip10LinkParams {
+    metadata?: DocMetadata
+    /**
+     * deterministic is a tri-state. True means means always create the document deterministically,
+     * false means always force the document to be unique, undefined means use the default behavior
+     */
+    deterministic?: boolean
     content?: Record<string, unknown>;
+
+    [index: string]: any // allow arbitrary properties
 }
 
 const throwReadOnlyError = (): Promise<void> => {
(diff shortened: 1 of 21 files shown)


packages/doctype-caip10-link/src/caip10-link-doctype.ts
```diff
@@ -3,18 +3,27 @@ import {
...
+  DocMetadata,
 } from '@ceramicnetwork/common';
 
 export const DOCTYPE_NAME = 'caip10-link'
 
 /**
  * Caip10Link parameters
  */
-export interface Caip10LinkParams extends DocParams {
+export interface Caip10LinkParams {
```
stbrody, line 17: Temporarily folding the fields from `DocParams` into this. My next PR will remove this type entirely as part of revamping the whole Caip10Link API

packages/doctype-tile/src/tile-doctype.ts
```diff
@@ -255,4 +180,41 @@ export class TileDoctype<T = Record<string, any>> extends Doctype {
...
 
+    /**
+     * Create genesis commit.
+     * Placed on the TileDoctype class as a static method only for use in testing.
+     * @param did - DID making (and signing) the commit
+     * @param content - genesis content
+     * @param metadata - genesis metadata
+     * @private
+     */
+    static async _makeGenesis<T>(did: DID, content: T | null, metadata?: TileMetadataArgs): Promise<CeramicCommit> {
```
oed, line 191: Any reason to keep this @private? Seems like it could be a useful function in general.

stbrody: Yeah that makes sense.  Updated

oed (approved):
Ty

packages/doctype-tile/src/tile-doctype.ts
```diff
@@ -264,4 +183,42 @@ export class TileDoctype<T = Record<string, any>> extends Doctype {
         return _signDagJWS(commit, did, this.controllers[0])
     }
 
+    /**
+     * Create genesis commit.
+     * @param did - DID making (and signing) the commit
+     * @param content - genesis content
+     * @param metadata - genesis metadata
+     */
+    static async makeGenesis<T>(did: DID, content: T | null, metadata?: TileMetadataArgs): Promise<CeramicCommit> {
```
ukstv, line 192: @stbrody I am too late to the party, but maybe it returns `Promise<SignedCommitContainer>`

stbrody: I'll see if I can do that in a follow up, but I think it'd need to be `Promise<SignedCommit|SignedCommitContainer>` since the genesis commit can come back unsigned if there's no content

ukstv: Oomps, did not know that. My suggestion then does not make sense.

PR #1199, opened 2021-04-04
feat(doctype-tile): Log when DID is authenticated

stbrody:
Also changes TileDoctype's `makeCommit` and `makeGenesis` methods to take a `CeramicAPI` rather than taking the `DID` directly


packages/doctype-tile/src/tile-doctype.ts
```diff
@@ -162,12 +172,11 @@ export class TileDoctype<T = Record<string, any>> extends Doctype {
...
     /**
      * Make a commit to update the document
-     * @param did - DID making (and signing) the commit
+     * @param ceramic - Ceramic API instance containing the DID making (and signing) the commit
      * @param newContent
      * @param newMetadata
-     * @private
      */
-    async _makeCommit(did: DID, newContent: T | undefined, newMetadata?: TileMetadataArgs): Promise<CeramicCommit> {
+    async makeCommit(ceramic: CeramicApi, newContent: T | undefined, newMetadata?: TileMetadataArgs): Promise<CeramicCommit> {
```
stbrody, line 179: I figured if `makeGenesis` is public then `makeCommit` aught to be also

oed:
Ugh, not sure how I feel about this actually. It's kind of nice to be able to create commits without having access to the CeramicApi. Is there any way we can just optionally pass a logger?

stbrody:
> Ugh, not sure how I feel about this actually. It's kind of nice to be able to create commits without having access to the CeramicApi. Is there any way we can just optionally pass a logger?

What's the use case for creating a GenesisCommit without access to a CeramicApi?  What would you do with it?

While I agree it's not totally ideal, this approach seems better than the alternatives.  Just adding an optional logger param isn't great because A) what's the incentive for apps to go through the work to figure out how to provide the logger on each call?  I feel like most apps will just not bother in which case there's no point in doing the logging at all.  And B) it's not very extensible.  If we wind up needing access to some other internal component in order to generate commits in the future, we'll need to add another argument to the `createGenesis` method.  The old version of `makeGenesis` took in a handle to the `Context` object.  We could do that again, but I like that less.  I'm hoping to get the `Context` entirely out of the public-facing API so it can just be an implementation detail.  I feel like the future-proof handle that should be passed around everywhere and gets us access to any internal components needed is the `CeramicApi`, and so we should lean toward every static method taking one as its first arg, and every object returned to the application storing a handle to one internally.

oed:
Ping @PaulLeCam you might have some relevant context about how to construct these apis ☝️

PaulLeCam:
Looks like the main reason for this change is we need the logger here (https://github.com/ceramicnetwork/js-ceramic/pull/1199/files#diff-39fb467a29692ce95079e3ff4bf4eb09af506d263308c047ca7edd49610ced3fR59-R74) no?
If so, maybe we can add a `ensureAuthenticatedDID(did: DID)` function and update `makeGenesis(ceramicOrDID: CeramicApi | DID, ...)` with associated logic? This way whenever we have a CeramicApi we can use it directly, but in cases we just want to create genesis commits using a DID it's still possible.

stbrody:
> Looks like the main reason for this change is we need the logger here ([#1199 (files)](https://github.com/ceramicnetwork/js-ceramic/pull/1199/files#diff-39fb467a29692ce95079e3ff4bf4eb09af506d263308c047ca7edd49610ced3fR59-R74)) no?
> If so, maybe we can add a `ensureAuthenticatedDID(did: DID)` function and update `makeGenesis(ceramicOrDID: CeramicApi | DID, ...)` with associated logic? This way whenever we have a CeramicApi we can use it directly, but in cases we just want to create genesis commits using a DID it's still possible.

Yes, we could do that, though I'm still not totally convinced it's worth it?  It makes the API a bit more complicated to understand - now users will look at it and wonder why the first argument can be a `CeramicApi` or a `DID` which at first glance seem very different.  Also doesn't really solve the problem of future-proofing the API if we don't actually require the `CeramicAPI` arg.

If we feel strongly that requiring a `CeramicAPI` to be able to generate genesis commits is a problem then I can try going down that path, but I still don't really understand the use case we're trying to appeal to, or think it's that's onerous to require users to have a `CeramicAPI` in order to do operations with Ceramic?  In mind it makes the most sense to be consistent to always require a CeramicAPI for all public APIs. Gives us a consistent and intuitive developer experience, and helps future proof us since we know anything we'll ever need will always be available via the CeramicAPI

PaulLeCam:
Yeah I agree it makes APIs more complicated than necessary.
I think part of the issue is that going from just providing a DID to a full `CeramicAPI` interface seems like quite a gap. I think it makes sense to provide a Ceramic instance when interacting with the network but just to provide a DID that feels strange, especially since the DID may not even be present on the Ceramic instance, it creates a level of indirection that is kind of misleading.

I like the abstractions in [ethers.js](https://docs.ethers.io/v5/api/) with the separation between signers and providers, to me that could similarly apply to Ceramic so we could have interfaces like:

```ts
interface CeramicCommon {
  logger?: CeramicLogger
}

interface CeramicSigner extends CeramicCommon {
  did: DID
}

class CeramicAuthenticatedClient implements CeramicSigner {
  ...
}
```

Then we can have `makeGenesis(signer: CeramicSigner, ...)` where the signer can be a full Ceramic client, but can also be a simple object with the DID and optionally the logger?

stbrody:
> Then we can have `makeGenesis(signer: CeramicSigner, ...)` where the signer can be a full Ceramic client, but can also be a simple object with the DID and optionally the logger?

I like this.  I'll give it a shot.  Thanks for the suggestion!

stbrody:
okay @oed, new version up using the new CeramicSigner interface

packages/doctype-tile/src/tile-doctype.ts
```diff
@@ -31,6 +32,21 @@ export interface TileMetadataArgs {
...
+interface CeramicCommon {
+    logger?: DiagnosticsLogger
+}
+
+/**
+ * Interface for an object that contains a DID that can be used to sign Ceramic commits.
+ * Any implementation of CeramicAPI will match this interface, though if no CeramicAPI instance is
+ * available users can provide any object containing an authenticated DID instance.
+ */
+interface CeramicSigner extends CeramicCommon {
```
stbrody, line 44: I left this unexported and local to tile-doctype.ts for now.  I figure we can move it if/when we need it somewhere else.  If you'd like me to move it somewhere more reusable now though just lmk

stbrody: I changed my mind on this and moved it into `common` so that `CeramicApi` can extend `CeramicSigner` to make it super clear that a `CeramicApi` instance can be used anywhere that asks for a `CeramicSigner`

stbrody:
Merged in develop to make tests pass, and switched to using LoggerProvider instead of DiagnosticsLogger directly to future proof things more, and be compatible with another PR Sergey has in flight.

Anyway I think this should be ready for a real review now @oed

packages/common/src/ceramic-api.ts
```diff
@@ -35,12 +36,28 @@ export interface PinApi {
...
+}
+
 /**
  * Describes Ceramic node API
  */
-export interface CeramicApi {
+export interface CeramicApi extends CeramicSigner {
     pin: PinApi;
-    did?: DID;
+    did: DID | null;
```
oed, line 59: I don't think you need to add this here since it's already in `CeramicSigner`.

packages/doctype-tile/src/tile-doctype.ts
```diff
@@ -56,22 +56,28 @@ function headerFromMetadata(metadata?: TileMetadataArgs | DocMetadata): CommitHe
     return header
 }
 
+async function _ensureAuthenticated(signer: CeramicSigner) {
+    if (signer.did == null) {
+        throw new Error('No DID authenticated')
```
oed, line 61:
```suggestion
        throw new Error('No DID provided')
```

oed (approved):
lgtm

PR #1200, opened 2021-04-05
fix(cli): set a DID with resolvers in daemon

oed:
When running the cli daemon no did resolvers where provided.


stbrody (approved):
Oops, sorry about that, should have tested the daemon before merging.  Thanks for fixing

PR #1202, opened 2021-04-05
feat(core): Limit total number of the tasks executed concurrently

ukstv:
Add one more config parameter: `concurrentTasksLimit`. By default it is the same as `docCacheLimit`.
Sets the upper bound of concurrently processing documents.

### DIFF
--- a/packages/core/src/ceramic.ts
+++ b/packages/core/src/ceramic.ts
@@ -95,6 +95,7 @@ export interface CeramicConfig {
   pubsubTopic?: string;
 
   docCacheLimit?: number;
+  concurrentTasksLimit?: number;
   cacheDocCommits?: boolean; // adds 'docCacheLimit' additional cache entries if commits can be cached as well
 
   useCentralizedPeerDiscovery?: boolean;
@@ -126,6 +127,7 @@ export interface CeramicModules {
 export interface CeramicParameters {
   cacheDocumentCommits: boolean,
   docCacheLimit: number,
+  concurrentTasksLimit: number,
   networkOptions: CeramicNetworkOptions,
   supportedChains: string[],
   validateDocs: boolean,
@@ -374,15 +376,17 @@ class Ceramic implements CeramicApi {
     }
 
     const docCacheLimit = config.docCacheLimit ?? DEFAULT_DOC_CACHE_LIMIT
+    const concurrentTasksLimit = config.concurrentTasksLimit ?? docCacheLimit
 
     const ipfsTopology = new IpfsTopology(ipfs, networkOptions.name, logger)
     const pinStoreFactory = new PinStoreFactory(ipfs, pinStoreOptions)
-    const repository = new Repository(docCacheLimit, logger)
+    const repository = new Repository(docCacheLimit, logger, concurrentTasksLimit)
     const dispatcher = new Dispatcher(ipfs, networkOptions.pubsubTopic, repository, logger, pubsubLogger)
 
-    const params = {
+    const params: CeramicParameters = {
       cacheDocumentCommits: config.cacheDocCommits ?? true,
       docCacheLimit: docCacheLimit,
+      concurrentTasksLimit: concurrentTasksLimit,
       networkOptions,
       supportedChains,
       validateDocs: config.validateDocs ?? true,
--- a/packages/core/src/ceramic.ts
+++ b/packages/core/src/ceramic.ts
@@ -483,8 +483,7 @@ class Ceramic implements CeramicApi {
    */
   async applyCommit<T extends Doctype>(docId: string | DocID, commit: CeramicCommit, opts: CreateOpts | UpdateOpts = {}): Promise<T> {
     opts = { ...DEFAULT_APPLY_COMMIT_OPTS, ...opts };
-    const state$ = await this._loadDoc(normalizeDocID(docId), opts as CreateOpts)
-    await this.repository.stateManager.applyCommit(state$, commit, opts)
+    const state$ = await this.repository.stateManager.applyCommit(normalizeDocID(docId), commit, opts as CreateOpts)
     return doctypeFromState<T>(this.context, this._doctypeHandlers, state$.value, this.repository.updates$)
   }
 
--- a/packages/core/src/ceramic.ts
+++ b/packages/core/src/ceramic.ts
@@ -95,7 +95,7 @@ export interface CeramicConfig {
   pubsubTopic?: string;
 
   docCacheLimit?: number;
-  concurrentTasksLimit?: number;
+  concurrentRequestsLimit?: number;
   cacheDocCommits?: boolean; // adds 'docCacheLimit' additional cache entries if commits can be cached as well
 
   useCentralizedPeerDiscovery?: boolean;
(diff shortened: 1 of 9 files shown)


packages/core/src/state-management/execution-queue.ts
```diff
@@ -44,15 +51,15 @@ export class ExecutionQueue {
         return this.tasks.add(docId.toString(), async () => {
           const doc = await this.get(docId);
```
stbrody, line 52: Should the semaphore be held around the call to `.get()`?  `get` might have to do real i/o after all

packages/core/src/state-management/repository.ts
```diff
@@ -50,12 +50,18 @@ export class Repository {
    */
   stateManager: StateManager;
 
-  constructor(limit: number, private readonly logger: DiagnosticsLogger) {
+  /**
+   * @param cacheLimit - Maximum number of documents to store in memory cache.
+   * @param logger - Where we put diagnostics messages.
+   * @param concurrencyLimit - Maximum number of concurrently running tasks on the documents.
+   */
+  constructor(cacheLimit: number, private readonly logger: DiagnosticsLogger, concurrencyLimit?: number) {
```
stbrody, line 58: Any reason for `concurrencyLimit` to be optional?  I generally prefer to avoid optional parameters in internal-only APIs so that behavior is clearer at the call site and developers have to consider all available options whenever calling something

packages/core/src/state-management/repository.ts
```diff
@@ -50,12 +50,18 @@ export class Repository {
...
     this.loadingQ = new NamedTaskQueue((error) => {
       logger.err(error);
     });
-    this.executionQ = new ExecutionQueue(logger, (docId) => this.get(docId));
-    this.inmemory = new StateCache(limit, (state$) => state$.complete());
+    const effectiveConcurrencyLimit = cacheLimit || concurrencyLimit;
```
stbrody, line 62: Should this be the `min()`?  What if concurrencyLimit is less than cacheLimit?

stbrody (approved):
LGTM
Left some suggestions inline, but consider them all optional.

Only thing that really matters to figure out is the name of the new param, since that's harder to change once it's out there.  CC @oed in case you have any thoughts on `concurrentTasksLimit` vs `concurrentRequestsLimit` vs something else

packages/core/src/ceramic.ts
```diff
@@ -121,6 +122,7 @@ export interface CeramicModules {
 export interface CeramicParameters {
   cacheDocumentCommits: boolean,
   docCacheLimit: number,
+  concurrentTasksLimit: number,
```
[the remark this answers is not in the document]
ukstv: Okay. Was following `docCacheLimit` lead :)

packages/core/src/state-management/execution-queue.ts
ukstv: True. My bad.

packages/core/src/ceramic.ts
```diff
@@ -90,6 +90,7 @@ export interface CeramicConfig {
   pubsubTopic?: string;
 
   docCacheLimit?: number;
+  concurrentTasksLimit?: number;
```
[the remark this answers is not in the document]
ukstv: +1 for suggestions.

A bit torn between requests and tasks. `request` has strong connotations to network and I/O, and this is good, but tasks are all over internal code: in the execution queue, in doc-comments, in NamedTaskQueue, so maybe this would serve as an argument for tasks? (Not a strong opinion)

packages/core/src/state-management/repository.ts
ukstv: These two limits are orthogonal, really like dimensions. If concurrencyLimit is less than cacheLimit, perceived node performance would suffer, as only `concurrencyLimit` documents can be _processed_ concurrently. Other `cacheLimit - concurrencyLimit` documents would sit idle waiting for their turn.

packages/core/src/state-management/repository.ts
ukstv: Yup. Concur here.

packages/core/src/state-management/repository.ts
stbrody: Yep, but in that case isn't that the behavior the user asked for?  What's the rationale for always choosing `cacheLimit` over `concurrencyLimit`?  Feels like that makes the `concurrencyLimit` param meaningless except when you have an infinite sized cache (I'm assuming that `cacheLimit` of 0 means "no limit"?  otherwise the `concurrenclyLimit` would literally never be used)

I could imagine a scenario where running with a lower concurrencyLimit than cacheLimit could be valuable if for example you really want to avoid going to pubsub by keeping lots of docs in cache, but don't want a large amount processing in parallel

stbrody: Actually, instead of `min` or `||`, should we just not be considering `cacheLimit` here at all and just always take the `concurrencyLimit` as given?  There's already logic at a higher level to default `concurrencyLimit` to `cacheLimit` if `concurrencyLimit` isn't explicitly specified.  So maybe if the user configures an explicit `concurrencyLimit` we should just use it, regardless of what the `cacheLimit` is?

ukstv: Holy cow. It should be `concurrencyLimit || cacheLimit`. And not even here, since this is set during Ceramic init process.

packages/core/src/ceramic.ts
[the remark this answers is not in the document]
stbrody: My assumption is that most users won't read the internal code, so having a mismatch between the user-visible names and the internal names doesn't bother me as much as making sure that the public-facing names make the most sense.  So just considering the name and its behavior completely independently of the implementation details, "requests" seems slightly better to me. But it's also not a super strong opinion for me either.

packages/core/src/state-management/execution-queue.ts
ukstv: Reduced responsibility of executionQ to just concurrency control. Can not make it accommodate for `Repository#get` and `Repository#load`.

packages/core/src/state-management/state-manager.ts
```diff
@@ -30,6 +30,8 @@ export class StateManager {
     public anchorService: AnchorService,
     public conflictResolution: ConflictResolution,
     private readonly logger: DiagnosticsLogger,
+    private readonly get: (docId: DocID) => Promise<RunningState | undefined>,
```
stbrody, line 33: Would be nice to have a comment clarifying what `get` and `load` are for and how they're different

stbrody: Maybe just renaming it to `getFromRepository` would make it clearer?

packages/core/src/state-management/execution-queue.ts
```diff
@@ -45,22 +23,16 @@ export class ExecutionQueue {
   /**
    * Return execution lane for a document.
    */
-  forDocument(docId: DocID): ExecutionLane {
+  forDocument(docId: DocID): TaskQueueLike {
```
stbrody, line 26: Yeah, this is much simpler to understand now.  Thanks!

stbrody (approved):
Great cleanups!  Still LGTM.

One small suggestion around clarifying the role of the functions passed to the StateManager

packages/core/src/state-management/state-manager.ts
ukstv: `fromMemoryOrStore` ?

stbrody: sure, that works

ukstv: Done

PR #1205, opened 2021-04-05
feat(doctype-caip10-link): Add new API methods to Caip10LinkDoctype

stbrody:
I can't easily test this or remove the old API methods until this is merged and an RC release made so that the `doctype-caip10-link-handler` package can be updated to depend on the new version.


stbrody:
Opened prematurely, closing for now

PR #1208, opened 2021-04-06
fix(key-did-resolver): Unbreak tests by pinning to specific versions of multibase and uint8arrays

stbrody:


oed (approved):

stbrody:
FYI this PR doesn't seem to be sufficient to resolve the issue.  I've updated every single package.json that mentions `multibase` or `uint8arrays`, but it looks like we're still picking up the newer version of `multibase` via `transitive` dependencies from `cids`.

Alex from PL pointed out on https://github.com/achingbrain/uint8arrays/issues/17 that `TextDecoder` should be globally available to all modern node versions, so it's unclear why this is failing in CI as well as on my local machine, both of which use node v14 I believe.

@zachferland @PaulLeCam @ukstv, any ideas what might be going on here?

stbrody:
It seems like I might be able to use `npm-shrinkwrap` to pin `multibase` to the older version even from transitive dependencies, but that seems not ideal if we can avoid it.  Would be better to understand why we the `TextDecoder` error is happening with the new `multibase` in the first place

stbrody:
It looks like this is likely an issue with jest: https://github.com/achingbrain/uint8arrays/issues/17#issuecomment-814393633

::sigh::

stbrody:
Filed https://github.com/ceramicnetwork/js-ceramic/issues/1211.  Looks like that is more likely the proper way to fix this

PaulLeCam:
@stbrody this seems related more specifically to jsdom, have you tried using the node environment for jest or a setup file as suggested in https://stackoverflow.com/questions/57712235/referenceerror-textencoder-is-not-defined-when-running-react-scripts-test?

stbrody:
> have you tried using the node environment for jest or a setup file as suggested in [stackoverflow.com/questions/57712235/referenceerror-textencoder-is-not-defined-when-running-react-scripts-test](https://stackoverflow.com/questions/57712235/referenceerror-textencoder-is-not-defined-when-running-react-scripts-test)?

Is that not what the ticket I filed is describing to do?

PaulLeCam:
Could be just a config option if it works using the node env, cf https://jestjs.io/docs/configuration#testenvironment-string 
jsdom is the default but node is supported out of the box by jest.

stbrody:
> Could be just a config option if it works using the node env, cf [jestjs.io/docs/configuration#testenvironment-string](https://jestjs.io/docs/configuration#testenvironment-string)
> jsdom is the default but node is supported out of the box by jest.

Okay yeah, adding `--env=node` to the `test` script in all the relevant package.json files does fix most of the tests.  The only problem is the `http-client` package.  Using `--env=node` there causes it to fail with a different error:
```
/home/spencer/js-ceramic3/node_modules/rxjs/internal/util/hostReportError.js:4
    setTimeout(function () { throw err; }, 0);
                             ^

TypeError: Only absolute URLs are supported
    at getNodeRequestOptions (/home/spencer/js-ceramic3/node_modules/node-fetch/lib/index.js:1305:9)
    at /home/spencer/js-ceramic3/node_modules/node-fetch/lib/index.js:1410:19
    at new Promise (<anonymous>)
    at fetch (/home/spencer/js-ceramic3/node_modules/node-fetch/lib/index.js:1407:9)
    at fetch (/home/spencer/js-ceramic3/node_modules/cross-fetch/dist/node-ponyfill.js:10:20)
    at fetchJson (/home/spencer/js-ceramic3/packages/http-client/src/utils.ts:561:13)
    at Document._syncState (/home/spencer/js-ceramic3/packages/http-client/src/document.ts:883:17)
    at ThrottleSubscriber.durationSelector (/home/spencer/js-ceramic3/packages/http-client/src/document.ts:812:21)
    at ThrottleSubscriber.Object.<anonymous>.ThrottleSubscriber.tryDurationSelector (/home/spencer/js-ceramic3/node_modules/rxjs/src/internal/operators/throttle.ts:132:19)
    at ThrottleSubscriber.Object.<anonymous>.ThrottleSubscriber.throttle (/home/spencer/js-ceramic3/node_modules/rxjs/src/internal/operators/throttle.ts:124:27)
```

PaulLeCam:
Seems like an issue in the test or even the http-client code itself if there's any logic relying on relative URLs?

stbrody:
Closing in favor of https://github.com/ceramicnetwork/js-ceramic/pull/1212

PR #1210, opened 2021-04-06
Daemon error handling

ukstv:
- LoggerProvider is exposed as a field of Ceramic core node (Context ≠ DI)
- Beautification for async request handlers. This makes all the errors fall to the generic error handler.
- Generic error handler dumps error message to DiagnosticsLogger and response back with 500 error
- Morgan now exposes caught error as `error_message` and `error_code` tokens.
- Drop health checks

Screenshot 2021-04-06 at 21 56 44

### DIFF
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -1,16 +1,13 @@
-import express, { Request, Response, NextFunction } from 'express'
+import express, { Request, Response } from 'express'
 import Ceramic from '@ceramicnetwork/core'
 import type { CeramicConfig} from "@ceramicnetwork/core"
 import { RotatingFileStream } from "@ceramicnetwork/logger"
 import { buildIpfsConnection } from "./build-ipfs-connection.util";
 import { S3StateStore } from "./s3-state-store";
 import {
-  DiagnosticsLogger,
   DoctypeUtils,
-  RootLogger,
   MultiQuery,
   LoggerConfig,
-  LogLevel,
   LoggerProvider,
 } from "@ceramicnetwork/common"
 import { LogToFiles } from "./ceramic-logger-plugins"
@@ -19,8 +16,9 @@ import ThreeIdResolver from '@ceramicnetwork/3id-did-resolver'
 import KeyDidResolver from 'key-did-resolver'
 import { DID } from 'dids'
 import cors from 'cors'
-import * as core from "express-serve-static-core"
-import morgan from 'morgan';
+import { errorHandler } from './daemon/error-handler';
+import { addAsync, ExpressWithAsync } from '@awaitjs/express'
+import { logRequests } from './daemon/log-requests';
 
 const DEFAULT_PORT = 7007
 const toApiPath = (ending: string): string => '/api/v0' + ending
@@ -46,34 +44,11 @@ export interface CreateOpts {
   pubsubTopic?: string;
 }
 
-interface HttpLog {
-  request: Record<string, unknown>;
-  response?: Record<string, unknown>;
-}
-
 interface MultiQueries {
   queries: Array<MultiQuery>
 }
 
-const ACCESS_LOG_FMT = 'ip=:remote-addr ts=:date[iso] method=:method original_url=:original-url base_url=:base-url path=:path http_version=:http-version req_header:req[header] status=:status content_length=:res[content-length] content_type=":res[content-type]" ref=:referrer user_agent=:user-agent elapsed_ms=:total-time[3]';
-
-const makeExpressMiddleware = function (loggerProvider: LoggerProvider) {
-  morgan.token<Request, Response>('original-url', function (req, res): any {
-    return req.originalUrl;
-  });
-  morgan.token<Request, Response>('base-url', function (req, res): any {
-    return req.baseUrl;
-  });
-  morgan.token<Request, Response>('path', function (req, res): any {
-    return req.path;
-  });
-
-  const logger = loggerProvider.makeServiceLogger("http-access")
-
-  return [morgan(ACCESS_LOG_FMT, { stream: logger })]
-};
-
-const makeCeramicConfig = function (opts: CreateOpts): CeramicConfig {
+function makeCeramicConfig (opts: CreateOpts): CeramicConfig {
   const loggerProvider = new LoggerProvider(opts.loggerConfig, (logPath: string) => { return new RotatingFileStream(logPath, true)})
   const ceramicConfig: CeramicConfig = {
     loggerProvider,
@@ -118,62 +93,23 @@ const makeCeramicConfig = function (opts: CreateOpts): CeramicConfig {
  */
 class CeramicDaemon {
   private server: any
-  private readonly logger: DiagnosticsLogger
 
   constructor (public ceramic: Ceramic, opts: CreateOpts) {
-    this.logger = ceramic.context.loggerProvider.getDiagnosticsLogger()
-
-    const app: core.Express = express()
+    const diagnosticsLogger = ceramic.loggerProvider.getDiagnosticsLogger()
+    const app = addAsync(express());
     app.set('trust proxy', true)
     app.use(express.json())
     app.use(cors({ origin: opts.corsAllowedOrigins }))
 
-    const expressMiddleware = makeExpressMiddleware(ceramic.context.loggerProvider)
-    app.use(expressMiddleware)
+    app.use(logRequests(ceramic.loggerProvider))
 
     this.registerAPIPaths(app, opts.gateway)
 
-    const loggerOld = RootLogger.getLogger(CeramicDaemon.name)
-    if (opts.loggerConfig?.logLevel <= LogLevel.debug) {
-      app.use((req: Request, res: Response, next: NextFunction): void => {
-        const requestStart = Date.now()
-
-        let requestError: string = null;
-        let body: string | any = [];
-        req.on("data", chunk => {
-          body.push(chunk)
-        })
-        req.on("end", () => {
-          body = Buffer.concat(body)
-          body = body.toString()
-        });
-        req.on("error", error => {
-          requestError = error.message
-        });
-
-        res.on("finish", () => {
-          const httpLog = this._buildHttpLog(requestStart, req, res, {requestError, body})
-          const logString = JSON.stringify(httpLog)
-          loggerOld.debug(logString)
-        })
-        next()
-      })
-    }
-
-    // next is required in function signature
-    app.use((err: Error, req: Request, res: Response, next: NextFunction): void => {
-      loggerOld.error(err)
-      if (res.statusCode < 300) { // 2xx indicates error has not yet been handled
-        res.status(500)
-      }
-      res.send({error: err.message})
-      // TODO: Get real request start
-      loggerOld.error(JSON.stringify(this._buildHttpLog(Date.now(), req, res)))
-    })
+    app.use(errorHandler(diagnosticsLogger))
 
     const port = opts.port || DEFAULT_PORT
     this.server = app.listen(port, () => {
-      this.logger.imp('Ceramic API running on port ' + port)
+      diagnosticsLogger.imp('Ceramic API running on port ' + port)
     })
     this.server.keepAliveTimeout = 60 * 1000
   }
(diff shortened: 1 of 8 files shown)


packages/cli/src/ceramic-daemon.ts
```diff
@@ -215,89 +143,43 @@ class CeramicDaemon {
...
-    app.get(toApiPath('/commits/:docid'), this.commits.bind(this))
-    app.get(toApiPath('/records/:docid'), this.commits.bind(this))
-    app.post(toApiPath('/multiqueries'), this.multiQuery.bind(this))
-    app.get(toApiPath('/documents/:docid'), this.state.bind(this))
-    app.get(toApiPath('/pins/:docid'), this.listPinned.bind(this))
-    app.get(toApiPath('/pins'), this.listPinned.bind(this))
-    app.get(toApiPath('/node/chains'), this.getSupportedChains.bind(this))
-    app.get(toApiPath('/node/healthcheck'), this.healthcheck.bind(this))
+  registerAPIPaths (app: ExpressWithAsync, gateway: boolean): void {
+    app.getAsync(toApiPath('/commits/:docid'), this.commits.bind(this))
```
valmack, line 147: Why do we need to use `ExpressWithAsync`? Doesn't express already handle async routes on its own?

packages/cli/src/ceramic-daemon.ts
```diff
@@ -215,89 +143,43 @@ class CeramicDaemon {
...
     const { doctype, genesis, docOpts } = req.body
-    try {
-      const doc = await this.ceramic.createDocumentFromGenesis(doctype, DoctypeUtils.deserializeCommit(genesis), docOpts)
-      res.json({ docId: doc.id.toString(), state: DoctypeUtils.serializeState(doc.state) })
-    } catch (e) {
-      return next(e)
-    }
-    next()
+    const doc = await this.ceramic.createDocumentFromGenesis(doctype, DoctypeUtils.deserializeCommit(genesis), docOpts)
+    res.json({ docId: doc.id.toString(), state: DoctypeUtils.serializeState(doc.state) })
```
valmack, line 182: Here as an example, can we do a check to ensure that `doctype`, `gensis`, and `docOpts` are not empty? Or is that already handled by the methods below?

Also is there any potential attack vector here we can mitigate with sanitization?

packages/cli/src/ceramic-daemon.ts
```diff
@@ -307,154 +189,105 @@ class CeramicDaemon {
...
    * Apply one commit to the existing document
    */
-  async applyCommit (req: Request, res: Response, next: NextFunction): Promise<void> {
+  async applyCommit (req: Request, res: Response): Promise<void> {
     const { docId, commit, docOpts } = req.body
     if (!(docId && commit)) {
-      res.json({ error: 'docId and commit are required in order to apply commit' })
-      next()
-      return
+      throw new Error('docId and commit are required in order to apply commit')
```
valmack, line 229: Here we can replace the throw with below?

`res.status(400).send('docId and commit are required in order to apply commit')`

packages/cli/src/ceramic-daemon.ts
```diff
@@ -307,154 +189,105 @@ class CeramicDaemon {
...
-      const pinnedDocIds = []
-      const iterator = await this.ceramic.pin.ls(docId)
-      for await (const id of iterator) {
-        pinnedDocIds.push(id)
-      }
-      res.json({ pinnedDocIds: pinnedDocIds })
-    } catch (e) {
-      return next(e)
+    const pinnedDocIds = []
+    const iterator = await this.ceramic.pin.ls(docId)
```
valmack, line 277: Can we send 400 if we don't have the docId? Maybe just add an else statement above

packages/cli/src/ceramic-daemon.ts
```diff
@@ -307,154 +189,105 @@ class CeramicDaemon {
...
+    for await (const id of iterator) {
+      pinnedDocIds.push(id)
     }
-    next()
+    res.json({ pinnedDocIds: pinnedDocIds })
   }
 
-  async _notSupported (req: Request, res: Response, next: NextFunction): Promise<void> {
+  async _notSupported (req: Request, res: Response): Promise<void> {
     res.status(400).json({ status: 'error', message: 'Method not supported by read only Ceramic Gateway' })
```
valmack, line 285: I think this json should match the format that we have in the `errorHandler` middleware.

packages/cli/src/daemon/error-handler.ts
```diff
@@ -0,0 +1,20 @@
+import { ErrorRequestHandler, NextFunction, Request, Response } from 'express';
+import { DiagnosticsLogger } from '@ceramicnetwork/common';
+
+/**
+ * Generic error handling middleware for the daemon.
+ */
+export function errorHandler(logger: DiagnosticsLogger): ErrorRequestHandler {
+  return (err: Error, req: Request, res: Response, next: NextFunction) => {
+    (req as any).error = err // Allow other middlewares access the error
```
valmack, line 9: Not sure this is accurate. It seems like the error should be on the response instead.

packages/cli/src/daemon/error-handler.ts
```diff
@@ -0,0 +1,20 @@
...
+    logger.err(err);
+    if (res.headersSent) {
+      return next(err);
```
valmack, line 12: Does this mean express will handle the error?

packages/cli/src/daemon/log-requests.ts
```diff
@@ -0,0 +1,27 @@
+import { Request, Response } from 'express';
+import { LoggerProvider } from '@ceramicnetwork/common';
+import morgan from 'morgan';
+
+const ACCESS_LOG_FMT = 'ip=:remote-addr ts=:date[iso] method=:method original_url=:original-url base_url=:base-url path=:path http_version=:http-version req_header:req[header] status=:status content_length=:res[content-length] content_type=":res[content-type]" ref=:referrer user_agent=:user-agent elapsed_ms=:total-time[3] error_message=:error-message error_code=:error-code';
+
+export function logRequests(loggerProvider: LoggerProvider): any[] {
```
valmack, line 7: I think this is a misnomer. This function is just returning middleware, not actually logging anything.

I think we should call it `buildHttpLog` (or something similar) and it can return the middleware as any object instead of an array.

valmack (commented):
Doing some sanitization still seems to be low hanging fruit here. We can put in some 4xx errors and get a clearer signal on what is going wrong.

This can be streamlined by putting similar paths behind a route. For example
```
app.route('/pins/:docid')
  .all(function (req, res, next) {
    // sanitize and/or return 4xx here
    next()
  })
  .post(function (req, res, next) {
    // ...
  })
  .delete(function (req, res, next) {
    // ...
  })
```
But refactoring the routers like above is not strictly necessary (though I will probably do it anyway this week for some metrics clean up).

packages/cli/src/ceramic-daemon.ts
ukstv: It is express 4. It does not handle async routes. https://gist.github.com/ukstv/14377c8ed437aa6243be94d0c532cb25

Express 5 is in alpha.

ukstv: > Starting with Express 5, route handlers and middleware that return a Promise will call next(value) automatically when they reject or throw an error

https://expressjs.com/en/guide/error-handling.html

packages/cli/src/ceramic-daemon.ts
ukstv: Yes.

packages/cli/src/daemon/log-requests.ts
ukstv: It returns a middleware, that _logs requests_ :) One discovers existence of this thing I guess by reading `app.use` clause first.

How do you feel about `app.use(buildHttpLog(loggerProvider))` vs `app.use(logRequests(loggerProvider))`?

packages/cli/src/ceramic-daemon.ts
valmack: Worth it to upgrade? I'm guessing not but probably worth considering?

packages/cli/src/ceramic-daemon.ts
valmack: Or more accurately we should use the same json response format in the error handler middleware to be consistent

packages/cli/src/daemon/error-handler.ts
valmack: "This" was in reference to `next(err)`

packages/cli/src/daemon/log-requests.ts
valmack: I guess it should be `app.use(loggerMiddleware(...))`

packages/cli/src/daemon/error-handler.ts
ukstv: Semantically, yes, better put it with response.

packages/cli/src/daemon/log-requests.ts
ukstv: It is always middleware put to `app.use` :) I was following the lead of a canonical list https://expressjs.com/en/resources/middleware.html There is no one middleware there, btw. Names are either idiomatic nouns (Morgan, session, timeout, slash), or agent nouns (body-parser, cookie-parser, errorHandler) or plain verbs (serve-static, expiry, uncapitalize).

`requestLogger` as the middle ground? It is a petty battle anyway.

packages/cli/src/ceramic-daemon.ts
ukstv: Vote for ditching Express in favour of [Fastify](https://www.fastify.io), if it ever comes to it.

Yet, it is also worth considering how many breaking or (worse) potentially breaking changes one tries to introduce at a unit of time, and to what effect. I have no idea how that ever-alpha version behaves. Documentation is sparse, mildly speaking. If the goal is to ship a working system, I better stick with the oldie here. As a purely internal detail, it can be upgraded at any time later.

packages/cli/src/daemon/log-requests.ts
ukstv: Like, after your next proposition, I give up and accept anything.

packages/cli/src/daemon/error-handler.ts
ukstv: > If you call next() with an error after you have started writing the response (for example, if you encounter an error while streaming the response to the client) the Express default error handler closes the connection and fails the request.

> So when you add a custom error handler, you must delegate to the default Express error handler, when the headers have already been sent to the client

https://expressjs.com/en/guide/error-handling.html

ukstv: TLDR It will close the connection and fail the request. Most probably never happens.

packages/cli/src/ceramic-daemon.ts
ukstv: Actually, this is a perfect example to take a strong stance "no sanitation at http daemon level". `pin.ls` accepts both docId and `undefined`. Both are valid. The latter means "list everything pinned".

Here, the only problematic case is invalid DocID passed. This is handled in `DocID.fromString` which errors if the input can not be deserialised into `DocID`.

`loadDocumentCommits` AKA [`commits`](https://github.com/ceramicnetwork/js-ceramic/pull/1210/files/436c246640c917e21ada94412cfa3ff96840af68#diff-29a806cda83d5289fde2f9b08500d1f2335980307791376095ed3d7c600da98eR210) is a bit more interesting. It tries to parse DocID in the Express handler, but the underlying code also tries to perform deserialisation. Effectively, either DocID deserialization in [`commits` handler](https://github.com/ceramicnetwork/js-ceramic/pull/1210/files/436c246640c917e21ada94412cfa3ff96840af68#diff-29a806cda83d5289fde2f9b08500d1f2335980307791376095ed3d7c600da98eR211) or DocID deserialization inside `loadDocumentCommits` is redundant.

What makes sense to me is a proper serialization/deserialization mechanism to strictly delineate formatting (400) errors from other ones while reducing redundancy with the underlying code. The _other errors_ could be mapped to whatever http errors, including 400. As discussed, first step on the way there is getting that mapping done. Then, few discussions have to be made with regards to serialization/deserialization tooling. It better be a single thing, like a codec, it better be composable, otherwise it introduces maintenance headache. But, it is a different story from the one we have now.

packages/cli/src/ceramic-daemon.ts
ukstv: This kind of checks would eliminate some execution steps required to achieve same checks in the underlying code, which is good. How good? Have no data. Probably, minuscule, but this is just a guess. I am pretty sure with error-to-http mapping, UX/DX is the same, with a benefit of having a single place where the checks are executed.

packages/cli/src/ceramic-daemon.ts
valmack: Ok great! Agree

packages/cli/src/daemon/error-handler.ts
valmack: Aha thanks

packages/cli/src/daemon/log-requests.ts
valmack: Yeah I think you are right actually because it just takes a function. Up to you!

packages/cli/src/ceramic-daemon.ts
valmack: Ok let's punt for now

packages/cli/src/ceramic-daemon.ts
valmack: Ok so docid here is optional. Wasn't clear to me. I would prefer being more verbose. Such as `pin.ls(docId && docId || undefined)`. Or at least something so it is clear that docId is optional here.

Agree on the mapping and the fact that it is out of scope here.

packages/cli/src/ceramic-daemon.ts
valmack: Let's punt on this change too and just put it in project.

valmack (commented):
One ask before approval. Can you confirm what files the http logs are being written to? Need to ensure the error logs are written to a different file. That should be the case but I'm not positive the loggers used here do that so want to check before we finalize this.

ukstv:
> One ask before approval. Can you confirm what files the http logs are being written to? Need to ensure the error logs are written to a different file. That should be the case but I'm not positive the loggers used here do that so want to check before we finalize this.

@valmack Confirm. http-access.log contains only morgan-formatted errors. And, diagnostics logger receives an actual error.

valmack (approved):
Awesome. Can you start a project for the "http-error" mapping?

ukstv:
> Awesome. Can you start a project for the "http-error" mapping?

Yup.

PR #1213, opened 2021-04-07
feat(doctype-caip10-link): Update Caip10LinkDoctype API

stbrody:


stbrody:
I'm not sure why the tests are hanging.  I did confirm that the issue is indeed the new `caip10-link.test.ts`, and I confirmed that if I run just the `invalid account id` test (which never successfully makes a `Caip10LinkDoctype` instance) then it doesn't hang, but if I run just the `Create from valid account id` test (the simplest of the tests that create a `Caip10LinkDoctype` instance), then the test hangs.  So it seems to be that the mere act of creating a `Caip10LinkDoctype` causes the test to hang, but I have no idea why or what to do about it, and I've never successfully gotten jest's `--detectOpenHandles` flag to do anything useful.  I'm open to suggestions about how to go about debugging this further

@oed @ukstv

stbrody:
Okay I fixed the test issue, it had to do with how I was using MockDate.  This is now ready for review

packages/doctype-caip10-link/src/caip10-link-doctype.ts
```diff
@@ -36,80 +22,107 @@ const throwReadOnlyError = (): Promise<void> => {
 @DoctypeStatic<DoctypeConstructor<Caip10LinkDoctype>>()
 export class Caip10LinkDoctype extends Doctype {
```
oed, line 23: Did we want to call this class just `Caip10Link`, or is that a future PR?

oed (approved):
A few minor comments, but looks good.

packages/doctype-caip10-link/src/caip10-link-doctype.ts
stbrody: Future PR, under https://github.com/ceramicnetwork/js-ceramic/issues/1115

packages/doctype-caip10-link/src/caip10-link-doctype.ts
```diff
@@ -36,80 +22,107 @@ const throwReadOnlyError = (): Promise<void> => {
...
 
+    static DOCTYPE_NAME = 'caip10-link'
+    static DOCTYPE_ID = 1
```
ukstv, line 26: Doctype ID here with TODO comment above for HandlersMap. Uff, this is quite a change coming. Would vote for talking through this.

Non-actionable for here, just in case.

packages/doctype-caip10-link-handler/src/caip10-link-doctype-handler.ts
```diff
@@ -87,6 +81,16 @@ export class Caip10LinkDoctypeHandler implements DoctypeHandler<Caip10LinkDoctyp
         return state
     }
 
+    /**
+     * Thin wrapper around 'validateLink' function from blockchain-utils-validation to allow
+     * injecting a mock implementation for ease of unit testing.
+     * @param proof
+     * @private
+     */
```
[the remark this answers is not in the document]
stbrody: I guess my concern with that would be that I'd have to mock the whole module for all consumers, rather than just for a specific Caip10Link instance at a time.  But for the test as I have it now it doesn't really matter.  Mostly I just didn't have a good example of how to do that and didn't want to spend too long figuring it out, but I can look into it today.  It does seem kind of gross and silly to add this just as a crutch for testing

packages/doctype-caip10-link/src/caip10-link-doctype.ts
```diff
@@ -36,80 +22,107 @@ const throwReadOnlyError = (): Promise<void> => {
...
-     * @param opts - Initialization options
+     * Creates a Caip10Link for the given CAIP10 address. Initially created without a link to any DID,
+     *   use 'setDid' to create the public link between the given CAIP10 account and a DID.
+     * @param ceramic - Instance of CeramicAPI used to communicate with the Ceramic network
+     * @param accountId - Blockchain account id in CAIP10 format.
+     * @param opts - Additional options.
      */
-    async change(params: Caip10LinkParams, opts?: DocOpts): Promise<void> {
-        const { content, metadata } = params
+    static async fromAccount(ceramic: CeramicApi,
```
ukstv, line 35: sweet

packages/doctype-caip10-link/src/caip10-link-doctype.ts
stbrody: Yeah happy to talk about it.  I think we should be relying on the doctype ID over the name whenever possible within internal-only code, since that's what shows up in the protocol messages and will future-proof us in case we ever decide to rename an existing doctype.

packages/core/src/__tests__/caip10-link.test.ts
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  beforeAll(async () => {
+    ipfs = await createIPFS()
+    MockDate.set('2018-10-01')
+  })
+
+  afterAll(async () => {
+    MockDate.reset()
+    await ipfs.stop(() => console.log('IPFS stopped'))
```
ukstv, line 51: Minor point here. console.log callback can be removed.

ukstv (commented):
🏇

ukstv (approved):

PR #1214, opened 2021-04-08
Ipfs daemon docker

valmack:
### Motivation

On infra the migration from 9 to 10 is failing. The very odd thing here is the the image built from `develop` branch results in v9. Images built from `master` are properly using v10. The docker image we are using for the build just pulls from npm so this PR addresses that issue. No guarantees it will fix the migration problem unfortunately but submitted an issue here: https://github.com/ipfs/js-ipfs-repo/issues/299

### Changes
- Replaces Dockerfile with Dockerfile.ipfs-daemon in root
- Builds ipfs-daemon with lerna
- Uses ipfs path specified in config by default

### DIFF
--- a/packages/ipfs-daemon/src/ipfs-daemon.ts
+++ b/packages/ipfs-daemon/src/ipfs-daemon.ts
@@ -86,7 +86,7 @@ export class IpfsDaemon {
             bucket: configuration.awsBucketName,
             accessKeyId: configuration.awsAccessKeyId,
             secretAccessKey: configuration.awsSecretAccessKey,
-        }) : null
+        }) : configuration.ipfsPath
 
         const ipfs = await IPFS.create({
             start: false,
--- a/Dockerfile.ipfs-daemon
+++ b/Dockerfile.ipfs-daemon
@@ -1,0 +1,39 @@
+FROM node:14
+
+RUN apt-get update && apt-get install -y netcat
+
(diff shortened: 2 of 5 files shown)


packages/ipfs-daemon/src/ipfs-daemon.ts
```diff
@@ -86,7 +86,7 @@ export class IpfsDaemon {
             bucket: configuration.awsBucketName,
             accessKeyId: configuration.awsAccessKeyId,
             secretAccessKey: configuration.awsSecretAccessKey,
-        }) : null
+        }) : configuration.ipfsPath
```
ukstv, line 89: AAAA!! YESS!

ukstv (approved):
If it builds, and runs, I see no reason to stop the spice flowing. LGTM

@valmack In a fantasy universe, you could add pruning dependencies and use alpine as a base image, to make the image lighter, but here and now it does not matter.

🏇

valmack:
> If it builds, and runs, I see no reason to stop the spice flowing. LGTM
> 
> @valmack In a fantasy universe, you could add pruning dependencies and use alpine as a base image, to make the image lighter, but here and now it does not matter.
> 
> 🏇

Yeah! Definitely want to make it lighter. Leaving packages out made the build fail (no surprise) so I'll look into how dependency pruning works at some point.

PR #1216, opened 2021-04-08
feat(doctype-caip10-link): Rename 'content' to 'did' for Caip10Link

stbrody:


packages/core/src/state-management/state-validation.ts
```diff
@@ -35,7 +36,7 @@ export class RealStateValidation implements StateValidation {
     } catch {
       throw new Error('Commit missing when loading schema document');
     }
-    return this.loadDocument(commitId).then((doc) => doc.content);
+    return this.loadDocument<TileDoctype<T>>(commitId).then((doc) => doc.content);
```
oed, line 39: Wow, schema validation should really take place within the TileDocumentHandler!

packages/core/src/utils.ts
```diff
@@ -1,6 +1,7 @@
 import ajv from "ajv"
 
 import { Doctype } from "@ceramicnetwork/common"
+import {TileDoctype} from "@ceramicnetwork/doctype-tile";
```
oed, line 4: maybe we can just import the type here?

packages/doctype-caip10-link/src/caip10-link-doctype.ts
```diff
@@ -59,7 +67,7 @@ export class Caip10LinkDoctype extends Doctype {
      * create a verifiable link from the CAIP10 address to the DID.
      * @param did - The DID being linked to the CAIP10 address that this Caip10Link object represents.
      *   If the 'did' provided is an instance of the DID type, the DID must already be authenticated
-     *   so that the did string it represents is available.
+     *   so that the did string it represents is avapilable.
```
oed, line 70: Wut? 😆

oed (approved):
lgtm

packages/core/src/state-management/state-validation.ts
stbrody: Yep agreed.  I just noticed this and actually added this comment in my last commit: https://github.com/ceramicnetwork/js-ceramic/blob/afcf35426582bbc6aa0a5b2181feb5bf5c5016f9/packages/core/src/conflict-resolution.ts#L266

packages/core/src/utils.ts
stbrody: what do you mean?  You mean using `import type { TileDoctype } from ..."?  I've never been clear on what `import type` actually does

packages/doctype-caip10-link/src/caip10-link-doctype.ts
stbrody: oops!

stbrody: In my defense I hadn't actually sent this for review yet, I was still gonna do a proofreading pass.  But you're just too quick for me!  :laughing:

packages/core/src/utils.ts
stbrody: Okay just did some reading.  I guess `import type` means the import only really happens at compile time to do type checking, but it doesn't actually import the code behind the type into the generated javascript at the end?

oed: Yep exactly!

packages/core/src/state-management/state-validation.ts
stbrody: FYI I just filed https://github.com/ceramicnetwork/js-ceramic/issues/1219 for this.  Could be a good starting issue for our upcoming new hire to get familiar with the codebase?

Issue #1123, opened 2021-03-13
Change "sync" DocOpt to an enum


stbrody:
with 3 values.  One that means don't sync the tip from pubsub at all, one that means sync the tip from pubsub only if we didn't previously have any state for this document in the Repository (this is the current behavior of sync:true), and one that means always sync even if we have state for the doc in the Repository (this will be new behavior).

```
enum SyncOptions {
  DONT_SYNC,
  SYNC_ON_LOAD, // Open to ideas for a better name for this.  Maybe just `SYNC` and call the other option `FORCE_SYNC`?
  SYNC_ALWAYS, // or FORCE_SYNC?
}
```

oed:
> and one that means always sync even if we have state for the doc in the Repository (this will be new behavior).

This would only happen on startup I presume? Would be bad if we try to sync it every time we load it from disk (we should still listen to updates for it when it's not loaded).

stbrody:
> > and one that means always sync even if we have state for the doc in the Repository (this will be new behavior).
> 
> This would only happen on startup I presume? Would be bad if we try to sync it every time we load it from disk (we should still listen to updates for it when it's not loaded).

It would be a flag that could be set any time a document is loaded. So no, not just at startup.  But now when a user/application calls loadDocument, they can specify to do no syncing (get whatever contents is already stored in the cache or the state store or get back nothing if there is no state already stored in the Repository), have the current syncing behavior (get the state from the cache or state store, but if there is no existing state then request the tip from pubsub and wait 3 seconds for it), or use the new "force sync" behavior (request the tip from pubsub and wait 3 seconds for a response, even if we already have state stored in the cache or state store for this doc)

I assume the CLI would continue to have the same default behavior as it does today, though we may choose to expose this option via a new cli flag at some point.  Something like `--force-sync`.

oed:
Ok, thanks for clarifying!

stbrody:
As part of this, will also add a new field to `DocOpts` called `syncDelay`.  The value controls the number of milliseconds to wait for a pubsub response when `sync` is true. Default will be 3000, as it is today.

PR #1225, opened 2021-04-09
feat(http-client, cli): Enable specifying DocOpts in CeramicClient.loadDocument

stbrody:


packages/cli/src/__tests__/ceramic-daemon.test.ts
```diff
@@ -161,6 +158,32 @@ describe('Ceramic interop: core <> http-client', () => {
         expect(DoctypeUtils.serializeState(doc3.state)).toEqual(DoctypeUtils.serializeState(doc4.state))
     })
 
+    // TODO Enable this test when its possible to use 'sync' flag to disable querying pubsub entirely
```
stbrody, line 161: This test will become possible to write correctly as part of https://github.com/ceramicnetwork/js-ceramic/issues/1123, which will make it possible to load a doc without querying pubsub at all.  Right now `sync` only controls whether or not we wait for a reply, but we always do the query, which introduces a race condition around how fast the response comes back and if it comes back before we've built the Doctype instance and taken the current version of the state from the RunningState

packages/cli/src/ceramic-daemon.ts
```diff
@@ -177,8 +177,8 @@ class CeramicDaemon {
    * @dev Useful when the docId is unknown, but you have the genesis contents
    */
   async createDocFromGenesis (req: Request, res: Response): Promise<void> {
-    const { doctype, genesis, docOpts } = req.body
-    const doc = await this.ceramic.createDocumentFromGenesis(doctype, DoctypeUtils.deserializeCommit(genesis), docOpts)
+    const { doctype, genesis, opts } = req.body
```
stbrody, line 180: This is technically a breaking change to the HTTP api, but a minor one.  It seemed worth doing to avoid having an arg called `docOpts` when we're moving away from calling all doctypes (streamtypes) documents.

stbrody: Pulled this change out to its own PR: https://github.com/ceramicnetwork/js-ceramic/pull/1227

packages/cli/src/__tests__/ceramic-daemon.test.ts
```diff
@@ -18,26 +18,31 @@ import dagJose from 'dag-jose'
 import { sha256 } from 'multiformats/hashes/sha2'
```
stbrody, line 18: No real changes in this file, just code cleanup

packages/cli/src/__tests__/ceramic-daemon.test.ts
stbrody: I was wrong about why this test was failing, and was able to get it working

oed (approved):
lgtm

PR #1227, opened 2021-04-10
feat(http-client,cli): Rename docOpts to opts in http api

stbrody:
This is technically a breaking change to the HTTP api, but a minor one. It seemed worth doing to avoid having an arg called docOpts when we're moving away from calling all doctypes (streamtypes) documents.

### DIFF
--- a/packages/3id-did-resolver/src/__tests__/vectors.json
+++ b/packages/3id-did-resolver/src/__tests__/vectors.json
@@ -159,13 +159,13 @@
       }
     },
     "http://localhost:7007/api/v0/documents": [{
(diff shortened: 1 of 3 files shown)


oed (changes requested):
Can we please add backwards compatibility here? This could otherwise cause issues if we launch ELP without all packages updated everywhere.

oed:
Couldn't we just keep this api as is and rename the param when we introduce the `streams` endpoint?

stbrody:
> Couldn't we just keep this api as is and rename the param when we introduce the `streams` endpoint?

That's a good idea, I'll do that

Issue #1179, opened 2021-03-30
Rename DocID to StreamID


stbrody:
There may be some extra complexity here around handling Symbols used for equality comparisons

PR #1237, opened 2021-04-12
feat(cli,http-client): Update HTTP API to use streams terminology

stbrody:
Also updates the MultiQueries API since that also affects the http API.

Did some simple manual testing of a new daemon with an old http-client. I used an old CLI to create and update a tile document, and to pin, unpin, and list pins.  Anything beyond what's covered by that doesn't have test coverage of backwards compatibility.


oed (commented):
Let's maybe wait and rebase on #1179 once merged?

PR #1249, opened 2021-04-13
feat(core,http-client,cli): Update config options from document to stream-based terminology

stbrody:


packages/cli/src/__tests__/ceramic-daemon.test.ts
```diff
@@ -89,7 +89,7 @@ describe('Ceramic interop: core <> http-client', () => {
         const port = await getPort()
         const apiUrl = 'http://localhost:' + port
         daemon = new CeramicDaemon(core, { port })
-        client = new CeramicClient(apiUrl, { docSyncInterval: 500 })
+        client = new CeramicClient(apiUrl, { streamSyncInterval: 500 })
```
oed, line 92: Maybe just call it `syncInterval`?

oed (approved):

PR #1250, opened 2021-04-13
DocState contains type as number

ukstv:
- DocState#doctype is gone in favour of DocState#type (which is number)
- Concern with HTTP and StateStore is handled by DoctypeUtils: serialises with both type and doctype, deserialises with just `type` (assuming mapping name->id is in stream types registry). So, no breaking changes expected.

### DIFF
--- a/packages/streamid/src/stream-type.ts
+++ b/packages/streamid/src/stream-type.ts
@@ -3,7 +3,7 @@ const registry: Record<string, number | undefined> = {
   'caip10-link': 1,
 };
 
-export function indexByName(name: string): number {
+function indexByName(name: string): number {
   const index = registry[name];
   if (typeof index !== 'undefined') {
     return index;
@@ -12,11 +12,16 @@ export function indexByName(name: string): number {
   }
 }
 
-export function nameByIndex(index: number): string {
+function nameByIndex(index: number): string {
   const pair = Object.entries(registry).find(([, v]) => v === index);
   if (pair) {
     return pair[0];
   } else {
     throw new Error(`No stream type registered for index ${index}`);
   }
+}
+
+export class StreamType {
+  static nameByIndex = nameByIndex;
+  static indexByName = indexByName;
 }
--- a/packages/streamid/src/stream-type.ts
+++ b/packages/streamid/src/stream-type.ts
@@ -3,7 +3,7 @@ const registry: Record<string, number | undefined> = {
   'caip10-link': 1,
 };
 
-function indexByName(name: string): number {
+function codeByName(name: string): number {
   const index = registry[name];
   if (typeof index !== 'undefined') {
     return index;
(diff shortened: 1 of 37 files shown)


packages/common/src/doctype.ts
```diff
@@ -134,11 +134,11 @@ export abstract class Doctype extends Observable<DocState> implements DocStateHo
...
 
     get id(): StreamID {
-        return new StreamID(this.state$.value.doctype, this.state$.value.log[0].cid)
+        return new StreamID(this.state$.value.type, this.state$.value.log[0].cid)
     }
 
-    get doctype(): string {
-        return this.state$.value.doctype
+    get doctype(): number {
+        return this.state$.value.type
```
oed, line 141: I think me and @stbrody discussed this and wanted to keep the doctype as string here for the time being.

oed (changes requested):
Would want @stbrody review here as well.

We really need to make sure that this doesn't break the http api in any way.

packages/common/src/doctype.ts
ukstv: Oke.

stbrody: Did we discuss this?  I think the only thing we discussed was making sure it stayed in the http client output, which is already handled by the changes to DoctypeUtils.serializeState().

I don't really see any particular reason this *needs* to be included as a string.  But I also don't really feel strongly either way.  I guess it's a little nicer for applications to be able to get the human-readable string name without having to do an extra lookup of StreamType.nameByIndex, so the way it is in the most recent version looks fine. But again, I don't feel strongly either way

oed: hm In that case I think we could just remove this getter?

packages/streamid/src/stream-type.ts
```diff
@@ -3,7 +3,7 @@ const registry: Record<string, number | undefined> = {
   'caip10-link': 1,
 };
 
-export function indexByName(name: string): number {
+function indexByName(name: string): number {
```
stbrody, line 6: Can we call these `idByName` and `nameById`?

stbrody (approved):
LGTM mod renaming the StreamType conversion functions

packages/common/src/doctype.ts
stbrody: Yeah, seems fine to remove.  You can always get the streamtype ID and name from the StreamID

packages/streamid/src/stream-type.ts
oed: `typeByName`, `nameByType`?

stbrody: It feels a little weird to me to call `StreamType.nameByType()` and pass it a number.  It seems like a function on an object called `StreamType` that says it takes a 'type' should take a `StreamType` instance.  But I don't feel strongly and still like using 'type' better than 'index'

oed: Ah, maybe `codeByName` and `nameByCode`.

oed: But I'm quite ambivalent at this point :)

PR #1252, opened 2021-04-14
fix(3id-did-resolver): Use the correct url for json-ld context

oed:
Apparently we were using an incorrect url for the `@context` property for the json-ld encoded DID document.

Feel free to merge this without my oversight.


ukstv (approved):
🏇

PR #1253, opened 2021-04-14
fix: update docker images for streamid package

valmack:
Image builds were failing because docid package no longer exists. This pr cleans up the docker image code and avoids such issues in the future.


ukstv (approved):
If Docker image size is not a constraint, LGTM.

🏇

valmack:
> If Docker image size is not a constraint, LGTM.
> 
> 🏇

I do look forward to the day when we can optimize these though :)

PR #1261, opened 2021-04-14
Create a secp256r1 resolution method

bshambaugh:
I used JsonWebKey2020 which uses publicJwk to map to an object? This is the first use of publicJwk. secp256k1 and ed25519 uses publicKeyBase58 which uses a string.

### DIFF
--- a/packages/key-did-resolver/src/__tests__/secp256r1.test.ts
+++ b/packages/key-did-resolver/src/__tests__/secp256r1.test.ts
@@ -416,9 +416,17 @@ test('show how to convert a raw public key in hex and return an uncompressed key
  */
 
 
-test('test a null string to ECPointCompress', () => {
+test('test a null coordinates in ECPointCompress', () => {
    const x = null;
    const y = null;
+   expect(() => {
+      ECPointCompress(x,y);
+   }).toThrowError('input cannot be null or undefined.');
+});
+
+test('test a undefined coordinates in ECPointCompress', () => {
+   const x = undefined;
+   const y = undefined;
    expect(() => {
       ECPointCompress(x,y);
    }).toThrowError('input cannot be null or undefined.');
@@ -429,6 +437,13 @@ test('test null in pubKeyHexToUint8Array', () => {
    expect(() => {
       pubKeyHexToUint8Array(inputPublicKeyHex);
    }).toThrowError('input cannot be null or undefined.');   
+});
+
+test('test undefined in pubKeyHexToUint8Array', () => {
+   const inputPublicKeyHex = undefined;
+   expect(() => {
+      pubKeyHexToUint8Array(inputPublicKeyHex);
+   }).toThrowError('input cannot be null or undefined.');
 });
 
 test('test an empty string in pubKeyHexToUint8Array', () => {
@@ -452,7 +467,13 @@ test('test compresedKeyInHex with null', () => {
    expect(() => {
       compresedKeyInHex(inputPublicKeyHex);
    }).toThrowError('input cannot be null or undefined.');
+});
 
+test('test compresedKeyInHex with undefined', () => {
+   const inputPublicKeyHex = undefined;
+   expect(() => {
+      compresedKeyInHex(inputPublicKeyHex);
+   }).toThrowError('input cannot be null or undefined.');
 });
 
 //**** end of tests for functions withing the test file
@@ -475,18 +496,20 @@ function ECPointCompress( x: Uint8Array, y: Uint8Array )
 }
 
 function pubKeyHexToUint8Array(publicKeyHex: string) {
-   if(publicKeyHex !== null) {
+  if(publicKeyHex == null) {
+   throw new TypeError('input cannot be null or undefined.');
+  }
     if(publicKeyHex.length % 2 == 0) {
           return u8a.fromString(publicKeyHex,'base16');          	 
       } else {
           return u8a.fromString(('0'+publicKeyHex),'base16');
     }
-   }
-   throw new TypeError('input cannot be null or undefined.');
 }
 
 function compresedKeyInHex(publicKeyHex: string) {
- if(publicKeyHex !== null) {
+  if(publicKeyHex == null) {
+    throw new TypeError('input cannot be null or undefined.');
+  }
   const xHex = publicKeyHex.slice(0,publicKeyHex.length/2);
   const yHex = publicKeyHex.slice(publicKeyHex.length/2,publicKeyHex.length);
 
@@ -496,6 +519,4 @@ function compresedKeyInHex(publicKeyHex: string) {
   const compressedPoint = ECPointCompress( xOctet , yOctet );
   const compressedPointHex = u8a.toString(compressedPoint,'base16');
   return compressedPointHex;
-  }
-  throw new TypeError('input cannot be null or undefined.');
 }
--- a/packages/key-did-resolver/src/__tests__/secp256r1.test.ts
+++ b/packages/key-did-resolver/src/__tests__/secp256r1.test.ts
@@ -40,18 +40,6 @@ describe('Secp256r1 mapper', () => {
 
 })
 
-// this test fails
-/*
-test('convert a public key x,y as nulls', () => {
-   const ecpoint = {
-          x: null,
-          y: null
-   };
-
-   const u8aPoint = mapper.publicKeyIntToUint8ArrayPointPair(ecpoint);
-});
-*/
-
 test('empty key string to should not evaluate to null, or should it??', () => {
    const inputPublicKeyHex = '';
    const u8aPoint = mapper.publicKeyHexToUint8ArrayPointPair(inputPublicKeyHex);
@@ -66,16 +54,6 @@ test('public key as empty string', () => {
   const base64urlPoint = mapper.publicKeyToXY(inputPublicKeyHex);
   expect(base64urlPoint).toEqual(output);
   });
-
-
-// this test fails
-/*
-test('public key as null', () => {
-  const inputPublicKeyHex = null;
-  const base64urlPoint = mapper.publicKeyToXY(inputPublicKeyHex);
-  //  expect(base64urlPoint).toEqual(output);
- });
- */
 
 test('test a null string', () => {
    const inputPublicKeyHex = null;
@@ -348,16 +326,6 @@ function ECPointCompress( x: Uint8Array, y: Uint8Array )
     return out;
 }
 
-/*
-function pubKeyHexToUint8Array(publicKeyHex: string) {
-    let publicKey_u8a = null;
-    if(publicKeyHex.length % 2 == 0) {
-        publicKey_u8a = u8a.fromString(publicKeyHex,'base16');
-    }
-    return publicKey_u8a;
-}
-*/
-
 function pubKeyHexToUint8Array(publicKeyHex: string) {
 let publicKey_u8a = null;
    if(publicKeyHex !== null) {
@@ -369,30 +337,6 @@ let publicKey_u8a = null;
    }
     return publicKey_u8a;
 }
-
-/*
-function pubKeyHexToUint8Array(publicKeyHex: string) {
-    let publicKey_u8a = [];
-    if(publicKeyHex !== null) {
-       if(publicKeyHex.length % 2 == 0) {
-          publicKey_u8a = u8a.fromString(publicKeyHex,'base16');
-        } else {
-          publicKey_u8a = u8a.fromString(('0'+publicKeyHex),'base16');
-	  }
-     }
-     if(publicKeyHex !== null) {
-        publicKey_u8a = null;
-     }
-    return publicKey_u8a;
-}
-*/
-
-/*
-function pubKeyHexToUint8Array(publicKeyHex: string) {
-     const publicKey_u8a = u8a.fromString(publicKeyHex,'base16');
-     return publicKey_u8a;
-}
-*/
 
 function compresedKeyInHex(publicKeyHex: string) {
   const xHex = publicKeyHex.slice(0,publicKeyHex.length/2);
--- a/packages/key-did-resolver/src/__tests__/secp256r1.test.ts
+++ b/packages/key-did-resolver/src/__tests__/secp256r1.test.ts
@@ -40,78 +40,77 @@ describe('Secp256r1 mapper', () => {
 
 })
 
- test('expect publicKeyIntToXY to throw an error for incorrect type', () => {
+test('expect publicKeyIntToXY to throw an error for incorrect type', () => {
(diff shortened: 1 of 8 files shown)


packages/key-did-resolver/src/secp256r1.ts
```diff
@@ -0,0 +1,51 @@
...
+ */
+export function keyToDidDoc (pubKeyBytes: Uint8Array, fingerprint: string): any {
+  const did = `did:key:${fingerprint}`
+  const keyId = `${did}#${fingerprint}`
+  const key = pubKeyBytesToXY(pubKeyBytes);
+  return {
+    id: did,
+    verificationMethod: [{
+      id: keyId,
+      type: 'JsonWebKey2020',
```
oed, line 16: What's the reasoning behind using `JsonWebKey2020` instead of `publicKeyBase58`?

packages/key-did-resolver/src/__tests__/secp256r1.test.ts
```diff
@@ -0,0 +1,19 @@
+import varint from "varint"
+import multibase from "multibase"
+import * as mapper from "../secp256r1"
+
+describe('Secp256r1 mapper', () => {
+
+    it('successfully resolves the document from did', async () => {
+        const id = "zruuPojWkzGPb8sVc42f2YxcTXKUTpAUbdrzVovaTBmGGNyK6cGFaA4Kp7SSLKecrxYz8Sc9d77Rss7rayYt1oFCaNJ"
```
oed, line 8: Afaik the Key DIDs should only contain the X value (see secp256k1 and ed25519 examples).

packages/key-did-resolver/src/secp256r1.ts
```diff
@@ -0,0 +1,51 @@
...
+    capabilityInvocation: [keyId],
+  }
+  }
+
+function pubKeyBytesToHex(pubKeyBytes: Uint8Array) {
+ const bbf = u8a.toString(pubKeyBytes,'base16')
+ return bbf;
+}
+
+function publicKeyToXY(publicKeyHex: string) {
```
oed, line 37: This would break if the public key in the did string only contains the x value.

oed (commented):
Not sure this is completely right. From the other examples we always just include the X value in the public key part of the did string. Seems like you are including both X and Y.

package.json
```diff
@@ -37,5 +37,8 @@
     "printWidth": 120,
     "trailingComma": "all",
     "singleQuote": true
+  },
+  "dependencies": {
+    "@mapbox/node-pre-gyp": "^1.0.3"
```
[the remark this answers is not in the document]
bshambaugh: Okay, fair enough. I needed it install this dependency to get js-ceramic's tests for this package to work on my own machine.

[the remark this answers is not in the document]
bshambaugh: I can remove it from the PR.

packages/key-did-resolver/src/secp256r1.ts
bshambaugh: I will ask.

bshambaugh: See https://lists.w3.org/Archives/Public/public-credentials/2021Apr/0139.html .

oed: Ok, thanks for clarifying. I think this makes sense then.

packages/key-did-resolver/src/__tests__/secp256r1.test.ts
oed: does the transmute-industries/did-key.js also use this format?

oed: If so this should be alright.

package.json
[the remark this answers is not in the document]
oed: Yes please :)

packages/key-did-resolver/src/__tests__/secp256r1.test.ts
bshambaugh: Transmute industries may be inconsistent with the present state. I believe Orie is updating a part before becoming consistent with the did method key spec and this issue.
https://github.com/w3c-ccg/did-method-key/issues/29#issuecomment-786028270

However, the did key is the same length as in: https://w3c-ccg.github.io/did-method-key/#p-256 .

bshambaugh: edit: Orie is waiting on bls12381 which is not working with node 15. https://github.com/transmute-industries/did-key.js/pull/78#issuecomment-806200291

bshambaugh: Fwiw, I am going back to see how I created the did:key. The code works with the did:key examples from the ccg spec and the ones that I created on my own.

bshambaugh: This is my code to create the did:key :: https://github.com/bshambaugh/key-did-provider-secp256r1

bshambaugh: Yes. The hex that I used to create the key is 64 bytes. It is a hex string that has a uint8array of bytes that looks like this: https://www.youtube.com/watch?v=_MjjF211BM8&t=263s .

Looking at the code and output suggests that 64 bytes is the full X,Y key. 
https://gist.github.com/bshambaugh/36aa03e0d79b630678c558f32c6ed936

packages/key-did-resolver/src/secp256r1.ts
bshambaugh: I believe that I have the X,Y pair. My did:key is of the same length as the did:key spec.

packages/key-did-resolver/src/__tests__/secp256r1.test.ts
bshambaugh: I can ask if the compressed public key would be better for did:key

oed: Yeah, I'm mostly just surprised why `secp256k1` is compressed while `r1` is not. I'm fine with leaving it as is though since it seems like that's what the did:key spec says.

bshambaugh: Kyle Hartog believed the spec was the best thing to look at. He also believed that the work was stalled. 
https://lists.w3.org/Archives/Public/public-credentials/2021Apr/0171.html
I am currently investigating how difficult it would be to modify the function to convert the compressed form into the uncompressed form for resolving the did doc.

bshambaugh: There is some conversation going on over here about compressed vs. uncompressed did:keys. I am linking to it to keep track of it all: https://github.com/w3c-ccg/did-method-key/issues/32

oed: Thanks. Should we wait for that to be resolved before we merge this?

bshambaugh: The last link [1] that dlongley linked to in his Post is interesting.

[1] https://stackoverflow.com/questions/17171542/algorithm-for-elliptic-curve-point-compression/30431547#30431547
[2] https://github.com/w3c-ccg/did-method-key/issues/32#issuecomment-827681999

I found out that modPow is used for the decompression. Presently I am trying to perform the same functions by not using the JavaScript Big Integer Library. modPow is for modular exponentiation [4].

[3] https://github.com/peterolson/BigInteger.js
[4] https://en.wikipedia.org/wiki/Modular_exponentiation

BigInt [5] is included natively in JavaScript now, which allows me to emulate some of the functionality of the [1] link.

However, I am struggling to implement modPow. Assuming:

var yBig = xBig.pow(3).sub( xBig.multiply(3) ).add( b ).modPow( pIdent, prime );

is the same as

var yBig = ((xBig**3n - xBig*3n + b)**pIdent) % prime;

gives me an error: "RangeError: Maximum BigInt size exceeded"

To solve this error, I looked for implementations of modPow with JavaScript. I found this one [6] that works with BigInt.

[5] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt
[6] https://github.com/juanelas/bigint-mod-arith

or something like this [7] if the library is to be avoided

[7] https://gist.github.com/krzkaczor/0bdba0ee9555659ae5fe

Should I include a library, or should I add a function for implementing BigInt to my code?

oed: > Should I include a library, or should I add a function for implementing BigInt to my code?

Which library are you referring to? If it's not too big then I think it's fine to include.

bshambaugh: I have raw (64 byte), compressed (33 byte), and uncompressed (65 byte) keys in my code now. My tests pass, but I do not have full coverage yet.
scr-1620860401

oed:
@bshambaugh please let us know when this is ready for a full review again :)

packages/key-did-resolver/src/__tests__/secp256r1.test.ts
bshambaugh: Ahh...line 61 was for the function ECPointDecompress . I added an extra test and it looks to be covered. The conditional if block that line 61 is in only executes if the parity of the compressed key is opposite (of what is should be?).

bshambaugh: scr-1621014068

Now I need to address the pubKeyBytesToXY function. I believe the function will fail if the bytesCount is 65 and the 1st byte is not 03, the bytes count is 33 and the 1st bytes is not 03 or 02, or the bytes count is not 64, 65, or 33. I guess I could have the function return null if this is the case. I am not sure how the did:key document will resolve if the function returns null. I could try to report null in place of the key, or give an error message instead of resolving the document will nulls.

bshambaugh: scr-1621027899

I have full coverage. Unfortunately, I found some tests that I can create that can fail. I have not included them yet since they use functions included to enable the tests, not functions that are in the code. I think that the same bug may be in the code too. I am going to try to see if I can fix the bug in the test function and the code functions.

test that fail:

test('test a hex string shorter than 33 bytes', () => {
   const inputPublicKeyHex = '36f8964623378bdc068d4bce07ed17c8fa486f9ac0c2613ca3c8c306d7bb6'
   const publicKey_u8a = pubKeyHexToUint8Array(inputPublicKeyHex);
   const pubKeyBytesToXY = mapper.pubKeyBytesToXY(publicKey_u8a);
   expect(pubKeyBytesToXY).toEqual(null);
});


test('test a hex string between 64 and 33 bytes', () => {
   const inputPublicKeyHex = '36f8964623378bdc068d4bce07ed17c8fa486f9ac0c2613ca3c8c306d7bb61cd36717b8ac5e4fea8ad23dc8d0783c2318ee4ad7a80db6e0026ad0b072a24f'
   const publicKey_u8a = pubKeyHexToUint8Array(inputPublicKeyHex);
   const pubKeyBytesToXY = mapper.pubKeyBytesToXY(publicKey_u8a);
   expect(pubKeyBytesToXY).toEqual(null);
});

bshambaugh:
> @bshambaugh please let us know when this is ready for a full review again :)

I'm dipping my toes into the case where I have nulls coming into the functions:

https://gist.github.com/bshambaugh/333a4d28e3fdb275ed20422e97030e04

oed:
There's something wrong with this PR theres 269 changed files. Looks like you included build outputs in the src folder somehow. Please fix this and I'll do a final review!

bshambaugh:
scr-1621607332
scr-1621607383

There is not full test coverage for index.ts

oed (changes requested):
This looks great. Just add the missing dependency `bigint-mod-arith` to the package.json. Will merge after this is done.

Tests are passing locally when adding the missing dependency.

oed:
Still seeing issues with this branch. Can you try to do the following:

**Merge develop into this branch:**
```
git merge develop
```

**Make sure build and lint passes:**
```
npm run lint
// and
npm run build
```

That should be it.

bshambaugh:
> Still seeing issues with this branch. Can you try to do the following:
> 
> **Merge develop into this branch:**
> 
> ```
> git merge develop
> ```
> 
> **Make sure build and lint passes:**
> 
> ```
> npm run lint
> // and
> npm run build
> ```
yes, I still need to work through lint and build
> 
> That should be it.

bshambaugh:
In addition to keyToDidDoc which is used by 
https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/key-did-resolver/src/index.ts , the following functions are exported for testing purposes:

https://github.com/bshambaugh/js-ceramic/blob/develop/packages/key-did-resolver/src/secp256r1.ts#L54
ECPointDecompress

https://github.com/bshambaugh/js-ceramic/blob/develop/packages/key-did-resolver/src/secp256r1.ts#L80
publicKeyToXY

https://github.com/bshambaugh/js-ceramic/blob/develop/packages/key-did-resolver/src/secp256r1.ts#L87
publicKeyHexToUint8ArrayPointPair

https://github.com/bshambaugh/js-ceramic/blob/develop/packages/key-did-resolver/src/secp256r1.ts#L95
publicKeyIntToXY

https://github.com/bshambaugh/js-ceramic/blob/develop/packages/key-did-resolver/src/secp256r1.ts#L102
publicKeyIntToUint8ArrayPointPair

https://github.com/bshambaugh/js-ceramic/blob/develop/packages/key-did-resolver/src/secp256r1.ts#L110
pubKeyBytesToXY

With the exception of pubKeyBytesToXY, all of these functions are known to fail for null and undefined.

bshambaugh:
To my knowledge, everything looks good.

oed (approved):
Looks great 👍 

Btw, I made a PR to the multicodecs table to clarify that the compressed format should be used: https://github.com/multiformats/multicodec/pull/214

oed:
@bshambaugh looks like the docs are not building. My bad for not mentioning this to you. Could you make another PR to make sure that `npm run docs` passes?

bshambaugh:
key-did-resolver$ npm run docs
npm ERR! Missing script: "docs"
npm ERR! 
npm ERR! To see a list of scripts, run:
npm ERR!   npm run

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/ubuntu/.npm/_logs/2021-05-25T13_23_41_860Z-debug.lo

oed:
@bshambaugh We had to revert the change because the docs script didn't pass. Can you please submit a new PR for this feature which includes working docs? 🙏

PR #1262, opened 2021-04-14
fix(core): Cache providers per network

ukstv:
Dirty, but should go.


stbrody (approved):

oed:
Should we add this as a hotfix to master and make a release? @ukstv @stbrody

ukstv:
Vote for it @oed

PR #1266, opened 2021-04-14
feat(common): Rename Doctype to Stream

stbrody:
builds on https://github.com/ceramicnetwork/js-ceramic/pull/1264, also mostly done via sed script


packages/common/src/ceramic-api.ts
```diff
@@ -59,25 +59,25 @@ export interface CeramicApi extends CeramicSigner {
     // loggerProvider: LoggerProvider; // TODO uncomment once logger is available on http-client
 
     /**
-     * Register Doctype handler
-     * @param doctypeHandler - DoctypeHandler instance
+     * Register Stream handler
+     * @param doctypeHandler - StreamHandler instance
      */
-    addDoctypeHandler<T extends Doctype>(doctypeHandler: DoctypeHandler<T>): void;
+    addStreamHandler<T extends Stream>(doctypeHandler: StreamHandler<T>): void;
```
stbrody, line 65: Most places I felt made sense to rename `Doctype` to `Stream`.  This is the one place I'm not 100% on.  I could see calling the class `StreamTypeHandler` and this method `addStreamTypeHandler`, if others prefer that.

oed: Don't have a strong preference. StreamHandler should be fine.

oed (approved):

PR #1269, opened 2021-04-15
feat(common): Change 'sync' option to an enum and refine sync behaviors

stbrody:


ukstv:
Pls, `syncTimeoutMillis` -> `syncTimeout` as seconds. Milliseconds have kind of real-time connotations, and this is not happening here.

stbrody:
> Pls, `syncTimeoutMillis` -> `syncTimeout` as seconds. Milliseconds have kind of real-time connotations, and this is not happening here.

This is a worthwhile suggestion to discuss and one I am genuinely conflicted about.  When I added this flag I did think about making it have second-level granularity, because I agree with you that right now getting sub-second responses in unrealistic, and for the foreseeable future second-level granularity is probably all that could be meaningfully used.

That said, I decided to go with milliseconds here for 2 reasons.  The first is future-proofing.  Networks are getting faster all the time, and there could very well be a future where expecting to sync a stream in less than a second is reasonable.  Restricting users from even being able to express such a request feels shortsighted.  You can always express second-granularity timeouts with milliseconds, but you can't go the other way around.

The other, possibly less relevant reason, is that I can imagine scenarios even today where expressing a sub-second timeout could be desirable to some users.  For example if they are running ceramic in a closed environment disconnected from the global network.  The other, more likely case could be in testing environments, between nodes that might be running on the same machine and communicating via localhost.


All of this is to say, while in general I agree with trying to protect users from shooting themselves in the foot by keeping APIs restricted to minimal possible surface area, this feels like a case we should allow greater expressiveness in the API, and allow users to make their own decisions about what behavior makes sense for their use case.


That said, I do see the argument for changing this to be in seconds, and I could definitely be swayed if there is consensus that that is the better way to go

packages/cli/src/ceramic-daemon.ts
```diff
@@ -95,6 +96,22 @@ function parseQueryObject(opts: Record<string, any>): Record<string, string | bo
   return typedOpts
 }
 
+/**
+ * Converts 'sync' option sent as a bool by outdated http-clients to the current format of an enum.
+ * The old behaviors don't map directly to the new behaviors, so we take the best approximation.
+ * TODO remove this once we no longer need to support clients older than v1.0.0
```
oed, line 102: 🙏🏻

oed (approved):
Lgtm as far as I can tell. Don't have a strong opinion on s vs ms, but I don't imagine that it will ever be below 1s. Note that if the response comes faster the stream will be returned faster.

stbrody:
> Lgtm as far as I can tell. Don't have a strong opinion on s vs ms, but I don't imagine that it will ever be below 1s. Note that if the response comes faster the stream will be returned faster.

Okay, I'll update to `syncTimeoutSeconds`.  I guess worst case scenario is at some point in the future we add a `syncTimeoutMillis` and deprecate `syncTimeoutSeconds`.  Not the end of the world if we have to do that in the future, and lets us set the proper expectations in the meantime

PR #1273, opened 2021-04-15
fix(store): web browsers don't have access to fs

Geo25rey:
When using `js-ceramic` with webpack, nodes can't be used due to the use of `fs` in the Level store. Since `fs` is `undefined` in web browsers, a simple solution is to only use `fs` when it's defined. There seems to only be 1 occurance in `js-ceramic` of this, and I updated it accordingly.


stbrody (approved):
Cool, thanks for the fix!

PR #1281, opened 2021-04-16
chore: Bump ajv to ^8.1.0

ukstv:


packages/core/src/utils.ts
```diff
@@ -1,4 +1,6 @@
-import ajv from "ajv"
+import Ajv from "ajv"
+import addFormats from "ajv-formats"
```
oed, line 2: What does this function do?

oed (commented):
Just a question.

packages/core/src/utils.ts
ukstv: It adds "custom" formats defined in later versions of JSON schema. For example, "date-time". Earlier it was part of the core `ajv` package.

ukstv: > From version 7 Ajv does not include formats defined by JSON Schema specification - these and several other formats are provided by ajv-formats (opens new window) plugin.

https://ajv.js.org/guide/formats.html

oed: ok

oed (approved):

PR #1285, opened 2021-04-16
feat(common): Rename createDocumentFromGenesis to createStreamFromGenesis

stbrody:
Also includes some additional doctype renames

PR #1286, opened 2021-04-16
feat(common): Update createStreamFromGenesis to take 'type' number instead of 'doctype' string

stbrody:
Builds on https://github.com/ceramicnetwork/js-ceramic/pull/1285


oed (approved):
great

PR #1291, opened 2021-04-18
feat(core): Update pubsub messages to use 'stream' instead of 'doc'

stbrody:


oed (changes requested):
This is a protocol breaking change. I think we should wait with this until after ELP and think about a real upgrade path.

stbrody:
> This is a protocol breaking change. I think we should wait with this until after ELP and think about a real upgrade path.

Well I did include backwards compatibility logic in this, but I guess there's no reason this has to happen now.  We could wait and do this as part of the protocol versioning project as the first live protocol message format update.

oed:
> Well I did include backwards compatibility logic in this

Did you forget to push this? Definitely no there afaict.

stbrody:
> > Well I did include backwards compatibility logic in this
> 
> Did you forget to push this? Definitely no there afaict.

It's there.  Check out the `serialize` and `deserialize` functions in pubsub-message.ts

packages/core/src/pubsub/pubsub-message.ts
```diff
@@ -64,7 +64,8 @@ export function serialize(message: PubsubMessage): string {
     case MsgType.QUERY: {
       return JSON.stringify({
         ...message,
-        doc: message.doc.toString(),
+        doc: message.stream.toString(), // todo remove once we no longer support interop with nodes older than v1.0.0
+        stream: message.stream.toString(),
```
oed, line 68: 👀

oed (approved):
Sneaky! Alright my bad. This should be fine then :)

PR #1298, opened 2021-04-19
chore: Bump Filecoin and Polkadot dependency

ukstv:
- Zondax get the act together, so the new packages provide now signatures fully-compatible with Filecoin Lotus
- Quirky Jest loader can not handle [conditional exports](https://nodejs.org/docs/latest-v14.x/api/packages.html#packages_conditional_exports), which are used extensively (Polkadot, multiformats). To handle these, I added custom loader based on WebPack [enhanced-resolve](https://github.com/webpack/enhanced-resolve) and Jest default loader. Although it makes tests a little slower, it eliminates manual module-mapping that is required otherwise.


stbrody:
Seems reasonable to me but I don't think I can meaningfully review the resolver change.  Subbing myself for Paul who might have more context to review

PaulLeCam (approved):
I'm not familiar with `enhanced-resolve` and how Jest's resolver works so I don't have any specific feedback, seems all good if it makes the setup work!

oed (approved):

PR #1302, opened 2021-04-19
chore: Path to jest and eslint

ukstv:


stbrody (approved):
Thanks for the quick fix!

PR #1309, opened 2021-04-21
chore: Bump datastore-s3 to 5.0.0

ukstv:


valmack:
Before this goes in we need to double check which repos are sharded already. I don't think this will work if we start sharding something that previously wasn't. Will report back shortly

packages/ipfs-daemon/src/s3-lock.ts
```diff
@@ -0,0 +1,120 @@
+import path from 'path';
+import S3Store from 'datastore-s3';
+
+/**
+ * Uses an object in an S3 bucket as a lock to signal that an IPFS repo is in use.
+ * When the object exists, the repo is in use. You would normally use this to make
+ * sure multiple IPFS nodes don’t use the same S3 bucket as a datastore at the same time.
+ */
+export class S3Lock {
+  readonly s3: S3Store;
```
valmack, line 10: Why do we need this S3 lock, given we have been operating without it?

packages/ipfs-daemon/src/create-repo.ts
```diff
@@ -0,0 +1,79 @@
...
+        ...storeConfig,
+        sharding: false,
+      },
+      datastore: {
+        ...storeConfig,
+        sharding: false,
+      },
+      pins: {
+        ...storeConfig,
+        sharding: true,
```
valmack, line 74: Not positive this was previously sharded

valmack (commented):
Will check about sharding (after I feed the cat 😄)

packages/ipfs-daemon/src/create-repo.ts
ukstv: It was not.

packages/ipfs-daemon/src/create-repo.ts
```diff
@@ -0,0 +1,74 @@
...
+      keys: S3Store,
+      datastore: S3Store,
+      pins: S3Store,
+    },
+    storageBackendOptions: {
+      root: {
+        ...storeConfig,
+      },
+      blocks: {
+        ...storeConfig,
```
valmack, line 60: We actually do want sharding here for blocks, as that is what we currently have.

valmack (commented):
Great just need to add sharding back to blocks. Confirmed that this is the only store that is sharded in our buckets currently

packages/ipfs-daemon/src/create-repo.ts
ukstv: okay.

ukstv: done.

ukstv: heck, committed in a different branch

ukstv: now it is back.

valmack (approved):
👍

PR #1310, opened 2021-04-21
IPFS S3 block only repo

ukstv:
For s3-block-only mode to work one has to provide the following environment variables:
- `IPFS_LOCAL` - prefix folder for local ipfs data, actual ipfs repo then is `<IPFS_LOCAL>/<IPFS_PATH>` (handles case of `IPFS_PATH` being a first-tier folder like `/ipfs`),
- `IPFS_CREATE_IF_MISSING` - signals to create repo folders if not present.

By default both variables are undefined. If `IPFS_LOCAL` is set, the daemon treats it as a signal for s3-block-only mode.

@zachferland @valmack  Should createIfMissing flag be `true` by default?


packages/ipfs-daemon/src/ipfs-daemon.ts
```diff
@@ -80,6 +84,8 @@ export class IpfsDaemon {
 
         const repo = configuration.ipfsS3RepoEnabled ? createRepo({
             path: configuration.ipfsPath,
+            local: configuration.ipfsLocal,
+            createIfMissing: configuration.ipfsCreateIfMissing
```
valmack, line 88: The default here for `createIfMissing` should be the same as whatever the current default is, which I think is true.

packages/ipfs-daemon/src/make-storage-backends.ts
```diff
@@ -0,0 +1,40 @@
+import DatastoreLevel from 'datastore-level';
```
valmack, line 1: I think the use of the word `local` is too ambiguous the way you have it here.

It seems like the most straightforward thing is an env var for each data store with enumerated options. Something like...

```sh
IPFS_ROOT_STORE=default
IPFS_BLOCKS_STORE=s3
IPFS_KEYS_STORE=default
IPFS_PINS_STORE=default
IPFS_DATASTORE_STORE=default
```

If one of these is omitted, it becomes default. If one of these is s3, the s3 related env vars are required or we throw. In this case it is probably better to remove `IPFS_S3_REPO_ENABLED` as well.

Maybe we name it `IPFS_<datastore>_BACKEND` instead?

ukstv: @valmack `IPFS_BACKEND_<datastore>` sounds fine?

Now configuration is segregated. Local prefix is fully optional. I find it useful still, as a first-tier directory can be unavailable.

packages/ipfs-daemon/src/create-repo.ts
```diff
@@ -16,6 +19,95 @@ const notALock = {
...
+
+function LocalDatastoreLevel(pathPrefix: string | undefined) {
+  return class LocalDatastoreLevel extends DatastoreLevel {
+    constructor(repoPath: string, options: any) {
+      super(localPath(pathPrefix, repoPath), options);
+    }
+  };
+}
+
+function makeBackend(config: any, name: string, options: Options, s3: () => S3, defaultBackend: any) {
```
valmack, line 86: Why pass s3 as function instead of object?

packages/ipfs-daemon/src/ipfs-daemon.ts
```diff
@@ -60,31 +80,40 @@ export class IpfsDaemon {
...
-            healthcheckEnabled: props.healthcheckEnabled ?? (process.env.HEALTHCHECK_ENABLED === 'true'),
+            healthcheckEnabled: props.healthcheckEnabled ?? fromBooleanInput(process.env.HEALTHCHECK_ENABLED, false),
             healthcheckPort: props.healthcheckPort || (process.env.HEALTHCHECK_PORT != null ? parseInt(process.env.HEALTHCHECK_PORT) : 8011),
             logger: props.logger ?? new DiagnosticsLogger(LogLevel.important, false),
         }
 
-        const repo = configuration.ipfsS3RepoEnabled ? createRepo({
+        const repo = createRepo({
             path: configuration.ipfsPath,
+            localPrefix: configuration.ipfsLocalPrefix,
```
valmack, line 103: Can we rename to `localPathPrefix`? That way it is clear exactly what this option is for.

packages/ipfs-daemon/src/create-repo.ts
valmack, line 86: Can we name the function `setIpfsRepoBackend`?
And rename the options param to `ipfsRepoOptions`?

It took me a bit to understand all these moving pieces and more descriptive names will make it quicker.

valmack (approved):
Minor naming requests but this is awesome!

packages/ipfs-daemon/src/ipfs-daemon.ts
valmack: Actually it would be nicer if we eliminate this options and just allow this to work

`IPFS_PATH=/my_dir_prefix/my_dir`

That should work with both s3 and local

packages/ipfs-daemon/src/create-repo.ts
ukstv: It is not required for purely local deployment. Treat `s3` function as a lazy variable.

packages/ipfs-daemon/src/ipfs-daemon.ts
ukstv: `IPFS_PATH` currently means two things at once:
1. Path to a local repository,
2. Bucket name on S3.

In case of purely local or purely s3 deployment, prefix is an unnecessary complication.

For mixed deployment, it might happen that `IPFS_PATH` points to a directory unavailable on a local machine. For example: `IPFS_PATH=/ipfs` in mixed setting makes use of `/ipfs` which might be unavailable on [certain](https://github.com/coreos/fedora-coreos-tracker/issues/270) [operating](https://cloud.google.com/container-optimized-os/docs/concepts/security#filesystem) [systems](https://support.apple.com/en-gb/HT210650).

ukstv: So, I am up for a rename, but I argue it has to stay with us.

packages/ipfs-daemon/src/create-repo.ts
ukstv: Agree. I just omit `ipfs` then, as it is _all_ about IPFS.

PR #1313, opened 2021-04-22
chore: add timestamp to service log

ukstv:


packages/common/src/loggers.ts
```diff
@@ -43,7 +43,10 @@ export class ServiceLogger extends ServiceLoggerBase {
     super.write(message)
 
     if (this.logToFiles) {
-      this.stream.write(util.format(message, '\n').replace(/\n\s*\n$/, '\n'));
+      const now = new Date();
+      // RFC1123 timestamp
+      message = `[${now.toUTCString()}] service=${this.service} ${util.format(message, '\n').replace(/\n\s*\n$/, '\n')}`;
```
stbrody, line 48: ooph, we were doing it for diagnostics logger but not service logger. Fun.

I also don't know why we don't add the timestamp for the console log but only add it to the file log?

Anyway LGTM if Val is happy

valmack (approved):

PR #1317, opened 2021-04-23
chore(core): Small logging improvements

stbrody:


packages/core/src/ceramic.ts
```diff
@@ -483,7 +483,6 @@ class Ceramic implements CeramicApi {
     opts = { ...DEFAULT_LOAD_OPTS, ...opts };
     const streamRef = StreamRef.from(streamId)
     const base$ = await this._loadStream(streamRef.baseID, opts)
-    this._logger.verbose(`Stream ${streamId.toString()} successfully loaded`)
```
stbrody, line 486: This is way too verbose, even for the verbose log level.  I'm seeing this showing up dozens of times per second for the same streamId, in logs from a node being used for the integration tests.  This is a bit worrisome on its own as it suggests we're calling loadStream dozens of times in a second for some reason, on a node that has very low traffic and is likely just trying to serve a single request related to that stream.

So at some point we should look into why `loadStream` is being called so much.  But for now, this log message isn't necessary.  The combination of the log message for when the genesis commit is loaded and when the stream gets a new tip gives us enough info about when we're loading streams from the network, and we probably don't need or want to log every time we get a stream from the cache.

packages/core/src/state-management/state-manager.ts
```diff
@@ -196,7 +197,8 @@ export class StateManager {
           }
         }),
         catchError((error) => {
-          this.logger.err(error);
+          this.logger.err(`Error while anchoring stream ${state$.id.toString()}: ${error.message}`)
+          this.logger.err(error); // Log full error to get the stack trace
```
valmack, line 201: Probably better to put the stack trace in the same logger call, as these are not guaranteed to be sequential

```js
this.logger.err(`Error while anchoring stream ${state$.id.toString()}: ${error.message}\n${error}`)
```

valmack (approved):

PR #1324, opened 2021-04-27
fix(cli): Allow large requests to http API

stbrody:


packages/cli/src/ceramic-daemon.ts
```diff
@@ -124,7 +124,7 @@ class CeramicDaemon {
     this.diagnosticsLogger = ceramic.loggerProvider.getDiagnosticsLogger()
     this.app = addAsync(express());
     this.app.set('trust proxy', true)
-    this.app.use(express.json())
+    this.app.use(express.json({limit: '1mb'}))
```
stbrody, line 127: 1mb was chosen mostly arbitrarily.  It seemed like we might have http requests larger than 256kb despite that being the commit size limit.  Wasn't really sure how much padding to add.  Other examples I saw online were setting this option to `50mb`, so `1mb` still seemed pretty conservative despite being 4x the commit size limit. Happy to turn this down to something smaller though if you think it prudent

oed (approved):

PR #1329, opened 2021-04-28
feat(cli): allow daemon hostname to be configured

valmack:
For some reason our integration tests starting failing with this

```
FetchError: request to http://localhost:7007/api/v0/streams failed, reason: socket hang up

FetchError: request to http://localhost:7007/api/v0/streams failed, reason: connect ECONNREFUSED 127.0.0.1:7007
```

*Potentially* we need to serve the daemon from 0 address when running it in docker.


packages/cli/src/ceramic-daemon.ts
```diff
@@ -137,8 +139,9 @@ class CeramicDaemon {
   async listen(): Promise<void> {
     return new Promise<void>((resolve) => {
       const port = this.opts.port || DEFAULT_PORT
-      this.server = this.app.listen(port, () => {
-        this.diagnosticsLogger.imp('Ceramic API running on port ' + port)
+      const hostname = this.opts.hostname || DEFAULT_HOSTNAME
```
stbrody, line 142: By setting this to `localhost` by default, now are we going to start rejecting connections not from localhost by default?

stbrody (approved):
Per offline discussion, it seems like this doesn't make things any worse then they were already.  LGTM

PR #1330, opened 2021-04-29
feat(cli): use 0 address for default hostname

valmack:
Less friction to serve it from any address so that a proxy is not required


stbrody (approved):
:pray:

PR #1348, opened 2021-05-03
fix: accept multiple pubsub responses

ukstv:


packages/core/src/state-management/state-manager.ts
```diff
@@ -50,10 +50,12 @@ export class StateManager {
    */
   async sync(state$: RunningState, timeoutMillis: number): Promise<void> {
     const tip$ = this.dispatcher.messageBus.queryNetwork(state$.id);
-    const tip = await tip$.pipe(timeoutWith(timeoutMillis, of(undefined))).toPromise();
-    if (tip) {
-      await this._handleTip(state$, tip);
-    }
+    await tip$
```
oed, line 53: Should this variable be called `tips$` now?

oed (commented):
Can we talk through this on the voice channel after meetings? :)

packages/core/src/state-management/state-manager.ts
ukstv: It is `Observable<CID>`,  not `Observable<CID[]>` so still `tip$`.

oed: I c

packages/core/src/__tests__/state-manager.test.ts
```diff
@@ -295,3 +303,84 @@ test('should announce change to network', async () => {
...
+    });
+  });
+  test('not handle delayed', async () => {
+    const stateManager = ceramic.repository.stateManager;
+    const amount = 10;
+    const response = responseTips(amount);
+    ceramic.dispatcher.messageBus.queryNetwork = () =>
+      from(response).pipe(
+        concatMap(async (value, index) => {
+          await new Promise((resolve) => setTimeout(resolve, index * MAX_RESPONSE_INTERVAL * 0.3));
```
stbrody, line 361: could you maybe add a comment to the start of this test explaining what it does and how the timing math works out?  It's not obvious to me reading this why we're multiplying this by 0.3

packages/core/src/pubsub/__tests__/message-bus.test.ts
```diff
@@ -74,30 +74,69 @@ test('unsubscribe', async () => {
...
 
-  test('timeout: return undefined', async () => {
+  test('late first message', async () => {
+    const pubsub = (new Observable<ResponseMessage>((subscriber) => {
+      setTimeout(() => {
+        subscriber.next(responseMessage1);
+        setTimeout(() => {
+          subscriber.next(responseMessage2);
+          subscriber.complete();
+        }, MAX_RESPONSE_INTERVAL * 2);
```
stbrody, line 129: So I believe the test above ('delayed message') is testing what happens when there's plenty of time in the outer 3 second syncTimeoutSeconds, but the time between the two responses is larger than the smaller 300ms inner MAX_RESPONSE_INTERVAL, so the second response doesn't get processed because it came in too late.

There's another case, which I *think* is what this test is trying to test? That case is where a response comes in close to the end of the outer 3 second syncTimeoutSeconds, and then another response comes in in *less* than the 300ms MAX_RESPONSE_INTERVAL, but by that point the 3 second syncTimeoutSeconds time limit is up, so even though the second response came in less than 300ms after the first, it doesn't get processed because we've exhausted our outer time limit.

Maybe that's not what this test was trying to test, but in any case I don't think this tests that since the second response comes in more than MAX_RESPONSE_INTERVAL after the first. Seems like this would be a good case to test, but I think maybe it would need to happen in state-manager.test.ts, since it's the state manager that enforces the outer 3 second syncTimeoutSeconds?

Hopefully I explained that right and it's clear what I'm referring to, please let me know if anything isn't clear

stbrody: I guess the 'stop after timeout' test in the state manager tests basically tests the case I'm talking about.  So then I guess I just don't understand what case this 'late first message' test here is trying to test?

stbrody (approved):
Some of these tests have some tricky logic to exercise some very specific cases.  I think some comments to explain what case they are trying to exercise would make things clearer to future readers of the tests. But functionality change and test coverage both look good, so this PR LGTM.  If you have time to clarify some of the questions I asked in comments great, but don't consider it blocking to getting this merged in.

oed (approved):

packages/core/src/pubsub/__tests__/message-bus.test.ts
ukstv: As stated in the name, it tests for the case when the first received message is late, that is arrives after the in-between timeout has passed from the start time. The timeout must only be counted between the messages, not between start time and the first message.

stbrody: Ah got it, that makes sense.  Thanks for clarifying.

PR #1352, opened 2021-05-04
fix(stream-tile): TileDocument content can always be null or undefined

stbrody:
Makes the rest of the TileDocument APIs consistent with that of `TileDocument.create()`, which accepts `T | null | undefined`

### DIFF
--- a/packages/stream-tile/src/tile-document.ts
+++ b/packages/stream-tile/src/tile-document.ts
@@ -186,7 +186,7 @@ export class TileDocument<T = Record<string, any>> extends Stream {
      * @param metadata - Changes to make to the metadata.  Only fields that are specified will be changed.
      * @param opts - Additional options
      */
-    async update(content: T, metadata?: TileMetadataArgs, opts: UpdateOpts = {}): Promise<void> {
+    async update(content: T | null | undefined, metadata?: TileMetadataArgs, opts: UpdateOpts = {}): Promise<void> {
         opts = { ...DEFAULT_UPDATE_OPTS, ...opts };
         const updateCommit = await this.makeCommit(this.api, content, metadata)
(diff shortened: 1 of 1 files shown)


packages/stream-tile/src/tile-document.ts
```diff
@@ -186,7 +186,7 @@ export class TileDocument<T = Record<string, any>> extends Stream {
      * @param metadata - Changes to make to the metadata.  Only fields that are specified will be changed.
      * @param opts - Additional options
      */
-    async update(content: T, metadata?: TileMetadataArgs, opts: UpdateOpts = {}): Promise<void> {
+    async update(content: T | null | undefined, metadata?: TileMetadataArgs, opts: UpdateOpts = {}): Promise<void> {
```
oed, line 189: You could define a type that is `T | null | undefined` so that you don't have to have it thrice in the code.

oed (approved):
Looks good. Suggested an improvement.

packages/stream-tile/src/tile-document.ts
stbrody: I just took a crack at this, you can see it here: https://github.com/ceramicnetwork/js-ceramic/commit/3f6a9af11314fec50def1f3a6613f7fb34c99be3

I don't really love it, I think it just adds more indirection and makes it harder to see what's actually going on.

Maybe if I could declare the type alias within the body of TileDocument so that the type aliases directly to the `T` that TileDocument is specialized on, without needing to further specialize the TileContent type to `T` as well. If I could do it that way then I'd like it a bit more, but I couldn't figure out a way to do that. Quick google search suggests type aliases aren't allowed within class definitions.

oed: Ah, ok didn't think about this limitation. Agree that it's a bit weird.

PR #1357, opened 2021-05-06
feat: Expose CeramicDaemon from cli package

ukstv:
I would like to use CeramicDaemon in 3id-connect for testing purposes.


oed (approved):
Makes sense. This should likely be exposed as a separate package in the future.

PR #1371, opened 2021-05-11
feat(ipfs-topology): update base peerlist

valmack:
Updates clay subdomains and adds mainnet multiaddresses


valmack:
Need to add mainnet to peer file url as well

ukstv (approved):

PR #1374, opened 2021-05-12
feat(ipfs-topology): map elp network to mainnet peerlists

valmack:
Support for `elp` network. Using switch statements to avoid code duplication (aka human error)

### DIFF
--- a/packages/ipfs-topology/src/ipfs-topology.ts
+++ b/packages/ipfs-topology/src/ipfs-topology.ts
@@ -2,7 +2,7 @@ import fetch from "cross-fetch";
 import { Networks } from "@ceramicnetwork/common";
 import type { DiagnosticsLogger, IpfsApi } from "@ceramicnetwork/common";
 
-const PEER_FILE_URLS = (ceramicNetwork: string): string | null => {
+const PEER_FILE_URLS = (ceramicNetwork: Networks): string | null => {
   switch (ceramicNetwork) {
     case Networks.MAINNET:
     case Networks.ELP:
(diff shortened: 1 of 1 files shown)


packages/ipfs-topology/src/ipfs-topology.ts
```diff
@@ -1,27 +1,40 @@
...
+const PEER_FILE_URLS = (ceramicNetwork: string): string | null => {
+  switch (ceramicNetwork) {
+    case Networks.MAINNET:
+    case Networks.ELP:
+      return "https://raw.githubusercontent.com/ceramicnetwork/peerlist/main/mainnet.json";
+    case Networks.TESTNET_CLAY:
+      return "https://raw.githubusercontent.com/ceramicnetwork/peerlist/main/testnet-clay.json";
+    case Networks.DEV_UNSTABLE:
+      return "https://raw.githubusercontent.com/ceramicnetwork/peerlist/main/dev-unstable.json";
+    default:
```
stbrody, line 14: If there's a `default` case then you don't get compile time checking that no case was missed, at which point using a switch statement doesn't seem to add value over the object form that was there previously.  Probably should drop the default case and add a case for in-memory and local networks that just explicitly returns null

valmack: Updated this to add a default unhandled case.

Missing a case

> src/ipfs-topology.ts:18:13 - error TS2322: Type 'import("/Users/val/Code/js-ceramic/packages/common/lib/networks").Networks' is not assignable to type 'never'.
> 18       const unhandledCase: never = ceramicNetwork;
               ~~~~~~~~~~~~~
> Found 1 error.

Not missing a case, it builds successfully.

stbrody: weird.  Oh javascript, always so wonky. Glad you were able to figure out how to make it work though!

stbrody (approved):
LGTM

PR #1377, opened 2021-05-12
fix(core): Increase max anchor poll timeout

stbrody:


package-lock.json
```diff
@@ -1,18200 +1,7 @@
 {
   "name": "root",
-  "lockfileVersion": 2,
```
stbrody, line 3: I'm not really sure why such a huge change to package-lock.json got included here, or if that's a problem? @oed @ukstv

ukstv: It's me and npm. During work with 3id-connect I had to use npm v7 on Node v14. On the latest PR with TS bump I did not downgrade the npm to v6. Since js-ceramic repo still uses npm v6, this PR would revert the package-lock.json to its proper state.

Unless, of course, there is a decision to move to npm v7.

ukstv (approved):

PR #1380, opened 2021-05-13
fix (core,http-client): Don't parse failed HTTP responses

stbrody:
Goal is to provide more useful failure messages when we see failures like this one:
```
'Anchor failed for commit bagcqceradpc2jmxmesy4rd3xrocapfu7qfyny6k5jd3w3kvhuwiaec26e34q of stream kjzl6cwe1jw148ot31rno7z5s1pnpg7gl9g0rit45rjrcbwcbixj2zc5xdlj22h: invalid json response body at https://cas-dev-internal.3boxlabs.com/api/v0/requests/bagcqceradpc2jmxmesy4rd3xrocapfu7qfyny6k5jd3w3kvhuwiaec26e34q reason: Unexpected token < in JSON at position 0'
```

I believe this was due to places like [this](https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/core/src/anchor/ethereum/ethereum-anchor-service.ts#L202-L205) where we call `.json()` on an HTTP response object even when the HTTP status may be non-ok.

This PR also takes the opportunity to de-dupe and clean up the various places places where we use cross-fetch to make HTTP requests


oed (approved):
lgtm

ukstv (approved):

PR #1391, opened 2021-05-17
fix(core,http-client): Add setter for 'did' property on CeramicApi implmentations

stbrody:


stbrody:
> The setter, or property, needs to be present on the _CeramicApi_ as well!

It already is, via the CeramicSigner interface: https://github.com/ceramicnetwork/js-ceramic/blob/cbad156a225a405876db177ec928a51253c0e348/packages/common/src/ceramic-api.ts#L49

oed (approved):
> It already is, via the CeramicSigner interface: 

Ah, my bad!

PR #1392, opened 2021-05-18
Feat/#1343: TileDocument.create should error if 'deterministic' is true and content is provided

smrz2001:


packages/stream-tile-handler/src/__tests__/tile-document.test.ts
```diff
@@ -243,8 +251,8 @@ describe('TileDocumentHandler', () => {
     })
 
     it('creates genesis records deterministically if deterministic:true is specified', async () => {
-        const record1 = await TileDocument.makeGenesis(context.api, RECORDS.genesis.data, { deterministic: true })
-        const record2 = await TileDocument.makeGenesis(context.api, RECORDS.genesis.data, { deterministic: true })
+        const record1 = await TileDocument.makeGenesis(context.api, null, { deterministic: true })
```
stbrody, line 254: Nitpick: it would be nice to put *something* into the metadata just so we see that the deterministic documents have something in them.  I'd just add a family and a tag

packages/stream-tile/src/tile-document.ts
```diff
@@ -265,6 +265,8 @@ export class TileDocument<T = Record<string, any>> extends Stream {
         const header = headerFromMetadata(metadata)
         if (!metadata?.deterministic) {
             header.unique = uint8arrays.toString(randomBytes(12), 'base64')
+        } else if (content) {
+            throw new Error('Content specified for deterministic creation')
```
stbrody, line 269: How about `Initial content must be null when creating a deterministic document`?

ukstv: Though about mentioning that it is now a stream, but really it is a _tile document_

+1

ukstv:
LGTM apart failing CircleCI tests

packages/core/src/__tests__/state-manager.test.ts
```diff
@@ -233,13 +233,13 @@ describe('rewind', () => {
   });
 
   test('return read-only snapshot', async () => {
-    const stream1 = await TileDocument.create<any>(ceramic, INITIAL_CONTENT, { deterministic: true }, { syncTimeoutSeconds: 0 });
-    await anchorUpdate(ceramic, stream1);
+    const metadata = { deterministic: true, family: "family", tags: ["x", "y"], controllers: controllers}
+    const stream1 = await TileDocument.create<any>(ceramic, null, metadata, { syncTimeoutSeconds: 0 });
```
stbrody, line 237: maybe add `anchor:false` to the opts here if the test isn't going wait for the anchor to succeed. Otherwise the test can fail nondeterministically due to a race between the anchor and the update

packages/core/src/__tests__/state-manager.test.ts
```diff
@@ -233,13 +233,13 @@ describe('rewind', () => {
...
     await stream1.update({ abc: 321, def: 456, gh: 987 });
     await anchorUpdate(ceramic, stream1);
 
     const ceramic2 = await createCeramic(ipfs, { anchorOnRequest: false });
-    const stream2 = await TileDocument.create(ceramic, INITIAL_CONTENT, { deterministic: true }, { anchor: false, syncTimeoutSeconds: 0 });
+    const stream2 = await TileDocument.create(ceramic, null, metadata, { anchor: false, syncTimeoutSeconds: 0 });
```
stbrody, line 242: I kinda feel like all this messing with deterministic documents is distracting for this test given that that isn't what this test is supposed to be about.  What if you just made the stream a normal non-deterministic stream and changed this line to `TileDocument.load(ceramic, stream1.id)`?

packages/core/src/state-management/__tests__/cache-eviction.test.ts
```diff
@@ -7,10 +7,12 @@ import { delay } from '../../pubsub/__tests__/delay';
 
 let ipfs: IpfsApi;
 let ceramic: Ceramic;
+let metadata: any;
 
 beforeAll(async () => {
   ipfs = await createIPFS();
   ceramic = await createCeramic(ipfs, { streamCacheLimit: 1, anchorOnRequest: false });
+  metadata = { deterministic: true, controllers: [ceramic.did.id], family: "family", tags: ["x", "y"] }
```
stbrody, line 15: If more than 1 test uses the same deterministic metadata to create streams then the actions of one test can have impact on the other tests.  I think each test that needs to use deterministic document lookups should make their own unique metadata.  Could add a helper for generating the metadata to make that cleaner though

packages/core/src/state-management/__tests__/cache-eviction.test.ts
```diff
@@ -106,6 +109,7 @@ test('RunningState stops updating after evicted', async () => {
   expect(runningState1.isStopped).toBeTruthy(); // RunningState is stopped after eviction
 
   const stream2 = await createDocument();
+  await stream2.update(INITIAL);
```
stbrody, line 112: this update is unnecessary

packages/core/src/state-management/__tests__/cache-eviction.test.ts
```diff
@@ -106,6 +109,7 @@ test('RunningState stops updating after evicted', async () => {
   expect(runningState1.isStopped).toBeTruthy(); // RunningState is stopped after eviction
 
   const stream2 = await createDocument();
```
stbrody, line 111: This test too I feel like we could make unique documents and just use explicit document loads, rather than deterministic creates, without affecting what the test is actually trying to test

stbrody: Actually it looks like this is probably the case for all the tests in this file.  Nothing about this test is explicitly about testing the behavior of deterministic document lookups.  I'm pretty sure they were just written this way because it was convenient and easy at the time. Now I think we'd be better served by explicit load operations rather than jumping through hoops to keep using deterministic documents when nothing about the core of the test needs that

packages/stream-tile-handler/src/__tests__/tile-document.test.ts
```diff
@@ -185,27 +185,35 @@ describe('TileDocumentHandler', () => {
     })
 
     it('makes genesis record correctly', async () => {
-        const record1 = await TileDocument.makeGenesis(context.api, RECORDS.genesis.data, { deterministic: true })
-        expect(record1).toBeDefined()
+        const record = await TileDocument.makeGenesis(context.api, RECORDS.genesis.data)
```
stbrody, line 188: don't worry about changing this now, but for future reference if you're ever working on an old test and see a reference to "records" on a test you're changing anyway it'd be nice to use that opportunity to change it to "commits" which is the current terminology for these things

stbrody (commented):
a few comments around some unintended consequences to some of the updated tests.  Also some suggestions on replacing usage of deterministic document creation/lookup with non-deterministic documents and explicit loads, but those should be considered optional.

packages/core/src/__tests__/ceramic.test.ts
```diff
@@ -153,16 +155,20 @@ describe('Ceramic integration', () => {
...
 
     await anchorUpdate(ceramic1, stream1)
 
     // Through a different ceramic instance create a new stream with the same contents that will
     // therefore resolve to the same genesis record and thus the same streamId.  Make sure the new
     // Document object can see the updates made to the first Document object since they represent
     // the same Document in the network.
-    const stream3 = await TileDocument.create(ceramic3, {test: 321}, {deterministic: true})
+    const stream3 = await TileDocument.create(ceramic3, null, metadata)
+    await stream3.update({ test: 321 })
```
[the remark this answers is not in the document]
smrz2001: Ahh, yes, I see.

packages/core/src/state-management/__tests__/cache-eviction.test.ts
smrz2001: Ok, even though we're creating fresh Ceramic/IPFS instances before each test and closing them after? Is it because of the parallelism of the test execution?

stbrody: We don't create a new ceramic for each test.  `beforeAll`/`afterAll` run once per suite. `beforeEach`/`afterEach` run for each test, but that's not what we're using here.

smrz2001: Darn, you're right. I always get confused by the names.

smrz2001: Based on [this](https://github.com/ceramicnetwork/js-ceramic/pull/1392#discussion_r638117147) comment, I just removed `metadata` and used `TileDocument.load()` with non-deterministic documents.

packages/core/src/__tests__/state-manager.test.ts
smrz2001: What's the right principle to follow when calling `anchorUpdate()`? Does your comment apply just to this one test case or in more places where `anchorUpdate()` is being called?

stbrody: `anchorUpdate` works bests for tests that have their ceramic node set to `anchorOnRequest: false`, since `anchorUpdate` [instructs the anchor service](https://github.com/ceramicnetwork/js-ceramic/blob/62e87b19d36c9ce8dce76323f61004980c030b6e/packages/core/src/state-management/__tests__/anchor-update.ts#L14) to perform an on-demand anchor.  So the `anchor` opt to `TileDocument{create,update}` controls whether that specific operation sends an anchor request to the anchor service.  If `anchorOnRequest` is set to true (the default behavior), then the anchor service will automatically perform an anchor at some point after receiving the request. If `anchorOnRequest` is false, then the anchor service will only perform the anchor if the `anchor()` method is explicitly called against the anchor service.

So the combination of `anchorOnRequest: false` + using `anchorUpdate` (which manually triggers the anchor), is the current best practice for writing reliably deterministic tests that perform anchors. There are still, however, lots of tests out there that are not written so deterministically, so you need to be careful when copy+pasting from or updating existing tests.

packages/core/src/state-management/__tests__/cache-eviction.test.ts
```diff
@@ -191,10 +175,10 @@ describe('evicted then subscribed', () => {
   });
 
   test('pinned', async () => {
-    const stream1 = await TileDocument.create(ceramic, null, metadata, { syncTimeoutSeconds: 0 });
+    const stream1 = await TileDocument.create(ceramic, { foo: Math.random().toString() }, null, { syncTimeoutSeconds: 0 });
```
stbrody, line 178: nitpick: you don't really need the Math.random since you're not making a deterministic document anymore, the document will automatically have a random number added to the metadata to make it unique

stbrody (approved):
LGTM. Nice work!

packages/core/src/__tests__/state-manager.test.ts
smrz2001: Thanks for explaining!!

packages/core/src/state-management/__tests__/cache-eviction.test.ts
smrz2001: Yes, makes sense. Fixing that.

smrz2001: Changing it to `await TileDocument.create(ceramic, { foo: 'bar' })` since `syncTimeoutSeconds` gets defaulted to 0 when `deterministic` is absent (i.e. no `metadata` is passed) or is `false`.

smrz2001: I guess that applies elsewhere in this file as well lol, but I'll come around to those tests some other time :)

Issue #1394, opened 2021-05-20
Create a raw_data endpoint to display current stream contents


dysbulic:
For uses where the accessing system doesn't speak Ceramic, like NFTs, it would be useful to have an endpoint that returns the current state of the stream as a standalone document.

The current suggested name for such an endpoint is `/raw_data/:streamid`.

PR #1395, opened 2021-05-20
feat(daemon): add raw_data endpoint

dysbulic:
This PR addresses #1394 and adds a `/raw_data` endpoint to the Ceramic daemon which returns the current `content` of a stream.

The purpose is to allow programs without knowledge of Ceramic to access the data in streams.


packages/cli/src/ceramic-daemon.ts
```diff
@@ -388,6 +392,20 @@ export class CeramicDaemon {
     res.json(response)
   }
 
+  /**
+   * Render the most recent version of a stream's contents
+   */
+  async rawData (req: Request, res: Response): Promise<void> {
+    const opts = parseQueryObject(req.query)
+    const stream = await this.ceramic.loadStream(req.params.streamid, opts)
+    const state = StreamUtils.serializeState(stream.state)
```
stbrody, line 401: Instead of this, I think it would make more sense to add a `raw_data` accessor to the [Stream type](https://github.com/ceramicnetwork/js-ceramic/blob/881d7f0f17de820290ba6b5b7f4b19e00d2eed6c/packages/common/src/stream.ts#L130).  The `raw_data` accessor can work very similarly to the [metadata accessor](https://github.com/ceramicnetwork/js-ceramic/blob/881d7f0f17de820290ba6b5b7f4b19e00d2eed6c/packages/common/src/stream.ts#L145), just returning the `content` field instead of the `metadata` field.

packages/cli/src/__tests__/ceramic-daemon.test.ts
```diff
@@ -290,6 +290,27 @@ describe('Ceramic interop: core <> http-client', () => {
         expect(StreamUtils.serializeState(docV5Core.state)).toEqual(StreamUtils.serializeState(docV5Client.state))
     })
 
+    it('can get stream contents from /raw_data', async () => {
+      const content1 = {test: 123}
+      const content2 = {test: 456, test2: 'abc'}
+      const content3 = {test2: 'def'}
+      const doc = await TileDocument.create(core, content1)
+
+      await anchorDoc(doc)
```
stbrody, line 299: Since you're not testing anything related to anchoring in this test, you can remove the calls to anchorDoc and add `{ anchor: false }` to the `create` and `update` calls, which will speed up the test

packages/cli/src/__tests__/ceramic-daemon.test.ts
```diff
@@ -290,6 +290,27 @@ describe('Ceramic interop: core <> http-client', () => {
...
+      await doc.update(content2)
+      await anchorDoc(doc)
+      await doc.update(content3)
+      await anchorDoc(doc)
+
+      let returnValue
+      await daemon.rawData(
```
stbrody, line 306: It might be a more realistic test to use [fetchJson](https://github.com/ceramicnetwork/js-ceramic/blob/881d7f0f17de820290ba6b5b7f4b19e00d2eed6c/packages/common/src/utils/http-utils.ts#L9) to hit the HTTP endpoint rather than calling into the daemon object directly.

stbrody (changes requested):
Basic approach looks good, a few suggestions inline.  Thanks for taking on the work to do this and make a PR yourself!

packages/cli/src/__tests__/ceramic-daemon.test.ts
dysbulic: I tried [adding `anchor: false` to the various calls](https://github.com/ceramicnetwork/js-ceramic/pull/1395/files#diff-077b6994383a23f7d2210366626822554ebbcde4d9de8e8a5f94a93a7d1b1141R298-R300) and the test fails as through the updates are not applied:

```
● Ceramic interop: core <> http-client › can get stream contents from /raw_data

    expect(received).toEqual(expected) // deep equality

    - Expected  - 1
    + Received  + 1

      Object {
    -   "test2": "def",
    +   "test": 123,
      }
```

stbrody: Most likely because the current version lives in `state.next.content` not in `state.content`.  If you switch the raw_data endpoint to use the new `rawData` accessor you added to the Stream class, that should fix it.

ukstv:
IMO, the endpoint is better be `/:stream_id/content`, part of `streamsRouter`.

ukstv:
And +1 to @stbrody comments.

dysbulic:
> IMO, the endpoint is better be `/:stream_id/content`, part of `streamsRouter`.

I thought that `content` made more sense for the name given the names of the elements of the objects. @oed said that is [too closely tied to existing stream types](https://discordapp.com/channels/682786569857662976/801539259353923655/843422531059056650), thus the name `raw_data` was used.

packages/cli/src/__tests__/ceramic-daemon.test.ts
dysbulic: Done. I needed to [expose the port](https://github.com/dysbulic/js-ceramic/blob/feat/raw_data-endpoint/packages/cli/src/ceramic-daemon.ts#L142) in the daemon in order to know where to connect. I also [exposed the hostname](https://github.com/dysbulic/js-ceramic/blob/feat/raw_data-endpoint/packages/cli/src/ceramic-daemon.ts#L143) b/c the two generally go together and it looked lopsided w/ just one.

packages/cli/src/__tests__/ceramic-daemon.test.ts
dysbulic: Done. Tests are passing now.

packages/cli/src/ceramic-daemon.ts
dysbulic: Done.

oed:
> I thought that content made more sense for the name given the names of the elements of the objects. @oed said that is too closely tied to existing stream types, thus the name raw_data was used.

Yeah, `content` doesn't really make sense for Caip10Links for example. I think `only-data` might be more descriptive though!

stbrody:
> > I thought that content made more sense for the name given the names of the elements of the objects. @oed said that is too closely tied to existing stream types, thus the name raw_data was used.
> 
> Yeah, `content` doesn't really make sense for Caip10Links for example. I think `/only-data` might be more descriptive though!

Personally I like `raw_data` better than `only_data`.  I like it because the important part in my mind isn't that we're getting the data instead of the metadata, it's that we're getting the data in the raw form from the state object, without any of the streamtype-specific presentation of that data that the more standard APIs provide. The object returned from `raw_data` might require some streamtype specific knowledge to know how to interpret/transform into something useful.

But I also don't feel super strongly.  If you prefer `only_data` @oed that's fine by me.

stbrody (approved):
LGTM

@oed, care to make a final ruling on `raw_data` vs `only_data` or something else?

stbrody:
After a lengthy discussion this morning @oed and I decided to expose `content` as a primary accessor on the Stream type, and change this new endpoint to be at `/api/v0/streams/contents`.  I have gone ahead and updated this PR to incorporate those changes, as well as to update it to current develop branch.

@oed, can you take a final look at this PR and merge it if it looks good?

packages/cli/src/ceramic-daemon.ts
```diff
@@ -137,10 +139,10 @@ export class CeramicDaemon {
 
   async listen(): Promise<void> {
     return new Promise<void>((resolve) => {
-      const port = this.opts.port || DEFAULT_PORT
-      const hostname = this.opts.hostname || DEFAULT_HOSTNAME
-      this.server = this.app.listen(port, hostname, () => {
-        this.diagnosticsLogger.imp(`Ceramic API running on ${hostname}:${port}'`)
+      this.port = this.opts.port || DEFAULT_PORT
```
ukstv, line 142: @dysbulic Nope. Properties like this should be initialized in the constructor.

packages/cli/src/ceramic-daemon.ts
```diff
@@ -187,10 +189,33 @@ export class CeramicDaemon {
...
+
+    app.use('/api/v0', baseRouter)
+    baseRouter.use('/commits', commitsRouter)
+    baseRouter.use('/documents', documentsRouter)
+    baseRouter.use('/multiqueries', multiqueriesRouter)
+    baseRouter.use('/node', nodeRouter)
+    baseRouter.use('/pins', pinsRouter)
+    baseRouter.use('/records', recordsRouter)
+    baseRouter.use('/streams', streamsRouter)
+    streamsRouter.use('/contents', contentsRouter) // sub-endpoint of 'streams'
```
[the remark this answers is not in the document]
ukstv: Non-actionable, probably: +1 to the current `/:streamid/content`, +1 to `contents -> content` rename.

ukstv (approved):
Pls, @dysbulic fix initialisation of the properties, otherwise LTGM.

I would really, really love to have url endpoint as singular `content`, but I defer to @oed and @stbrody here. Having "content" in code and plural "contents" in url is a confusing to me.

packages/cli/src/__tests__/ceramic-daemon.test.ts
```diff
@@ -290,6 +290,22 @@ describe('Ceramic interop: core <> http-client', () => {
         expect(StreamUtils.serializeState(docV5Core.state)).toEqual(StreamUtils.serializeState(docV5Client.state))
     })
 
+    it('can get stream contents from /streams/contents', async () => {
```
oed, line 293:
```suggestion
    it('can get stream contents from /streams/content', async () => {
```

packages/cli/src/__tests__/ceramic-daemon.test.ts
```diff
@@ -290,6 +290,22 @@ describe('Ceramic interop: core <> http-client', () => {
...
+      const content1 = {test: 123}
+      const content2 = {test: 456, test2: 'abc'}
+      const content3 = {test2: 'def'}
+
+      const doc = await TileDocument.create(core, content1, null, { anchor: false })
+      await doc.update(content2, null, { anchor: false })
+      await doc.update(content3, null, { anchor: false })
+
+      const json = await fetchJson(
+        `http://localhost:${daemon.port}/api/v0/streams/contents/${doc.id}`
```
oed, line 303:
```suggestion
        `http://localhost:${daemon.port}/api/v0/streams/content/${doc.id}`
```

packages/cli/src/ceramic-daemon.ts
```diff
@@ -187,10 +189,33 @@ export class CeramicDaemon {
...
+    pinsRouter.use(errorHandler(this.diagnosticsLogger))
+    recordsRouter.use(errorHandler(this.diagnosticsLogger))
+    streamsRouter.use(errorHandler(this.diagnosticsLogger))
+    contentsRouter.use(errorHandler(this.diagnosticsLogger))
+
 
     commitsRouter.getAsync('/:streamid', this.commits.bind(this))
     multiqueriesRouter.postAsync('/', this.multiQuery.bind(this))
     streamsRouter.getAsync('/:streamid', this.state.bind(this))
+    contentsRouter.getAsync('/:streamid', this.contents.bind(this))
```
oed, line 218:
```suggestion
    contentsRouter.getAsync('/:streamid', this.content.bind(this))
```

packages/cli/src/ceramic-daemon.ts
```diff
@@ -388,6 +394,15 @@ export class CeramicDaemon {
     res.json(response)
   }
 
+  /**
+   * Render the most recent version of a stream's contents
+   */
+  async contents(req: Request, res: Response): Promise<void> {
```
oed, line 400:
```suggestion
  async content(req: Request, res: Response): Promise<void> {
```

oed (commented):
These are just suggestions, need to be verified to be working.

stbrody:
Updates based on last round of comments made.  Tests passed locally.  Back to you @oed for final approval

dysbulic:
> Updates based on last round of comments made.

Thanks for doing this. I was getting a failing test in another part of the system and am happy to not have to figure out what's going on right now.

v-stickykeys (commented):
I'd like to second @ukstv and propose that we generally aim for REST so that things are predictable/standardized rather than haphazard.

He already mentioned above but this would (1) use the streams router and add a path there instead of creating a new router and (2) change `/streams/content/<streamId>` to `/streams/<streamId>/content`.

Subtle but I think will be effective moving forward.

Additionally, want to check on the ordering of routes to ensure errors continue to propagate correctly.

packages/cli/src/ceramic-daemon.ts
```diff
@@ -187,10 +190,33 @@ export class CeramicDaemon {
...
+    pinsRouter.use(errorHandler(this.diagnosticsLogger))
+    recordsRouter.use(errorHandler(this.diagnosticsLogger))
+    streamsRouter.use(errorHandler(this.diagnosticsLogger))
+    contentRouter.use(errorHandler(this.diagnosticsLogger))
+
 
     commitsRouter.getAsync('/:streamid', this.commits.bind(this))
     multiqueriesRouter.postAsync('/', this.multiQuery.bind(this))
     streamsRouter.getAsync('/:streamid', this.state.bind(this))
+    contentRouter.getAsync('/:streamid', this.content.bind(this))
```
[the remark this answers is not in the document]
stbrody: Done

packages/cli/src/ceramic-daemon.ts
```diff
@@ -187,10 +190,33 @@ export class CeramicDaemon {
...
+    baseRouter.use(errorHandler(this.diagnosticsLogger))
+    commitsRouter.use(errorHandler(this.diagnosticsLogger))
+    documentsRouter.use(errorHandler(this.diagnosticsLogger))
+    multiqueriesRouter.use(errorHandler(this.diagnosticsLogger))
+    nodeRouter.use(errorHandler(this.diagnosticsLogger))
+    pinsRouter.use(errorHandler(this.diagnosticsLogger))
+    recordsRouter.use(errorHandler(this.diagnosticsLogger))
+    streamsRouter.use(errorHandler(this.diagnosticsLogger))
+    contentRouter.use(errorHandler(this.diagnosticsLogger))
+
```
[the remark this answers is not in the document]
stbrody: FWIW all the tests passed. I prefer this ordering as I think it's a little clearer to the reader what's happening as we go from more general things to more specific things.  But if it breaks things then of course we should change it.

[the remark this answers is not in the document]
v-stickykeys: Good point that we should have tests here. I can work on that this sprint as I test the metrics logs again

v-stickykeys (approved):
Confirmed that this works afaict! Nitpick if you don't mind @dysbulic 
let's move `baseRouter.use(errorHandler(this.diagnosticsLogger))` under `app.use('/api/v0', baseRouter)`

oed:
Looks like this can be merged after a rebase 🎉

stbrody:
Rebased and merged!  Thank you for your patience @dysbulic!

PR #1406, opened 2021-05-24
fix(http-client): Allow just one stream of sync http calls

ukstv:
Previously, two subscriptions on a stream state resulted in two streams of HTTP update calls. Say, one HTTP call is marked `U`, and sync interval is four dashes (~seconds). With two independent `.subscribe` calls for one `TileDocument` object we would have: `UU----UU----UU----`.

This PR makes it back to `U----U----U----` regardless of how many subscriptions are active. The stream of HTTP calls starts when the first observer arrives, and stops when the last observer unsubscribes.


stbrody (approved):
Nice!

PR #1408, opened 2021-05-24
fix(ipfs-topology): Don't swarm connect to ourselves if we are in the peerlist

stbrody:


packages/ipfs-topology/src/ipfs-topology.ts
```diff
@@ -111,8 +111,16 @@ export class IpfsTopology {
     ipfs: IpfsApi,
     bootstrapList: string[]
   ): Promise<void> {
+    // Don't want to swarm connect to ourself
+    const myPeerId = (await ipfs.id()).id
+    const filteredBootstrapList = bootstrapList.filter((addr) => {
+      const arr = addr.split('/')
+      const peerId = arr[arr.length - 1]
+      return myPeerId != peerId
```
v-stickykeys, line 119: Can replace this with `endsWith` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith

v-stickykeys, line 119:
```suggestion
      return !addr.endsWith(myPeerId)
```

v-stickykeys (approved):
👍 small suggestion

ukstv (changes requested):
+1 to @v-stickykeys

packages/ipfs-topology/src/ipfs-topology.ts
stbrody: oh yeah, that's much nicer, thanks!

PR #1417, opened 2021-05-27
fix(core): only sync pinned streams the first time they are loaded

stbrody:


packages/common/src/docopts.ts
```diff
@@ -13,6 +13,12 @@ export enum SyncOptions {
      *   for the response, regardless of whether or not the stream is found in the node's cache
      */
     SYNC_ALWAYS,
+
+    /**
+     * Do not query pubsub for the new tip in any circumstance. This means that if the stream
+     * is not in cache or the pin store, then only the genesis commit for the stream will be returned
+     */
+    NEVER_SYNC,
```
stbrody, line 21: If we also did the change where the CAS causes the AnchorCommit to be published, we could get out of the habit of loading all pinned streams at startup entirely, which would let us avoid adding this new sync mode. Might be worth doing, although it's a larger change with some implications we would need to think about.

stbrody: Actually, nvm, it's not quite that straightforward, since we need to worry about what happens if by the time the stream is loaded again the AnchorCommit no longer exists on the CAS' ceramic node. I guess we'd have to make the CAS pin every stream it ever anchors, which is probably not ideal

oed (approved):
lgtm

packages/core/src/state-management/state-manager.ts
```diff
@@ -15,14 +15,22 @@ import {
...
 
 export class StateManager {
 
+  /**
+   * Keeps track of every pinned StreamID that has had its state 'synced' (i.e. a query was sent to
+   * pubsub requesting the current tip for that stream) since the start of this process. This set
+   * only grows over time, in line with how many pinned streams get queried.
+   * @private
+   */
+  private readonly syncedPinnedStreams: Set<string>
```
ukstv, line 32: Not important. Could be `private readonly syncedPinnedStreams: Set<string> = new Set()`, and no logic in constructor.

Issue #730, opened 2020-12-30
DocState should store inner CIDs for SignedCommit link records


stbrody:
The document state log contains the CID of every record applied to a document.  But in the case of SignedRecords, there is an additional CID that is important to the document log: the `link` record that contains the payload of the actual data.  This leads [Document._findIndex](https://github.com/ceramicnetwork/js-ceramic/blob/c7561344c619f72a243d1f27978393830bf49f56/packages/core/src/document.ts#L374-L377) to need to load this `link` record for every signed record in the log when looking for a CID.  Given that `Document._findIndex` is used quite often as part of handling incoming tips, this results in a very real perf hit (I imagine, I haven't actually measured this directly).  If we cached both the outer SignedRecord and the inner 'link' record CIDs in the document state, all those checks could be done purely in memory without any network i/o, and I suspect this would give us a major perf improvement.

In addition to the perf improvement, this could be used to clean up and simplify a lot of code that has to have special case handling for SignedCommits in many different places.

ukstv:
Hey, @stbrody may we close this one in favour of #1444 and #1445 ?

stbrody:
It does seems likely that some combination of #1444 and #1434 probably remove the need for this, but given as there's still some open questions around #1434 and whether or not we can safely remove the checks on link CIDs from _findIndex, I'd like to leave this open at least until we've confirmed that.  If #1434 winds up removing the need to load link CIDs from _findIndex, then I think it'd be fine to go ahead and close this one out.

stbrody:
Obsolete now due to https://github.com/ceramicnetwork/js-ceramic/issues/1434

Issue #1421, opened 2021-05-28
Optimize commit application to minimize calls to IPFS


stbrody:
To see the issue, add this test to `state-manager.test.ts`:
```
test('handleTip for commit already in log', async () => {
  const newContent = {foo: 'bar'}
  const stream1 = await TileDocument.create<any>(ceramic, INITIAL_CONTENT, null, { anchor:false });
  await stream1.update(newContent, null, { anchor: false })

  const ceramic2 = await createCeramic(ipfs);
  const retrieveCommitSpy = jest.spyOn(ceramic2.dispatcher, 'retrieveCommit');

  const stream2 = await ceramic2.loadStream<TileDocument>(stream1.id, { syncTimeoutSeconds:0 });
  const streamState2 = await ceramic2.repository.load(stream2.id, {});


  retrieveCommitSpy.mockClear()
  await (ceramic2.repository.stateManager as any)._handleTip(streamState2, stream1.state.log[1].cid);

  expect(streamState2.state).toEqual(stream1.state);
  expect(retrieveCommitSpy).toBeCalledTimes(2) // actually is 7

  // Now re-apply the same commit and don't expect any additional calls to IPFS
  await (ceramic2.repository.stateManager as any)._handleTip(streamState2, stream1.state.log[1].cid);
  expect(retrieveCommitSpy).toBeCalledTimes(2) // actually is 8

  await ceramic2.close();
});
```

Notice how applying a single SignedCommit results in 7 calls to `ipfs.dag.get`.  I believe this should really only be 2 - one for the base signed commit and one for the dag-jose payload from the `link` commit.  But there are many inefficiencies in the codepaths for applying commits that results in the same CID being loaded multiple times.

stbrody:
FYI @ukstv since this seems relevant to some of the perf issues you've been investigating

ukstv:
Closing in favour of actionable #1444

stbrody:
I'd like to leave this open as I think there are alternate approaches to fixing this problem besides #1444 that are worth considering.  Namely the approach that doesn't involve caching any more information in memory, but just reworking the codepaths and control flow so that we don't re-do work during the process of a single stream load

PR #1422, opened 2021-05-28
fix(core): Don't refetch CID from IPFS when re-applying commits already in the log

stbrody:
https://github.com/ceramicnetwork/js-ceramic/issues/730 is the better/more complete fix, but this simpler code change optimizes for the most common cases of trying to re-apply tips already in the log

### DIFF
--- a/packages/core/src/__tests__/state-manager.test.ts
+++ b/packages/core/src/__tests__/state-manager.test.ts
@@ -145,6 +145,31 @@ test('handleTip', async () => {
   await ceramic2.close();
 });
 
+test('handleTip for commit already in log', async () => {
+  const newContent = {foo: 'bar'}
+  const stream1 = await TileDocument.create<any>(ceramic, INITIAL_CONTENT, null, { anchor:false });
+  await stream1.update(newContent, null, { anchor: false })
+
+  const ceramic2 = await createCeramic(ipfs);
+  const retrieveCommitSpy = jest.spyOn(ceramic2.dispatcher, 'retrieveCommit');
+
+  const stream2 = await ceramic2.loadStream<TileDocument>(stream1.id, { syncTimeoutSeconds:0 });
+  const streamState2 = await ceramic2.repository.load(stream2.id, {});
+
+
+  retrieveCommitSpy.mockClear()
+  await (ceramic2.repository.stateManager as any)._handleTip(streamState2, stream1.state.log[1].cid);
+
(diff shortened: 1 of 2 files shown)


packages/core/src/__tests__/state-manager.test.ts
```diff
@@ -145,6 +145,32 @@ test('handleTip', async () => {
...
+
+  const stream2 = await ceramic2.loadStream<TileDocument>(stream1.id, { syncTimeoutSeconds:0 });
+  const streamState2 = await ceramic2.repository.load(stream2.id, {});
+
+
+  retrieveCommitSpy.mockClear()
+  await (ceramic2.repository.stateManager as any)._handleTip(streamState2, stream1.state.log[1].cid);
+
+  expect(streamState2.state).toEqual(stream1.state);
+  expect(retrieveCommitSpy).toBeCalledTimes(7) // TODO(1421): This should be 2!
```
stbrody, line 164: this is pretty scary. Filed https://github.com/ceramicnetwork/js-ceramic/issues/1421 to clean this up

oed (commented):
I don't understand why we have a need to retrieve the CID of the content. When do we need to find the index of the signed content CID?

Seems to me like we could just remove the code that makes a call to ipfs here.

oed:
Places this method is called:

* [`loadAtCommit`](https://github.com/ceramicnetwork/js-ceramic/blob/c7561344c619f72a243d1f27978393830bf49f56/packages/core/src/document.ts#L170) - doesn't need content CID check
* [`_applyLog`](https://github.com/ceramicnetwork/js-ceramic/blob/c7561344c619f72a243d1f27978393830bf49f56/packages/core/src/document.ts#L471) - doesn't need content CID check
* [`_isCidIncluded`](https://github.com/ceramicnetwork/js-ceramic/blob/c7561344c619f72a243d1f27978393830bf49f56/packages/core/src/document.ts#L389) - now this one I'm not certain about if it needs the additional content CID check


Seems like we should at least have a flag to skip content CID check for the first two methods.

stbrody:
@oed I filed https://github.com/ceramicnetwork/js-ceramic/issues/1434 to revisit this.  In the meantime I'd like to get this PR merged as it makes the worst case no worse than it is today, but improves the common case a lot

oed (approved):

ukstv (approved):

PR #1433, opened 2021-05-31
feat: Restore streams one by one

ukstv:


packages/core/src/ceramic.ts
```diff
@@ -598,11 +598,13 @@ class Ceramic implements CeramicApi {
    */
   restoreStreams() {
     this.repository.listPinned().then(async list => {
+      let item = list.pop()
       let n = 0
-      await Promise.all(list.map(async streamId => {
-        await this._loadStream(StreamID.fromString(streamId), { sync: SyncOptions.NEVER_SYNC })
+      while (item) {
+        await this._loadStream(StreamID.fromString(item), { sync: SyncOptions.NEVER_SYNC })
         n++;
```
oed, line 605: Why do we have `n` here?

ukstv: To present it below: `this._logger.verbose(`Successfully restored ${n} pinned streams`)`
Otherwise no reason.

packages/core/src/ceramic.ts
```diff
@@ -598,11 +598,13 @@ class Ceramic implements CeramicApi {
    */
   restoreStreams() {
     this.repository.listPinned().then(async list => {
+      let item = list.pop()
       let n = 0
-      await Promise.all(list.map(async streamId => {
-        await this._loadStream(StreamID.fromString(streamId), { sync: SyncOptions.NEVER_SYNC })
+      while (item) {
```
stbrody, line 603: I'm worried this could take a really long time to finish, especially if some of the streams time out loading some CID during the load process.

The simplest middle ground would be to pick a batch size, let's say 100, and load 100 at a time, waiting for all 100 to finish before starting the next batch.  The slightly more complicated but better fix would be to pick some number of streams we are comfortable to have loading in parallel, and then whenever one stream finishes loading have another stream start loading to keep the load buffer full.

stbrody: Actually, now that I think about it, we kind of already have the second in place due to the loading queue.  Even if we kick off 10000+ load streams at the same time, we're only going to wind up with a smaller fixed number running at once, based on the size of the loading queue.

@ukstv, what did we wind up picking as the concurrency limit for the loading queue?  Is it the same as the execution queue?

stbrody: Oh wait, I'm misremembering how the loading queue works.  I guess the loading queue is per-streamId with a concurrency limit of 1, just to limit simultaneous loading of the *same* stream.

I wonder if this is still the right idea for a fix though. Rather than doing something special at startup, what if we set a process-wide limit on how many streams can be loaded simultaneously, just like we have for the execution queue?

ukstv: True, LoadingQ just serializes loading of the _same_ stream, but do not limit how many streams one could attempt to load at once. It could be changed so that at most N streams are loaded at once indeed. And, TBH, it seems like an idea worth exploring. I mean, I see no direct harm and we discussed this as a probable evolution.

ukstv: Linking #1443

PR #1453, opened 2021-06-02
feat(core): Limit the number of concurrently loading streams

stbrody:
Achieved by removing the loadingQ and using the executionQ for both loads and updates.


ukstv:
This changes semantics of operations. As loadingQ is removed, instead of having independently serialised loading and execution operations, now we have a single queue. It has following implications.

**1. Loading is blocked by execution.** With the single queue, `applyCommit`, `handleTip` and the likes (together named `exec` below) block the load (`load` below).
Say, for operations order like [`exec`, `exec`, `load`], the latter gets executed only after two `exec`s run. Effectively, this makes `load` operation to appear slow: "We know we have some things to do, so please wait a bit, so that we can send you the most recent version available". There are no guarantees on how long single `exec` could take, as it does IPFS calls, so in the worst case `load` has to wait for `N*IPFS_GET_TIMEOUT`, where `N` is number of `exec` operations initiated before `load` call appears.

**2. Probability of bugs increases.** With processing queues it takes some care to not introduce a deadlock situation: one operation wrapped in `executionQ.run` waits for another operation that is also wrapped in `executionQ.run`. This will stale indefinitely. In the upstream, executionQ-wrapped operations are concentrated in StateManager, while loadingQ operations are concentrated in Repository, no intersections, so it is somewhat easy to spot potential deadlocks. With this change, Repository also has executionQ-wrapped operations. One might easily miss a  deadlock here.

I can not approve this PR.

stbrody:
Okay @ukstv, new approach up that uses a separate, independent ExecutionQueue for loading

ukstv (approved):
🏇

PR #1474, opened 2021-06-08
fix(core): Don't submit an anchor request for an AnchorCommit

stbrody:


packages/core/src/state-management/state-manager.ts
```diff
@@ -228,6 +228,9 @@ export class StateManager {
    * Request anchor for the latest stream state
    */
   anchor(state$: RunningState): Subscription {
+    if (state$.value.anchorStatus == AnchorStatus.ANCHORED) {
+      return empty().subscribe()
```
ukstv, line 232:
```suggestion
      return Subscription.EMPTY;
```

ukstv (approved):
🏇

PR #1475, opened 2021-06-09
fix: Flaky doc test

ukstv:


oed (approved):

packages/http-client/src/__tests__/document.test.ts
```diff
@@ -106,10 +106,10 @@ describe('periodic subscription', () => {
       .toPromise();
     subscription.unsubscribe();
 
-    // Invocations should happen every SYNC_INTERVAL (with some error bounded to 10% of SYNC_INTERVAL)
+    // Invocations should happen every SYNC_INTERVAL (with some error bounded to 20% of SYNC_INTERVAL)
     const deltas = pairs(invocations).map((pair) => pair[1] - pair[0]);
     deltas.forEach((delta) => {
-      expect(Math.abs(delta - SYNC_INTERVAL)).toBeLessThan(SYNC_INTERVAL * 0.1);
+      expect(Math.abs(delta - SYNC_INTERVAL)).toBeLessThan(SYNC_INTERVAL * 0.8);
```
stbrody, line 112: Shouldn't this be `0.2`?  With the SYNC_INTERVAL at 200ms and a 20% error window, I would expect the test should be testing that poll times should be between 160ms and 240ms apart.  As written I believe it's testing that the polls are between 40ms and 360ms apart?

ukstv: Changed the number in https://github.com/ceramicnetwork/js-ceramic/pull/1480

It really has to be that much.

PR #1477, opened 2021-06-10
chore(core): Add tests of interplay between stream cache and the various SyncOptions

stbrody:
To make sure that the PREFER_CACHE sync behavior won't prefer an old version in the cache over getting the current version from pubsub


smrz2001 (approved):
I understand caching a _little_ better now 😁 

:shipit:

PR #1484, opened 2021-06-15
feat(core): Loading a CommitID makes the node handle the commit CID as a potential new tip

stbrody:
This basically makes loading a stream with a CommitID function as a backdoor way to expose "handleTip" to the public API.

### DIFF
--- a/packages/core/src/state-management/state-manager.ts
+++ b/packages/core/src/state-management/state-manager.ts
@@ -12,6 +12,7 @@ import {
   UnreachableCaseError,
   RunningStateLike,
   DiagnosticsLogger,
+  StreamUtils,
 } from '@ceramicnetwork/common';
 import { RunningState } from './running-state';
 import CID from 'cids';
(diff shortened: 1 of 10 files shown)


stbrody:
FYI @smrz2001

oed (approved):
makes sense to me 👍

packages/core/src/state-management/state-manager.ts
```diff
@@ -89,14 +90,24 @@ export class StateManager {
...
+   * for this stream, emit the new state.
    *
-   * @param state$ - Stream state to rewind.
+   * @param state$ - Currently known state of the stream.
    * @param commitId - Requested commit.
    */
-  async rewind(state$: RunningStateLike, commitId: CommitID): Promise<SnapshotState> {
-    const snapshot = await this.conflictResolution.rewind(state$.value, commitId);
-    return new SnapshotState(snapshot);
+  async atCommit(state$: RunningStateLike, commitId: CommitID): Promise<SnapshotState> {
```
stbrody, line 99: this is the core of the change

ukstv (approved):
🏇

PR #1487, opened 2021-06-16
feat(core): Allow lazy initializtaion of anchor service

stbrody:
This PR allows Ceramic to start up with the CAS unavailable. If any operations are attempted which try to perform an anchor, the anchor will fail and a warning will be logged to the logs.  The operation that attempted the anchor, however, will **not** get an exception thrown.  That means that if you are using the CLI to create or update streams, you will not see any error or any indication that the anchor failed unless you read the logs or check the state of the stream and notice that the AnchorStatus is FAILED. This is consistent with the way we handle all anchor failures currently, although the behavior may be surprising to some users that not just when the anchor fails on the CAS, but even when we fail to even *submit the request* to the CAS, they still see no error in their app.

I started work on a patch to change things so that if we fail while initially requesting the anchor we throw an exception in the operation that requested it, while leaving anchor failures later on in the process as happening in the background and just setting the anchorStatus to FAILED.  The problem with that change is that now if you try to create a stream and requesting the anchor fails, you get an exception thrown, which is good for informing you that something went wrong, but means you don't get back the StreamID. This is a problem since the stream actually was created, but now you have no handle to it or any way to get a handle to it.

So I guess I'll just leave things as-is on this front for now, but it's not very satisfying. It feels like developers will want fast feedback if their anchor requests are failing. The fact that you can configure a ceramic node with no available CAS, then create and update streams against it from the CLI without seeing any indication that anything is wrong is fairly worrisome.

### DIFF
--- a/packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
+++ b/packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
@@ -117,7 +117,7 @@ export default class EthereumAnchorService implements AnchorService {
    * @param streamId - Stream ID
    * @param tip - Tip CID of the stream
    */
-  requestAnchor(streamId: StreamID, tip: CID): Observable<AnchorServiceResponse> {
+  async requestAnchor(streamId: StreamID, tip: CID): Promise<Observable<AnchorServiceResponse>> {
     const cidStreamPair: CidAndStream = { cid: tip, streamId };
     return concat(
       this._announcePending(cidStreamPair),
--- a/packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
+++ b/packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
@@ -86,23 +86,27 @@ export default class EthereumAnchorService implements AnchorService {
   }
 
   async init(): Promise<void> {
-    // Get the chainIds supported by our anchor service
-    const response = await fetchJson(this.chainIdApiEndpoint)
-    if (response.supportedChains.length > 1) {
+    try {
+      // Get the chainIds supported by our anchor service
+      const response = await fetchJson(this.chainIdApiEndpoint)
+      if (response.supportedChains.length > 1) {
         throw new Error("Anchor service returned multiple supported chains, which isn't supported by js-ceramic yet")
-    }
-    const chainId = response.supportedChains[0]
+      }
+      const chainId = response.supportedChains[0]
 
-    // Confirm that we have an eth provider that works for the same chain that the anchor service supports
-    const provider = this._getEthProvider(chainId)
-    const provider_chain_idnum = (await provider.getNetwork()).chainId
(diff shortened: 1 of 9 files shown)


ukstv (changes requested):
I do not get it. I understand the intent, but how it works I do not understand. I tried to run the scenario of unavailable anchor with HTTP client. HTTP Client is the most used way to communicate with a node, and it does not throw an error. Maybe the error is thrown by CLI. In my case, no error was thrown. Even if it does throw an error (maybe I did something wrong), it introduces certain incompatibility between CLI and client behaviour, which is not good.

The solution conflates an error being thrown in a control flow with an error that happens in (possibly) out-of-direct-control-flow processing. One can not rely on an error like this to happen _inside_ a control flow. Considering possible field of optimisation, it makes sense to think about stream processing in CQRS paradigm, and the change here ties the hands with regards to possible optimisations.

Another thing, the proposed solution conflates node misconfiguration and client behaviour. It is a situation of a node being misconfigured, not a client doing something wrong. IMO, it makes sense to straight away halt a node process, if during the runtime it can not connect to a CAS instance. Right now the node just hangs until a human kills it.

At the same time, I feel like the intent is valuable, so maybe as another way of thinking about how to deliver failure reason to a developer, one could extend single "FAILED" status with error details. In this case, anchorStatus is not just a string. It is a full-fledged object that contains whatever details needed based on the actual status.

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -86,28 +86,44 @@ export default class EthereumAnchorService implements AnchorService {
...
+      this._logger.warn(`Error while initializing connection to anchor service '${this.url}'. Will try again on demand.`)
     }
   }
 
+  private async _initIfNeeded(): Promise<void> {
+    if (this._chainId) {
+      return // already initialized
+    }
+    await this.init()
+  }
```
oed, line 117: 🙏

oed (approved):
lgtm. This setup makes sense to me. Warnings in the console should be enough.

v-stickykeys:
@stbrody What about a flag that explicitly says `--disable-anchors`?

stbrody:
> @stbrody What about a flag that explicitly says `--disable-anchors`? (Can change wording)
> 
> So you can still run Ceramic without CAS but if you don't include that flag you will get errors for an unreachable CAS. Seems much more clear to me and achieves what we want still.
> 
> The areas that feel suboptimal to me are: 1-users not knowing that they need CAS; 2-making requests to CAS when we know it is unreachable on start up

Just spoke with @ukstv and we reached a similar conclusion. The fact that you can be running in a misconfigured state without realizing it for an extended period of time is worrisome. I filed https://github.com/ceramicnetwork/js-ceramic/issues/1489 and will close this PR in favor of a new one to come that tries that approach instead

PR #1491, opened 2021-06-17
fix(core): Remove loading inner CID from SignedCommits in ConflictResolution.findIndex

smrz2001:


stbrody (approved):
LGTM! :tada: 

I also did a quick scan on the uses of findIndex and nothing looked like it depends on being able to find inner link CIDs that I could see. It seems to be exclusively used for Ceramic commits from the log, which will always be the outer dag-jose envelope CID

PR #1498, opened 2021-06-25
feat(cli): add global sync option and disable restore stream to options

zachferland:
Adds to cli following options: 
--no-restore-streams
--sync    "prefer-cache", "sync-always", or "never-sync"

Adds syncOptions to core, which will force sync option on all requests 

Both these options are required for gateway, but also could have been passed with env vars.

TODO:

Probably make sense to split sync options into something like Default Sync Options and Global Sync Options (or other names). Where default is fallback, given individual call does not include option, while global is a force/override. This would also allow this option to be added to http client with consistency (not added now). Where default is more useful for http client (likely no use case for global), and core/daemon will often use either one. 

Another note, all these boolean options dont behave as expected, or as we use them in terraform. Ie --log-to-files --debug etc. Any of these options when given any value return true. `ceramic daemon --log-to-files false` will set log to files true. 

Reason why --no-restore-streams (vs restore-streams) is used here for consistency, but can also change all these booleans to strings, to accept true/false values, rather than just flag. Might be breaking though if any one just using flag without value, but didnt check behavior, should be able to implement anyways.

### DIFF
--- a/packages/cli/src/bin/ceramic.ts
+++ b/packages/cli/src/bin/ceramic.ts
@@ -24,6 +24,8 @@ program
     .option('--max-healthy-memory <decimal>', 'Fraction of total memory usage considered healthy. Defaults to 0.7')
     .option('--cors-allowed-origins <list>', 'Space-separated list of strings and/or regex expressions to set for Access-Control-Allow-Origin . Defaults to all: "*"')
     .option('--disable-anchors', 'Allows Ceramic Daemon to start up without a configured anchor service. Any anchor requests made will fail.')
+    .option('--no-restore-streams', 'Disable loading pinned stream state into memory at startup')
+    .option('--sync <string>', 'Global modes for syncing streams. One of: "prefer-cache", "sync-always", or "never-sync". Defaults to "prefer-cache"')
     .description('Start the daemon')
     .action(async ({
         ipfsApi,
@@ -44,6 +46,8 @@ program
         pubsubTopic,
         corsAllowedOrigins,
         disableAnchors,
+        restoreStreams,
+        sync
     }) => {
         if (stateStoreDirectory && stateStoreS3Bucket) {
           throw new Error("Cannot specify both --state-store-directory and --state-store-s3-bucket. Only one state store - either on local storage or on S3 - can be used at a time")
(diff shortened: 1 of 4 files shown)


oed:
😬 We really need that config file!

packages/cli/src/bin/ceramic.ts
```diff
@@ -24,6 +24,8 @@ program
     .option('--max-healthy-memory <decimal>', 'Fraction of total memory usage considered healthy. Defaults to 0.7')
     .option('--cors-allowed-origins <list>', 'Space-separated list of strings and/or regex expressions to set for Access-Control-Allow-Origin . Defaults to all: "*"')
     .option('--disable-anchors', 'Allows Ceramic Daemon to start up without a configured anchor service. Any anchor requests made will fail.')
+    .option('--no-restore-streams', 'Disable loading pinned stream state into memory at startup')
```
stbrody, line 27: I wonder if we need this option, or if we should just disable the stream restoration behavior whenever we're running in gateway mode?  Now that we don't sync the streams at startup as part of this, the "restore" functionality is just to poll the CAS for pending anchors that were created by this node, but if the node is a gateway it shouldn't have created any anchors. If you had a node that was writeable and then got restarted as a gateway you could miss some anchor commits, but that seems like a pretty niche case that maybe it's fine if we don't support anyway.

In retrospect, I wonder if I should get rid of the "--disble-anchors" flag also and just make that the behavior of the gateway mode.  The gateway should never have need of a CAS.

packages/cli/src/bin/ceramic.ts
```diff
@@ -24,6 +24,8 @@ program
...
+    .option('--sync <string>', 'Global modes for syncing streams. One of: "prefer-cache", "sync-always", or "never-sync". Defaults to "prefer-cache"')
```
stbrody, line 28: Can we call this `--sync-override`?  Or something else to clarify that it overrides the behavior of whatever the client requests?  Would also be good to clarify that in the flag description.

stbrody (changes requested):

packages/cli/src/bin/ceramic.ts
zachferland: oh ok, yeah i thought i remembered the sync streams at start changed, agree that this makes senes to fold into gateway mode, dont have any concerns with that. Looks like this covers all that already, right? - https://github.com/ceramicnetwork/js-ceramic/pull/1513/files

zachferland:
Closing in favor of https://github.com/ceramicnetwork/js-ceramic/pull/1516

PR #1502, opened 2021-06-28
feat: Pass time-information when checking a signature

ukstv:


packages/core/src/conflict-resolution.ts
```diff
@@ -174,31 +167,45 @@ export async function fetchLog(
...
-  log.push(cid); // Should be unshift [O(N)], but push [O(1)] + reverse [O(N)] seem better
-  if (await stateLog.includes(prevCid)) {
+  let nextEntry: LogEntry;
+  if (isSignedCommit) {
+    nextEntry = {
+      type: CommitType.SIGNED,
+      cid: cid,
+    };
+  } else {
+    // If not signed, then anchor commit
```
oed, line 197: Ehm, it could also be a genesis commit? Or is that handled by line 172?

packages/core/src/conflict-resolution.ts
```diff
@@ -253,10 +260,11 @@ export class ConflictResolution {
         // it's an anchor commit
         // TODO: Anchor validation should be done by the StreamHandler as part of applying the anchor commit
         await verifyAnchorCommit(this.dispatcher, this.anchorService, commit);
-        state = await handler.applyCommit(commit, cid, this.context, state);
+        state = await handler.applyCommit(commit, { cid: cid }, this.context, state);
       } else {
         // it's a signed commit
-        const tmpState = await handler.applyCommit(commit, cid, this.context, state);
+        const timestamp = StreamUtils.commitTimestamp(log, commit)
```
oed, line 266: Seems inefficient to add an `O(n)` operation here. We should use the recursive nature of `fetchLog` instead!

packages/core/src/conflict-resolution.ts
```diff
@@ -174,31 +167,45 @@ export async function fetchLog(
   dispatcher: Dispatcher,
   cid: CID,
   stateLog: HistoryLog,
-  log: CID[] = [],
-): Promise<CID[]> {
-  if (await stateLog.includes(cid)) {
+  log: LogEntry[] = [],
+): Promise<LogEntry[]> {
```
oed, line 171: Add optional parameter `timestamp`.

packages/core/src/conflict-resolution.ts
```diff
@@ -174,31 +167,45 @@ export async function fetchLog(
...
     return [];
   }
-  log.push(cid); // Should be unshift [O(N)], but push [O(1)] + reverse [O(N)] seem better
-  if (await stateLog.includes(prevCid)) {
+  let nextEntry: LogEntry;
+  if (isSignedCommit) {
+    nextEntry = {
+      type: CommitType.SIGNED,
+      cid: cid,
+    };
```
oed, line 195:
```suggestion
    nextEntry = {
      type: CommitType.SIGNED,
      cid: cid,
      timestamp,
    };
```

packages/core/src/conflict-resolution.ts
```diff
@@ -174,31 +167,45 @@ export async function fetchLog(
...
+  } else {
+    // If not signed, then anchor commit
+    const proof = await dispatcher.retrieveFromIPFS(cid);
+    const timestamp = proof.blockTimestamp;
```
oed, line 199: Just assigned to function wide `timestamp` variable.

packages/core/src/conflict-resolution.ts
```diff
@@ -174,31 +167,45 @@ export async function fetchLog(
...
+      cid: cid,
+      timestamp: timestamp,
+    };
+  }
+  // Should be unshift [O(N)], but push [O(1)] + reverse [O(N)] seem better
+  log.push(nextEntry);
+  if (stateLog.includes(prevCid)) {
     // we found the connection to the canonical log
     return log.reverse();
   }
```
oed, line 211: Can't comment on the next line but include `timestamp` variable to the `fetchLog` function call.

oed (changes requested):
So this doesn't pass the timestamp to `verifyJWS` yet. I think this can be simplified to remove the `commitTimestamp` function!

packages/core/src/conflict-resolution.ts
ukstv: It is combination of 172 plus a caller side. State that is passed there is already constructed from the genesis.

packages/stream-tile-handler/src/tile-document-handler.ts
```diff
@@ -70,7 +70,7 @@ export class TileDocumentHandler implements StreamHandler<TileDocument> {
         const isSigned = StreamUtils.isSignedCommit(commit)
         if (isSigned) {
             payload = (await context.ipfs.dag.get(commit.link, { timeout: IPFS_GET_TIMEOUT })).value
-            await this._verifySignature(commit, context, payload.header.controllers[0])
+            await this._verifySignature(commit, {cid: cid}, context, payload.header.controllers[0])
```
oed, line 73: Wouldn't this mean that a signed genesis will fail if you rotate keys? i.e. we need timestamp for genesis commit as well.

ukstv: Right. Just passed time information there as well.

There is a one more issue with that though. For a genesis with no anchor, there is no way to deduce its timestamp. It could be created any time `(-∞; now]`. So, if we always lean towards `now`, there is a possibility of accidental signature invalidation for non-anchored genesis. Probably, one could enforce creating an anchor for a signed genesis.

oed: Seems like this is exactly the same as non-anchored signed commits in general.

ukstv: I was going to disagree, telling that the interval now is `(prevAnchorTimestamp; now]`, but it does not make any difference. The same, agree.

packages/core/src/conflict-resolution.ts
ukstv: Working note: This is true only for Tiles. Caip10 has a completely different notion of a signed commit.

packages/core/src/conflict-resolution.ts
ukstv: Yup. Changed. Now fetchLog provides timestamps.

oed (commented):
This looks good now, but I think we are missing a case here because when we load a stream we always apply the genesis commit before fetching the log which will cause issues if the `kid` used to sign the genesis is no longer valid. We need to find a way around that!

packages/stream-tile-handler/src/tile-document-handler.ts
```diff
@@ -169,14 +170,15 @@ export class TileDocumentHandler implements StreamHandler<TileDocument> {
...
      * @param context - Ceramic context
      * @param did - DID value
      * @private
      */
-    async _verifySignature(commit: any, context: Context, did: string): Promise<DIDResolutionResult> {
+    async _verifySignature(commit: any, meta: CommitMeta, context: Context, did: string): Promise<DIDResolutionResult> {
         let result
         try {
-            result = await context.did.verifyJWS(commit)
+            result = await context.did.verifyJWS(commit, {atTime: meta.timestamp})
```
stbrody, line 181: wow, all that for this one magical line.  Amazing

stbrody (approved):
LGTM if Joel's happy

CC @smrz2001 as this touches a lot of the code he is working on

packages/stream-tile-handler/src/tile-document-handler.ts
oed: 😆

oed: There was a PR in JWS to support this fwiw.

oed (changes requested):
Just to be clear, since Spencer approved :P

stbrody (changes requested):
I think it'd be great to have a version of the key-revocation test you added to the integration tests repo that lives here in js-ceramic and uses the InMemoryAnchorService (in a style akin to the tests from `ceramic.test.ts`).  Probably gives us almost as good coverage but will run much faster and run before any commit to the repo

ukstv:
> I think it'd be great to have a version of the key-revocation test you added to the integration tests repo that lives here in js-ceramic and uses the InMemoryAnchorService (in a style akin to the tests from `ceramic.test.ts`). Probably gives us almost as good coverage but will run much faster and run before any commit to the repo

I would love to, but not this time.
1. **Minor trouble: circular dependency.** To test this properly, and if test is to be put in `core` package, we should instantiate ceramic-core, and we do have to use 3id-did-provider. And, 3id-did-provider dev-depends on core. There is currently no problem with it, but it might introduce a real dependency-related problem down the line. If put into tile, or tile-handler package, this one would have to depend on core, which is a circular dependency too.
2. **Major trouble: IDX definitions**. Testing uses in-memory network, and in-memory network has no knowledge of IDX definitions/schemas. This makes the test stuck.

oed:
@ukstv good points, have some thoughts:

1. Maybe we should create a package in this repo that only does testing i.e. `packages/integration-tests`?

2. I don't think we actually need 3id-did-provider to test this. We could simply mock a DID resolver and create static test vectors!

stbrody:
> 1. Maybe we should create a package in this repo that only does testing i.e. `packages/integration-tests`?

I've wanted a repo like this to put tests that depend on multiple packages within js-ceramic for a while.  If we did that it could probably even use the real 3id-did-provider

oed:
> I've wanted a repo like this to put tests that depend on multiple packages within js-ceramic for a while. If we did that it could probably even use the real 3id-did-provider

I meant more inside of the monorepo, which would mean no external deps.

stbrody:
> > I've wanted a repo like this to put tests that depend on multiple packages within js-ceramic for a while. If we did that it could probably even use the real 3id-did-provider
> 
> I meant more inside of the monorepo, which would mean no external deps.

Yep, I also meant inside the monorepo, sorry if that wasn't clear

packages/core/src/conflict-resolution.ts
```diff
@@ -233,14 +243,24 @@ export class ConflictResolution {
...
     handler: StreamHandler<T>,
-    log: Array<CID>,
+    log: LogEntry[],
     state?: StreamState,
     breakOnAnchor?: boolean,
   ): Promise<StreamState> {
+    // When we have genesis state only, and add some commits on top, we should check a signature at particular timestamp.
+    // Most probably there is a timestamp information there. If no timestamp found, we consider it to be _now_.
+    // `fetchLog` provides the timestamps.
+    if (state && state.log.length === 1) {
```
ukstv, line 253: @oed Please, pay attention to this. If a stream has just a genesis commit, we try to find time info in the log, and only here check signature wrt its timestamp.

ukstv: If you find the overall approach ok, I will go forward with documentation and tests.

packages/core/src/conflict-resolution.ts
```diff
@@ -253,10 +273,10 @@ export class ConflictResolution {
         // it's an anchor commit
         // TODO: Anchor validation should be done by the StreamHandler as part of applying the anchor commit
         await verifyAnchorCommit(this.dispatcher, this.anchorService, commit);
-        state = await handler.applyCommit(commit, cid, this.context, state);
+        state = await handler.applyCommit(commit, { cid: cid }, this.context, state);
```
oed, line 276: Can't we just pass `entry.value[1]` here?
```suggestion
        state = await handler.applyCommit(commit, entry.value[1], this.context, state);
```

packages/core/src/conflict-resolution.ts
```diff
@@ -253,10 +273,10 @@ export class ConflictResolution {
...
       } else {
         // it's a signed commit
-        const tmpState = await handler.applyCommit(commit, cid, this.context, state);
+        const tmpState = await handler.applyCommit(commit, { cid: cid, timestamp: timestamp }, this.context, state);
```
oed, line 279: Same thing? 
```suggestion
        const tmpState = await handler.applyCommit(commit, entry.value[1], this.context, state);
```

packages/core/src/conflict-resolution.ts
oed: Maybe we just create a variable `const commitMeta = entry.value[1]`?

packages/core/src/conflict-resolution.ts
oed: I think this makes sense, but seems like one case is missing. What if someone steals an old key, and uses it to create new genesis commits, but with no updates? Seems like we would incorrectly consider the stream valid if so?

ukstv: It would be considered valid indeed. Currently, the code retrieves a genesis commit and creates a Stream state of it. Then, it queries the network, and applies a tip. Whole conflict resolution code is built on assumption, that we can create a state from genesis. 

To handle the case you mentioned, we have to re-do the way a node creates a running state from network, and applies commits.

Having said that, I would like to argue, that the case of a stolen key is indistinguishable from a case of genesis being signed with a non-stolen key, but never updated. Do you feel like it is a good idea to suddenly invalidate the latter stream?

oed: > Having said that, I would like to argue, that the case of a stolen key is indistinguishable from a case of genesis being signed with a non-stolen key, but never updated. Do you feel like it is a good idea to suddenly invalidate the latter stream?

I think genesis records that have not been anchored at some point (update or not) should not be considered valid if the key has been revoked. Otherwise we might as well not sign the genesis at all (ok, there is still added security from it, but this limits it).

Would rather us coming up with a way to fix it. Maybe introducing a new [`SignatureStatus`](https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/common/src/stream.ts#L14-L16) that signals that the genesis is not validated might help reducing the amount of refactoring needed here?

oed: Any thoughts on this @stbrody ?

stbrody: I agree that if the key used to sign the commits has been revoked, and there's no anchors in the stream to provide time information, then the stream should be considered invalid.  I also think in that world there's no need to compare any notion of "now" to any block timestamp from an anchor (which I would have concerns about).


As to how to actually implement this, I would have to really sit down and think about it.  If you have an idea in mind @ukstv I'd love to see a proposal to think about. Or feel free to code up a POC if that's easier to demonstrate the idea than writing it down.

oed: @ukstv I would actually be fine with merging this PR as is if we follow it up with a separate one to address this last issue. Unless most of the code here would change.
This PR unlocks work for @smrz2001 since his changes hash similar semantics on the same code paths.

ukstv: Agree on unblocking @smrz2001

packages/core/src/state-management/repository.ts
```diff
@@ -152,7 +153,7 @@ export class Repository {
       }
 
       if (!stream) {
-        stream = await this.fromNetwork(streamId, opts);
+        stream = await this.fromNetwork(streamId);
```
stbrody, line 156: I think we need to make sure in the `sync` call below that if the sync never learns of any newer tip than the genesis commit that we still go and do the signature verification that we opted out of in `fromNetwork`

stbrody (changes requested):
I guess in an ideal world we would wait to load the genesis commit until after we've already synced the log so we have the timestamp information available, but that would be a much larger refactoring. In the meantime it seems like it can be done just by making sure that we go back and do signature verification for the genesis commit if the `sync` call never gets a newer tip.

packages/core/src/state-management/repository.ts
```diff
@@ -107,14 +107,15 @@ export class Repository {
...
+  private async fromNetwork(streamId: StreamID): Promise<RunningState> {
     const handler = this.#deps.handlers.get(streamId.typeName);
     const genesisCid = streamId.cid;
     const commit = await this.#deps.dispatcher.retrieveCommit(genesisCid, streamId);
     if (commit == null) {
       throw new Error(`No genesis commit found with CID ${genesisCid.toString()}`);
     }
-    const state = await handler.applyCommit(commit, streamId.cid, this.#deps.context);
+    // Do not check timestamp here
+    const state = await handler.applyCommit(commit, {cid: streamId.cid, timestamp: null}, this.#deps.context);
```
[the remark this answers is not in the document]
oed: Note that we don't really want to disable signature verification here. Just disable verifying whether or not the key used to sign has been revoked.

packages/core/src/conflict-resolution.ts
ukstv: done.

packages/core/src/conflict-resolution.ts
```diff
@@ -233,14 +243,24 @@ export class ConflictResolution {
...
     breakOnAnchor?: boolean,
   ): Promise<StreamState> {
+    // When we have genesis state only, and add some commits on top, we should check a signature at particular timestamp.
+    // Most probably there is a timestamp information there. If no timestamp found, we consider it to be _now_.
+    // `fetchLog` provides the timestamps.
+    if (state && state.log.length === 1) {
+      const timestamp = log[0].timestamp;
+      const genesisCid = state.log[0].cid;
+      const genesis = await this.dispatcher.retrieveCommit(genesisCid);
+      await handler.applyCommit(genesis, { cid: genesis, timestamp: timestamp }, this.context);
```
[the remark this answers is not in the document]
ukstv: good catch!!

ukstv:
This PR is done. I am doing yet another to include @stbrody suggestion.

oed (approved):
lgtm

packages/core/src/state-management/repository.ts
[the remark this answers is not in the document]
stbrody: Well, it would be safe to completely skip signature verification here so long as we definitely do it later. And I'm not sure it will be straightforward to disable the key revocation check without skipping all of signature verification.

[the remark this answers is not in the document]
ukstv: Actually, it is simpler then it sounds. `timestamp = 1` makes the trick. See https://github.com/ceramicnetwork/js-ceramic/pull/1518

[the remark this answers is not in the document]
stbrody: Just made a comment there - I'm not sure that actually works

ukstv:
Again, some unrelated test failures :(

Going to merge this to unblock @smrz2001. The case of lone genesis, is tracked in https://github.com/ceramicnetwork/js-ceramic/pull/1518

ukstv:
Ping @smrz2001

PR #1503, opened 2021-06-28
feat: Use ~/.jsipfs as default ipfs repo path

ukstv:


stbrody (approved):
LGTM, though this is a breaking change for users running their own nodes so we should make sure to message it clearly

v-stickykeys (approved):
+1 Spencer's comment

PR #1505, opened 2021-06-28
feat(core): Split AnchorService from AnchorValidator

stbrody:
This is important so that Ceramic nodes running with --disableAnchors can still validate AnchorCommits

### DIFF
--- a/packages/stream-tile-handler/src/tile-document-handler.ts
+++ b/packages/stream-tile-handler/src/tile-document-handler.ts
@@ -137,13 +137,6 @@ export class TileDocumentHandler implements StreamHandler<TileDocument> {
         // TODO: Assert that the 'prev' of the commit being applied is the end of the log in 'state'
         const proof = (await context.ipfs.dag.get(commit.proof, { timeout: IPFS_GET_TIMEOUT })).value;
 
-        const supportedChains = await context.api.getSupportedChains()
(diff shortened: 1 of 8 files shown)


ukstv (approved):
🏇

oed (approved):
😬

packages/stream-tile-handler/src/tile-document-handler.ts
```diff
@@ -137,13 +137,6 @@ export class TileDocumentHandler implements StreamHandler<TileDocument> {
         // TODO: Assert that the 'prev' of the commit being applied is the end of the log in 'state'
         const proof = (await context.ipfs.dag.get(commit.proof, { timeout: IPFS_GET_TIMEOUT })).value;
 
-        const supportedChains = await context.api.getSupportedChains()
```
stbrody, line 140: This isn't actually necessary and isn't really the right place to be doing this check anyway. The AnchorValidator already has a check for if the chainId changes from what the CAS has configured, this check is redundant and doesn't work when there is no CAS available

stbrody (commented):
Sorry, I did a poor job testing this PR initially. I just pushed a few small fixes that make it actually run successfully in `--disable-anchor` mode.  I'm going to go ahead and merge this, but if you have any concerns with anything I added post-review please let me know and I can do a follow-up patch if needed.  @ukstv @oed

PR #1506, opened 2021-06-28
Direct swarm connect with topology

v-stickykeys:
This is helpful for testing/debugging swarm connections

### DIFF
--- a/packages/ipfs-topology/src/ipfs-topology.ts
+++ b/packages/ipfs-topology/src/ipfs-topology.ts
@@ -76,10 +76,12 @@ export class IpfsTopology {
   ) {}
   
   /**
-   * Connect to custom peerList
+   * Connect to custom peer list
+   * @param peerList List of multiaddresses to swarm connect to
    */
-  async connect(peerList: Multiaddr[]): Promise<void> {
-    await this._forceBootstrapConnection(this.ipfs, peerList);
+  async connect(peerList: Array<string>) {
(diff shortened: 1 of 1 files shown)


packages/ipfs-topology/src/ipfs-topology.ts
```diff
@@ -74,6 +74,15 @@ export class IpfsTopology {
     readonly logger: DiagnosticsLogger,
     readonly period: number = DEFAULT_PEER_DISCOVERY_PERIOD
   ) {}
+  
+  /**
+   * Connect to custom peer list
+   * @param peerList List of multiaddresses to swarm connect to
+   */
+  async connect(peerList: Array<string>) {
```
stbrody, line 82: is this function meant to be used somewhere?

ukstv: +1 to this.

ukstv (changes requested):
Unclear how it fits.

packages/ipfs-topology/src/ipfs-topology.ts
```diff
@@ -74,6 +74,15 @@ export class IpfsTopology {
...
+    const addressList = peerList.map((peer) => { return new Multiaddr(peer) });
```
ukstv, line 83:
```suggestion
    const addressList = peerList.map((peer) => new Multiaddr(peer));
```

ukstv (approved):

PR #1510, opened 2021-06-29
fix(core): Add information for validating transactions on rinkeby

stbrody:
Without this change I get an error when trying to load a stream on dev-unstable that has been anchored against rinkeby.
```
Error loading transaction info for transaction 0xd05dd4d973596d766e9d1f2ae62aa7ca740480c96cd763d48d77cc19b8680518 from Ethereum: Error: No ethereum provider available for chainId eip155:4
Error: No ethereum provider available for chainId eip155:4
```

The fact that we're *not* seeing these errors in the integration tests has me pretty confused as to why not...

### DIFF
--- a/packages/core/src/anchor/ethereum/ethereum-anchor-validator.ts
+++ b/packages/core/src/anchor/ethereum/ethereum-anchor-validator.ts
@@ -26,8 +26,9 @@ interface EthNetwork {
  * Maps some of Ethereum chain IDs to network configuration
  */
 const ETH_CHAIN_ID_MAPPINGS: Record<string, EthNetwork> = {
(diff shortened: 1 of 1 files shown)


oed (approved):

stbrody:
> @stbrody No idea why we haven't been seeing errors in tests related to this. Probably worth investigating before merging this in?

Just read the code more closely and I think it makes sense assuming you updated the `ethereumRpcEndpoint` infura url. In that case it'll just use whatever URL is given.  My testing wasn't setting the endpoint so was relying on the default providers which come from this ETH_CHAIN_ID_MAPPINGS 
map.

I think this makes sense and I'm going to go ahead and merge this in.

PR #1516, opened 2021-07-01
feat(cli): add global sync override option

zachferland:
Follow up to - https://github.com/ceramicnetwork/js-ceramic/pull/1498

Adds sync override flag, as options to globally force sync mode on all requests. Required by gateway.


packages/core/src/ceramic.ts
```diff
@@ -121,8 +122,15 @@ export interface CeramicParameters {
   gateway: boolean,
   networkOptions: CeramicNetworkOptions,
   validateStreams: boolean,
+  syncOverride: SyncConfig
 }
 
+/**
+ * Global Sync Config Options
+ */
+export interface SyncConfig {
```
stbrody, line 131: why do we need this extra interface?  Couldn't `syncOverride` just be a `SyncOptions`?

stbrody: If you're trying to future-proof to a world where we may want to add other options to the override, such as `syncTimeoutSeconds` for example, you could have this be a `LoadOpts` and call the field `loadOptsOverride`

stbrody (commented):
looks good, one question about the new interface

packages/core/src/ceramic.ts
zachferland: Yeah was leaning towards that, just change/updated. Basically imagine options as something like this after
```
opts = { ...DEFAULT_CREATE_FROM_GENESIS_OPTS, this._loadOptsDefault, ...opts, ...this._loadOptsOverride}
```

stbrody (approved):
LGTM

zachferland:
Not sure whats up with tests right now, forgot to merge last week, merged/fixed with dev again, @ukstv or @stbrody please merge if tests are fine

ukstv:
@zachferland rebased on the current prettified develop in https://github.com/ceramicnetwork/js-ceramic/pull/1541

PR #1518, opened 2021-07-02
feat: Check the signature timestamp for genesis if either have a tip, or it is a lone genesis

ukstv:


packages/core/src/state-management/repository.ts
```diff
@@ -114,8 +114,8 @@ export class Repository {
     if (commit == null) {
       throw new Error(`No genesis commit found with CID ${genesisCid.toString()}`);
     }
-    // Do not check timestamp here
-    const state = await handler.applyCommit(commit, {cid: streamId.cid, timestamp: null}, this.#deps.context);
+    // Assume here genesis is always earlier than possible key revocation, so do not check a timestamp.
```
stbrody, line 117:
```suggestion
    // Do not check for possible key revocation here, as we will do so later after loading the tip (or learning that the genesis commit *is* the current tip), when we will have timestamp information for when the genesis commit was anchored. In order to disable this check for key rotation, set a timestamp that will be considered earlier than any other timestamp so it will always appear to have happened earlier than any possible key revocation.
```

stbrody: Tried fleshing out this comment a bit as I think it's tricky and likely to be confusing to future readers without the context, so merits a longer comment.  LMK if you think this is clear?

packages/core/src/state-management/repository.ts
```diff
@@ -114,8 +114,8 @@ export class Repository {
...
+    const state = await handler.applyCommit(commit, {cid: streamId.cid, timestamp: 1}, this.#deps.context);
```
stbrody, line 118: Actually thinking about this more, I'm not sure this actually works.  What if the version of the 3id that signed the message was neither the original version nor the current version, but was some version in the middle?  Then setting the timestamp to be so early will still make it look like it was revoked and cause the signature verification here to fail.  I think we need to skip signature verification here entirely rather than trying to opt-out of just the key revocation check (or we would need to add an arg to DID.verifyJWS to disable the key revocation check, but that seems like not something we would want to add)

packages/core/src/state-management/state-manager.ts
```diff
@@ -80,11 +80,14 @@ export class StateManager {
...
-      .toPromise().then(() => {
-        if (pinned) {
-          this.syncedPinnedStreams.add(state$.id.toString())
-        }
-      });
+      .toPromise()
+    if (pinned) {
+      this.syncedPinnedStreams.add(state$.id.toString())
+    }
+    // Still lone genesis, no time information in tips, so we do check the signature against _now_.
```
stbrody, line 87: We have to be very careful here - what if this `sync()` method never gets called?  For instance, right now if you look at `StateManager.load()`, when you load with `SyncOptions.NEVER_SYNC` the `sync()` method doesn't get called, which leaves a security hole since we never do the signature verification for the genesis commit.

stbrody (changes requested):
I'm pretty anxious about how subtle this stuff is and how easy it is to mess up and miss ever verifying the signature.  I'm worried that even if we get this right now that the likelihood is high of it breaking accidentally down the line.  It makes me wonder if we should bite the bullet and do the refactor to load the stream log from the tip before we ever load the genesis commit data.

In either case, high-level test coverage here that tests the behavior independent of the specific implementation is going to be key.

ukstv:
> I'm pretty anxious about how subtle this stuff is and how easy it is to mess up and miss ever verifying the signature. I'm worried that even if we get this right now that the likelihood is high of it breaking accidentally down the line. It makes me wonder if we should bite the bullet and do the refactor to load the stream log from the tip before we ever load the genesis commit data. That would give us a single code path where all the verification happens and should be much easier to understand and less likely to miss something (I hope).
> 
> In either case, high-level test coverage here that tests the behavior independent of the specific implementation is going to be key.

It is extremely fragile, I agree, that's why it is a separate PR as a _possible_ shortcut before doing it properly.

packages/core/src/state-management/repository.ts
ukstv: OK. 3id had some keys: v0, vA, vB. It signed a genesis with vA. We set `timestamp:1` here. Because of key revocation, vA has `nextUpdate` set to some date.

js-did would go into `if(nextUpdate)` section. Then `isEarlier = timestamp && timestamp < nextUpdate` ≡ `isEarlier = 1 && 1 < nextUpdate` ≡ [`1 < nextUpdate ≝ true`] ≡ `isEarlier = 1 && true` ≡ `isEarlier = true` ⟹ no throw.

ukstv: Here:
- `≡` means equivalence transformation,
- `≝` means "we can reasonable assume",
- square brackets are used to indicate a reason for transformation.

packages/core/src/state-management/state-manager.ts
ukstv: good catch.

packages/core/src/state-management/repository.ts
stbrody: Hmm... I think you're right with how the code is in js-did right now it won't throw, but that actually feels like a bug?  You don't just care that the timestamp of the signature was before the key was rotated, you care that the signature was during the time window when that key was valid.  In the example you gave above, we probably shouldn't accept a signature using vA that was made before vA was even added to the 3id, since that signature wasn't actually valid at that moment in time.

packages/core/src/state-management/state-manager.ts
stbrody: I think we could probably move the `verifyLoneGenesis` call up to `StateManager.load()` and that would probably be a better/more clear place for it.

packages/core/src/state-management/repository.ts
oed: I agree that if we are nit picky a 3ID version should only be valid between the `updated` and `nextUpdated` metadata properties. However, since it's unlikely someone had access to a key before it was added to the 3ID document I don't see any real harm by not checking this.

ukstv: Actually, it is an impossible situation for 3id to use a key before it is added, if only anchored versions are resolved.

ukstv: It's one thing to possess a private key, and another one is to reference it. The latter includes versionId, which is basically impossible to know beforehand. The version is a CID of the anchor commit, which depends on anchor transaction id, and block timestamp, and the merkle tree state which are intractable to guess beforehand. Without ability to reference a key predictably, one can not really use it, even if the key is in possession.

So, if I am not mistaken here, the code should only check if a key is used after revocation.

stbrody: Does the signature include the DID string of the did that is performing the signature?  I'm thinking about a case where a user uses the same key across multiple DIDs (for example, maybe they use their eth wallet as a signing key for their 3id, but also sign some messages with a did:key of the same key), or where they use the key directly outside of a DID.  Definitely an edge case situation that maybe we don't care about, but could possibly lead to a world where a commit in a stream log was manually signed with a key before it got added as a signing key to the 3id.

But yeah, I'm definitely reaching here trying to find cases where this could break down.  It does seem like in the majority of standard operation the check as written here is probably sufficient.

ukstv: > Does the signature include the DID string of the did that is performing the signature?

The JWT includes `kid` field in the header (just in case, header + payload are signed as a single blob, so you can consider it "part of the signature"). `kid` is a DID URL to a key that has a form `did:3:kjzl6cwe1jw14ah8wjy8grgzl52sl18sbyirgi9bqy9yzu28kbxvjmhip99r14k?version-id=bafyreiebh2rwia4qq5v43dqlnofkczsvoghdxymahv4p5lkp5ivr5kgatm#C26y2DZeJ9EMYsF`. `version-id` parameter is quasi-random, as I explained in the previous comment, so even if you possess a key, you can not create a signature (=signed JWT) before you know this `version-id`.

stbrody: Okay great, thanks for the explaining!  So then it sounds like even if you have access to the same private key, you can't actually use it to make a commit that looks like it's part of the 3id before it's actually truly added to the 3id, so that seems good then and addresses my original concern

PR #1519, opened 2021-07-02
fix(stream-caip10-link): better genesis determinism

oed:
Saw in the convo space app that my mainnet account has two caip10-links for the same address. This PR fixes this problem.

image


ukstv:
Yay!

ukstv (approved):
🏇

packages/stream-caip10-link/src/caip10-link.ts
```diff
@@ -122,7 +122,11 @@ export class Caip10Link extends Stream {
      * @param accountId
      */
     static makeGenesis(accountId: AccountID): GenesisCommit {
-        return { header: { controllers: [accountId.toString()],
+        // Ethereum addresses specifically are sometimes encoded with mixed case and
+        // sometimes all lower case. In order to deal with this and not have different
+        // links for different addresses.
+        const controller = accountId.toString().toLowerCase()
```
stbrody, line 128: how confident are we that no blockchains will ever have case-sensitive addresses?

It would be a pretty hairy decision to make as a blockchain designer, so here's hoping it never happens...

stbrody (approved):
LGTM

ukstv:
Added relevant issue on caip repo: https://github.com/pedrouid/caip-js/issues/11

packages/stream-caip10-link/src/caip10-link.ts
oed: I'm pretty sure for bitcoin this is not the case: https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-10.md

I do think we need a better strategy mid/long term. Alternatively we add specific logic for each blockchain here, which is not ideal

stbrody: Yeah wait, the CAIP10 spec specifically says that the accountId is case-senstive.  Blindly translating it to lower case feels wrong then and like it's corrupting the links.  If eth addresses are case insensitive, then that needs to be handled in an eth-specific way, not blindly applied to all possible Caip10Links.

stbrody (changes requested):
revoking my lgtm after reviewing the caip10 spec

packages/stream-caip10-link/src/caip10-link.ts
oed: Ok, we should do it properly and encode eth addresses as specified by [EIP-55](https://eips.ethereum.org/EIPS/eip-55).

oed: On second thought, doing EIP-55 could potentially cause issues for 3IDs which have already been migrated/created on 3ID Connect. Would like @zachferland to weigh in here if this matters or not.

stbrody: > On second thought, doing EIP-55 could potentially cause issues for 3IDs which have already been migrated/created on 3ID Connect. Would like @zachferland to weigh in here if this matters or not.

holding off on final review/approval of this PR until this question is addressed.  It does seem like using EIP-55 encoding would be the "better" solution than just lower-casing it (as https://github.com/pedrouid/caip-js/issues/11 is probably the best long-term solution and I imagine that that's what it would do). But if that breaks 3id-connect that's obviously a problem

zachferland:
Looks good, were they not using blockchain-utils-linking library? we already normalize or lowercase addresses for links there, all created through 3id-connect should already be lower case

eip55 would break 3id-connect in current form, and if changed yeah would require some backwards supports to make sure migration paths works fine. In any case it would require creating a new link. So on the flipside, the earlier it is the changed the less people that need to resign a link proof (not huge overhead though)

Would lean towards this fix, since kinda of hard to prioritize changes in 3id-connect for this right now

burk3:
fwiw, we (DNS) have been using blockchain-utils-linking but not 3id-connect and have been creating caip10links and rows in cryptoAccounts w/ EIP-55 addresses. The main, awkward result of this is that when a user with a wallet makes a DID on dns.xyz, attempting to log into (clay.)self.id with the same wallet gives the user the new account flow, but results in the same DID and profile info from DNS still shows up just fine. The user will also end up with two cryptoAccount rows (one with EIP-55 addr, one with all lower) each with their own caip10link.

So yea, extra signature, but no big deal.

oed:
Interesting @burk3 I noticed something similar when I authenticated to another app that was using 3ID Connect. I got the migration flow, but in the end ended up with the same 3ID. If that's the only problem (i.e. no new 3ID is generated) then we should use EIP-55 for sure!
Thoughts @zachferland ?

Geo25rey:
@oed Are you suggesting to only allow EIP-55 addresses in the cryptoAccounts schema? Would that include altering the schema itself to restrict Ethereum based addresses (if that's even possible)?

oed:
@Geo25rey No I don't think that's possible as you say. This is more about a restriction in the caip10-link doctype.

Geo25rey:
@oed does that mean there should be both EIP-55 addresses and lowercase addresses pointing to the same caip10-link stream ID?

oed:
> @oed does that mean there should be both EIP-55 addresses and lowercase addresses pointing to the same caip10-link stream ID?

I think *if* we decide to go with eip55 we should strive to have the entries in the cryptoAccounts also eip55 encoded. There's no way to enforce this on the schema level however.

packages/stream-caip10-link/src/caip10-link.ts
oed: After a conversation with @zachferland we decided that we should go for the EIP-55 path. Will be updating this PR shortly.

stbrody (approved):
👍

stbrody:
Does this require a corresponding change to IDX to handle the cryptoAccounts definition? @oed @PaulLeCam

oed:
> Does this require a corresponding change to IDX and/or 3id-connect to handle the cryptoAccounts definition?

No.

stbrody:
> No for IDX definition. Might need a change in 3id-connect.

Cool, I was mostly just responding to this earlier comment from you and trying to make sure we didn't forget to follow up if we need to:

> I think if we decide to go with eip55 we should strive to have the entries in the cryptoAccounts also eip55 encoded

PR #1523, opened 2021-07-06
chore: Format all TS code using prettier

ukstv:
1. Install [eslint-config-3box](https://github.com/3box/eslint-config-3box)
2. Use its prettier config
3. `prettier --write ./**/src/*.ts && prettier --write ./**/src/**/*.ts`

No real code changes.


oed:
Nice, added @PaulLeCam as a reviewer as he created the eslint plugin.

ukstv:
Apparently, there are few changes to code (like removing semicolons here and there) that were necessary to satisfy ESLint.

PaulLeCam (approved):
The relevant changes are in `package.json` and the rest are the changes applied by ESLint, right? If so, looks all good to me!

ukstv:
@PaulLeCam It is not ESLint. It is `prettier --write` that I used.

ESLint configuration from 3box is not enabled here. If enabled, it results with too much errors.

oed:
> ESLint configuration from 3box is not enabled here. If enabled, it results with too much errors.

@ukstv does it still make sense to enable this at some point in the future?

ukstv:
> > ESLint configuration from 3box is not enabled here. If enabled, it results with too much errors.
> 
> @ukstv does it still make sense to enable this at some point in the future?

I have a strange feeling about the topic. Consistent code style is good. IMO "prettier" and strict TS compiler settings fit the bill just right. They do not require a lot of dependencies, and a lot of configuration, while achieving the same result as ESLint. I have yet to see a bug or inconsistency that can be prevented by ESLint but not TSC+prettier.

PR #1524, opened 2021-07-06
feat: Drop "created" metadata property from 3id did resolution result

ukstv:
"created" timestamp refers to a time of DID Create Operation. By design of 3id, create operation timestamp could be any time in the past, and there is no way to determine it via anchor commits. An anchor commit is only present after 3id update operation, which is effectively key rotation. So, "created" field has to be removed.

### DIFF
--- a/packages/3id-did-resolver/src/index.ts
+++ b/packages/3id-did-resolver/src/index.ts
@@ -97,11 +97,7 @@ function extractMetadata(
     latestVersionState.log.find(
       ({ timestamp }) => timestamp > updated || (!updated && timestamp)
     ) || {}
-  const created = latestVersionState.log.find(({ timestamp }) => Boolean(timestamp))?.timestamp
 
-  if (created) {
-    metadata.created = formatTime(created)
-  }
   if (updated) {
     metadata.updated = formatTime(updated)
   }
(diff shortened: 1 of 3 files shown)


stbrody:
I feel like I may have missed some context of a conversation about this.  Can you add to the description a brief explanation of what this is trying to accomplish @ukstv?

ukstv:
Added background info.

packages/3id-did-resolver/src/index.ts
```diff
@@ -14,6 +14,7 @@ import LegacyResolver from './legacyResolver'
 import * as u8a from 'uint8arrays'
 import { StreamID } from '@ceramicnetwork/streamid'
 import CID from 'cids'
+import { errorRepresentation, withResolutionError } from './error-representation'
```
oed, line 17: Nice

oed (approved):
Nice cleanups 👍

stbrody:
@oed Is there a reason we don't anchor 3id genesis commits?  We certainly could if it would be useful for giving us more timestamp information.

oed:
We certainly could. Not sure if there may be some logic in 3id-did-provider that would be broken because of doing that though. We would also need to consider v03IDs and if there are different considerations for them.

Issue #1520, opened 2021-07-03
3id-did-resolver reports un-anchored version


ukstv:
Scenario:
1. A person creates a 3id `did:3:foo` (`foo` here is a placeholder for the actual identifier).
2. 3id-did-resolver is called to resolve `did:3:foo`: it reports `versionId: "0"`.
3. A person initiates key rotation.
4. 3id-did-resolver is called to resolve `did:3:foo` (no version), before the respective tile gets anchored: it reports `versionId: "update-commit"`, where `update-commit` is a CID of the update commit.

I assumed, that 3id DID is versioned by anchor commits. If the assumption is true, on step 4 the resolver should respond with version 0, rather than un-anchored version. And only after the respective tile gets anchored, `did:3:foo` should resolve to a new version.

If it is indeed a bug, the fix is to resolve only the last _anchored_ version.

CC @oed

oed:
Interesting nuance. Don't have a strong opinion what the correct behavior should be. Most of the time we assume that the DID provider adds a `versionId` in the `kid`, so I don't think we would run into any issues in Ceramic. However the 3ID DID Provider doesn't use the new keys to sign until the update has been anchored, so it could still make sense to just resolve the latest anchored version as you are suggesting.

ukstv:
It is stated in the spec, it makes sense, so I definitely consider this a bug.

[Spec](https://github.com/ceramicnetwork/CIP/blob/main/CIPs/CIP-79/CIP-79.md#did-document-metadata): "versionId - should be equal to the commit CID from the most recent AnchorCommit"

PR #1527, opened 2021-07-08
feat: Return last anchored state

ukstv:
3id-did-resolver returns now only the latest anchored state, or genesis state for vanilla DID URL like `did:3:foo`. See #1520

### DIFF
--- a/packages/3id-did-resolver/src/index.ts
+++ b/packages/3id-did-resolver/src/index.ts
@@ -83,16 +83,15 @@ export function wrapDocument(content: any, did: string): DIDDocument | null {
  * Return last anchor log entry, or genesis if no anchors found
  */
 function lastAnchorOrGenesisEntry(log: LogEntry[]): LogEntry {
-  // Genesis
-  if (log.length === 1) {
-    return log[0]
+  // Look for last anchor
+  for (let index = log.length - 1; index >= 0; index--) {
+    const entry = log[index]
+    if (entry.type === CommitType.ANCHOR) {
+      return entry
+    }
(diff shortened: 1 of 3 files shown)


packages/3id-did-resolver/src/index.ts
```diff
@@ -78,6 +79,22 @@ export function wrapDocument(content: any, did: string): DIDDocument | null {
   }, startDoc)
 }
 
+/**
+ * Return last anchor log entry, or genesis if no anchors found
+ */
+function lastAnchorOrGenesisEntry(log: LogEntry[]): LogEntry {
```
oed, line 85: Is recursion better than a for loop here?

oed (commented):
Looks good, just one question.
Also please try to provide more info in the PR description. Other ppl have minimal context here.

packages/3id-did-resolver/src/index.ts
ukstv: It is more idiomatic.

ukstv:
@oed there is an issue linked to PR. I can copy-paste the description here, but what is the point of linking the issue then?

298px-Prap

packages/3id-did-resolver/src/index.ts
ukstv: changed to `for` loop anyway

oed (approved):
Approved once tests are passing. 
My bad @ukstv Didn't see the linked issue.

oed:
But a short description like the one you added is always nice ;)

PR #1528, opened 2021-07-08
fix(core): Optimize commit application to minimize calls to IPFS

smrz2001:


packages/core/src/conflict-resolution.ts
```diff
@@ -389,4 +377,19 @@ export class ConflictResolution {
...
+  /**
+   * Return the log entry with its commit and JWS envelope (if applicable) populated if they're not already present.
+   */
+  private async expandLogEntry(logEntry: LogEntry): Promise<LogEntry> {
+    if (!logEntry.commit) {
+      const commit = await this.dispatcher.retrieveCommit(logEntry.cid)
+      logEntry.commit = commit
+      if (StreamUtils.isSignedCommit(commit)) {
+        logEntry.commit = await this.dispatcher.retrieveCommit(commit.link)
+        logEntry.envelope = commit
```
oed, line 390: This seems backwards? Shouldn't the JWS be the envelope?

packages/core/src/conflict-resolution.ts
```diff
@@ -271,24 +267,21 @@ export class ConflictResolution {
...
-
-      if (payload.proof) {
-        // it's an anchor commit
+      if (logEntry.commit.proof) {
+        // It's an anchor commit
         // TODO: Anchor validation should be done by the StreamHandler as part of applying the anchor commit
-        await verifyAnchorCommit(this.dispatcher, this.anchorValidator, commit)
-        state = await handler.applyCommit(commit, logEntry, this.context, state)
+        await verifyAnchorCommit(this.dispatcher, this.anchorValidator, logEntry.commit)
+        state = await handler.applyCommit(logEntry.commit, commitMeta, this.context, state)
```
oed, line 278: Maybe we should modify `applyCommit` to take an logEntry object?
Like so:
```suggestion
        state = await handler.applyCommit(logEntry, this.context, state)
```

packages/core/src/conflict-resolution.ts
```diff
@@ -271,24 +267,21 @@ export class ConflictResolution {
...
       } else {
-        // it's a signed commit
-        const tmpState = await handler.applyCommit(commit, logEntry, this.context, state)
-        const isGenesis = !payload.prev
+        // It's a signed commit but may not be using DagJWS. If the JWS envelope is present, use that while applying the
+        // commit, otherwise use the commit itself (e.g. for CAIP-10 links).
+        const commitToApply = logEntry.envelope ? logEntry.envelope : logEntry.commit
+        const tmpState = await handler.applyCommit(commitToApply, commitMeta, this.context, state)
```
oed, line 283: Don't we need to pass both the envelope and commit here?

packages/common/src/stream.ts
```diff
@@ -118,6 +118,11 @@ export interface LogEntry {
   cid: CID
   type: CommitType
   timestamp?: number
+  // Holding on to these fields when fetching and processing new commits significantly reduces the number of IPFS
+  // lookups required. Use of these fields is currently limited to `conflict-resolution.ts` but further optimization of
+  // the code is possible by relaying them to code that looks the fields up again.
+  commit?: any
+  envelope?: DagJWS
```
ukstv, line 125: Why both `commit` and `envelope`?

packages/core/src/conflict-resolution.ts
```diff
@@ -190,32 +194,24 @@ export async function fetchLog(
...
-  if (isSignedCommit) {
-    payload = await dispatcher.retrieveCommit(commit.link)
-    if (!payload) throw new Error(`No commit found for CID ${commit.link.toString()}`)
-  }
-  const prevCid: CID = payload.prev
-  if (!prevCid) {
-    // this is a fake log
-    return []
-  }
+  let prevCid: CID = commit.prev
```
ukstv, line 197: Previous version of the code ensured `prevCid` is really a `CID`, this one makes it `CID | undefined` until line 203. Fishy. Is it necessary to do it this way? Types are quite important for understanding code.

packages/common/src/stream.ts
smrz2001: Yes, the `commit` would always be a plain raw or anchor commit. In the case of a `SignedCommit`, it will be the linked commit. The `envelope` would be the DagJWS, in case of a `SignedCommit`.

It made sense to separate those two concepts versus considering them all under the same term. Spencer did suggest calling it `signature` instead of `envelope` but I thought the former has a more specific meaning in the code as well.

packages/core/src/conflict-resolution.ts
smrz2001: My thinking was that even in the earlier code, if this was a "fake log" as the comment says (I don't know what that means) or a `GenesisCommit` (?), that `prevCid` could still be `undefined` and so this logic essentially doesn't change the spirit of the code. I can definitely rewrite this though.

packages/core/src/conflict-resolution.ts
oed: Ah, sorry I misread!

packages/core/src/conflict-resolution.ts
smrz2001: That was definitely one of the things I'd done in a previous iteration of the code before I trimmed it down. My thinking here was to avoid changing any interfaces and increasing the scope of the changes, since more code and tests would be affected. I plan to open a new ticket to update the interface because the IPFS lookup inside `applyCommit` can totally be removed too.

oed: cc @ukstv @stbrody

packages/core/src/conflict-resolution.ts
smrz2001: Eventually, yes. Since I've left the IPFS lookup inside `applyCommit` in there, passing the envelope is sufficient for the moment. In a subsequent update, if I just pass the `LogEntry` itself, I'll be able to remove the internal lookup as well.

oed: ok, fair enough :)

packages/core/src/conflict-resolution.ts
smrz2001: Hmm, isn't that what this code is doing? The naming here has definitely been very confusing for me to follow
 `commit` could be a `SignedCommit` and, if so, it is stored in `envelope`.

oed: You are right, I misread the code.

oed:
@smrz2001 Fyi adding "Fixes #1421:" to the commit message will close that issue when this PR is merged. This messes with our flow on Zenhub because we usually move tickets to *Staging* until it has been part of a release, then close it.

smrz2001:
> @smrz2001 Fyi adding "Fixes #1421:" to the commit message will close that issue when this PR is merged. This messes with our flow on Zenhub because we usually move tickets to _Staging_ until it has been part of a release, then close it.

Ahhh, ok, didn't think about that @oed. I was trying to figure out a way to automatically link the PR to the issue. Is the trick to just mention the ticket in a commit message without adding "Fixes"?

oed:
@smrz2001 I'm not sure if you can link it automatically with Zenhub tbh!

packages/common/src/stream.ts
```diff
@@ -118,6 +118,11 @@ export interface LogEntry {
   cid: CID
   type: CommitType
   timestamp?: number
+  // Holding on to these fields when fetching and processing new commits significantly reduces the number of IPFS
+  // lookups required. Use of these fields is currently limited to `conflict-resolution.ts` but further optimization of
+  // the code is possible by relaying them to code that looks the fields up again.
+  commit?: any
```
stbrody, line 124: I don't think we can put these into LogEntry as then we'll be keeping the contents of every commit around in the log within the state object for each stream, which could be a significant memory overhead. We need a separate structure that is just used for the log as we load and apply it, and then a smaller-subset of that data can go into the LogEntry's that go into the state object of the fully-loaded stream

packages/core/src/conflict-resolution.ts
```diff
@@ -190,32 +194,24 @@ export async function fetchLog(
...
-    nextEntry = {
-      type: CommitType.SIGNED,
-      cid: cid,
-      timestamp: timestamp,
-    }
+    // For all cases not using DagJWS for signing (e.g. CAIP-10 links)
+    nextEntry = { cid: cid, type: CommitType.SIGNED, commit: commit, timestamp: timestamp }
+  }
+  if (!prevCid) {
+    // this is a fake log
```
stbrody, line 213: any idea what this means?  Is this just supporting unit tests?

packages/core/src/conflict-resolution.ts
```diff
@@ -271,24 +267,21 @@ export class ConflictResolution {
     const itr = log.entries()
     let entry = itr.next()
     while (!entry.done) {
-      const logEntry = entry.value[1]
-      const commit = await this.dispatcher.retrieveCommit(logEntry.cid)
+      const logEntry = await this.expandLogEntry(entry.value[1])
```
stbrody, line 270: why do we need this call?  Won't this have already been done by `fetchLog`?

packages/core/src/conflict-resolution.ts
```diff
@@ -271,24 +267,21 @@ export class ConflictResolution {
...
       // TODO - should catch potential thrown error here
 
-      let payload = commit
-      if (StreamUtils.isSignedCommit(commit)) {
-        payload = await this.dispatcher.retrieveCommit(commit.link)
-      }
-
-      if (payload.proof) {
-        // it's an anchor commit
+      if (logEntry.commit.proof) {
```
stbrody, line 274:
```suggestion
      if (StreamUtils.isAnchorCommit(logEntry.commit)) {
```

packages/common/src/stream.ts
stbrody: could go the other way and make LogEntry be what ConflictResolution uses during the load process and make a new type for what goes in the state object, but the main point is that these two things need to be different types so that we're not holding all the raw IPFS commit data in memory for every stream

packages/core/src/conflict-resolution.ts
stbrody: Yeah, I think it's fine to do this in a follow-up PR, but should definitely be done.

stbrody (changes requested):
My main comment is about splitting LogEntry into two types - one that's used in ConflictResolution and one that's used in the StreamState - so that we don't wind up storing commit data in the state object for an entire stream's lifetime.

packages/core/src/conflict-resolution.ts
smrz2001: Opened #1532 and added it to the backlog.

packages/common/src/stream.ts
smrz2001: Yes, agreed about not storing the contents of all commits in the log. That won't actually happen at the moment because the two new fields (`commit` and `envelope`) are only populated and used during the application of a new `tip`, then discarded.

I was attempting to reduce IPFS lookups during tip application without changing adjacent code too much but if this seems like a hack then I can change the approach.

packages/core/src/conflict-resolution.ts
smrz2001: LOL I was wondering the same thing. Can this condition occur when the genesis commit is processed?

packages/core/src/conflict-resolution.ts
smrz2001: I did it this way because there were calls being made to `applyLogToState()` via  `snapshotAtCommit()` _without_ a prior call to `fetchLog()`. This seemed like a good reason to just do it everywhere. If the commit had already been unwrapped, the extra function call would just be a no-op.

Besides, now we have one place where commits are unwrapped that we can update/optimize in the future.

packages/common/src/stream.ts
smrz2001: Overloading `LogEntry` did seem a bit of a hack when I did it TBH but it was so tempting because the new structure would need all the same fields as `LogEntry` but one (`type: CommitType`) 😅

I wouldn't mind creating a new structure for loading/processing new commits. Something like the one you sent to me earlier:

```
interface CommitData {
  cid: CID
  commit: CeramicCommit
  envelope?: DagJWS
  timestamp?: number
}
```

smrz2001: In other words, `fetchLog` (and other functions) would change to something like this:

```
export async function fetchLog(
  dispatcher: Dispatcher,
  cid: CID,
  stateLog: HistoryLog,
-  log: LogEntry[] = [],
+  newCommits: CommitData[] = [],
  timestamp?: number
): Promise<LogEntry[]> {
```

smrz2001: I would then, however, have to translate `LogEntry` to `CommitData` because `snapshotAtCommit()` calls `applyLogToState()` using part of `StreamState.log` (array of `LogEntry`).

stbrody: interfaces can extend each other.  So maybe CommitData extends LogEntry and adds the `commit` and `envelope` fields.  Or maybe they both extend some smaller interface so LogEntry can add `type` without it being needed in `CommitData`

packages/core/src/conflict-resolution.ts
stbrody: okay fair enough, makes sense

packages/core/src/conflict-resolution.ts
oed: This means that someone sent a tip that is a fake log. I.e. a log that at some point does not point to a previous commit or genesis.

stbrody: Ah yes, that makes sense, thanks!  Maybe you could expand the comment a little while we're in there Mohsin?

packages/common/src/stream.ts
smrz2001: How does this look @stbrody? I like `CommitData` extending `LogEntry` even if `type` isn't quite necessary in `CommitData`.

packages/core/src/conflict-resolution.ts
```diff
@@ -389,4 +379,19 @@ export class ConflictResolution {
     const handler = this.handlers.get(initialState.type)
     return this.applyLogToState(handler, resetLog)
   }
+
+  /**
+   * Return `CommitData` with commit and JWS envelope, if applicable and not already present.
+   */
+  private async getCommitData(commitData: CommitData): Promise<CommitData> {
```
stbrody, line 386: the argument here should be of type `LogEntry`, right?  Since we want to be able to call this with or without the `commit` and `envelope` fields already present?

packages/core/src/conflict-resolution.ts
```diff
@@ -129,7 +129,12 @@ export async function pickLogToAccept(
 
 export class HistoryLog {
   static fromState(dispatcher: Dispatcher, state: StreamState): HistoryLog {
-    return new HistoryLog(dispatcher, state.log)
+    return new HistoryLog(
+        dispatcher,
+        state.log.map((logEntry) => {
```
stbrody, line 134: I don't think this should be necessary anymore now that the state log just holds LogEntry not CommitData

packages/core/src/conflict-resolution.ts
stbrody: Maybe you could wait to create `prevCid` until after building `nextCommitData`, then take `nextCommitData.commit.prev`?

stbrody (commented):
looking pretty good, a few small comments, mostly on things that I think got lost in the shuffle of the last few updates

packages/core/src/conflict-resolution.ts
smrz2001: Good idea!

smrz2001: And I think I finally see what @ukstv was trying to say 🤦🏼

packages/core/src/conflict-resolution.ts
smrz2001: Everywhere it's being used right now passes `CommitData` (`commit` and `envelope` are nullable) but I like the idea of passing `LogEntry`.

packages/core/src/conflict-resolution.ts
```diff
@@ -389,4 +373,26 @@ export class ConflictResolution {
     const handler = this.handlers.get(initialState.type)
     return this.applyLogToState(handler, resetLog)
   }
+
+  /**
+   * Return `CommitData` with commit and JWS envelope, if applicable and not already present.
+   */
+  private async getCommitData(logEntry: LogEntry): Promise<CommitData> {
```
smrz2001, line 380: How does this look @stbrody? Passing `LogEntry` is not quite necessary right now since `CommitData` is available wherever this method is called but could be useful during further optimizations.

smrz2001: If there's a cleaner way to do this let me know (cc @ukstv).

packages/core/src/conflict-resolution.ts
smrz2001: I tried that earlier but reverted it because the extra fields kept getting returned making tests fail. I realized today that that was because of a different bug in my code 😞 I've now fixed that bug and removed this change.

packages/core/src/conflict-resolution.ts
```diff
@@ -172,59 +172,51 @@ export class HistoryLog {
...
+    nextCommitData = { cid: cid, type: CommitType.SIGNED, commit: linkedCommit, envelope: commit, timestamp: timestamp }
+  } else if (StreamUtils.isAnchorCommit(commit)) {
     const proof = await dispatcher.retrieveFromIPFS(commit.proof)
     timestamp = proof.blockTimestamp
-    nextEntry = {
-      type: CommitType.ANCHOR,
-      cid: cid,
-      timestamp: timestamp,
-    }
+    nextCommitData = { cid: cid, type: CommitType.ANCHOR, commit: commit, timestamp: timestamp }
```
oed, line 201: One optimization (maybe for later?) we could do here is to also retrieve the merkle wittness and put it into the `nextCommitData` object.

packages/core/src/conflict-resolution.ts
```diff
@@ -172,59 +172,51 @@ export class HistoryLog {
...
-      cid: cid,
-      timestamp: timestamp,
-    }
+    // For all cases not using DagJWS for signing (e.g. CAIP-10 links)
+    nextCommitData = { cid: cid, type: CommitType.SIGNED, commit: commit, timestamp: timestamp }
+  }
+  let prevCid: CID = nextCommitData.commit.prev
+  if (!prevCid) {
+    // Someone sent a tip that is a fake log, i.e. a log that at some point does not refer to a previous or genesis
+    // commit.
```
oed, line 209: 👍

packages/core/src/conflict-resolution.ts
smrz2001: Cool, yes 👍🏼

packages/core/src/conflict-resolution.ts
smrz2001: Although, if the method took `CommitData` like it did originally, wouldn't it work just the same? `CommitData` extends `LogEntry` and so passing a `LogEntry` structure would just be equivalent to a `CommitData` structure without the `commit` and `envelope` fields (which are optional).

packages/core/src/conflict-resolution.ts
smrz2001: Although, if the method took `CommitData` like it did originally, wouldn't it work just the same? `CommitData` extends `LogEntry` and so passing a `LogEntry` structure would just be equivalent to a `CommitData` structure without the `commit` and `envelope` fields (which are optional).

packages/core/src/conflict-resolution.ts
```diff
@@ -172,59 +172,51 @@ export class HistoryLog {
...
-  }
-  const prevCid: CID = payload.prev
-  if (!prevCid) {
-    // this is a fake log
-    return []
-  }
-  let nextEntry: LogEntry
-  if (StreamUtils.isAnchorCommit(commit)) {
+  let nextCommitData: CommitData
+  if (StreamUtils.isSignedCommit(commit)) {
```
oed, line 194: Can we create a ticket for renaming this function to `isJWSCommit` as we may support other signature formats in the future.

packages/core/src/conflict-resolution.ts
oed: Mind creating a ticket to track this?

packages/core/src/conflict-resolution.ts
oed: Seems like you should just use the *CommitData* type here?
```suggestion
  private async getCommitData(logEntry: CommitData): Promise<CommitData> {
```

oed (approved):
lgtm!

packages/core/src/conflict-resolution.ts
smrz2001: Yes, that's how it was before. Will revert!

packages/core/src/conflict-resolution.ts
smrz2001: Done #1544

packages/core/src/conflict-resolution.ts
smrz2001: Done #1545

packages/core/src/conflict-resolution.ts
stbrody: right, that was kind of my point.  In my mind `getCommitData` converts a LogEntry (which doesn't have the two extra fields) to a CommitData (which does).  It's not a big deal either way though

packages/core/src/conflict-resolution.ts
```diff
@@ -172,59 +172,51 @@ export class HistoryLog {
...
-    }
+    nextCommitData = { cid: cid, type: CommitType.ANCHOR, commit: commit, timestamp: timestamp }
   } else {
-    nextEntry = {
-      type: CommitType.SIGNED,
-      cid: cid,
-      timestamp: timestamp,
-    }
+    // For all cases not using DagJWS for signing (e.g. CAIP-10 links)
+    nextCommitData = { cid: cid, type: CommitType.SIGNED, commit: commit, timestamp: timestamp }
```
stbrody, line 204: I wonder if we should have a `CommitType.RAW` or `CommitType.UNSIGNED` for these types of commits?  Anyway that's outside the scope of this PR

packages/core/src/conflict-resolution.ts
smrz2001: Ah, I see. It does also take `CommitData` sometimes and pass it through without changes if `commit` and `envelope` were already present (e.g. from `applyLog` after `fetchLog` has been called).

packages/core/src/conflict-resolution.ts
```diff
@@ -255,40 +247,37 @@ export class ConflictResolution {
...
-        state = await handler.applyCommit(commit, logEntry, this.context, state)
+        await verifyAnchorCommit(this.dispatcher, this.anchorValidator, logEntry.commit)
+        state = await handler.applyCommit(logEntry.commit, commitMeta, this.context, state)
       } else {
-        // it's a signed commit
-        const tmpState = await handler.applyCommit(commit, logEntry, this.context, state)
-        const isGenesis = !payload.prev
+        // It's a signed commit but may not be using DagJWS. If the JWS envelope is present, use that while applying the
+        // commit, otherwise use the commit itself (e.g. for CAIP-10 links).
+        const commitToApply = logEntry.envelope ? logEntry.envelope : logEntry.commit
```
stbrody, line 278: I assume this will go away with the follow-up PR that makes the handler utilize the CommitData more directly?

packages/core/src/conflict-resolution.ts
```diff
@@ -389,4 +373,27 @@ export class ConflictResolution {
     const handler = this.handlers.get(initialState.type)
     return this.applyLogToState(handler, resetLog)
   }
+
+  /**
+   * Return `CommitData` with commit and JWS envelope, if applicable and not already present.
+   */
+  private async getCommitData(_commitData: CommitData): Promise<CommitData> {
+    // Clone the `CommitData` so that the Stream state is not affected when commit/JWS data is added to the structure
+    const commitData: CommitData = {
```
stbrody, line 382: would `cloneDeep` suffice here?

stbrody (approved):
LGTM, nice work!

packages/core/src/conflict-resolution.ts
oed: What do you mean by RAW? To me it's very unclear what that word implies here.

packages/core/src/conflict-resolution.ts
oed: Did we end up creating a ticket that tracks this?

packages/core/src/conflict-resolution.ts
stbrody: yep, sometimes, but CommitData is LogEntry so that's fine.  In my mind CommitData is always LogEntry but LogEntry is not always CommitData.  It so happens that right now LogEntry *is* always CommitData, because the extra fields that CommitData adds are all nullable, but that's a happy accident for how things happen to work currently, not something that's necessarily true at the conceptual level regarding the relationship between the two types.  This is why I would have the arg as a LogEntry personally, but all of this is me being extremely pedantic and doesn't actually matter at all, feel free to leave it is as-is, if the structure of CommitData changes in the future in a way that breaks this, we can easily change it then.

packages/core/src/conflict-resolution.ts
stbrody: the same as it does here: https://github.com/ceramicnetwork/js-ceramic/blob/feat/%231421/packages/common/src/stream.ts#L49

We need a name for a commit that isn't an anchor commit but also isn't signed. Like what Caip10Link uses.  We have called them UnsignedCommits in the past, but I don't love defining them by what they're *not* instead of what they *are*. This very PR is about to rename `UnsignedCommit` to `RawCommit`, so that's where I got the idea for `CommitType.RAW`

packages/core/src/conflict-resolution.ts
smrz2001: Yes, that sounds good. Ty for the discussion 👍🏼

packages/core/src/conflict-resolution.ts
smrz2001: I did, yes, #1532

smrz2001: And to answer your question, @stbrody, yes, this will go away.

packages/core/src/conflict-resolution.ts
smrz2001: Hah, that was my first thought too. With `cloneDeep`, for some reason, stream state comparison keeps failing in one of the Ceramic tests. I didn't want to hold this PR up to debug but I'm pretty sure it's because the current way populates all the fields, even optional ones (which could be `undefined`) versus cloning, which does not.

I _hated_ leaving it like this but it's about time for this PR to go in so I'll treat this as tech debt and take a look with the next ticket.

stbrody: okay sounds good. Would be nice to understand why cloneDeep breaks things, as I don't think it should, but fine to not hold up this PR over it

packages/core/src/conflict-resolution.ts
oed: Yeah, don't lover *RawCommit* either. Raw could mean litterally anything, just the raw data. What's the reasoning for not calling it *UnsignedCommit* Seems like it would make sense in the context of this PR?

Fyi, a "raw transaction" in ethereum is a signed transaction encoded as hex. So to me something *raw* could easily mean both signed or unsigned.

stbrody: What about `BasicCommit`?

smrz2001: Want me to add this discussion to the renaming ticket I just opened (#1545)? Can mark it `maintenance`.

packages/core/src/conflict-resolution.ts
smrz2001: Also added this one to the maintenance ticket so I don't forget.

PR #1529, opened 2021-07-09
feat: Check signature of a lone genesis

ukstv:
There can be a case, when a stream has only genesis commit. Its signature has to be verified. On the other hand, a stream can has more commits, which can provide timestamp information important for signature verification. A Ceramic node has to verify only lone-genesis signature, while postponing verification for multi-commit streams until all the commits are available

This has to be the case, until proper full-log state reconstruction is available.
---

v2 approach after the global prettier-ification.

Replaces https://github.com/ceramicnetwork/js-ceramic/pull/1518

### DIFF
--- a/packages/core/src/state-management/repository.ts
+++ b/packages/core/src/state-management/repository.ts
@@ -122,7 +122,7 @@ export class Repository {
     // Do not check for possible key revocation here, as we will do so later after loading the tip (or learning that the genesis commit *is* the current tip), when we will have timestamp information for when the genesis commit was anchored. In order to disable this check for key rotation, set a timestamp that will be considered earlier than any other timestamp so it will always appear to have happened earlier than any possible key revocation.
     const state = await handler.applyCommit(
       commit,
-      { cid: streamId.cid, timestamp: 1 },
+      { cid: streamId.cid, disableTimecheck: true },
       this.#deps.context
     )
     await this.#deps.stateValidation.validate(state, state.content)
--- a/packages/core/src/state-management/repository.ts
+++ b/packages/core/src/state-management/repository.ts
@@ -119,7 +119,7 @@ export class Repository {
     if (commit == null) {
       throw new Error(`No genesis commit found with CID ${genesisCid.toString()}`)
     }
-    // Do not check for possible key revocation here, as we will do so later after loading the tip (or learning that the genesis commit *is* the current tip), when we will have timestamp information for when the genesis commit was anchored. In order to disable this check for key rotation, set a timestamp that will be considered earlier than any other timestamp so it will always appear to have happened earlier than any possible key revocation.
+    // Do not check for possible key revocation here, as we will do so later after loading the tip (or learning that the genesis commit *is* the current tip), when we will have timestamp information for when the genesis commit was anchored.
     const state = await handler.applyCommit(
       commit,
       { cid: streamId.cid, disableTimecheck: true },
--- a/packages/core/src/state-management/repository.ts
+++ b/packages/core/src/state-management/repository.ts
@@ -119,10 +119,10 @@ export class Repository {
     if (commit == null) {
       throw new Error(`No genesis commit found with CID ${genesisCid.toString()}`)
     }
-    // Do not check timestamp here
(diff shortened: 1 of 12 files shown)


packages/core/src/state-management/repository.ts
```diff
@@ -119,10 +119,10 @@ export class Repository {
     if (commit == null) {
       throw new Error(`No genesis commit found with CID ${genesisCid.toString()}`)
     }
-    // Do not check timestamp here
+    // Do not check for possible key revocation here, as we will do so later after loading the tip (or learning that the genesis commit *is* the current tip), when we will have timestamp information for when the genesis commit was anchored. In order to disable this check for key rotation, set a timestamp that will be considered earlier than any other timestamp so it will always appear to have happened earlier than any possible key revocation.
     const state = await handler.applyCommit(
       commit,
-      { cid: streamId.cid, timestamp: null },
+      { cid: streamId.cid, timestamp: 1 },
```
oed, line 125: Will this fail now that we check that a signature has to be made after the key was added to the DID doc using the `updated` property?

ukstv: Indeed. It will fail. Added a commit that passes `disableTimecheck` flag.

oed (approved):
lgtm, would be good for @stbrody to give this a quick eye

packages/core/src/state-management/repository.ts
```diff
@@ -119,10 +119,10 @@ export class Repository {
     if (commit == null) {
       throw new Error(`No genesis commit found with CID ${genesisCid.toString()}`)
     }
-    // Do not check timestamp here
+    // Do not check for possible key revocation here, as we will do so later after loading the tip (or learning that the genesis commit *is* the current tip), when we will have timestamp information for when the genesis commit was anchored. In order to disable this check for key rotation, set a timestamp that will be considered earlier than any other timestamp so it will always appear to have happened earlier than any possible key revocation.
```
stbrody, line 122:
```suggestion
    // Do not check for possible key revocation here, as we will do so later after loading the tip (or learning that the genesis commit *is* the current tip), when we will have timestamp information for when the genesis commit was anchored.
```

stbrody: Latter part isn't accurate anymore now that we have the `disableTimecheck` flag

packages/core/src/state-management/state-manager.ts
```diff
@@ -83,11 +83,21 @@ export class StateManager {
...
+      this.syncedPinnedStreams.add(state$.id.toString())
+    }
+  }
+
+  /**
+   * If it is a lone genesis, verify the signature.
+   * @param state$
+   */
+  async verifyLoneGenesis(state$: RunningState): Promise<RunningState> {
+    if (state$.value.log.length > 1) {
```
stbrody, line 96: what if when the stream is loaded it has multiple signed commits but still no anchor commits? Will we ever verify the genesis?

EDIT: Nevermind, I see now how this is handled in `ConflictResolution.applyLogToState`.

Phew, all of this is so subtle 😬

stbrody (approved):
This change "LGTM" in that I believe it is probably correct and functional right at this moment.  But this change makes me very nervous as it makes the code very subtle and tricky to get right.  I'm pretty nervous about how easy this would be to break.  I seriously think we should consider a bigger refactor to change the way streams are loaded so that we don't load the genesis commit in a separate special case, but rather as part of the standard path of fetching the log from the tip.  I think if we did that we could unify the control flow here for how the timestamp check is performed, and maybe could even get rid of the `disableTimecheck` flag entirely.

IMHO this would be worth doing now rather than waiting and leaving this to potentially blow up down the road.

I'm going to go ahead and approve the PR, as I think in any case it'd be good to get this in and then get the integration test passing and testing all the cases here, which will be valuable even if we change the underlying implementation, but I think we should seriously consider spending more time here to clean this up before moving on to other things and forgetting about it.

oed:
> IMHO this would be worth doing now rather than waiting and leaving this to potentially blow up down the road.

I rather we track it in our internal Tech debt system and deal with it at some other point (maybe bundled with other refactoring).

PR #1531, opened 2021-07-09
feat(core): Cache IPFS commit data

smrz2001:


packages/core/src/__tests__/ceramic-api.test.ts
```diff
@@ -238,7 +238,7 @@ describe('Ceramic API', () => {
         })
       }
 
-      expect(logRecords).toEqual(expected)
+      expect(JSON.stringify(logRecords)).toEqual(JSON.stringify(expected))
```
oed, line 241: Why is this change needed?

packages/core/src/dispatcher.ts
```diff
@@ -4,6 +4,7 @@ import { StreamUtils, IpfsApi, UnreachableCaseError } from '@ceramicnetwork/comm
 import StreamID from '@ceramicnetwork/streamid'
 import { DiagnosticsLogger, ServiceLogger } from '@ceramicnetwork/common'
 import { Repository } from './state-management/repository'
+import { LruCache } from './state-management/lru-cache'
```
oed, line 7: We already use a library for an lruCache.

packages/core/src/__tests__/ceramic-api.test.ts
smrz2001: The test kept failing on my machine and IntelliJ identified that as missing JSONification since the serialized objects were essentially the same. I'll take this change out to see if it works and if it was just Windows being Windows.

packages/core/src/dispatcher.ts
smrz2001: 😭

smrz2001: I'll use the existing one.

packages/core/src/__tests__/ceramic-api.test.ts
smrz2001: Could be because of cached records but this is what I'm seeing on my machine (even in WSL):
```
 FAIL  src/__tests__/ceramic-api.test.ts (76.816 s)
  ● Ceramic API › API › can list log records

    expect(received).toEqual(expected) // deep equality

    Compared values serialize to the same structure.
    Printing internal object structure without calling `toJSON` instead.

    - Expected  - 0
    + Received  + 1

    @@ -41,10 +41,11 @@
                  62,
                  114,
                  84,
                ],
                "version": 1,
    +           Symbol(@ipld/js-cid/CID): true,
              },
              "payload": "AXESICWDwB7sbj3IWGh3QnQLkoFxfjPhWa-m-AQc34wPPnJU",
              "signatures": Array [
                Object {
                  "protected": "eyJhbGciOiJFZERTQSIsImtpZCI6ImRpZDprZXk6ejZNa2pLZUg4U2dWQVlDdlRCb3l4eDd1UkpGR00yYTlIVWVGd2ZKZmQ2Y3R1QTNYI3o2TWtqS2VIOFNnVkFZQ3ZUQm95eHg3dVJKRkdNMmE5SFVlRndmSmZkNmN0dUEzWCJ9",

      239 |       }
      240 |
    > 241 |       expect(logRecords).toEqual(expected)
          |                          ^
      242 |     })
      243 |
      244 |     it('can store commit if the size is lesser than the maximum size ~256KB', async () => {

      at Object.<anonymous> (src/__tests__/ceramic-api.test.ts:241:26)
```

smrz2001: Yep, seeing the same failure in CircleCI after reverting this change and also switching to the existing LRU cache library.

packages/core/src/dispatcher.ts
```diff
@@ -14,10 +14,12 @@ import {
 import { Pubsub } from './pubsub/pubsub'
 import { Subscription } from 'rxjs'
 import { MessageBus } from './pubsub/message-bus'
+import { LRUMap } from 'lru_map'
 
 const IPFS_GET_TIMEOUT = 60000 // 1 minute
 const IPFS_MAX_RECORD_SIZE = 256000 // 256 KB
 const IPFS_RESUBSCRIBE_INTERVAL_DELAY = 1000 * 15 // 15 sec
+const IPFS_CACHE_SIZE = 1000000 // 1 million records
```
stbrody, line 22: This seems pretty large.  IPFS_MAX_RECORD_SIZE is 256kb, so 1 million records could be up to 256GB just for this cache.  I'd probably start this at like 1000 records, that should be enough for us to be able to load a small handful of streams concurrently and keep all their commits in the cache for normal-case mid-sized streams.  Loading lots of really large streams concurrently risks exhausting the cache, but then we're still no worse off then we are today.

stbrody (commented):
I would change the commit message.  You took the title of the ticket as the title of the PR, but what you're implementing here is pretty different than what that ticket describes.  I'd make the commit message describe more directly what you're implementing here: "feat(core): Cache IPFS commit data"

packages/core/src/dispatcher.ts
smrz2001: Whoops! For some reason I thought the max would end up being 256MB 😬

1000 records sounds reasonable.

smrz2001:
> I would change the commit message. You took the title of the ticket as the title of the PR, but what you're implementing here is pretty different than what that ticket describes. I'd make the commit message describe more directly what you're implementing here: "feat(core): Cache IPFS commit data"

Ok, yes, makes sense 👍🏼

oed:
@smrz2001 To add to what spencer said about commit msg: The release notes and semantic versions get autogenerated from our commit msgs if formatted properly. `feat(core): Cache IPFS commit data` would bump the minor version as well as add an entry *feat(core): Cache IPFS commit data* to the release notes of the core package. If you do `chore(core): blabla` it adds a release note and bumps the patch version.

smrz2001:
> @smrz2001 To add to what spencer said about commit msg: The release notes and semantic versions get autogenerated from our commit msgs if formatted properly. `feat(core): Cache IPFS commit data` would bump the minor version as well as add an entry _feat(core): Cache IPFS commit data_ to the release notes of the core package. If you do `chore(core): blabla` it adds a release note and bumps the patch version.

Ahh cool, makes sense, ty @oed

packages/core/src/__tests__/ceramic-api.test.ts
stbrody: We've seen issues like this before. I think it might have to do with mismatched versions of the CID library.  @ukstv may be able to provide some insight and guidance here

stbrody: does this test fail on your machine against a clean build of the `develop` branch?

packages/core/src/__tests__/dispatcher.test.ts
```diff
@@ -35,6 +35,16 @@ const ipfs = {
   id: async () => ({ id: 'ipfsid' }),
 }
 
+const recordCache = {
+  get: jest.fn(async (cid: CID, record: any) => {
+    // CID not found in cache
+    if (cid === FAKE_CID2) return null
```
stbrody, line 41: you need to use `.equals` to compare CIDs for equality.  `==` will use object equality - i.e. checking if you the pointers are to the same CID instance in memory.

packages/core/src/ceramic.ts
```diff
@@ -422,12 +425,14 @@ class Ceramic implements CeramicApi {
     const ipfsTopology = new IpfsTopology(ipfs, networkOptions.name, logger)
     const pinStoreFactory = new PinStoreFactory(ipfs, pinStoreOptions)
     const repository = new Repository(streamCacheLimit, concurrentRequestsLimit, logger)
+    const recordCache = new LRUMap<CID, any>(IPFS_CACHE_SIZE)
```
stbrody, line 428: nitpick, but fwiw I think it'd be fine for the record cache to be considered an implementation detail of the Dispatcher, and thus constructed within it rather than passed in.  Not a big deal either way though

packages/core/src/dispatcher.ts
```diff
@@ -37,18 +38,21 @@ function messageTypeToString(type: MsgType): string {
  */
 export class Dispatcher {
   readonly messageBus: MessageBus
+  readonly recordCache: LRUMap<CID, any>
```
stbrody, line 41: I think this cache might need to use strings as the key rather than CID instances, otherwise I think it might fail to detect two CID instances of the same CID contents as equivalent (though there may be a way to provide a custom equality checker to LRUMap?  I'm not sure about that though).

Should be fairly straightforward to do a simple test to confirm though

packages/core/src/__tests__/dispatcher.test.ts
```diff
@@ -35,6 +35,16 @@ const ipfs = {
   id: async () => ({ id: 'ipfsid' }),
 }
 
+const recordCache = {
```
stbrody, line 38: ah, I see, this is why you were passing in the recordCache rather than constructing it within the dispatcher.  For one thing, you can actually modify the contents of an object out from under it in javascript, so you could still have provided a mock record cache to the dispatcher even while having the dispatcher construct it itself.

But also, I don't understand why you're providing a fake recordCache at all?  It doesn't do any networking or I/O, so it shouldn't be hard to use the real thing, and mocking this out just weakens the value of the test IMO.  For instance, I'm wondering if you used the real cache if the tests would have caught the issue I mentioned about CID object equality not working in the cache (or at least, it *could* repro the issue if the test used multiple CID instances for the same CID)

packages/core/src/__tests__/ceramic-api.test.ts
smrz2001: No, it works otherwise and only fails with the cache. You might have hit it on the nose with the CID equality comment @stbrody. Let me either switch to strings or look at a custom comparator and try again.

packages/core/src/__tests__/dispatcher.test.ts
smrz2001: The modified test and the new test were mainly to test the cache hit and miss code paths. The problem with using the real thing is that I won't be able to check for a cache miss and force a lookup in IPFS 😕
 
I guess I could look for a CID I haven't stored yet and assume both calls will be performed but will return null? 

I was also going to add one more test to check whether eviction of the oldest entry caused IPFS to lookup the entry but that's probably unnecessary since that would just be testing `LRUMap`.

smrz2001: Actually, I won't be able to check the two code paths without a mocked cache.

stbrody: hmm okay, I don't totally follow but if you think it's important to test with a mocked cache that's fine by me.  Maybe we could *also* have a small test that uses the real thing?  That could also make sure that 2 different CID instances for the same data are handled properly by the cache

packages/core/src/__tests__/ceramic-api.test.ts
smrz2001: Ok, I figured it out. I changed code in `loadStreamCommits` in `ceramic.ts` like so:

```
-         const record = (await this.ipfs.dag.get(cid, { timeout: IPFS_GET_TIMEOUT })).value
+         const record = await this.dispatcher.retrieveCommit(cid)
```

I did this because the dispatcher is available and there's no reason not to use the cache. Once I updated the test code to also use `Dispatcher.retrieveCommit()` instead if `IpfsApi.dag` to obtain `expected`.

In other words, it wasn't the introduction of the cache causing this issue but a mismatched comparison.

packages/core/src/dispatcher.ts
smrz2001: Since this also reduces the size of the Map, I just went ahead and changed the key to `string`.

packages/core/src/dispatcher.ts
```diff
@@ -49,6 +52,7 @@ export class Dispatcher {
     const pubsub = new Pubsub(_ipfs, topic, IPFS_RESUBSCRIBE_INTERVAL_DELAY, _pubsubLogger, _logger)
     this.messageBus = new MessageBus(pubsub)
     this.messageBus.subscribe(this.handleMessage.bind(this))
+    this.recordCache = new LRUMap<string, any>(IPFS_CACHE_SIZE)
```
oed, line 55: I guess the name "record" here is from when commits were called records. I would name this something like `objectCache` or `dagNodeCache`.

packages/core/src/dispatcher.ts
```diff
@@ -95,8 +99,15 @@ export class Dispatcher {
...
     try {
       const asCid = typeof cid === 'string' ? new CID(cid) : cid
+
+      // Lookup CID in cache before looking it up IPFS
+      const cachedRec = await this.recordCache.get(asCid.toString())
+      if (cachedRec) return cloneDeep(cachedRec)
+
+      // Now lookup CID in IPFS and also store it in the cache
       const record = await this._ipfs.dag.get(asCid, { timeout: IPFS_GET_TIMEOUT })
       await this._restrictRecordSize(cid)
```
oed, line 109: This method should really be called `_restrictCommitSize`.

packages/core/src/__tests__/dispatcher.test.ts
smrz2001: Took a bit to wrap my head around your suggestion. I'm now testing with just the real cache and able to test both the cache hit/miss legs like I wanted.

packages/core/src/dispatcher.ts
smrz2001: And you were absolutely right here - `LRUMap<CID, any>` with an entry for a CID failed to look that entry up with a different instance of the same CID.

oed (approved):
lgtm

packages/core/src/dispatcher.ts
stbrody: I've been calling objects returned from the IPFS dag "ipfs records" - is that not proper terminology?

stbrody (commented):
LGTM. This is very nice, I'm excited for the benefit this will provide

PR #1539, opened 2021-07-14
chore: Make tests run again

ukstv:


oed (commented):
Not really sure why this would make the test run reliably again, but ok

ukstv:
@oed I have read some time ago a blog post that accused Jest with being unreliable for single big test files. I guess, this is it plus a fact that `ipfs.stop` does not really clean up all the resources when it returns the control flow.

PR #1542, opened 2021-07-15
feat: Pass issuer to verifyJWS

ukstv:
DID.verifyJWS accepts issuer parameter now. TileDocumentHandler does not have to verify kid-controller correspondence by itself.


packages/stream-tile-handler/src/tile-document-handler.ts
```diff
@@ -193,21 +193,15 @@ export class TileDocumentHandler implements StreamHandler<TileDocument> {
...
-      result = await context.did.verifyJWS(commit, {
+      const verification = await context.did.verifyJWS(commit, {
         atTime: meta.timestamp,
+        issuer: did,
         disableTimecheck: meta.disableTimecheck,
       })
+      return verification.didResolutionResult
     } catch (e) {
       throw new Error('Invalid signature for signed commit. ' + e)
     }
```
oed, line 205: Why is this wrapped in a try catch and an error is rethrown? Seems better to just let the error from *verifyJWS* propagate?

packages/stream-tile-handler/src/tile-document-handler.ts
```diff
@@ -193,21 +193,15 @@ export class TileDocumentHandler implements StreamHandler<TileDocument> {
...
   ): Promise<DIDResolutionResult> {
-    let result
     try {
-      result = await context.did.verifyJWS(commit, {
+      const verification = await context.did.verifyJWS(commit, {
         atTime: meta.timestamp,
+        issuer: did,
         disableTimecheck: meta.disableTimecheck,
       })
+      return verification.didResolutionResult
```
oed, line 202: I guess there's not really any reason to return anything from this method?

oed (commented):
makes sense

packages/stream-tile-handler/src/tile-document-handler.ts
ukstv: Was here before. Decided not to change this.

packages/stream-tile-handler/src/tile-document-handler.ts
ukstv: Was here before. Decided not to change this.

ukstv: Changed.

packages/stream-tile-handler/src/tile-document-handler.ts
ukstv: Changed.

oed (approved):
lgtm, why is this still a draft?

ukstv:
> lgtm, why is this still a draft?

Discovered uint8arrays rug pull, and made it a draft to not confuse, until I am sure tests pass all right with jest resolver fix.

ukstv:
Going to undraft it when tests are green.

ukstv:
Undrafted. All green.

PR #1548, opened 2021-07-15
feat(core): Sync Streams with cache before returning from multiQuery

stbrody:
This ensures returned Streams have considered tips learned about from CommitIDs in the same batch

This PR was inspired by this comment: https://github.com/ceramicnetwork/ceramic-anchor-service/pull/473#discussion_r667321534


packages/core/src/ceramic.ts
```diff
@@ -662,15 +662,27 @@ class Ceramic implements CeramicApi {
...
+    const results = queryResults.reduce((acc, res) => ({ ...acc, ...res }), {})
+    // Before returning the results, sync each Stream to its current state in the cache.  This is
+    // necessary to handle the case where there are two MultiQueries in this batch for the same stream,
+    // one on a specific CommitID and one on a base StreamID, and loading the CommitID tells this
+    // node about a tip it had not previously heard about that turns out to be the best current tip.
+    // This ensures the returned Streams always are as up-to-date as possible, and is behavior that
+    // the anchor service relies upon.
+    await Promise.all(Object.values(results).map((stream) => {
+      // We swallow errors because if any of the queries used CommitIDs or the `atTime` arg calling
+      // sync() on them will throw.
```
oed, line 682: Don't we have a way to check if a stream is read only? I thought so.

oed: If not it would make sense to add a `isReadOnly` property!

oed (commented):
Looks good, one suggestion.

packages/core/src/ceramic.ts
stbrody: yep, that's definitely better, good call

packages/common/src/stream.ts
```diff
@@ -228,7 +228,12 @@ export abstract class Stream extends Observable<StreamState> implements StreamSt
    * Makes this stream read-only. After this has been called any future attempts to call
    * mutation methods on the instance will throw.
    */
-  abstract makeReadOnly()
+  abstract makeReadOnly(): void
+
+  /**
+   * Returns true if 'makeReadOnly()' has been called previously.
+   */
+  abstract isReadOnly(): boolean
```
oed, line 236: Why is this a method vs a property with a getter?

oed (approved):
lgtm. Added a question about the api, but I'm fine either way.

packages/common/src/stream.ts
stbrody: honestly no reason. Can change

PR #1555, opened 2021-07-18
feat(tezos-blockchain-link): initial support for linking and validating DIDs to Tezos wallets

Geo25rey:
- implements the `AuthProvider` interface for Tezos
- adds a `BlockchainHandler` for Tezos
- the Tezos `LinkProof` validator uses a RPC client to try to resolve the public key associated with the  wallet address (AKA the public key hash)
- relevant tests are added, I can add more if needed
- the `AuthProvider` implementation uses the `WalletProvider` and `Signer` interfaces from [`@taquito/*` packages](https://tezostaquito.io/typedoc/modules.html)


packages/blockchain-utils-linking/package.json
```diff
@@ -26,6 +26,8 @@
   },
   "dependencies": {
     "@stablelib/sha256": "^1.0.0",
+    "@taquito/taquito": "^9.2.0",
+    "@taquito/utils": "^9.2.0",
```
oed, line 30: Please come up with a way to implement this without these packages. We are trying to keep the `blockchain-utils-linking` package small so that it can be easily bundled in browsers!

https://bundlephobia.com/package/@taquito/taquito@9.2.0 is 385.9 kB 
https://bundlephobia.com/package/@taquito/utils@9.2.0 is 69.5 kB

packages/blockchain-utils-linking/src/tezos.ts
```diff
@@ -0,0 +1,117 @@
...
+ * message with the active account's address.
+ *
+ * @param {TezosProvider} provider - the provider to use signing the link proof
+ * @param {string} address - the address to sign the link proof with
+ * @param {string} chainRef - the chain reference to link to
+ */
+export class TezosAuthProvider implements AuthProvider {
+  readonly isAuthProvider = true
+
+  constructor(private readonly provider: TezosProvider, private readonly chainRef: string) {}
```
oed, line 62: Maybe instead of taking a `TezosProvider` here you can accept an already initialized `TezosToolkit` instance?

packages/blockchain-utils-linking/src/tezos.ts
```diff
@@ -0,0 +1,117 @@
+import { AccountID } from 'caip'
+import { AuthProvider } from './auth-provider'
+import { getConsentMessage, LinkProof } from './util'
+import { hash } from '@stablelib/sha256'
+import * as uint8arrays from 'uint8arrays'
+import { SetProviderOptions, TezosToolkit } from '@taquito/taquito'
+import { char2Bytes } from '@taquito/utils'
```
oed, line 7: Please use `uint8arrays` to convert utf-8 to a bytestring.

packages/blockchain-utils-validation/src/blockchains/tezos.ts
```diff
@@ -0,0 +1,64 @@
+import { AccountID } from 'caip'
+import { RemoteSigner } from '@taquito/remote-signer'
+import { char2Bytes } from '@taquito/utils'
```
oed, line 3: `uint8arrays` here as well

packages/blockchain-utils-validation/src/blockchains/tezos.ts
```diff
@@ -0,0 +1,64 @@
...
+import { BlockchainHandler } from '../blockchain-handler'
+import { LinkProof } from '@ceramicnetwork/blockchain-utils-linking'
+import axios from 'axios'
```
oed, line 6: Please use `cross-fetch` as that's what we are using everywhere else.

packages/blockchain-utils-validation/src/blockchains/tezos.ts
```diff
@@ -0,0 +1,64 @@
...
+      return result
+    } else {
+      throw PUBLIC_KEY_NOT_PUBLISHED_ERROR
+    }
+  }
+}
+
+// can only validate proofs with wallet addresses (public key hashes) that have:
+// - at least one transaction on the blockchain
+// - their public key published to the blockchain
```
oed, line 42: Why is this needed? Seems wrong. With all other blockchains you can verify signatures on just the address (public key hash) because we're using recoverable signatures. Is Tezos really not doing the same (I'm pretty sure it has to!)?

packages/blockchain-utils-validation/src/blockchains/tezos.ts
```diff
@@ -0,0 +1,64 @@
+import { AccountID } from 'caip'
+import { RemoteSigner } from '@taquito/remote-signer'
```
oed, line 2: Wow, this package pulls in `libsodium`. That's stupid. What elliptic curve is Tezos using? I'm sure there's a better way to verify the signatures!

oed (commented):
Not super happy about the current state of this. I think we can improve it a lot! 

@ukstv can you provide feedback on the `npm` update done here?

packages/blockchain-utils-linking/package.json
Geo25rey: `@taquito/taquito` can probably just be a dev dependency since it is really only used for typing.

`@taqutio/utils` is required tho since `char2Bytes` is custom.

packages/blockchain-utils-linking/src/tezos.ts
Geo25rey: Sounds good. I was just trying to match how the `EthereumAuthProvider` works with its use of a raw provider. Sadly, Tezos doesn't have a great standard for this.

packages/blockchain-utils-linking/src/tezos.ts
Geo25rey: Cant't. `char2Bytes` does a custom `string` to `string` conversion.

packages/blockchain-utils-validation/src/blockchains/tezos.ts
Geo25rey: Got it.

packages/blockchain-utils-validation/src/blockchains/tezos.ts
Geo25rey: No other blockchain hashes the wallet's public key to generate the wallet address. Other blockchains just encode the public key as a string. So unless SHA2-256 is broken, the public key has to be published somewhere.

packages/blockchain-utils-linking/src/tezos.ts
oed: It's really simple: https://github.com/ecadlabs/taquito/blob/master/packages/taquito-utils/src/taquito-utils.ts#L249
Please use `uint8arrays`!

packages/blockchain-utils-linking/package.json
oed: Ok, nice

packages/blockchain-utils-validation/src/blockchains/tezos.ts
oed: > No other blockchain hashes the wallet's public key to generate the wallet address.

Yes they all do 😄

oed: See for example: https://www.quora.com/What-encoding-is-used-for-Ethereum-addresses?share=1 for ethereum

Geo25rey: > > No other blockchain hashes the wallet's public key to generate the wallet address.
> 
> Yes they all do 😄 

Whoops

Geo25rey: > See for example: https://www.quora.com/What-encoding-is-used-for-Ethereum-addresses?share=1 for ethereum

This is mapping from public key to hash tho. We need to map from hash to public key, and that still requires an on chain lookup.

> You can if and only if a transaction has been sent from the account. When you send a tx, you sign the transaction and it includes these v r and s values. You parse these from the signed tx and then pass these v r and s values and the hash of the transaction back into a function and it'll spit out the public key. This is actually how you get the from address of a transaction.
Source: [Ethereum StackExchange](https://ethereum.stackexchange.com/questions/13778/get-public-key-of-any-ethereum-account)

Tezos just has the public key directly in the transaction since it supports multiple types of keys.

packages/blockchain-utils-linking/package.json
Geo25rey: The actual compiled JS should only be a few 100 bytes due to [tree shaking](https://webpack.js.org/guides/tree-shaking/), but configuring this is outside the scope of this PR.

oed: Well tree shaking doesn't always work and if any developer that uses the ceramic http-client misconfigures their tree shaking they will end up blaming us. Please you create a simple type in your code rather than importing this library!

packages/blockchain-utils-validation/src/blockchains/tezos.ts
Geo25rey: They use many including p256, secp256k1, and ed25519.

packages/blockchain-utils-validation/src/blockchains/tezos.ts
oed: > and that still requires an on chain lookup

No this is not the case. If you have the message and the signature, you can recover the public key, then hash it to see if it matches the address.

packages/blockchain-utils-validation/src/blockchains/tezos.ts
oed: So it looks like this lib uses `elliptic` library for p256 and secp256k1 which is fine: https://github.com/ecadlabs/taquito/blob/eb6afed/packages/taquito-remote-signer/src/taquito-remote-signer.ts#L199

However the use of libsodium is not acceptable due to it's huge size: https://github.com/ecadlabs/taquito/blob/eb6afed/packages/taquito-remote-signer/src/taquito-remote-signer.ts#L188 when there exists much tinier libs such as this one: https://www.stablelib.com/modules/_ed25519_ed25519_.html#verify

Unfortunately I can not accept this PR as long as it's using libsodium. Maybe a PR to fix it in the taquito library is warranted?

oed (changes requested):

ukstv:
@Geo25rey 

> Note: I updated the package-lock.json with npm@7.19.0 and there were some conflicts so I replaced it. Do you use an older version of npm?

npm v6 and node v14 are still used here. Please, install it via [nvm](https://github.com/nvm-sh/nvm) or [fnm](https://github.com/Schniz/fnm).

packages/blockchain-utils-linking/src/tezos.ts
ukstv: `uint8arrays.toString(uint8arrays.fromString(input, 'utf8'), 'base16')` would work @Geo25rey

packages/blockchain-utils-linking/src/tezos.ts
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ * initializes the TezosToolkit with the given {TezosProvider} and returns the {TezosToolkit}
+ *
+ * @param {TezosProvider} provider - the provider to use for initializing the TezosToolkit
+ * @returns {TezosToolkit} the initialized TezosToolkit
+ */
+function getTezosToolkit(provider: TezosProvider) {
+  // initialize the Tezos RPC client
+  const Tezos = new TezosToolkit('https://mainnet-tezos.giganode.io')
```
ukstv, line 19: @Geo25rey Errm, are you sure it is a good idea to have depend on this HTTP endpoint??

ukstv:
I am also deeply worried about the fact one has to go unchain, to query a public key. This is so wrong. I hope Tezos guys are smarter than that, and there is a way to verify a signature without going onchain.

packages/blockchain-utils-linking/src/tezos.ts
Geo25rey: I want to confirm it produces the same output then I will fix. I will get to it soon.

packages/blockchain-utils-linking/src/tezos.ts
Geo25rey: It's a required argument. I don't think it's used in this context tho. This should be getting removed anyway. See https://github.com/ceramicnetwork/js-ceramic/pull/1555#discussion_r671802423

Geo25rey:
> I am also deeply worried about the fact one has to go unchain, to query a public key. This is so wrong. I hope Tezos guys are smarter than that, and there is a way to verify a signature without going onchain.

We could save it as part of the `LinkProof` since it is not encoded in the signature. Doing the query is pretty fast however since it is querying an indexed block explorer not the chain itself.

ukstv:
> > I am also deeply worried about the fact one has to go unchain, to query a public key. This is so wrong. I hope Tezos guys are smarter than that, and there is a way to verify a signature without going onchain.
> 
> We could save it as part of the `LinkProof` since it is not encoded in the signature. Doing the query is pretty fast however since it is querying an indexed block explorer not the chain itself.

Better save it as part of LinkProof somehow.

packages/blockchain-utils-linking/package.json
vaultec81: > @taquito/taquito can probably just be a dev dependency since it is really only used for typing.

Could we just copy the types file instead of needing to import the entire thing? Fairly straight forward to do

packages/blockchain-utils-linking/src/tezos.ts
Geo25rey: Works the same!

packages/blockchain-utils-validation/src/blockchains/tezos.ts
Geo25rey: Isn't size of `blockchain-utils-validation` irrelevant? It's only used in `core` which doesn't really depend on size.

Geo25rey: @oed Does this help? https://github.com/ecadlabs/taquito/pull/978

oed: > @oed Does this help? ecadlabs/taquito#978
 Yes, that PR looks great!

oed: > Isn't size of blockchain-utils-validation irrelevant? It's only used in core which doesn't really depend on size.

It's not irrelevant. Ideally `core` would also be small enough to run in a browser, it's a tradeoff though. Don't want to just carelessly add more deps however.

packages/blockchain-utils-linking/src/__tests__/tezos.test.ts
```diff
@@ -84,4 +87,12 @@ describe('Blockchain: Tezos', () => {
...
     })
   })
+
+  describe('char2Bytes', () => {
+    test('should match the @taquito/utils char2Bytes() implementation', () => {
+      // get a random string of length 32
+      const s = 'some random string. this should be fine.'
+      expect(char2Bytes(s)).toBe(utils.char2Bytes(s))
+    })
+  })
```
oed, line 97: Not sure why you would add this test here. It's a really straight forward conversion.

packages/blockchain-utils-linking/package.json
```diff
@@ -42,6 +40,8 @@
     "@polkadot/util-crypto": "^6.2.1",
     "@smontero/eosio-local-provider": "^0.0.3",
     "@taquito/signer": "^9.2.0",
+    "@taquito/taquito": "^9.2.0",
+    "@taquito/utils": "^9.2.0",
```
oed, line 44: This dep can be removed.

oed (commented):
Looking forward to seeing your PR to the tezos package merged and released :)

packages/blockchain-utils-validation/src/blockchains/tezos.ts
Geo25rey: A step that should be taken outside of this PR would probably be removing any dependencies that use WebAssembly, since they are rather large for no noticable performance gain. That is why `sodium` is so large.

packages/blockchain-utils-linking/src/__tests__/tezos.test.ts
Geo25rey: If `@taquito/utils` ever changes their implementation, it would be nice to know rather than having a weird bug that's hard to debug.

Geo25rey:
@oed just got the stablelib version of taquito-remote-signer

oed (approved):
Alright, this lgtm. Thanks @Geo25rey. Want a confirmation from @ukstv as well!

packages/blockchain-utils-linking/src/__tests__/tezos.test.ts
ukstv: @Geo25rey It is not `@taquito/utils` that is important, it is `TezosProvider.signer.sign` that one should care about. As long as the latter accepts hex, `@taquito/utils` can be removed. And, if by some chance `TezosProvider.signer.sign` stops accepting hex, the tests will warn us. So, please, remove it from the dependencies list. It brings nothing useful here.

packages/blockchain-utils-linking/src/tezos.ts
```diff
@@ -0,0 +1,119 @@
...
+
+/**
+ * get the active address from the {TezosProvider}
+ *
+ * @param {TezosProvider} provider - the provider to use for getting the active account
+ * @returns {Promise<string>} - a promise that resolves to the active account's address
+ */
+async function getActiveAddress(provider: TezosProvider): Promise<string> {
+  const Tezos = provider
+  const activeAddress = await Tezos.wallet.pkh({ forceRefetch: true })
```
ukstv, line 48:
```suggestion
return Tezos.wallet.pkh({ forceRefetch: true })
```

packages/blockchain-utils-validation/src/blockchains/__tests__/tezos.test.ts
```diff
@@ -0,0 +1,183 @@
+import { TezosAuthProvider, TezosProvider } from '../../../../blockchain-utils-linking/src/tezos' //'@ceramicnetwork/blockchain-utils-linking'
```
ukstv, line 1: @Geo25rey What is that?!?

Geo25rey: For some reason tests, fail to resolve local changes to packages, so this is a work around

oed: This will break, plz change back and run `npm i` and `npm run build` top level!

packages/blockchain-utils-linking/src/tezos.ts
```diff
@@ -0,0 +1,119 @@
...
+  return activeAddress
```
Geo25rey, line 49:
```suggestion
```

packages/blockchain-utils-linking/src/__tests__/tezos.test.ts
Geo25rey: That makes sense to me. I am on it.

packages/blockchain-utils-linking/src/__tests__/tezos.test.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  describe('char2Bytes', () => {
+    test('should match the @taquito/utils char2Bytes() implementation', () => {
+      // get a random string of length 32
+      const s = 'some random string. this should be fine.'
+      expect(char2Bytes(s)).toBe(utils.char2Bytes(s))
+    })
+  })
```
Geo25rey, line 97:
```suggestion
```

packages/blockchain-utils-linking/src/__tests__/tezos.test.ts
```diff
@@ -0,0 +1,98 @@
+import { char2Bytes, TezosAuthProvider, TezosProvider } from '../tezos'
+import { InMemorySigner } from '@taquito/signer'
+import {
+  WalletDelegateParams,
+  WalletOriginateParams,
+  WalletTransferParams,
+  WalletProvider,
+  TezosToolkit,
+} from '@taquito/taquito'
+import utils from '@taquito/utils'
```
Geo25rey, line 10:
```suggestion
```

packages/blockchain-utils-linking/src/__tests__/tezos.test.ts
```diff
@@ -0,0 +1,98 @@
+import { char2Bytes, TezosAuthProvider, TezosProvider } from '../tezos'
```
Geo25rey, line 1:
```suggestion
import { TezosAuthProvider, TezosProvider } from '../tezos'
```

packages/blockchain-utils-linking/package.json
```diff
@@ -39,6 +39,9 @@
     "@polkadot/util": "^6.2.1",
     "@polkadot/util-crypto": "^6.2.1",
     "@smontero/eosio-local-provider": "^0.0.3",
+    "@taquito/signer": "^9.2.0",
+    "@taquito/taquito": "^9.2.0",
+    "@taquito/utils": "^9.2.0",
```
Geo25rey, line 44:
```suggestion
```

packages/blockchain-utils-linking/src/tezos.ts
```diff
@@ -0,0 +1,119 @@
...
+
+// should be the same as @taquito/utils char2Bytes()
+// exported for testing to ensure above
+/**
+ * converts a utf8 string to a hex string
+ *
+ * @param {string} input - the string to convert
+ * @returns {string} the converted hex string
+ */
+export function char2Bytes(input: string): string {
```
Geo25rey, line 22:
```suggestion
function char2Bytes(input: string): string {
```

packages/blockchain-utils-linking/src/tezos.ts
```diff
@@ -0,0 +1,119 @@
...
+  // the below are requried for the TezosProvider interface
+  wallet: Wallet
+  signer: Signer
+}
+
+// should be the same as @taquito/utils char2Bytes()
+// exported for testing to ensure above
+/**
+ * converts a utf8 string to a hex string
+ *
```
Geo25rey, line 18:
```suggestion
 *
 * @internal
 *
```

packages/blockchain-utils-validation/src/blockchains/__tests__/tezos.test.ts
Geo25rey: Sure thing.

Geo25rey, line 1:
```suggestion
import { TezosAuthProvider, TezosProvider } from '@ceramicnetwork/blockchain-utils-linking'
```

packages/blockchain-utils-validation/src/blockchains/__tests__/tezos.test.ts
```diff
@@ -0,0 +1,183 @@
...
+import {
+  WalletDelegateParams,
+  WalletOriginateParams,
+  WalletTransferParams,
+  WalletProvider,
+  TezosToolkit,
+} from '@taquito/taquito'
+import { char2Bytes, validateLink } from '../tezos'
+import mockFetch from 'jest-fetch-mock'
+import { LinkProof } from '../../../../blockchain-utils-linking/src/util'
```
Geo25rey, line 12:
```suggestion
import { LinkProof } from '@ceramicnetwork/blockchain-utils-linking'
```

Geo25rey:
@oed @ukstv I believe that is all the requested changes addressed. I am ready for a "squash and merge"

ukstv:
Not so fast @Geo25rey. Unless @oed overrules, I do believe that going on chain to validate a vanilla signature is a wrong idea. As we discussed a week ago, please, adjust the signature format so that it includes a public key, or otherwise makes the signature verification available offline. I guess, it also makes sense to raise the issue in a wider Tezos community as well.

Apart from technical wrongness, this makes Ceramic unavailable to fresh Tezos accounts, that just created their keys. I have just created a Tezos account on Kukai. Most probably I can sign something with it. The current go-on-chain method makes the signature (and caip-10 link) invalid, because obviously it has no funds on the account.

Geo25rey:
> Not so fast @Geo25rey. Unless @oed overrules, I do believe that going on chain to validate a vanilla signature is a wrong idea. As we discussed a week ago, please, adjust the signature format so that it includes a public key, or otherwise makes the signature verification available offline. I guess, it also makes sense to raise the issue in a wider Tezos community as well.

@oed already verified that this is fine for now. https://github.com/ceramicnetwork/js-ceramic/pull/1555#discussion_r671847242

> Apart from technical wrongness, this makes Ceramic unavailable to fresh Tezos accounts, that just created their keys. I have just created a Tezos account on Kukai. Most probably I can sign something with it. The current go-on-chain method makes the signature (and caip-10 link) invalid, because obviously I have no funds on the account.

It's implied that users have at least 1 transaction on the blockchain, so it's fine.

ukstv:
@Geo25rey Ok. Is there something specific about this type of signature? Can one deduce at a later time somehow that a signature includes pub key or not? Prefix, maybe? Length? Is there a way to incorporate a distinction like this, again at a later time?

Geo25rey:
> @Geo25rey Ok. Is there something specific about this type of signature?

This is included as the signature's prefix.

> Can one deduce at a later time somehow that a signature includes pub key or not? Prefix, maybe? Length? Is there a way to incorporate a distinction like this, again at a later time?

Currently, based on what I can tell, the public key is not encoded in the signature, only the type of public key used.

ukstv:
> as the signature's prefix

good.

packages/blockchain-utils-validation/src/blockchains/__tests__/tezos.test.ts
```diff
@@ -0,0 +1,183 @@
...
+  })
+
+  describe('char2Bytes', () => {
+    test('should match the @taquito/utils char2Bytes() implementation', () => {
+      // get a random string of length 32
+      const s = 'some random string. this should be fine.'
+      expect(char2Bytes(s)).toBe(utils.char2Bytes(s))
+    })
+  })
+})
```
Geo25rey, line 183:
```suggestion
```

PR #1560, opened 2021-07-19
fix(core): Properly cache IPFS lookups with paths

stbrody:
This is a regression that was causing my CAS consensus test to fail


smrz2001 (approved):
:shipit:

PR #1561, opened 2021-07-20
feat(cli): ethr-did-resolver support for mainnet

simonas-notcat:
This PR adds support for `did:ethr` in the CLI daemon.

To create ceramic documents using `did:ethr` one can use [Veramo](https://github.com/uport-project/veramo/pull/627)

Current implementation assumes that ceramic node is being run with `--ethereum-rpc` parameter. E.x.:
```
./bin/ceramic.js daemon --network inmemory --debug --ethereum-rpc https://mainnet.infura.io/v3/5ffc47f65c4042ce847ef66a3fa70d4c
```

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -49,6 +49,7 @@
     "dag-jose": "^0.3.0",
     "did-resolver": "^3.1.0",
     "dids": "^2.4.0",
+    "ethr-did-resolver": "4.3.4",
     "express": "^4.17.1",
     "flat": "^5.0.2",
     "ipfs-http-client": "~50.1.1",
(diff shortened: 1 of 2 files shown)


oed (commented):
Nice to see a PR for `did:ethr`, thanks @simonas-notcat :)

Sidenote: noticed that the `ethr-did-resolver` package is quite large. Interestingly it pulls in the [entire ABI](https://github.com/uport-project/ethr-did-registry/blob/develop/build/contracts/EthereumDIDRegistry.json) for the Ethr contract. This might not be needed for the resolution?

As for the config, I think this is good for now. I guess it would only work on ethereum mainnet though. We are in the process of specing out what using a config file for Ceramic would looks like. @ukstv it would be nice to have method specific configs for did resovlers, something like this:

```js
{
  "resolvers": {
    "ethr": {
       networks: [{
         name: 'mainnet',
         rpcUrl: ceramicConfig.ethereumRpcUrl,
       }],
      }
    }
    "nft": ...
  }
}
```

Would like a review from @stbrody and @ukstv as well.

stbrody (approved):
lgtm

ukstv (approved):
🏇

PR #1563, opened 2021-07-21
Publish build events to webhook

v-stickykeys:
Makes a curl request to our lambda whenever a new image is built and published on docker hub. This is part of the flow to achieve automated deployments.


smrz2001 (approved):
🚀

PR #1574, opened 2021-07-21
chore(core): Add retry logic when fetching CIDs from IPFS

stbrody:


oed (commented):
lgtm

packages/core/src/dispatcher.ts
```diff
@@ -128,23 +121,49 @@ export class Dispatcher {
...
-      // Lookup CID in cache before looking it up IPFS
-      const cidAndPath = path ? asCid.toString() + path : asCid.toString()
-      const cachedDagNode = await this.dagNodeCache.get(cidAndPath)
-      if (cachedDagNode) return cloneDeep(cachedDagNode)
-
-      // Now lookup CID in IPFS and also store it in the cache
-      const dagNode = await this._ipfs.dag.get(asCid, { timeout: IPFS_GET_TIMEOUT, path })
-      await this.dagNodeCache.set(cidAndPath, dagNode.value)
-      return cloneDeep(dagNode.value)
+      return this._getFromIpfs(cid, path)
```
smrz2001, line 124: Is there a reason we don't call `_restrictCommitSize()` here? If we should, then perhaps it can be called from inside `_getFromIpfs()`.

packages/core/src/dispatcher.ts
```diff
@@ -98,17 +99,9 @@ export class Dispatcher {
...
-      const asCid = typeof cid === 'string' ? new CID(cid) : cid
-
-      // Lookup CID in cache before looking it up IPFS
-      const cachedDagNode = await this.dagNodeCache.get(asCid.toString())
-      if (cachedDagNode) return cloneDeep(cachedDagNode)
-
-      // Now lookup CID in IPFS and also store it in the cache
-      const dagNode = await this._ipfs.dag.get(asCid, { timeout: IPFS_GET_TIMEOUT })
+      const result = await this._getFromIpfs(cid)
       await this._restrictCommitSize(cid)
```
smrz2001, line 103: Can this IPFS API call also fail? If so, should it also be retried? Should we have generic retry handling that gets commonly applied across all IPFS API calls?

packages/core/src/dispatcher.ts
```diff
@@ -16,7 +16,8 @@ import { Subscription } from 'rxjs'
 import { MessageBus } from './pubsub/message-bus'
 import { LRUMap } from 'lru_map'
 
-const IPFS_GET_TIMEOUT = 60000 // 1 minute
+const IPFS_GET_RETRIES = 3
+const IPFS_GET_TIMEOUT = 30000 // 30 seconds per retry, 3 retries = 90 seconds total timeout
```
smrz2001, line 20: The timeout should probably be ~3 seconds instead of 30 seconds since this could be called during user API processing and normally a couple of seconds is considered the maximum a user should have to wait for an API call to return.

What are your thoughts?

packages/core/src/dispatcher.ts
```diff
@@ -128,23 +121,49 @@ export class Dispatcher {
...
+
+    // Now lookup CID in IPFS, with retry logic
+    // Note, in theory retries shouldn't be necessary, as just increasing the timeout should
+    // allow IPFS to use the extra time to find the CID, doing internal retries if needed.
+    // Anecdotally, however, we've seen evidence that IPFS sometimes finds CIDs on retry that it
+    // doesn't on the first attempt, even when given plenty of time to load it.
+    let dagResult = null
+    for (let retries = IPFS_GET_RETRIES - 1; retries >= 0 && dagResult == null; retries--) {
+      try {
+        dagResult = await this._ipfs.dag.get(asCid, {timeout: IPFS_GET_TIMEOUT, path})
```
smrz2001, line 150: It's generally considered good practice to use exponential backoff with jitter when retrying.

The reason for the exponential backoff is so that a struggling network/system doesn't get slammed in tight succession with retries. The reason for the jitter is to make sure that retries from different parts of the system don't all get aligned at a single point in time and cause the network/system to get slammed when the backoff kicks in.

One or both of these are optional but would be good to consider if they would be useful to add. Neither will make the logic significantly more complicated.

packages/core/src/dispatcher.ts
stbrody: I think 3 seconds would be way too short when we're thinking about navigating the ipfs DHT to do content lookups

packages/core/src/dispatcher.ts
stbrody: Yes in theory, but since if we get this far that means we've already loaded the CID into our local IPFS node, it should be extremely unlikely for this to fail

packages/core/src/dispatcher.ts
stbrody: Because this isn't used for loading commits.  That is the primary difference between `retrieveFromIPFS` and `retrieveCommit`.  The latter is specifically used to load Ceramic commits and restricts the commit size accordingly, while `retrieveFromIPFS` is for loading other random data from IPFS (primarily used for loading the anchor merkle tree currently).

packages/core/src/dispatcher.ts
stbrody: with the timeout being so high (30 seconds) I'm not too worried about hammering the IPFS node with these retries.

I'd rather keep things simple for now and add more advanced retry logic here if/when we see it being a problem

packages/core/src/dispatcher.ts
stbrody: I'd rather err on the side of keeping the timeouts longer than needed for now, rather than risk timing out operations that might have succeeded if they were given more time.

packages/core/src/dispatcher.ts
smrz2001: Ah gotcha.

smrz2001 (approved):
🚢

PR #1576, opened 2021-07-22
feat(ci): Add AWS creds to build events

smrz2001:


.github/workflows/docker-image.yml
```diff
@@ -77,7 +81,7 @@ jobs:
       -
         name: Publish build event
         run: |
-          res=$(curl -X POST ${{ env.EVENT_WEBHOOK_URL }} -H "Content-Type: application/json" -d '{ "event_type": "new_image", "client_payload": {"env_tag": "${{ env.ENV_TAG }}", "sha_tag": "${{ env.SHA_TAG }}" } }') || exit 1
+          res=$(docker run --rm curlimages/curl curl -X POST ${{ env.EVENT_WEBHOOK_URL }} -H "Content-Type: application/json" --user ${{ env.AWS_ACCESS_KEY_ID }}:${{ env.AWS_SECRET_ACCESS_KEY }} --aws-sigv4 "aws:amz:${{ env.AWS_DEFAULT_REGION }}:execute-api" -d '{ "event_type": "new_image", "client_payload": {"env_tag": "${{ env.ENV_TAG }}", "sha_tag": "${{ env.SHA_TAG }}" } }') || exit 1
```
v-stickykeys, line 84: @smrz2001 Why did we need to use curlimages here?

Also could just use secrets.<KEY> instead of setting them in env first, right?

smrz2001: True, could just have used secrets directly. Will do that in the next PR.

The new `aws-sig` option is only available in cURL 7.75.0 onwards :( The highest "official" release used in Ubuntu is 7.68.0. I could have used some other method but this was the most straightforward. I used the latest cURL Docker image instead of needing to muck around with updating cURL.

PR #1597, opened 2021-07-28
feat(core): Invalid commits don't prevent loading a stream

stbrody:
If an error is encountered while applying commits as part of loading a stream, the application process stops and the stream is returned with as much of the state as could be validly constructed

### DIFF
--- a/packages/common/src/docopts.ts
+++ b/packages/common/src/docopts.ts
@@ -47,7 +47,7 @@ export interface InternalOpts {
    * This option is used internally but is not designed to be set by user applications.
    * @private
    */
-  throwOnApplyCommitError?: boolean
+  throwOnInvalidCommit?: boolean
 }
 
 /**
--- a/packages/common/src/docopts.ts
+++ b/packages/common/src/docopts.ts
@@ -21,6 +21,9 @@ export enum SyncOptions {
   NEVER_SYNC,
 }
 
+/**
+ * Options shared between LoadOpts and CreateOpts
+ */
 interface BasicLoadOpts {
   /**
    * Controls the behavior related to syncing a stream to the most recent tip.
@@ -31,6 +34,20 @@ interface BasicLoadOpts {
    * How long to wait for a response from pubsub when syncing a stream.
    */
   syncTimeoutSeconds?: number
+}
+
+/**
+ * Options that are used internally but aren't designed to be set by end users.
+ */
+export interface InternalOpts {
+  /**
+   * If true, when loading a stream log will throw an exception if any commit fails to apply.
+   * If false, the error will be logged, but the log application operation will return as much
+   * of the log as it was able to successfully apply.
+   * This option is used internally but is not designed to be set by user applications.
+   * @private
+   */
+  throwOnApplyCommitError?: boolean
 }
 
 /**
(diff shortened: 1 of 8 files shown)


packages/common/src/docopts.ts
```diff
@@ -33,6 +36,20 @@ interface BasicLoadOpts {
...
+ */
+export interface InternalOpts {
+  /**
+   * If true, when loading a stream log will throw an exception if any commit fails to apply.
+   * If false, the error will be logged, but the log application operation will return as much
+   * of the log as it was able to successfully apply.
+   * This option is used internally but is not designed to be set by user applications.
+   * @private
+   */
+  throwOnApplyCommitError?: boolean
```
stbrody, line 50: I'm not thrilled with putting this here where it's potentially exposed to users, but the only alternative I could come up with is to add a new parameter to `CeramicAPI.applyCommit()`, which will change both the core and HTTP client api, and still be visible to users anyway just in a different place.

The best long term solution, but one that would be a much bigger refactor, would be to have an `OperationContext` object threaded through just about every function in the code base.  We could use such an object to attach additional information and options related to the current top-level operation that we are executing.  That would let us distinguish in low level code whether we're doing something on behalf of a message that was received via pubsub, via an HTTP request, via a direct call into the core API, etc.

packages/core/src/conflict-resolution.ts
```diff
@@ -260,34 +294,24 @@ export class ConflictResolution {
...
-        state = tmpState // if validation is successful
+
+    for (const entry of unappliedCommits) {
+      try {
+        state = await this.applyLogEntryToState(entry, state, handler)
+      } catch(err) {
+        // TODO(1586): include StreamID in log message
+        this.context.loggerProvider.getDiagnosticsLogger().warn(
+          `Error while applying commit ${entry.cid.toString()}: ${err}`)
+        if (opts.throwOnApplyCommitError) {
```
stbrody, line 305: I'm open to other names for this parameter.  `requireSuccessfulCommitApplication` maybe?

stbrody:
CC @smrz2001

packages/core/src/conflict-resolution.ts
ukstv: Vote for `throwOnInvalidCommit`, but this is a weak opinion. Like `throw` prefix.

ukstv (approved):
🏇

packages/core/src/conflict-resolution.ts
oed: I like @ukstv suggestion

packages/core/src/conflict-resolution.ts
```diff
@@ -242,14 +243,47 @@ export class ConflictResolution {
...
+   * If an error is encountered while applying a commit, commit application stops and the state
+   * that was built thus far is returned, unless 'opts.throwOnApplyCommitError' is true.
    */
   private async applyLogToState<T extends Stream>(
     handler: StreamHandler<T>,
     unappliedCommits: CommitData[],
-    state?: StreamState,
-    breakOnAnchor?: boolean
+    state: StreamState | null,
+    breakOnAnchor: boolean,
```
oed, line 285: Can `breakOnAnchor` be folded into opts?

oed (approved):
seems fine

packages/core/src/conflict-resolution.ts
stbrody: It could be, but since it's only used in a single place I'd rather not as moving it to `InternalOpts` exposes it to a lot more layers of the stack unnecessarily.

PR #1600, opened 2021-07-30
docs(ipfs-daemon): change dockerhub url to public

v-stickykeys:
The original URL points to the private dashboard so I swapped it for the public url that anyone can view.


ukstv (approved):

oed (commented):
Nice

PR #1609, opened 2021-07-31
feat: Allow stream controller to differ from signer

ukstv:
Necessary for did:nft and did:safe

### DIFF
--- a/packages/core/src/__tests__/tile-controller.test.ts
+++ b/packages/core/src/__tests__/tile-controller.test.ts
@@ -1,0 +1,63 @@
+import { IpfsApi } from '@ceramicnetwork/common'
+import { createIPFS } from './ipfs-util'
+import Ceramic from '../ceramic'
+import { createCeramic } from './create-ceramic'
+import { Ed25519Provider } from 'key-did-provider-ed25519'
+import * as random from '@stablelib/random'
+import { Resolver } from 'did-resolver'
+import KeyDidResolver from 'key-did-resolver'
+import { DID } from 'dids'
+import { TileDocument } from '@ceramicnetwork/stream-tile'
+
+let ipfs: IpfsApi
+let ceramic: Ceramic
+let alice: DID
+let bob: DID
+
+const resolver = new Resolver({
+  ...KeyDidResolver.getResolver(),
+})
+
+function makeKeyDid(): DID {
+  return new DID({
+    resolver: resolver,
+    provider: new Ed25519Provider(random.randomBytes(32)),
+  })
+}
+
+beforeAll(async () => {
+  ipfs = await createIPFS()
+  alice = makeKeyDid()
+  bob = makeKeyDid()
+  await alice.authenticate()
+  await bob.authenticate()
+}, 12000)
+
+afterAll(async () => {
+  await ipfs.stop()
+})
+
+beforeEach(async () => {
+  ceramic = await createCeramic(ipfs)
+})
+
+afterEach(async () => {
+  await ceramic.close()
+})
+
+describe('throw if controller is different from signer', () => {
+  test('create', async () => {
+    ceramic.did = alice
+    await expect(
+      TileDocument.create(ceramic, { foo: 'blah' }, { controllers: [bob.id] })
+    ).rejects.toThrow(/invalid_jws/)
+  })
+  test('update', async () => {
+    ceramic.did = alice
+    const tile = await TileDocument.create(ceramic, { foo: 'blah' })
+    expect(tile.controllers.length).toEqual(1)
+    expect(tile.controllers[0]).toEqual(alice.id)
+    ceramic.did = bob
+    await expect(tile.update({ foo: 'bar' })).rejects.toThrow(/invalid_jws/)
+  })
+})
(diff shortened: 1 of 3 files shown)


packages/stream-tile/src/tile-document.ts
```diff
@@ -101,16 +99,14 @@ async function _ensureAuthenticated(signer: CeramicSigner) {
  * Sign Tile commit
```
oed, line 99:
```suggestion
 * Sign a Tile commit with the currently authenticated DID.
```

oed (commented):
I think this change makes sense. However it might cause issues with the current implementations of our DID providers. I think at first we should make changes to these package to support signing when no DID is passed!

* `3id-did-provider` expects it's own DID: https://github.com/ceramicstudio/js-3id-did-provider/blob/main/src/did-provider.ts#L53
* `key-did-provider` as well: https://github.com/ceramicnetwork/key-did-provider-ed25519/blob/master/src/index.ts#L75

ukstv:
> I think this change makes sense. However it might cause issues with the current implementations of our DID providers. I think at first we should make changes to these package to support signing when no DID is passed!

@oed this is a non-issue, if dids package still populates `did` field to be current DID: https://github.com/ceramicnetwork/js-did/blob/master/src/did.ts#L187

IMO, no more changes required.

oed (approved):
Ah, nice!

packages/stream-tile/src/tile-document.ts
```diff
@@ -173,9 +169,7 @@ export class TileDocument<T = Record<string, any>> extends Stream {
       // document as there shouldn't be any existing state for this doc on the network.
       opts.syncTimeoutSeconds = 0
     }
-    const commit = genesisCommit.data
-      ? await _signDagJWS(ceramic, genesisCommit, genesisCommit.header.controllers[0])
-      : genesisCommit
+    const commit = genesisCommit.data ? await _signDagJWS(ceramic, genesisCommit) : genesisCommit
```
stbrody, line 172: What's the behavior now if the currently authenticated DID isn't the controller of this stream?  We should make sure that an exception is thrown rather than having the commit just silently fail to apply (or worse actually get applied, but I assume that somewhere else in commit application code will prevent that)

stbrody: Can you also check to confirm that there's a test for trying to author commits on streams you don't own?  I assume there must be one already but would be good to make sure

stbrody (commented):
One question about the user-visible behavior when trying to author a commit on a stream for which the authenticated DID is not a controller.  Assuming that still throws an exception that the end user dev will see in their code, then this looks good to me.

packages/stream-tile/src/tile-document.ts
oed: Agreed that it would be good to have this test. My assumption is that we should throw an error when the commit is applied! @ukstv

oed: Ah, that's what you said Spencer :)

ukstv: > What's the behavior now if the currently authenticated DID isn't the controller of this stream?

It throws. Just tried that with my test harness. After signing commit, control flow goes through `applyCommit` function that verifies the signature.

ukstv: Can not add a test here, because there is no resolver that can be used to demonstrate the behaviour.

oed: > Can not add a test here, because there is no resolver that can be used to demonstrate the behaviour.

Sure you can. Just create a Tile with key-did-1 then try to update it with key-did-2?

ukstv: > Sure you can. Just create a Tile with key-did-1 then try to update it with key-did-2?

Hmm, this does not get deep enough for 'controllers' code path, but yeah, makes sense.

stbrody: Ooh, we should also test this for *creating* a stream with the wrong controller, not just for updates.  I just realized that I suspect my fix from https://github.com/ceramicnetwork/js-ceramic/pull/1597 may not have handled the stream **creation** case properly and may cause the exception to be swallowed.  If so the fix would be to do the same for stream creation as we do for stream updates and pass the `throwOnInvalidCommit` flag in `opts`

ukstv: Done.

packages/core/src/__tests__/tile-controller.test.ts
```diff
@@ -0,0 +1,63 @@
+import { IpfsApi } from '@ceramicnetwork/common'
```
stbrody, line 1: wow, what a nice looking test file!

packages/stream-tile/src/tile-document.ts
stbrody: nice!  Glad to see I didn't accidentally break stream creation 😅.  And I see now why it works so it makes sense too.

stbrody (approved):
LGTM, thanks!

PR #1616, opened 2021-08-02
chore(cli): Rename default client config file

stbrody:


oed (commented):
Why are we renaming this file?

ukstv:
> Why are we renaming this file?

The specification tells us to do this.

As per spec, there are two default config files: `~/.ceramic/client.config.json` and `~/.ceramic/daemon.config.json`.  The latter by default contains all the current default parameters of `ceramic daemon` command.

Current config file is named `config.json`. The name does not imply it only applies to CLI _client_ configuration. One can be confused, what it really means. So, we go with changing the name of that file, in a backwards-compatible fashion.

packages/cli/src/ceramic-cli-utils.ts
```diff
@@ -454,6 +455,26 @@ export class CeramicCliUtils {
     } catch (e) {
       // TODO handle invalid configuration
     }
+
+    // If nothing found in default config file path, check legacy path too
+    // TODO(1615): Remove this backwards compatibility code
+    const legacyCliConfigPath = path.join(DEFAULT_CLI_CONFIG_PATH, LEGACY_CLI_CONFIG_FILE)
+    try {
+      await fs.access(legacyCliConfigPath)
+      const config = await JSON.parse(await fs.readFile(legacyCliConfigPath, { encoding: 'utf8' }))
```
ukstv, line 464: JSON.parse is synchronous.

packages/cli/src/ceramic-cli-utils.ts
```diff
@@ -454,6 +455,26 @@ export class CeramicCliUtils {
     } catch (e) {
       // TODO handle invalid configuration
     }
+
+    // If nothing found in default config file path, check legacy path too
+    // TODO(1615): Remove this backwards compatibility code
+    const legacyCliConfigPath = path.join(DEFAULT_CLI_CONFIG_PATH, LEGACY_CLI_CONFIG_FILE)
+    try {
+      await fs.access(legacyCliConfigPath)
```
ukstv, line 463: Arrgh, really dislike this abuse of exception handling for control flow purposes, but I know of no better way :(

ukstv (approved):
🏇

oed:
> The specification tells us to do this.

Ok, I have not had a chance to read though the spec. Will do now!

oed:
Ok, read the spec. Still not sure why this is really needed. Imo it's more confusing to have two separate config files. Not sure what we gain by separating them really?

ukstv:
> Ok, read the spec. Still not sure why this is really needed. Imo it's more confusing to have two separate config files. Not sure what we gain by separating them really?

Daemon config is predominantly used in cloud environment, where client config, and whole notion of client is meaningless, as deployment these days heavily uses automation. Combining the two would bring quite a lot of daily confusion: on this environment, these fields are not used, while on other environment they are used. Say, you tried some configuration locally, then for deployment purposes you have to strip down unused client fields.

oed:
Yeah, that makes sense @ukstv. At first glance I didn't think it would be so problematic to combine them, but I see your point 👍

oed (approved):

packages/cli/src/ceramic-cli-utils.ts
oed: I'm sure fs has some way to check if a file is present?

ukstv: `fs.access` is richer, and serves a replacement for deprecated `fs.exists`. In addition to existence checking, it also makes sure you can read that. But its API sucks :(

stbrody: yeah, I'm also not a fan of exceptions for control flow but this code is pretty isolated so I'm not that worried about it at the moment.  There are a lot of ways this code could be cleaned up at some point, but that feels low priority at the moment

PR #1619, opened 2021-08-04
feat: Allow updating tile immediately after controller change

ukstv:
Right now, controller change comes into effect only after a tile is anchored. Here we allow a new controller to edit a tile without waiting for an anchor.

Prompted by HackFS discord message.


oed (approved):
lgtm

stbrody (approved):
`state.next` strikes again!  One of the most subtle things about our implementation currently that I fully expect to cause more bugs down the line

PR #1622, opened 2021-08-05
feat(core): Add API to request an anchor

stbrody:


stbrody:
CC @oed just a heads up that this adds a new HTTP endpoint

packages/common/src/stream.ts
```diff
@@ -224,6 +224,10 @@ export abstract class Stream extends Observable<StreamState> implements StreamSt
     this.state$.next(stream.state)
   }
 
+  async requestAnchor(): Promise<AnchorStatus> {
+    return this.api.requestAnchor(this.id)
+  }
```
oed, line 229: Should we add the same thing to caip10-link?

oed (commented):
Thanks, is was needed.

packages/common/src/stream.ts
stbrody: This is in the Stream base class, so it's available to all StreamType implementations

oed: Ah, my bad!

ukstv (approved):

PR #1624, opened 2021-08-06
feat: pkh-did-resolver

oed:
A quick and dirty implementation of `did:pkh`. Right now it only supports plain caip10 accountIds. Roughly based on [this spec](https://github.com/spruceid/ssi/blob/main/did-pkh/did-pkh-method-draft.md), both the caip10 spec and the `did:pkh` spec are in flight and may change soon.


oed:
This should be ready for review!

PR #1629, opened 2021-08-09
feat(cli): Add daemon config file

stbrody:
builds on https://github.com/ceramicnetwork/js-ceramic/pull/1628

### DIFF
--- a/packages/cli/src/bin/ceramic.ts
+++ b/packages/cli/src/bin/ceramic.ts
@@ -4,6 +4,7 @@ import { CeramicCliUtils } from '../ceramic-cli-utils'
 
 program
   .command('daemon')
+  .option('--config <path>', 'Path to the Ceramic Daemon config file')
   .option('--ipfs-api <url>', 'The ipfs http api to use')
   .option(
     '--ethereum-rpc <url>',
@@ -57,6 +58,7 @@ program
   .description('Start the daemon')
   .action(
     async ({
+      config,
       ipfsApi,
       ethereumRpc,
       anchorServiceApi,
@@ -82,6 +84,7 @@ program
         )
       }
       await CeramicCliUtils.createDaemon(
+        config,
         ipfsApi,
         ethereumRpc,
         anchorServiceApi,
--- a/packages/cli/src/bin/ceramic.ts
+++ b/packages/cli/src/bin/ceramic.ts
@@ -235,21 +235,21 @@ config
   .command('show')
   .description('Show CLI Ceramic configuration')
   .action(async () => {
-    await CeramicCliUtils.showConfig()
+    await CeramicCliUtils.showCliConfig()
   })
 
 config
   .command('set <variable> <value>')
   .description('Set variable value')
   .action(async (variable, value) => {
-    await CeramicCliUtils.setConfig(variable, value)
+    await CeramicCliUtils.setCliConfig(variable, value)
   })
 
 config
   .command('unset <variable>')
   .description('Unset configuration variable')
   .action(async (variable) => {
-    await CeramicCliUtils.unsetConfig(variable)
+    await CeramicCliUtils.unsetCliConfig(variable)
   })
 
 program.parse(process.argv)
--- a/packages/cli/src/ceramic-cli-utils.ts
+++ b/packages/cli/src/ceramic-cli-utils.ts
@@ -139,7 +139,7 @@ export class CeramicCliUtils {
     deterministic: boolean,
     schemaStreamId: string = null
   ): Promise<void> {
-    await CeramicCliUtils._runWithCeramic(async (ceramic: CeramicClient) => {
+    await CeramicCliUtils._runWithCeramicClient(async (ceramic: CeramicClient) => {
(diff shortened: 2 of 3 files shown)


packages/cli/src/bin/ceramic.ts
```diff
@@ -4,6 +4,7 @@ import { CeramicCliUtils } from '../ceramic-cli-utils'
 
 program
   .command('daemon')
+  .option('--config <path>', 'Path to the Ceramic Daemon config file')
```
oed, line 7: Is only daemon config file configurable? Would it make more sense to just specify the config directory?

packages/cli/src/ceramic-cli-utils.ts
```diff
@@ -25,10 +25,14 @@ import KeyDidResolver from 'key-did-resolver'
...
 const DEFAULT_CLI_CONFIG_FILE = 'client.config.json'
-const LEGACY_CLI_CONFIG_FILE = 'config.json' // todo(1615): Remove this backwards compatibily support
-const DEFAULT_CLI_CONFIG_PATH = path.join(os.homedir(), '.ceramic')
-const DEFAULT_NETWORK = Networks.TESTNET_CLAY
+const LEGACY_CLI_CONFIG_FILE = 'config.json' // todo(1615): Remove this backwards compatibility support
+const DEFAULT_CONFIG_PATH = path.join(os.homedir(), '.ceramic')
+
+const DEFAULT_DAEMON_CONFIG: DaemonConfig = {
+  network: Networks.TESTNET_CLAY,
+}
```
oed, line 35: Is this the only thing that is needed in the default config?

packages/cli/src/bin/ceramic.ts
stbrody: this flag is specific to the `daemon` command so there's no reason to take anything other than the path to the daemon config file.  I also don't think it would make sense to make the location of the config file for the cli client be configurable, as you'd have to add the flag to every cli command it doesn't really make sense to try to use the CLI with passing a flag for the config file on every single command you try to run through the cli

packages/cli/src/ceramic-cli-utils.ts
stbrody: I added https://github.com/ceramicnetwork/js-ceramic/issues/1632 to revisit what the defaults should be

packages/cli/src/ceramic-cli-utils.ts
```diff
@@ -364,17 +382,17 @@ export class CeramicCliUtils {
   }
 
   /**
-   * Open Ceramic and execute function
+   * Open Ceramic http client and execute function
    * @param fn - Function to be executed
    * @private
    */
-  static async _runWithCeramic(fn: (ceramic: CeramicClient) => Promise<void>): Promise<void> {
+  static async _runWithCeramicClient(fn: (ceramic: CeramicClient) => Promise<void>): Promise<void> {
```
ukstv, line 389: IMO no need to specify it is ceramic client. Just `runWithClient` seems to fit.

packages/cli/src/ceramic-cli-utils.ts
```diff
@@ -477,20 +495,37 @@ export class CeramicCliUtils {
...
-    return await this._saveCliConfig({})
+    await this._saveConfig({}, DEFAULT_CLI_CONFIG_FILE)
+    return {}
+  }
+
+  /**
+   * Load configuration file for the Ceramic Daemon.
+   * @private
+   */
+  static async _loadDaemonConfig(): Promise<DaemonConfig> {
```
ukstv, line 506: I guess, it is the next iteration to make it `ConfigFile.load`.

ukstv (approved):

packages/cli/src/ceramic-cli-utils.ts
stbrody: I actually think it's generally good practice to separate data storage objects from logic to load said data.  I could see a `ConfigFileLoader.loadConfig()` if we wanted to put the load logic into a class (so that it could be mocked, for example), but I wouldn't want to actually change `DaemonConfig` to a class and put logic into it.  In any case I feel like the simple thing that is done here is probably fine for now.  If we need to be able to mock the load behavior in the future we can clean it up then.

stbrody: Actually, with javascript it would be possible to mock this behavior even with it just being a static function like it is here.  So I'm not really sure what having a separate class for encapsulating the load logic would gain us?

ukstv: > it's generally good practice to separate data storage objects from logic to load said data.

Indeed, and it does not involve `DaemonConfig` interface in any way. `ConfigFile` is to read a file, and provide proper safe accessors like `get('resolvers.nft-did-resolver')`, so that a reader does not do `resolvers?.nft-did-resolver?.whatever` on raw json.

stbrody: ah okay, yeah that makes sense.  Seems not that important at the moment though?

ukstv: Not atm.

PR #1633, opened 2021-08-10
feat(cli): Add hierarchy to daemon config

stbrody:
Adds basic structure to the DaemonConfig.

One thing that is still missing from this PR relative to the spec is that for things like the state store that has 2 modes it can run in, and where some options only apply to one mode or the other, this PR does not introduce an explicit mode/type/kind flag.  I like the idea of adding such an explicit flag to select which mode to run in, but I wasn't sure the best way to do so.  Ideally this flag would be an enum, so it could have a fixed set of values, but then the JSON representation for the enum that goes into the config file would wind up being an integer, instead of the string enum name.  Is there any way to iterate over the parsed JSON and for every `enum`-typed variable parse the string name of the enum value into the enum itself?  I'm not sure of any way to iterate over fields in an interface and check which of them are of type `enum`.  Without such a construct, then every enum typed field in the config will need to be handled explicitly, which seems brittle and onerous to maintain.

Curious if you have any thoughts on how best to approach this @ukstv


ukstv:
Enum in TS can be a string all right:
```typescript
    
  enum Power {
      MID = 'mid',
      HIGH = 'high',
  }

const input: string = 'mid'
const value = Object.values(Power).find(v => v == input)
if (!value) throw new Error(`no value`)
const v: Power = value
console.log(v)
```

ukstv (dismissed):

packages/cli/src/daemon-config.ts
```diff
@@ -0,0 +1,46 @@
+import { LoggerConfig, SyncOptions } from '@ceramicnetwork/common'
+
+export interface DaemonIpfsConfig {
```
oed, line 3: Can we add detailed TS docs in this file? Will be super nice to be able to just point to the generated docs for how to configure your node!

oed (changes requested):

packages/cli/src/daemon-config.ts
```diff
@@ -0,0 +1,46 @@
...
+  syncOverride?: SyncOptions
+  validateStreams?: boolean
+}
+
+/**
+ * Daemon create options
+ */
+export interface DaemonConfig {
+  anchor?: DaemonAnchorConfig
+  httpApi?: DaemonHTTPApiConfig
```
oed, line 40: Json normally uses `- ` instead of camelCase, e.g. `http-api`.

oed (approved):

oed (changes requested):
ugh, not supposed to be an approval

packages/cli/src/daemon-config.ts
stbrody: I don't think that's a valid field name in typescript.  When I try it I get compile failures

stbrody: oh, I guess it works if I put quotes around the field name.  That's pretty annoying though and would need to be done throughout the code anywhere the field is referenced.  Is it worth it?

packages/cli/src/ceramic-cli-utils.ts
```diff
@@ -106,42 +100,74 @@ export class CeramicCliUtils {
...
-        Object.entries(configFromCli).filter(([k, v]) => v !== undefined)
-      )
+      configFromCli = this.removeUndefinedFields(configFromCli)
     }
 
     const config = Object.assign(configFromFile, configFromCli)
     return CeramicDaemon.create(config)
   }
 
+  static removeUndefinedFields(obj: Record<string, any>): Record<string, any> {
```
stbrody, line 157: I need to go on record that I hate javascript for making me write this.

@ukstv, did I miss some built-in function that would have done this for me?

packages/cli/src/daemon-config.ts
oed: I think it is worth conforming to the widely used json formatting rules.

packages/cli/src/daemon-config.ts
oed: Thanks for adding this!

packages/cli/src/daemon-config.ts
stbrody: Done, although it really makes things way uglier, and also seems to remove the compile-time checks that the fields being accessed are part of the defined interface, making the code more error-prone.

oed (approved):
Approving. The json kebab-case with typescript is quite annoying. If there is an easy way to convert between the two it might be worth doing that?

ukstv:
Now it got ugly with kebab-cased fields. Why? The major point of the spec was to separate translation of json from representation in code. Now it is all conflated, and fragile.

@stbrody can you start with the `ConfigFile` reader first, maybe? Then construct `DaemonConfig` instance based on the fields in config. Then add CLI overrides. I really do not want to go back to this config story again in the future, redoing the current iteration.

ukstv (changes requested):
Please, separate JSON structure from the internal data structure construction.

packages/cli/src/ceramic-daemon.ts
```diff
@@ -164,15 +152,25 @@ export class CeramicDaemon {
...
       ceramicConfig.loggerProvider.getDiagnosticsLogger(),
-      opts.ipfsHost
+      opts.ipfs?.host
     )
 
     const [modules, params] = Ceramic._processConfig(ipfs, ceramicConfig)
-
-    if (opts.s3StateStoreBucket) {
-      const s3StateStore = new S3StateStore(opts.s3StateStoreBucket)
+    modules.loggerProvider
```
stbrody, line 162: Added this log with the daemon config, and removed the log from within Ceramic itself that logs the config.  This seems more useful, and doesn't require any special work to massage complex objects into a format where they can be logged.

packages/cli/src/daemon-config.ts
```diff
@@ -0,0 +1,285 @@
...
+ * Ceramic Daemon options for configuring behavior related to IPFS.
+ */
+@jsonObject
+@toJson
+export class DaemonIpfsConfig {
+  /**
+   * Whether the daemon should start its own bundled in-process ipfs node, or if it should connect
+   * over HTTP to an existing remote IPFS node.
+   */
+  @jsonMember(String)
```
stbrody, line 23: I'm disappointed at the need of these type specifiers in the decorators.  They're only there for tests.  The regular ceramic daemon compiles and runs just fine without them - the `typedjson` library uses `reflect-metadata` to infer the types on its own just fine.  But when running in jest with babel it couldn't do the type inference unless I put these here explicitly.

stbrody (commented):
Okay new, more complete and robust version is up.  Please take another look

oed (approved):
lgtm

ukstv (approved):

PR #1634, opened 2021-08-10
fix(core): Periodically publish keepalive pubsub message

stbrody:


packages/core/src/dispatcher.ts
```diff
@@ -147,10 +163,12 @@ export class Dispatcher {
     let dagResult = null
     for (let retries = IPFS_GET_RETRIES - 1; retries >= 0 && dagResult == null; retries--) {
       try {
-        dagResult = await this._ipfs.dag.get(asCid, {timeout: IPFS_GET_TIMEOUT, path})
+        dagResult = await this._ipfs.dag.get(asCid, { timeout: IPFS_GET_TIMEOUT, path })
       } catch (err) {
         if (err.code == 'ERR_TIMEOUT') {
-          console.warn(`Timeout error while loading CID ${cid.toString()} from IPFS. ${retries} retries remain`)
+          console.warn(
```
stbrody, line 169: Looks like some formatting changes snuck in here, not sure why.  @ukstv, did we ever run prettier against the entire codebase so that everything would be formatted the same way?

packages/core/src/pubsub/pubsub-message.ts
```diff
@@ -34,7 +35,13 @@ export type ResponseMessage = {
   tips: Map<string, CID>
 }
 
-export type PubsubMessage = UpdateMessage | QueryMessage | ResponseMessage
+// All nodes will always ignore this message
+export type KeepaliveMessage = {
+  typ: MsgType.KEEPALIVE
+  ts: number // current time
```
oed, line 41: Why do we include a timestamp?

packages/core/src/pubsub/pubsub.ts
```diff
@@ -98,4 +110,21 @@ export class Pubsub extends Observable<PubsubMessage> {
...
+   * we'll publish one so that we never go longer than MAX_PUBSUB_PUBLISH_INTERVAL without
+   * publishing a pubsub message.  This is to work around a bug in IPFS where peer connections
+   * get dropped if they haven't had traffic in too long.
+   */
+  publishPubsubKeepaliveIfNeeded(): void {
+    const now = Date.now()
+    if (now - this.lastPublishedMessageDate < this.maxPubsubPublishInterval / 2) {
+      // We've published a message recently enough, no need to publish another
+      return
+    }
```
oed, line 125: Why do it like this when we can just trigger it on a regular interval from `setInterval`?

packages/core/src/dispatcher.ts
ukstv: Git blame tells me that 15 days ago your commit did not include proper formatting.

https://github.com/ceramicnetwork/js-ceramic/commit/99da05c94cfdbf8472c7f7cf14a3b68227766b4a#diff-e522280df0e1c4c8fe4f157393e23872535ee7bab6d33a2930c26b18f7411d19R150

packages/core/src/pubsub/pubsub.ts
```diff
@@ -48,11 +48,14 @@ function ipfsToPubsub(
  */
 export class Pubsub extends Observable<PubsubMessage> {
   private readonly peerId$: Observable<string>
+  private readonly pubsubKeepaliveInterval
+  private lastPublishedMessageDate: number = Date.now() - this.maxPubsubPublishInterval
```
ukstv, line 52: Add another thing on top, please. Too much responsibility in here already. Let's not make it a stuffed turkey.

packages/core/src/pubsub/pubsub.ts
```diff
@@ -66,6 +69,14 @@ export class Pubsub extends Observable<PubsubMessage> {
     // Textually, `this.peerId$` appears after it is called.
     // Really, subscription is lazy, so `this.peerId$` is populated before the actual subscription act.
     this.peerId$ = from<Promise<string>>(this.ipfs.id().then((_) => _.id))
+    this.pubsubKeepaliveInterval = setInterval(
+      this.publishPubsubKeepaliveIfNeeded.bind(this),
+      this.maxPubsubPublishInterval / 2
+    )
+  }
+
+  shutdown(): void {
```
ukstv, line 78: `Pubsub` is `Observable`, do not stuff it with an alien responsibility. See how `MessageBus` treats a similar case.

This retry feature should be a thing between MessageBus and Pubsub, similar to MessageBus in how it treats "I am subscribed to" state.

ukstv (changes requested):

packages/core/src/pubsub/pubsub-message.ts
```diff
@@ -34,7 +35,13 @@ export type ResponseMessage = {
   tips: Map<string, CID>
 }
 
-export type PubsubMessage = UpdateMessage | QueryMessage | ResponseMessage
+// All nodes will always ignore this message
```
v-stickykeys, line 38: What do you mean by nodes will ignore this message? Should this not be tracked as a message with event type "received"?

v-stickykeys (commented):
+1 questions from others

packages/core/src/dispatcher.ts
stbrody: ah okay, that was before I set up my IDE to do it automatically. Thanks

packages/core/src/pubsub/pubsub-message.ts
stbrody: Not strictly needed.  Just keeps each keepalive message unique, and figured it might be useful in debugging if we have connectivity issues in the future, it would be easier to see when we last published a message

packages/core/src/pubsub/pubsub.ts
stbrody: You mean just unilaterally publish a message on an interval?  That would create additional pubsub traffic unnecessarily for a well-used node.  I'm trying to only publish these keepalive messages if the node isn't publishing other messages from normal use.  So basically only idle nodes will generate keepalive messages

packages/core/src/pubsub/pubsub.ts
stbrody: Sorry, I don't actually understand what you're asking for here

packages/core/src/pubsub/pubsub.ts
stbrody: so I have to admit I don't 100% understand what MessageBus is doing with `betweenTimeout`, but from how it's used in `queryNetwork`, I think it's for doing the timeout so that we keep waiting an extra 300ms after the first response is received, to make sure we don't get a better response.  That seems fairly different than what we're trying to do here which is do something on an interval, but skip doing the thing if we've done it already within some interval.

To be honest, I find much of the more complex rxjs logic pretty inscrutable.  The advantage of doing it the way I did it here is it's actually understandable to new readers, since the logic is pretty simple.  If you really think this should be done with rxjs operations, then I might need to hand this ticket off to you, or maybe we can pair program it, because I can't imagine myself being able to figure out how to do it on my own without taking days to do so.

packages/core/src/pubsub/pubsub-message.ts
stbrody: It should get logged the same as any pubsub message, but the ceramic node doesn't do any processing of it

packages/core/src/pubsub/pubsub.ts
```diff
@@ -66,6 +69,14 @@ export class Pubsub extends Observable<PubsubMessage> {
     // Textually, `this.peerId$` appears after it is called.
     // Really, subscription is lazy, so `this.peerId$` is populated before the actual subscription act.
     this.peerId$ = from<Promise<string>>(this.ipfs.id().then((_) => _.id))
+    this.pubsubKeepaliveInterval = setInterval(
+      this.publishPubsubKeepaliveIfNeeded.bind(this),
+      this.maxPubsubPublishInterval / 2
```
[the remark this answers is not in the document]
stbrody: Either we set the interval, then have to do math to figure out what the max delay between messages becomes based on that interval.  Or we set the max delay between messages, and do math to set the interval.  Since the thing we actually care about is how long it is between messages in the worst case, that seemed like the thing that made sense to set, and then work out the interval timing from that.

packages/core/src/pubsub/pubsub-message.ts
oed: Ok, just thinking this adds to bandwidth overhead, but maybe just marginal?

packages/core/src/pubsub/pubsub.ts
oed: Oh I see! Makes sense ty!

packages/core/src/pubsub/pubsub.ts
ukstv: Proposed changes to `Pubsub` add here one more responsibility - posting a keep-alive message when idle. Current code is complex enough, better handle that responsibility somewhere else. It is proper time now to recall a single-responsibility principle.

Upstream implementation uses `new MessageBus(new Pubsub())` construction. We can wrap a Pubsub instance with `KeepAlive` class that provides the an interface similar to Pubsub (i.e. Observable + next) but also sends a keep-alive message (at most) every N seconds. Like this: https://github.com/ceramicnetwork/js-ceramic/blob/su/keep-alive-wrap/packages/core/src/pubsub/keep-alive.ts; usage: https://github.com/ceramicnetwork/js-ceramic/blob/su/keep-alive-wrap/packages/core/src/dispatcher.ts#L57

Note: few console-logs here and there for visibility purposes.

packages/core/src/pubsub/pubsub.ts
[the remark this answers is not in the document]
v-stickykeys: I think the key thing I was missing here was in effect "we are doing this to reduce the amount of pubsub traffic to only what is necessary as much as possible". May be worth including in the existing comments.

stbrody:
ping @ukstv

ukstv (approved):

PR #1648, opened 2021-08-15
chore(cli): Use consistent type for 'corsAllowedOrigins'

stbrody:
In order to introduce a config file for the daemon with automatic parsing, I need all the config options to have a single type - 'corsAllowedOrigins' is the one field causing issues given it can be either a single string or an array of RegExps.  I don't totally follow how express uses this field, but as far as I can tell converting the input to a RegExp shouldn't be necessary?  Could also try to go in the direction of moving to always being an array of RegExp, but when I tried to represent the default case as `new RegExp('*')` I get an Eslint error: `Invalid regular expression: /*/: Nothing to repeat.(no-invalid-regexp)`.

### DIFF
--- a/packages/cli/src/bin/ceramic.ts
+++ b/packages/cli/src/bin/ceramic.ts
@@ -49,7 +49,7 @@ program
   )
   .option(
     '--cors-allowed-origins <list>',
-    'Space-separated list of strings and/or regex expressions to set for Access-Control-Allow-Origin . Defaults to all: "*"'
+    'Space-separated list of strings and/or regex expressions to set for Access-Control-Allow-Origin . Defaults to all: ".*"'
   )
   .option(
     '--sync-override <string>',
--- a/packages/cli/src/ceramic-cli-utils.ts
+++ b/packages/cli/src/ceramic-cli-utils.ts
@@ -101,9 +101,9 @@ export class CeramicCliUtils {
 
     let configFromCli: DaemonConfig
     {
-      let _corsAllowedOrigins: string[] = ['*']
+      let _corsAllowedOrigins: RegExp[] = [new RegExp('.*')]
       if (corsAllowedOrigins != null) {
-        _corsAllowedOrigins = corsAllowedOrigins.split(' ')
+        _corsAllowedOrigins = corsAllowedOrigins.split(' ').map((origin) => new RegExp(origin))
       }
       const logLevel = verbose ? LogLevel.verbose : debug ? LogLevel.debug : LogLevel.important
       const loggerConfig: LoggerConfig = {
--- a/packages/cli/src/ceramic-cli-utils.ts
+++ b/packages/cli/src/ceramic-cli-utils.ts
@@ -101,9 +101,9 @@ export class CeramicCliUtils {
 
     let configFromCli: DaemonConfig
     {
-      let _corsAllowedOrigins: string | RegExp[] = '*'
-      if (corsAllowedOrigins != null && corsAllowedOrigins != '*') {
-        _corsAllowedOrigins = corsAllowedOrigins.split(' ').map((origin) => new RegExp(origin))
+      let _corsAllowedOrigins: string[] = ['*']
+      if (corsAllowedOrigins != null) {
+        _corsAllowedOrigins = corsAllowedOrigins.split(' ')
       }
       const logLevel = verbose ? LogLevel.verbose : debug ? LogLevel.debug : LogLevel.important
       const loggerConfig: LoggerConfig = {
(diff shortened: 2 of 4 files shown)


packages/cli/src/ceramic-cli-utils.ts
```diff
@@ -101,9 +101,9 @@ export class CeramicCliUtils {
 
     let configFromCli: DaemonConfig
     {
-      let _corsAllowedOrigins: string | RegExp[] = '*'
-      if (corsAllowedOrigins != null && corsAllowedOrigins != '*') {
-        _corsAllowedOrigins = corsAllowedOrigins.split(' ').map((origin) => new RegExp(origin))
+      let _corsAllowedOrigins: string[] = ['*']
```
stbrody, line 104: This is the change I'm most worried about - I don't want to accidentally break the common case where all origins should be allowed.  I'm also not really sure how to test these changes.

stbrody (commented):
I'm definitely a little bit flailing in the dark with this change.  Definitely could use you attention on this one @v-stickykeys.  The goal is just to have some single type for this field - could be a string array, a RegExp array, a single string, etc.  Whatever if easiest and won't break anything.

packages/cli/src/ceramic-cli-utils.ts
v-stickykeys: @stbrody this works so we can use RegExp[]
```
a = new RegExp('^\\*$')
> /^\*$/
a.test('*')
> true
```

v-stickykeys: Oh actually I take that back I guess the default here should be to match anything

`.*`

stbrody: ah yes, of course, thanks.

Updated to use array of RegExp as the type

v-stickykeys (approved):
LGTM

v-stickykeys:
Oh sorry @stbrody we need to update here as well https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/cli/src/bin/ceramic.ts#L52

stbrody:
> Oh sorry @stbrody we need to update here as well [https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/cli/src/bin/ceramic.ts#L52](https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/cli/src/bin/ceramic.ts?rgh-link-date=2021-08-17T14%3A35%3A15Z#L52)

What kind of change do you think is needed?  That still looks right to me? @v-stickykeys

stbrody:
> Oh sorry @stbrody we need to update here as well [https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/cli/src/bin/ceramic.ts#L52](https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/cli/src/bin/ceramic.ts?rgh-link-date=2021-08-17T14%3A35%3A15Z#L52)

Updated comment to clarify that the default is `".*"` not `"*"`

packages/cli/src/bin/ceramic.ts
```diff
@@ -49,7 +49,7 @@ program
   )
   .option(
     '--cors-allowed-origins <list>',
-    'Space-separated list of strings and/or regex expressions to set for Access-Control-Allow-Origin . Defaults to all: "*"'
+    'Space-separated list of strings and/or regex expressions to set for Access-Control-Allow-Origin . Defaults to all: ".*"'
```
v-stickykeys, line 52:
```suggestion
    'Space-separated list of strings and/or regex expressions to set for Access-Control-Allow-Origin . Defaults to all: "/.*/"'
```

v-stickykeys (commented):
After testing, modified suggestion for option help text. Looks ready to go

packages/cli/src/bin/ceramic.ts
stbrody: hmm... interesting. In my testing, if I specify `".*"` as the input string, then when it's parsed it winds up as a RegExp that displays as `"/.*/"`.  If I pass `"/.*/"` as input then it shows up as `"\\/.*\\/"` when parsed

v-stickykeys: > hmm... interesting. In my testing, if I specify `".*"` as the input string, then when it's parsed it winds up as a RegExp that displays as `"/.*/"`. If I pass `"/.*/"` as input then it shows up as `"\\/.*\\/"` when parsed

Yes you are correct we don't need the slashes because it is a string input that gets converted to regex.

stbrody: okay great, then I'll leave it as-is.  Thanks!

PR #1656, opened 2021-08-19
feat: configure nft-did-resolver via config file

ukstv:
Pass `resolvers.nft-did-resolver` to nft-did-resolver.

In an ideal universe, one could iterate over keys of `resolvers`, and initialize (and dynamically include!) the appropriate resolvers, but this IMO is for the next time, when dynamic non-http modules are a thing.

### DIFF
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -176,13 +176,13 @@ export class CeramicDaemon {
 
     const ceramic = new Ceramic(modules, params)
     await ceramic._init(true, true)
-
     const did = new DID({
       resolver: {
         ...KeyDidResolver.getResolver(),
         ...ThreeIdResolver.getResolver(ceramic),
         ...NftDidResolver.getResolver({
           ceramic: ceramic,
+          ...opts.resolvers['nft-did-resolver']
         }),
         ...(ceramicConfig.ethereumRpcUrl &&
           EthrDidResolver.getResolver({
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -182,7 +182,7 @@ export class CeramicDaemon {
         ...ThreeIdResolver.getResolver(ceramic),
         ...NftDidResolver.getResolver({
           ceramic: ceramic,
-          ...opts.resolvers?.['nft-did-resolver']
+          ...opts.resolvers?.nftDidResolver
         }),
         ...(ceramicConfig.ethereumRpcUrl &&
           EthrDidResolver.getResolver({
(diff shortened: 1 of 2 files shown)


packages/cli/src/daemon-config.ts
```diff
@@ -262,6 +269,9 @@ export class DaemonConfig {
   @jsonMember(DaemonStateStoreConfig, { name: 'state-store' })
   stateStore: DaemonStateStoreConfig
 
+  @jsonMember(DaemonCeramicResolversConfig)
+  resolvers?: DaemonCeramicResolversConfig;
```
oed, line 273:
```suggestion
  didResolvers?: DaemonCeramicResolversConfig;
```

This would be more clear imo

oed (commented):
Would be nice to include configuration pass though for ethr-did-resolver as well in this PR 🙏

packages/cli/src/daemon-config.ts
ukstv: okay

ukstv: done.

packages/cli/src/daemon-config.ts
```diff
@@ -155,6 +155,22 @@ export class DaemonAnchorConfig {
   ethereumRpcUrl?: string
 }
 
+@jsonObject
+@toJson
+export class DaemonCeramicResolversConfig {
```
stbrody, line 160: nitpick: Maybe call this `DaemonDidResolversConfig`

packages/cli/src/daemon-config.ts
```diff
@@ -155,6 +155,22 @@ export class DaemonAnchorConfig {
...
+  /**
+   * Configuration for nft-did-resolver. Its README contains appropriate documentation.
```
stbrody, line 162: please add line to the doc comments for all the new fields you added that explain that the name of the field is different in the config file than it is in code.  Something like "When specifying in a config file, use the name 'nft-did-resolver'".

This is the one annoying downside of using the typedjson library to do the field name renames automatically - it makes the doccomments more confusing.

stbrody (approved):
lgtm mod adding note about field names to doc comments

packages/cli/src/ceramic-daemon.ts
```diff
@@ -176,23 +176,17 @@ export class CeramicDaemon {
...
         }),
-        ...(ceramicConfig.ethereumRpcUrl &&
-          EthrDidResolver.getResolver({
-            networks: [
-              {
-                name: 'mainnet',
-                rpcUrl: ceramicConfig.ethereumRpcUrl,
-              },
-            ],
-          })),
```
oed, line 195: Hm, what happened to the default here now? Are there no default anymore?

packages/cli/src/daemon-config.ts
ukstv: ok

packages/cli/src/ceramic-daemon.ts
ukstv: Got it back.

oed (approved):
Thanks!

PR #1664, opened 2021-08-23
feat: named exports

PaulLeCam:
Following on https://github.com/ceramicnetwork/js-ceramic/pull/1661 this adds named exports for the DID resolvers and the core package.

### DIFF
--- a/packages/3id-did-resolver/src/index.ts
+++ b/packages/3id-did-resolver/src/index.ts
@@ -216,7 +216,7 @@ const resolve = async (
   }
 }
 
-export default {
+export const threeId =  {
   getResolver: (ceramic: CeramicApi): ResolverRegistry => ({
     '3': (
       did: string,
@@ -254,3 +254,5 @@ export default {
     },
   }),
 }
+
+export default threeId
--- a/packages/3id-did-resolver/src/index.ts
+++ b/packages/3id-did-resolver/src/index.ts
@@ -216,8 +216,8 @@ const resolve = async (
   }
 }
 
-export const threeId =  {
-  getResolver: (ceramic: CeramicApi): ResolverRegistry => ({
+export function getResolver(ceramic: CeramicApi): ResolverRegistry {
+  return {
     '3': (
       did: string,
       parsed: ParsedDID,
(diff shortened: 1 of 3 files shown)


packages/3id-did-resolver/src/index.ts
```diff
@@ -216,7 +216,7 @@ const resolve = async (
   }
 }
 
-export default {
+export const threeId =  {
```
oed, line 219:
```suggestion
export const threeIdDidResolver =  {
```

packages/key-did-resolver/src/index.ts
```diff
@@ -19,7 +19,7 @@ const prefixToDriverMap: any = {
   0xed: ed25519,
 }
 
-export default {
+export const key = {
```
oed, line 22:
```suggestion
export const keyDidResolver = {
```

packages/3id-did-resolver/src/index.ts
```diff
@@ -254,3 +254,5 @@ export default {
     },
   }),
 }
+
+export default threeId
```
oed, line 258:
```suggestion
export default threeIdDidResolver
```

packages/key-did-resolver/src/index.ts
```diff
@@ -55,3 +55,5 @@ export default {
     },
   }),
 }
+
+export default key
```
oed, line 59:
```suggestion
export default keyDidResolver
```

oed (approved):
Thanks. Made a naming suggestion.

PaulLeCam:
@oed thanks for the review and suggestions, i changed it to exporting `getResolver()` directly, I think it's more consistent this way.

oed:
Hm, so you would still have to use the default export if you are importing multiple resolvers then?

ukstv (approved):
And now it is proper `import * as KeyDidResolver from 'key-did-resolver'`. Nice.

PaulLeCam:
> Hm, so you would still have to use the default export if you are importing multiple resolvers then?

No just need to alias the imports:

```js
import { getResolver as get3IDResolver } from '@ceramicnetwork/3id-did-resolver'
import { getResolver as getKeyResolver } from 'key-did-resolver'
```

PR #1665, opened 2021-08-23
fix(core): Increase timeout to stabilize test

stbrody:


packages/core/src/__tests__/ceramic-anchor.test.ts
```diff
@@ -50,7 +50,7 @@ describe('Ceramic anchoring', () => {
   let ipfs3: IpfsApi
 
   beforeAll(async () => {
-    [ipfs1, ipfs2, ipfs3] = await Promise.all(Array.from({ length: 3 }).map(() => createIPFS()))
+    ;[ipfs1, ipfs2, ipfs3] = await Promise.all(Array.from({ length: 3 }).map(() => createIPFS()))
```
stbrody, line 53: kind of obnoxious formatting being imposed by prettier

stbrody: ugh, and that actually makes it fail lint.  Any ideas on how to make prettier not do this @ukstv?

ukstv: I fought with this for some time, and then gave up. The only way forward I see is to do something like
```
const instances = await Promise.all(...)
ipfs1 = instances[0]
ipfs2 = instances[1]
ipfs3 = instances[2]
```
And this is far from perfect too.

ukstv (approved):

PR #1668, opened 2021-08-25
feat: caip10 ethereum+eos integration test

ukstv:
Note: Consider it as a part of the whole caip10-integration thing.

Add caip10-link integration test, starting with ethereum.

Suggestions are welcome on renaming canary-integration.

One idea I have in mind is to run `canary-integration` only on master by changing `test` script to `test:full` inside the package, and adding a CircleCI rule that runs `test:full` on master only.


packages/canary-integration/src/create-ipsf.ts
```diff
@@ -0,0 +1,77 @@
+import dagJose from 'dag-jose'
```
stbrody, line 1: typo in filename: create-ipsf

stbrody (commented):
Why would we only run this on master branch?  That would mean its only tested after a release is already made.  Why not run these tests on every commit like any other test?

packages/canary-integration/src/create-ipsf.ts
ukstv: Oops. Thanks!

ukstv:
> That would mean its only tested after a release is already made.

Yes, this is not good. I was thinking about conserving runners capacity, be it on CircleCI or GitHub Actions. Maybe not on master. Maybe better go with a button-initiated workflow, like for releases.

stbrody:
>  I was thinking about conserving runners capacity, be it on CircleCI or GitHub Actions. 

Yeah, personally I'd vote for running this all the time like all our other tests, at least for now.  If/when in the future our test execution times are too long, or our costs associated with running the tests is too high, then we can take a holistic look on how to break up tests to address it.  By that time, we may have many more tests which may be even slower or more expensive to run than these are, so this kind of feels like premature optimization to me.

packages/canary-integration/src/create-ceramic.ts
```diff
@@ -0,0 +1,38 @@
...
+  const threeIdResolver = ThreeIdResolver.getResolver(ceramic)
+  const resolver = new Resolver({
+    ...threeIdResolver,
+    ...keyDidResolver,
+  })
+  const did = new DID({ provider, resolver })
+  await ceramic.setDID(did)
+  await did.authenticate()
+
+  return ceramic
```
oed, line 37: FWIW, @PaulLeCam created a `jest-environment-ceramic` package. At some point we should migrate that package to the js-ceramic code base and deduplicate a bunch of code.

oed (approved):
lgtm!

Sorry for taking so long to review this.

PR #1679, opened 2021-08-26
WIP: Simple subscription API based on custom interfaces for each StreamType

stbrody:
Here's a WIP implementation to demonstrate the approach that it seemed like everyone wanted on the design document, even though I still prefer the other approach of reusing the types we already have rather than adding new ones.  The PR is still incomplete - there will be some work to update the existing tests to use the new API - but it should be enough to demonstrate how it would be used.

I added some comments inline about what I don't like about this approach.  I'm still interested in trying the approach where the feed() method returns the appropriate Stream subclass, but that's a bit trickier to implement so I don't want to spend a lot of time on it without at least some initial agreement that it's worthwhile to do so.

### DIFF
--- a/packages/cli/src/__tests__/ceramic-daemon.test.ts
+++ b/packages/cli/src/__tests__/ceramic-daemon.test.ts
@@ -12,6 +12,7 @@ import getPort from 'get-port'
 import { createIPFS } from './create-ipfs'
 import { makeDID } from './make-did'
 import { DaemonConfig } from '../daemon-config'
+import InMemoryAnchorService from '@ceramicnetwork/core/lib/anchor/memory/in-memory-anchor-service'
 
 const seed = 'SEED'
 const TOPIC = '/ceramic'
@@ -90,20 +91,8 @@ describe('Ceramic interop: core <> http-client', () => {
    * @param doc
    */
   const anchorDoc = async (doc: Stream): Promise<void> => {
-    // TODO update this to only use public apis
-    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-    // @ts-ignore
-    const changeHandle = doc.state$
-      .pipe(
-        filter((state) =>
-          [AnchorStatus.ANCHORED, AnchorStatus.FAILED].includes(state.anchorStatus)
-        ),
-        take(1)
-      )
-      .toPromise()
-    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-    // @ts-ignore
-    await daemon.ceramic.context.anchorService.anchor()
+    const changeHandle = doc.waitForAnchor()
+    await (daemon.ceramic.context.anchorService as InMemoryAnchorService).anchor()
     await changeHandle
   }
 
--- a/packages/common/src/stream.ts
+++ b/packages/common/src/stream.ts
@@ -243,7 +243,7 @@ export abstract class Stream implements StreamStateHolder {
     return this.api.requestAnchor(this.id)
   }
 
-  async waitForAnchor(): Promise<void> {
+  async waitForAnchor(timeout?: number): Promise<void> {
     const tillAnchored = this.state$
       .pipe(
         filter((state) =>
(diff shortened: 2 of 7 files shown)


stbrody:
Okay, I made some updates based on @PaulLeCam's suggestions, and I agree this is nicer and does address some of my concerns about code duplication.

I think if we're going to go this route, we should commit to moving in the direction of using the new StreamSnapshot type and subtypes more prominently throughout the codebase so that they are more of a first-class citizen rather than a random one-off just for the subscription API.  For example, when loading a Stream at a CommitID or via `atTime` (basically anytime we call [makeReadOnly()](https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/stream-tile/src/tile-document.ts#L249) on a Stream instance today), we should be returning the appropriate Snapshot type, rather than just overriding the mutation of Stream to throw.  That would allow us to lean on the type system to enforce immutability rather than the somewhat hacky runtime approach that we use now. Doing that would likely require creating an inheritance relationship between Stream and StreamSnapshot (most likely Stream should extend StreamSnapshot, TileDocument should extend TileSnapshot, etc), so that it's more natural for some method to return either a mutable Stream instance or an immutable StreamSnapshot instance depending on how the method is called.

If this makes sense to everyone and seems like a reasonable direction to be moving in, then I can do another pass to clean this up and get it more production-ready and into proper review.

@oed @ukstv @smrz2001

ukstv:
Oomph, I am really worried by your suggestion @PaulLeCam This creates quite a lot of issues with how to represent type hierarchy (see lots of `any` types), and breaks composability we have now (streams are no longer interchangeable with other `RunningStateLike` entities). I am more inclined to vote for the solution in the first commit here. It is _much_ cleaner to my taste. The original proposal made cleaner delineation between internal stuff, and developer-exposed API. Now it is all smashed together.

One note on the original proposal: `feed()` could be just a getter `snapshot$`, but this is a weak opinion.

PaulLeCam:
> This creates quite a lot of issues with how to represent type hierarchy (see lots of `any` types)

The `any` types are just fallback for explicit type parameters that are currently missing (ex `Stream<T = any>` rather than `Stream`). The point is on the contrary to avoid having `content: any` types in the stream or state.

> The original proposal made cleaner delineation between internal stuff, and developer-exposed API. Now it is all smashed together.

To me `Stream` and the classes extending it, as well as the snapshot types, are consumer-level APIs, though we can name them differently if you want to have a more explicit separation?
The state and related APIs are mostly internal, but should be accessible for advanced use cases.

We discussed this a bit with Spencer yesterday and Joel this morning, fundamentally what would be helpful would be a clear separation between client and server concerns, so we don't have to worry so much about these differences. Would be good to start defining this sooner than later to help inform these kinds of changes.

ukstv:
> The point is on the contrary to avoid having `content: any` types in the stream.

Okay, now I get it. IMO, it makes more sense to handle two distinct problems separately. Subscriptions have nothing to do with the `content: any`.

ukstv:
I dare to say, I had an experience moving from `content: any` to `content: T`. t was quite hard to set generics right, to make them a helping hand, not an obstacle. That's why I am concerned with what subscription proposal turns into.

stbrody:
I think Paul's original proposal had 2 unrelated changes related to type specifiers.  One was a type specifier for Stream for what the snapshot type is, another for what the type of Stream's `content` is.  I think whatever we decide to do we should consider those two suggestions independently.  I do not intend to add a type specifier for what the underlying `content` is as part of this PR, so let's not spend too much time talking about that now.  I'm much more interested in focusing on the subscription API part, which does in the current proposal introduce a type specifier on Stream for the type of the Snapshot that gets returned when you subscribe to Stream.  @ukstv, is your concern about the type specifier for the snapshot type, or for the content type (which was included in Paul's proposal but is not included in this PR currently)?  Just want to make sure we're on the same page about what we're discussing

stbrody:
Added a default type specifier for the snapshot type and cleaned up all the errant `Stream<any>`

stbrody:
Okay, new version is up.

The main difference between this and the previous version is that Stream no longer extends Observable, and instead there's a `updates$` accessor that returns an Observerable of StreamSnapshot.  That should solve @ukstv concern about overuse of type specifiers.

I also added a `waitForAnchor()` method to Stream, and updated the relevant test code to use it, as an example of how to add more specific subscription functionality based on properties of the StreamState that aren't exposed in the Snapshot types.

This version leaves the state$ StreamState Observerable protected so that the StreamState interface no longer shows up in any public APIs, but streamtype implementers will always have access to it.  If an application developer really wants to access the underlying StreamState, despite it not being a public API and the fact that we make no guarantees that its API won't break in the future, they can do so by accessing the `state$` field, despite it being protected, and suppressing the compile warning that comes from accessing a non-public member.

I also did nothing here to change how readOnly streams are handled, or to unify them with the new Snapshot types.  Cleaning up how readonly streams loaded at a timestamp or CommitID works is left for a future refactor.


lmk if this seems reasonable and I'll start work on productionizing it

PaulLeCam:
> I also did nothing here to change how readOnly streams are handled, or to unify them with the new Snapshot types. Cleaning up how readonly streams loaded at a timestamp or CommitID works is left for a future refactor.

Is this scheduled for the next chunk of work or would these changes be pushed "as-is" please?
If the stream and snapshot don't at least implement the same interface, it's going to be very annoying for consumers to handle these in places that could accept either a snapshot or stream.

Issue #1544, opened 2021-07-15
Store Merkle witness in `CommitData` while processing new commits


smrz2001:
https://github.com/ceramicnetwork/js-ceramic/pull/1528/files#r669404173

PR #1683, opened 2021-08-27
feat(core): optimize document handler IPFS lookups

smrz2001:
(#1532, #1544)


packages/core/src/__tests__/state-manager.test.ts
```diff
@@ -165,12 +165,14 @@ test('handleTip for commit already in log', async () => {
   await (ceramic2.repository.stateManager as any)._handleTip(streamState2, stream1.state.log[1].cid)
 
   expect(streamState2.state).toEqual(stream1.state)
-  expect(retrieveCommitSpy).toBeCalledTimes(3) // TODO(1421): This should be 2!
+  // 4 IPFS retrievals - 2 each (signed commit and linked commit) for CID of commit to be applied and CID of genesis
+  // commit already in the stream state.
+  expect(retrieveCommitSpy).toBeCalledTimes(4)
```
ukstv, line 170: @smrz2001 I am confused. The goal is to reduce calls, but now it seems we have one more call. Is it really an optimisation? Or, maybe I misunderstand something?

smrz2001: No, you're absolutely right.

The reason the number of calls went up here is because the previous count was incorrect.

In the previous version of the code, there was a direct call to `ipfs.dag.get` that was not being counted. I have now consolidated all IPFS lookups in `conflictResolution.ts` as well as both document handlers and 4 is the correct (and lowest) count.

packages/core/src/__tests__/state-manager.test.ts
```diff
@@ -165,12 +165,14 @@ test('handleTip for commit already in log', async () => {
   await (ceramic2.repository.stateManager as any)._handleTip(streamState2, stream1.state.log[1].cid)
 
   expect(streamState2.state).toEqual(stream1.state)
-  expect(retrieveCommitSpy).toBeCalledTimes(3) // TODO(1421): This should be 2!
+  // 4 IPFS retrievals - 2 each (signed commit and linked commit) for CID of commit to be applied and CID of genesis
```
stbrody, line 168: why do we need to lookup the genesis commit if we already have it in the state?

packages/core/src/conflict-resolution.ts
```diff
@@ -290,9 +306,8 @@ export class ConflictResolution {
     // `fetchLog` provides the timestamps.
     if (state && state.log.length === 1) {
       const timestamp = unappliedCommits[0].timestamp
-      const genesisCid = state.log[0].cid
-      const genesis = await this.dispatcher.retrieveCommit(genesisCid)
-      await handler.applyCommit(genesis, { cid: genesisCid, timestamp: timestamp }, this.context)
+      const genesis = await getCommitData(state.log[0], this.dispatcher)
+      await handler.applyCommit(genesis, { cid: genesis.cid, timestamp: timestamp }, this.context)
```
stbrody, line 310: I'm confused now the relationship between `CommitData` and `CommitMeta`.  They both seem to have a `cid` and `timestamp`.  Do we need both?

packages/stream-caip10-link-handler/src/caip10-link-handler.ts
```diff
@@ -46,7 +48,7 @@ export class Caip10LinkHandler implements StreamHandler<Caip10Link> {
       return this._applyGenesis(commit, meta.cid)
     }
 
-    if ((commit as AnchorCommit).proof) {
+    if ((commit as AnchorCommit).proof || (commit as CommitData).proof) {
```
stbrody, line 51: `StreamUtils.isAnchorCommit`?

packages/stream-caip10-link-handler/src/caip10-link-handler.ts
```diff
@@ -153,7 +157,9 @@ export class Caip10LinkHandler implements StreamHandler<Caip10Link> {
     state: StreamState
   ): Promise<StreamState> {
     // TODO: Assert that the 'prev' of the commit being applied is the end of the log in 'state'
-    const proof = (await context.ipfs.dag.get(commit.proof, { timeout: IPFS_GET_TIMEOUT })).value
+    const proof = StreamUtils.isExpandedCommit(commit)
+      ? commit.proof
+      : (await context.ipfs.dag.get(commit.proof, { timeout: IPFS_GET_TIMEOUT })).value
```
stbrody, line 162: We should go through the dispatcher rather than accessing IPFS directly

packages/stream-caip10-link-handler/src/caip10-link-handler.ts
```diff
@@ -153,7 +157,9 @@ export class Caip10LinkHandler implements StreamHandler<Caip10Link> {
     state: StreamState
   ): Promise<StreamState> {
     // TODO: Assert that the 'prev' of the commit being applied is the end of the log in 'state'
-    const proof = (await context.ipfs.dag.get(commit.proof, { timeout: IPFS_GET_TIMEOUT })).value
+    const proof = StreamUtils.isExpandedCommit(commit)
```
stbrody, line 160: Can this ever be false?  Would be nice not to have to have the fallback logic to loading the proof here.  It makes it hard to know what the actual control flow is going to be

stbrody: same question for the above checks for signed commits.  Can we just have a single type that is used throughout the StreamtypeHandlers?

packages/stream-caip10-link-handler/src/caip10-link-handler.ts
```diff
@@ -92,6 +95,7 @@ export class Caip10LinkHandler implements StreamHandler<Caip10Link> {
    * @private
    */
   async _applySigned(commit: any, cid: CID, state: StreamState): Promise<StreamState> {
```
stbrody, line 97: can we give `commit` a proper type now instead of `any`?

packages/stream-tile-handler/src/tile-document-handler.ts
```diff
@@ -68,11 +69,16 @@ export class TileDocumentHandler implements StreamHandler<TileDocument> {
    * @private
    */
   async _applyGenesis(commit: any, meta: CommitMeta, context: Context): Promise<StreamState> {
-    let payload = commit
-    const isSigned = StreamUtils.isSignedCommit(commit)
+    let payload = StreamUtils.isExpandedCommit(commit) ? commit.commit : commit
+    const isSigned = StreamUtils.isSignedCommit(commit) || (commit as CommitData).envelope
```
stbrody, line 73: if `StreamUtils.isSignedCommit` is no longer sufficient to properly answer this question then we should update it.  Shouldn't have to do an extra check after calling the function whose sole purpose is answering exactly this question

packages/stream-tile-handler/src/tile-document-handler.ts
```diff
@@ -68,11 +69,16 @@ export class TileDocumentHandler implements StreamHandler<TileDocument> {
...
     if (isSigned) {
-      payload = (await context.ipfs.dag.get(commit.link, { timeout: IPFS_GET_TIMEOUT })).value
-      await this._verifySignature(commit, meta, context, payload.header.controllers[0])
+      let envelope = commit
+      if (StreamUtils.isExpandedCommit(commit)) {
```
stbrody, line 76: I'm really concerned about all these conditional checks with different behavior depending on whether or not this information was already loaded.  Ideally we would make it so all code paths that lead here do the work to load the inner data, so we can always assume we have an expanded CommitData throughout all the StreamtypeHandlers.

If that really isn't possible today, then we should do this check once, up front at the very beginning of commit application (presumably in `applyCommit`, before calling into the commit-type specific helper functions), and expand it to included the loaded commit data so that all later code can assume it is already done.

packages/stream-tile-handler/src/tile-document-handler.ts
```diff
@@ -144,12 +157,14 @@ export class TileDocumentHandler implements StreamHandler<TileDocument> {
    */
   async _applyAnchor(
     context: Context,
-    commit: AnchorCommit,
+    commit: any,
```
stbrody, line 160: woah, moving in the wrong direction here.  More type data, not less please

stbrody (changes requested):

packages/stream-caip10-link-handler/src/caip10-link-handler.ts
stbrody: yeah, I think it would be fine to put the Dispatcher into the Context.  Ideally we'd get rid of the IPFS handle out of Context at the same time, but don't worry about that if it's used in too many places

packages/stream-tile-handler/src/tile-document-handler.ts
smrz2001: Agreed on both points.

The reason I didn't do the latter is because we don't always need to do all the same lookups - anchor commit handling, for example, needs to lookup the `proof` while signed commit handling looks up the commit and envelope. I could add this logic to `applyCommit` but then I'd be checking for commit type, doing specific lookups, then passing data into specific handlers also under commit type checks? The code structure looked weird to me.

The former would be ideal but we use `applyCommit` all over the place and while I can look at all places, this would not be a surgical PR anymore.

packages/stream-tile-handler/src/tile-document-handler.ts
smrz2001: Yes, agreed, but I'd have to go to less type data regardless: `CeramicData | CommitData`.

To do `AnchorCommit | CommitData` I would have to check whether this was an anchor commit in `applyCommit` then `as AnchorCommit` it but pass it through unchanged if it was `CommitData`, which made me cringe to think of implementing.

packages/core/src/__tests__/state-manager.test.ts
stbrody: got it, makes sense

packages/core/src/conflict-resolution.ts
smrz2001: Removed `CommitMeta`.

packages/common/src/utils/stream-utils.ts
```diff
@@ -221,4 +221,20 @@ export class StreamUtils {
   static isAnchorCommit(commit: CeramicCommit): commit is AnchorCommit {
     return commit && (commit as AnchorCommit).proof !== undefined
   }
+
```
smrz2001, line 224: Couldn't get rid of the 3 functions above and switch entirely to `CommitData` :( - open to suggestions.

packages/common/src/stream.ts
```diff
@@ -122,16 +110,21 @@ export enum CommitType {
 
 export interface LogEntry {
   cid: CID
-  type: CommitType
+  type?: CommitType // marked optional for the times when only the CID is available for the commit
```
smrz2001, line 113: Comment from Sergey:
For future-safety, should not loosen this interface.

packages/common/src/stream.ts
```diff
@@ -122,16 +110,21 @@ export enum CommitType {
...
   timestamp?: number
 }
 
+/**
+ * Includes additional fields that significantly reduce the number of IPFS lookups required while processing commits
+ */
 export interface CommitData extends LogEntry {
```
smrz2001, line 120: From Sergey:
Is `CommitData` ever used as an extension of `LogEntry`?

packages/stream-tile-handler/src/tile-document-handler.ts
```diff
@@ -100,29 +91,31 @@ export class TileDocumentHandler implements StreamHandler<TileDocument> {
...
+    commitData: CommitData,
     state: StreamState,
     context: Context
   ): Promise<StreamState> {
     // TODO: Assert that the 'prev' of the commit being applied is the end of the log in 'state'
     const controller = state.next?.metadata?.controllers?.[0] || state.metadata.controllers[0]
-    await this._verifySignature(commit, meta, context, controller)
 
-    const payload = (await context.ipfs.dag.get(commit.link, { timeout: IPFS_GET_TIMEOUT })).value
+    // Verify the JWS before performing any additional IPFS lookups
```
smrz2001, line 113: Update comment

packages/core/src/state-management/state-manager.ts
```diff
@@ -245,7 +245,7 @@ export class StateManager {
       try {
         await this.executionQ.forStream(state$.id).run(async () => {
           await this._handleTip(state$, anchorCommit)
-          if (state$.tip == anchorCommit) {
+          if (state$.tip.equals(anchorCommit)) {
```
smrz2001, line 248: The previous code was preventing a _published_ anchor commit from being applied in the failing tests.

stbrody (changes requested):
overall this is looking great.  I'm very happy to have killed off CommitMeta and merged it with CommitData.

My number 1 remaining concern is that we can't rely on the type system right now to know whether the underlying commit data has been loaded or not.  Whenever you get a CommitData argument, it might or might not have the underlying commit loaded into it.  Some places in the code (such as the streamtype handlers) now assume that what it's given has the underlying commit data in it, while other places that receive CommitData arguments does not, and calls `getCommitData` to get it even though the argument was already a CommitData.

I think if we make the relevant CommitData fields (`commit` and `type` at a minimum) required, then we can trust that any place we are given CommitData we actually have that data present, and then if there are places that sometimes have that data present and sometimes not, their type signature could use `LogEntry | CommitData` to note that.

packages/stream-caip10-link-handler/src/caip10-link-handler.ts
```diff
@@ -141,21 +135,16 @@ export class Caip10LinkHandler implements StreamHandler<Caip10Link> {
...
    * @private
    */
   async _applyAnchor(
     context: Context,
-    commit: any,
-    cid: CID,
+    commitData: CommitData,
     state: StreamState
   ): Promise<StreamState> {
-    // TODO: Assert that the 'prev' of the commit being applied is the end of the log in 'state'
```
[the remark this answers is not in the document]
smrz2001: Whoops, reverting.

packages/core/src/conflict-resolution.ts
```diff
@@ -253,18 +240,14 @@ export class ConflictResolution {
    */
   private async applyLogEntryToState<T extends Stream>(
     entry: CommitData, state: StreamState, handler: StreamHandler<T>): Promise<StreamState> {
-    const logEntry = await this.getCommitData(entry)
-    const commitMeta = { cid: logEntry.cid, timestamp: logEntry.timestamp }
-    if (StreamUtils.isAnchorCommit(logEntry.commit)) {
+    const logEntry = await Utils.getCommitData(entry, this.dispatcher)
```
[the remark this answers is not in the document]
smrz2001: Hmm, 

> overall this is looking great. I'm very happy to have killed off CommitMeta and merged it with CommitData.
> 
> My number 1 remaining concern is that we can't rely on the type system right now to know whether the underlying commit data has been loaded or not. Whenever you get a CommitData argument, it might or might not have the underlying commit loaded into it. Some places in the code (such as the streamtype handlers) now assume that what it's given has the underlying commit data in it, while other places that receive CommitData arguments does not, and calls `getCommitData` to get it even though the argument was already a CommitData.
> 
> I think if we make the relevant CommitData fields (`commit` and `type` at a minimum) required, then we can trust that any place we are given CommitData we actually have that data present, and then if there are places that sometimes have that data present and sometimes not, their type signature could use `LogEntry | CommitData` to note that.

Good thoughts, thanks for thinking about this @stbrody.

The problem with making `type` required is that in places like `ConflictResolution.fetchLog()` we don't know the `type` because we don't have the `commit` yet.

This is going to be a general problem - we need the `commit` to determine the `type` and the only way to make both required would be to fetch the `commit` before the call to `getCommitData()`, which would require leaving an IPFS lookup outside of `Utils.getCommitData()`.

`in-memory-anchor-service.ts` had the same problem of having just the `cid` but not `type` or `commit` but I was able to revert the file to use the old approach to fetch the `commit` separately and determine the `type`. Since it is a test file, this should be ok, although I don't like leaving any IPFS lookups lingering outside of `Utils.getCommitData()` if it can be helped.

Similarly, in `Repository.fromNetwork()`, we have the `cid` and the `type` (genesis) but not the `commit`. Retrieving the `commit` before constructing the `CommitData` object would defeat the purpose of collapsing all lookups inside `Utils.getCommitData()`.

Alternatively, we could made `Utils.getCommitData()` accept `LogEntry | CommitData` but then it would be confusing to reliably determine what information was already available and what needed to be looked up.

`commit` being optional in `commitData` allows us to reliably determine the latter is incomplete and that the `commit`, `envelope`, `proof`, `timestamp` need to be looked up.

smrz2001:
> overall this is looking great. I'm very happy to have killed off CommitMeta and merged it with CommitData.
> 
> My number 1 remaining concern is that we can't rely on the type system right now to know whether the underlying commit data has been loaded or not. Whenever you get a CommitData argument, it might or might not have the underlying commit loaded into it. Some places in the code (such as the streamtype handlers) now assume that what it's given has the underlying commit data in it, while other places that receive CommitData arguments does not, and calls `getCommitData` to get it even though the argument was already a CommitData.
> 
> I think if we make the relevant CommitData fields (`commit` and `type` at a minimum) required, then we can trust that any place we are given CommitData we actually have that data present, and then if there are places that sometimes have that data present and sometimes not, their type signature could use `LogEntry | CommitData` to note that.

Good thoughts, thanks for thinking about this @stbrody.

The problem with making `type` required is that in places like `ConflictResolution.fetchLog()` we don't know the `type` because we don't have the `commit` yet.

This is going to be a general problem - we need the `commit` to determine the `type` and the only way to make both required would be to fetch the `commit` before the call to `getCommitData()`, which would require leaving an IPFS lookup outside of `Utils.getCommitData()`.

`in-memory-anchor-service.ts` had the same problem of having just the `cid` but not `type` or `commit` but I was able to revert the file to use the old approach to fetch the `commit` separately and determine the `type`. Since it is a test file, this should be ok, although I don't like leaving any IPFS lookups lingering outside of `Utils.getCommitData()` if it can be helped.

Similarly, in `Repository.fromNetwork()`, we have the `cid` and the `type` (genesis) but not the `commit`. Retrieving the `commit` before constructing the `CommitData` object would defeat the purpose of collapsing all lookups inside `Utils.getCommitData()`.

Alternatively, we could made `Utils.getCommitData()` accept `LogEntry | CommitData` but then it would be confusing to reliably determine what information was already available and what needed to be looked up.

`commit` being optional in `commitData` allows us to reliably determine the latter is incomplete and that the `commit`, `envelope`, `proof`, `timestamp` need to be looked up.

packages/core/src/__tests__/state-manager.test.ts
smrz2001: Yes, the count is 4 because of `verifyLoneGenesis()`.

Is the TODO still required or just a note that it would normally be 2 but this special case causes it to be 4? The count is now not more than it would be expected to be for a typical, non-genesis case.

stbrody:
> Alternatively, we could made Utils.getCommitData() accept LogEntry | CommitData but then it would be confusing to reliably determine what information was already available and what needed to be looked up.

Yes, this is what I was imagining. I don't see why that's more confusing?  The goal would be that any time we might not have the commit fetched we accept LogEntry (possibly with an `| CommitData`), and then anywhere we have explicitly a CommitData we know we have everything we need

packages/core/src/__tests__/state-manager.test.ts
stbrody: I think we still expect that this can and should be 2, no?  There's no intrinsic reason why applying a new commit to existing state should require re-loading the genesis commit.

In fact, I'm *still* confused why this is happening.  I just re-read [verifyLoneGenesis](https://github.com/ceramicnetwork/js-ceramic/blob/e50898ffc87e6f556bb55ba2179b67c875abb72a/packages/core/src/state-management/state-manager.ts#L97) and it looks like it only does anything if the state only has a single commit.  So I'm still confused why applying the second commit to a stream is resulting in the genesis commit being reloaded.

smrz2001: Sorry, I misspoke - the extra call is not from `verifyLoneGenesis`, it's from [applyLogToState](https://github.com/ceramicnetwork/js-ceramic/pull/1683/files#diff-efa0d2e3a8268f3351f1c2bd6a47b92f790f9286a890052ffe695890b4e0d332R276) instead.

This (i.e. an extra lookup) should only happen when there is just a genesis commit in the stream prior to the commit being applied:
```
    if (state && state.log.length === 1) {
```

smrz2001:
> > Alternatively, we could made Utils.getCommitData() accept LogEntry | CommitData but then it would be confusing to reliably determine what information was already available and what needed to be looked up.
> 
> Yes, this is what I was imagining. I don't see why that's more confusing? The goal would be that any time we might not have the commit fetched we accept LogEntry (possibly with an `| CommitData`), and then anywhere we have explicitly a CommitData we know we have everything we need

I thought it would be more confusing because now `getCommitData` would need to handle both types separately - cast the passed structure into `CommitData` to see if it had the `commit` and look up the commit (and determine the `type`) but if the type was `LogEntry`, it would need to always lookup the commit. and assume the type was always available and specified?

Also, in `fetchLog`, we don't even have the `type` and so `type` can't be a required field regardless.

packages/core/src/utils.ts
```diff
@@ -81,34 +81,23 @@ export default class Utils {
    * Return expanded `CommitData` with commit and JWS envelope or anchor proof/timestamp, if applicable and not already
```
stbrody, line 81: update comment

stbrody (approved):
LGTM!
Thank you for all your patience working and reworking this.  I feel really good about where it ended up and think it will make it much easier to write correct and performant code going forward.  Great work!

PR #1688, opened 2021-08-30
feat(common,stream-tile,stream-tile-handler): Add metadata option to forbid controller changes

stbrody:


packages/stream-tile/src/tile-document.ts
```diff
@@ -52,6 +53,16 @@ export interface TileMetadataArgs {
    * identical documents will generate unique StreamIDs and be able to be updated independently.
    */
   deterministic?: boolean
+
+  /**
+   * If true, all changes to the 'controllers' array are disallowed.  This guarantees that the
+   * Stream will always have the same controller. Especially useful for Streams controlled by
```
stbrody, line 59: I'm not sure if the part about did:nft is too much detail to go into here or not.  Could just end the comment after the second sentence if we think this is too specific for the doccomments

packages/core/src/__tests__/__snapshots__/caip10-link.test.ts.snap
```diff
@@ -274,6 +274,7 @@ Object {
       "0x0544dCF4Fce959C6C4F3b7530190cb5E1BD67cB1@eip155:1",
     ],
     "family": "caip10-eip155:1",
+    "forbidControllerChange": true,
```
oed, line 277: This should not be here. It will break all existing caip10-links. This doctype doesn't allow controller change anyway,

packages/stream-caip10-link/src/caip10-link.ts
```diff
@@ -154,6 +154,7 @@ export class Caip10Link extends Stream {
       header: {
         controllers: [accountId.toString()],
         family: `caip10-${accountId.chainId.toString()}`,
+        forbidControllerChange: true,
```
oed, line 157: Revert.

packages/stream-tile/src/tile-document.ts
oed: I think an example is fine.

packages/stream-tile/src/tile-document.ts
```diff
@@ -52,6 +53,16 @@ export interface TileMetadataArgs {
...
+   * DIDs that can have ownership changes within the DID itself, such as did:nft, as setting this
+   * would prevent the current holder of the NFT from changing the Stream's controller to that
+   * user's personal DID, which would prevent the ownership of the Stream from changing the next
+   * time the NFT is traded.
+   */
+  forbidControllerChange?: boolean
```
oed, line 65: Could we call it something more concise, like `nonTransferable`?

oed (changes requested):

packages/stream-tile/src/tile-document.ts
stbrody: I'd be okay with calling it `nonTransferable` in the commit header, but I think the user-facing argument should be more descriptive.  It doesn't seem to me like it would be immediately clear to users what `nonTransferable` actually means

stbrody: ugh, actually having the argument have a different name that the value that shows up in the state object is likely to just cause more confusion I think.  Probably makes sense to keep the names the same wherever possible.

Given that, I still prefer `forbidControllerChange` to `nonTransferable`.  For one thing, if the controller is a did:nft, the stream still *is* transferable, even with this property set, which seems confusing.

I'm not thrilled with the name `forbidControllerChange`, but it's the best I've been able to come up with, and at least it's pretty descriptive of what it actually does

packages/core/src/__tests__/__snapshots__/caip10-link.test.ts.snap
stbrody: Good call, fixed

ukstv (approved):
Spent some time making sense of how the new flag affects Caip10Link. Was surprised to find that the flag does not affect it in any way. It is good, it was removed for Caip10Links.

🏇

oed (approved):

packages/stream-tile/src/tile-document.ts
```diff
@@ -78,6 +93,27 @@ function headerFromMetadata(metadata?: TileMetadataArgs | StreamMetadata): Commi
...
+    }
+    if (metadata?.forbidControllerChange) {
+      header.forbidControllerChange = true
+    }
+  } else {
+    // These throws aren't strictly necessary as we can just leave these fields out of the header
+    // object we return, but throwing here gives more useful feedback to users.
+
+    if (metadata?.deterministic !== undefined || (metadata as any)?.unique !== undefined) {
+      throw new Error("Cannot change 'deterministic' or 'unique' properties on existing Streams")
```
gertig, line 109: @stbrody I'm new to Ceramic, working through the basics, trying to do the following 

`await idx.set("basicProfile", { name: "testing " });` 

but am getting the above error when making this call. Have attempted to understand what could be causing this issue but so far no luck. As a new Ceramic user this error wasn't clear in what it means by deterministic and unique, perhaps that means something more specific to an experienced user?

gertig: My issue was resolved by updating to `@ceramicstudio/idx` version `0.12.2`

https://github.com/ceramicnetwork/js-ceramic/issues/1714

PR #1690, opened 2021-08-31
chore(core): Prevent using custom CAS on mainnet

stbrody:
Until we have the trusted set of anchor services project complete, we don't want anyone using any anchor service other than our official one


v-stickykeys (approved):
LGTM

stbrody:
>  Not sure why we need this ?

To make sure people don't run their own anchor services until we're ready to open the network up to more anchor services.  This is important for the "trusted set of anchor services" project we need to close the data withholding attack

PR #1691, opened 2021-09-01
chore: Bump ethers to the latest version

ukstv:
Required for 1559 support.

Ceramic uses old ethers packages. CAS uses some Ceramic packages. Due to hoisting, CAS can not really use new ethers packages in presence of the old ethers required by Ceramic packages. So, here we are, updating ethers in Ceramic.


stbrody (approved):
LGTM

PR #1693, opened 2021-09-01
feat(core): Allow pinning/unpinning alongside CRUD operations in a single request

stbrody:


packages/core/src/__tests__/ceramic-pinning.test.ts
```diff
@@ -52,10 +53,19 @@ async function createStream(
...
 }
 
+async function isPinned(ceramic: CeramicApi, streamId: StreamID): Promise<boolean> {
+  const iterator = await ceramic.pin.ls(streamId)
+  const arr = []
+  for await (const id of iterator) {
+    arr.push(id)
+  }
+  return arr.length == 1 && arr[0] == streamId.toString()
+}
```
oed, line 67: Maybe just use `it-first`? 
https://github.com/achingbrain/it

packages/core/src/state-management/repository.ts
```diff
@@ -199,6 +205,44 @@ export class Repository {
...
+   */
+  async applyWriteOpts(state$: RunningState, opts: CreateOpts | UpdateOpts) {
+    this.stateManager.applyWriteOpts(state$, opts)
+
+    await this.handlePinOpts(state$, opts)
+  }
+
+  async handlePinOpts(state$: RunningState, opts: CommonOpts) {
+    if (opts.pin) {
+      await this.pin(state$)
```
oed, line 240: Does this only pin if it's not already pinned? Otherwise it could slow things down?

oed (commented):
Looks good, just a question.

packages/core/src/state-management/repository.ts
stbrody: It behaves exactly the same as pinning an already-pinned stream does via the public pin API (minus the step of loading the stream state since we already have it here).  So yes, this probably does slow things down if you set it to `true` for a stream that is already pinned.  But that's a bigger problem with our pinning system in general that probably needs to be optimized for all pin requests, regardless of which API they come in from.

packages/core/src/__tests__/ceramic-pinning.test.ts
stbrody: done

packages/core/src/state-management/repository.ts
oed: Right, I asked because I was unsure of how our existing pin api is set up. I remember that there were some things that could be optimized. Maybe we should capture that somewhere? Seems like there's possibly a large performance gain there iirc?

oed (approved):

packages/core/src/state-management/repository.ts
stbrody: Yep good call.  Made https://www.notion.so/threebox/2785876b2d984540a9eeb4832ee0ca81?v=ff8566cdd0d548a299cf0ee24b41527a&p=3d888998788441ea90b826c4612b5937

PR #1704, opened 2021-09-07
chore(ipfs-daemon): outline potential solution for role based S3 access

0xm1kr:
This PR is to open a discussion regarding the use of role based access over `accessKeyId` and `secretAccessKey`. During our node deployment we encountered the issue that if we do not explicitly set an `accessKeyId` and `secretAccessKey` then the `create-repo` will throw the `Expect AWS credentials` error. This shouldn't really be the case because many deployments will have a role attached to the instance which contains the proper security policies, in fact, this is preferable to passing in credentials over environment variables from a security perspective.

I believe this code is close, however, it hasn't been fully tested and it needs refactoring to convert the `createRepo` function into an async function.

Let me know what you think and we can work toward a full implementation. It may have some large downstream impacts so I didn't want to go through the effort of a full solution if it doesn't make sense or if there is a better solution.

For the time being we simply created a new user/role (`ceramic-state-s3-access`) to generate keys specifically for this purpose and pass them in with custom environment variables which we manually pass in to the configuration object.

### DIFF
--- a/packages/ipfs-daemon/src/create-repo.ts
+++ b/packages/ipfs-daemon/src/create-repo.ts
@@ -1,4 +1,5 @@
 import S3 from 'aws-sdk/clients/s3'
+import AWS from 'aws-sdk'
 import IPFSRepo from 'ipfs-repo'
 import DatastoreLevel from 'datastore-level'
 import DatastoreFS from 'datastore-fs'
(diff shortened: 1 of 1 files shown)


stbrody:
this looks directionally reasonable to me, though I'm not super familiar with the AWS APIs.  Adding @v-stickykeys and @smrz2001 to take a look

v-stickykeys (commented):
Yes this looks correct to me but want to ensure I understand how it works. So with this code the AWS sdk grabs creds from the normal places? ~/.aws and env vars?

mikehroth:
> Yes this looks correct to me but want to ensure I understand how it works. So with this code the AWS sdk grabs creds from the normal places? ~/.aws and env vars?

Correct, I think there are actually six different ways it can obtain credentials.
See: https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/setting-credentials-node.html

For our purposes we are assigning a role to our ECS Fargate containers:
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html#attach-iam-role

ukstv:
@v-stickykeys @smrz2001 Any opinion on what to do about the PR?

smrz2001:
I'll take a look @ukstv @stbrody, we can potentially merge it in this week, I think. Looks fairly straightforward, just need to test it a bit.

ukstv:
How are things going @smrz2001 ?

smrz2001:
> How are things going @smrz2001 ?

Hehe, yeah, @ukstv, this one's been pending for a long time. Will take care of it this/next cycle.

stbrody:
js-ceramic is now deprecated: https://blog.ceramic.network/the-future-of-ceramic-focusing-on-recall/

PR #1705, opened 2021-09-07
chore(common): Break down DocOpts to be unions of single-purpose opts interfaces

stbrody:
In my next PR I need access to an `opts` object with just the `publish` field, without the `anchor` field which is currently lumped together with it in `BasicUpdateOpts`.  I feel like at this point it probably makes sense to just have a dedicated `Opts` interface for each field and then compose them into higher-level groupings where it makes sense, so I went ahead and moved them all in that direction


smrz2001 (approved):
🚢

PR #1718, opened 2021-09-13
fix(stream-caip10-link): use lowercase in caip10-link genesis

oed:
Using eip55 turned out to be a mistake that causes a lot of pain in 3ID Connect. Futhermore, since caip10 1.0 will use a new format it doesn't really make sense to try to enforce eip55 at the genesis layer. 

NOTE: this PR is against master!


ukstv (approved):

packages/stream-caip10-link/src/caip10-link.ts
```diff
@@ -148,7 +147,7 @@ export class Caip10Link extends Stream {
     // sometimes all lower case. In order to deal with this and not have different
     // links for different addresses we encode the address using eip55.
```
stbrody, line 148:
```suggestion
    // links for different addresses we convert the address to all lower case.
```

stbrody (approved):
so I guess any links created during the window where this change was in place will now be invalidated?

oed:
> so I guess any links created during the window where this change was in place will now be invalidated?

Yes, but all caip10-links created by 3ID Connect will start working again.

packages/stream-caip10-link/src/caip10-link.ts
```diff
@@ -146,9 +145,9 @@ export class Caip10Link extends Stream {
   static makeGenesis(accountId: AccountID): GenesisCommit {
     // Ethereum addresses specifically are sometimes encoded with mixed case and
     // sometimes all lower case. In order to deal with this and not have different
-    // links for different addresses we encode the address using eip55.
+    // links for different addresses we using the address using lowercase.
```
stbrody, line 148:
```suggestion
    // links for different addresses we convert the address to lowercase.
```

stbrody: "we using the address using lowercase" <- broken grammar

oed: lol, I'm tired

PR #1726, opened 2021-09-16
chore: Remove trailing slash from anchor url

ukstv:


packages/core/src/ceramic.ts
```diff
@@ -377,7 +377,7 @@ export class Ceramic implements CeramicApi {
     let anchorService = null
     if (!config.gateway) {
       const anchorServiceUrl =
-        config.anchorServiceUrl || DEFAULT_ANCHOR_SERVICE_URLS[networkOptions.name]
+        config.anchorServiceUrl.replace(/\/$/, '') || DEFAULT_ANCHOR_SERVICE_URLS[networkOptions.name]
```
stbrody, line 380: can you add a comment what the replace call is doing?  I find regexes incredibly opaque and difficult to read

v-stickykeys: Would be even better to store the regex in a variable with a helpful name. So no need for comment

ukstv: @stbrody @v-stickykeys Extracted variable.

stbrody (approved):
lgtm, thanks!

PR #1736, opened 2021-09-21
Support optional genesis property in multiquery API

haardikk21:
This PR is meant to address issue https://github.com/ceramicnetwork/js-ceramic/issues/1647

### What was done
- Introduced an optional `genesis?: Record<string, any>` property in the `MultiQuery` interface 
- The `multiquery` function in Ceramic core calls `_loadLinkedStreams` and passes forward the `genesis` property
- `_loadLinkedStreams` calls `loadStream` iteratively, and passes `query.genesis` as part of the `LoadOpts` interface
- `LoadOpts` now has two additional optional properties - `genesis?: Record<string, any>` and `streamFromGenesis?: Stream`
- `loadStream` checks if the `opts.genesis` property is present, and if so, sets `opts.streamFromGenesis` to the returned result from `createStreamFromGenesis`
- `loadStream` proceeds to call either `repository.loadAtCommit`, `repository.loadAtTime` or `repository.load`, the former 2 of which call `repository.load` internally
- `repository.load` checks that if the stream could not be loaded from memory or the store, and the `genesis` and `streamFromGenesis` options are present. If so, it syncs the stream from pubsub and returns it after calling `verifyLoneGenesis` on it

### Expectations
This PR is expected to help with the loading times when the genesis content of a stream is available client side, but the server does not have it in memory or store as querying from pubsub is faster than the worst case `3 * 30s = 90s` timeout set on loading from IPFS.

### Issues
In `loadStream` at line 608 (https://github.com/ceramicnetwork/js-ceramic/blob/05c9b3cd13c92bbd2cdcd7e2f9298912814a7919/packages/core/src/ceramic.ts#L608) I check that the provided StreamID is equal to the created StreamID of the stream created from provided genesis content. 

I am unsure if the following line is proper:
`if (!opts.streamFromGenesis.id.equals(streamRef.baseID))` because I'm not sure if `streamRef.baseID` is the same thing as the StreamID or more of an initial StreamID that changes with updates.

---

### Notes
- This is my first PR to the `js-ceramic` repo so I would appreciate any and all feedback and thoughts. 
- I've cc'd the Protocol team, but not sure if anyone from Platform or Glaze also needs to review this PR before merge

### DIFF
--- a/packages/common/src/ceramic-api.ts
+++ b/packages/common/src/ceramic-api.ts
@@ -136,6 +136,10 @@ export interface CeramicApi extends CeramicSigner {
 
 export interface MultiQuery {
   /**
+   * The genesis content for the queried stream
+   */
+  genesis?: Record<string, any>
+  /**
    * The StreamID of the stream to load
    */
   streamId: CommitID | StreamID | string
--- a/packages/common/src/streamopts.ts
+++ b/packages/common/src/streamopts.ts
@@ -1,3 +1,4 @@
+import { Stream } from './stream'
 /**
  * Options that are related to pinning streams.
  */
@@ -68,6 +69,14 @@ export interface LoadOpts extends SyncOpts, PinningOpts {
    * Load a previous version of the stream based on unix timestamp
    */
   atTime?: number
+  /**
+   * Provide the genesis content so it does not need to fetched from IPFS daemon
+   */
+  genesis?: Record<string, any>
+  /**
+   * The created stream from the genesis content. Only present is genesis is present.
+   */
+  streamFromGenesis?: Stream
 }
 
 /**
--- a/packages/core/src/ceramic.ts
+++ b/packages/core/src/ceramic.ts
@@ -381,8 +381,8 @@ export class Ceramic implements CeramicApi {
 
       if (
         (networkOptions.name == Networks.MAINNET || networkOptions.name == Networks.ELP) &&
-        (anchorServiceUrl !== "https://cas-internal.3boxlabs.com") &&
-        (anchorServiceUrl !== DEFAULT_ANCHOR_SERVICE_URLS[networkOptions.name])
+        anchorServiceUrl !== 'https://cas-internal.3boxlabs.com' &&
+        anchorServiceUrl !== DEFAULT_ANCHOR_SERVICE_URLS[networkOptions.name]
       ) {
         throw new Error('Cannot use custom anchor service on Ceramic mainnet')
       }
@@ -599,6 +599,22 @@ export class Ceramic implements CeramicApi {
   ): Promise<T> {
     opts = { ...DEFAULT_LOAD_OPTS, ...opts, ...this._loadOptsOverride }
     const streamRef = StreamRef.from(streamId)
+    if (opts.genesis) {
+      opts.streamFromGenesis = await this.createStreamFromGenesis(
+        streamRef.type,
+        opts.genesis,
+        opts
+      )
+      // Is using streamRef.baseID the correct comparison?
+      if (!opts.streamFromGenesis.id.equals(streamRef.baseID)) {
+        throw new Error(
+          `Created StreamID ${opts.streamFromGenesis.id.toString()} does not equal provided StreamID ${
+            streamRef.baseID
+          }`
+        )
+      }
+    }
+
     if (CommitID.isInstance(streamRef)) {
       const snapshot$ = await this.repository.loadAtCommit(streamRef, opts)
       return streamFromState<T>(this.context, this._streamHandlers, snapshot$.value)
@@ -632,7 +648,10 @@ export class Ceramic implements CeramicApi {
     const walkNext = async (node: TrieNode, streamId: StreamID | CommitID) => {
       let stream
       try {
-        stream = await promiseTimeout(timeout, this.loadStream(streamId, { atTime: query.atTime }))
+        stream = await promiseTimeout(
+          timeout,
+          this.loadStream(streamId, { atTime: query.atTime, genesis: query.genesis })
+        )
       } catch (e) {
         return Promise.resolve()
       }
--- a/packages/core/src/state-management/repository.ts
+++ b/packages/core/src/state-management/repository.ts
@@ -165,6 +165,12 @@ export class Repository {
         }
       }
 
+      if (!stream && opts.genesis && opts.streamFromGenesis) {
+        stream = new RunningState(opts.streamFromGenesis.state)
+        await this.stateManager.sync(stream, opts.syncTimeoutSeconds * 1000, fromStateStore)
+        return this.stateManager.verifyLoneGenesis(stream)
+      }
+
       if (!stream) {
         stream = await this.fromNetwork(streamId)
       }


packages/common/src/streamopts.ts
```diff
@@ -68,6 +69,14 @@ export interface LoadOpts extends SyncOpts, PinningOpts {
    * Load a previous version of the stream based on unix timestamp
    */
   atTime?: number
+  /**
+   * Provide the genesis content so it does not need to fetched from IPFS daemon
+   */
+  genesis?: Record<string, any>
```
stbrody, line 75: you shouldn't need to make any changes to LoadOpts.  Any fields added to LoadOpts would need to be supported in the primary loadStream API, and be able to be serialized for use in the http client

packages/core/src/ceramic.ts
```diff
@@ -632,7 +648,10 @@ export class Ceramic implements CeramicApi {
     const walkNext = async (node: TrieNode, streamId: StreamID | CommitID) => {
```
stbrody, line 648: I think all you need to do is put the genesis contents into IPFS before the call to walkNext

stbrody (changes requested):
This approach is overcomplicated and introduces a bunch of fields into the public API (in LoadOpts) that only exist for internal consumption as implementation details.

I think the core change can actually be much smaller and simpler - you can check the query for a `genesis` field at the beginning of `_loadLinkedStreams` and if it's present, put the data into IPFS, and make sure the resulting CID matches the CID from the StreamID in the query. Once that's done, you should be able to proceed with the normal process for loading the stream since you can be confident that the genesis commit exists so you won't hit an IPFS timeout.

Also, this needs tests.  This won't work through the HTTP client right now (you need to update the multiquery function in the http-client to serialize the `genesis` property and send it to the daemon), which proper testing would have caught.  You probably want more extensive testing in [ceramic-api.test.ts](https://github.com/ceramicnetwork/js-ceramic/blob/a608e2ae8c4c3ee85a9cfa60a7fa240063a2c35c/packages/core/src/__tests__/ceramic-api.test.ts#L256) and some basic testing in [ceramic-daemon.test.ts](https://github.com/ceramicnetwork/js-ceramic/blob/a608e2ae8c4c3ee85a9cfa60a7fa240063a2c35c/packages/cli/src/__tests__/ceramic-daemon.test.ts#L356) to make sure it works over the http client.

packages/core/src/ceramic.ts
```diff
@@ -599,6 +599,22 @@ export class Ceramic implements CeramicApi {
...
+      // Is using streamRef.baseID the correct comparison?
+      if (!opts.streamFromGenesis.id.equals(streamRef.baseID)) {
+        throw new Error(
+          `Created StreamID ${opts.streamFromGenesis.id.toString()} does not equal provided StreamID ${
+            streamRef.baseID
+          }`
+        )
+      }
+    }
+
```
oed, line 617:
```suggestion

```

packages/core/src/ceramic.ts
```diff
@@ -632,7 +648,10 @@ export class Ceramic implements CeramicApi {
     const walkNext = async (node: TrieNode, streamId: StreamID | CommitID) => {
       let stream
       try {
-        stream = await promiseTimeout(timeout, this.loadStream(streamId, { atTime: query.atTime }))
+        stream = await promiseTimeout(
+          timeout,
+          this.loadStream(streamId, { atTime: query.atTime, genesis: query.genesis })
+        )
```
oed, line 654:
```suggestion
        if (query.genesis) {
          const genesisCid = await this.ipfs.dag.put(query.genesis)
          // TODO - check that streamId.id.cid.equals(genesisCid), error if not!
        }
        stream = await promiseTimeout(timeout, this.loadStream(streamId, { atTime: query.atTime }))
```

packages/core/src/state-management/repository.ts
```diff
@@ -165,6 +165,12 @@ export class Repository {
         }
       }
 
+      if (!stream && opts.genesis && opts.streamFromGenesis) {
+        stream = new RunningState(opts.streamFromGenesis.state)
+        await this.stateManager.sync(stream, opts.syncTimeoutSeconds * 1000, fromStateStore)
+        return this.stateManager.verifyLoneGenesis(stream)
+      }
+
```
oed, line 173:
```suggestion
```

packages/common/src/streamopts.ts
```diff
@@ -68,6 +69,14 @@ export interface LoadOpts extends SyncOpts, PinningOpts {
...
+  /**
+   * The created stream from the genesis content. Only present is genesis is present.
+   */
+  streamFromGenesis?: Stream
```
oed, line 79:
```suggestion
```

oed (changes requested):
Suggested edits from pair programming.

packages/core/src/__tests__/ceramic.test.ts
```diff
@@ -439,6 +442,94 @@ describe('Ceramic integration', () => {
...
+        tags: ['x', 'y'],
+      }
+
+      // Create a deterministic TileDocument
+      const stream1 = await TileDocument.create(
+        ceramic1,
+        content,
+        { ...metadata, deterministic: true },
+        { anchor: false, publish: false }
+      )
```
oed, line 464: Can we add a test that does the same test but without `TileDocument.create`? We can use `StreamID.fromGenesis(genesisCommit)` to get the streamId for the multiquery.

oed (commented):
This looks good. One test I would like to see added!

packages/core/src/ceramic.ts
```diff
@@ -632,6 +633,12 @@ export class Ceramic implements CeramicApi {
     const walkNext = async (node: TrieNode, streamId: StreamID | CommitID) => {
       let stream
       try {
+        if (query.genesis) {
```
stbrody, line 636: So if a user tries to use this functionality with a non-deterministic genesis commit, it will be a signed commit. In that case I believe loading the stream will still time out because the payload won't be put into ipfs, just the jwt envelope will be.  We should add a test for that and reproduce the behavior, and then probably the thing to do is to check here if the commit looks like a signed commit (can use [StreamUtils.isSignedCommit](https://github.com/ceramicnetwork/js-ceramic/blob/6871b7dcd27d08a727ae492754440309a563efc3/packages/common/src/utils/stream-utils.ts#L213)) and if so throw an exception.

stbrody (commented):
this looks much better, nice job!  One comment about handling dag-jose signed commits

packages/core/src/ceramic.ts
haardikk21: I'm confused about how would we know if it is a signed commit before we do actually load it. But it will end up throwing a different error in that case as genesis CID =/= streamId CID

haardikk21: Nvm, we can do this. Pushing it out soon

packages/common/src/ceramic-api.ts
```diff
@@ -138,7 +138,7 @@ export interface MultiQuery {
   /**
    * The genesis content for the queried stream
    */
-  genesis?: Record<string, any>
+  genesis?: CeramicCommit
```
stbrody, line 141:
```suggestion
  genesis?: GenesisCommit
```

stbrody: can make the type even more specific

packages/core/src/ceramic.ts
```diff
@@ -634,6 +634,13 @@ export class Ceramic implements CeramicApi {
       let stream
       try {
         if (query.genesis) {
+          if (
+            StreamUtils.isSignedCommitContainer(query.genesis) ||
+            StreamUtils.isSignedCommit(query.genesis)
+          ) {
+            return Promise.reject('Given genesis commit is not deterministic')
```
stbrody, line 644:
```suggestion
            throw new Error('Given genesis commit is not deterministic')
```

stbrody: Better for the function's promise to resolve in an Error object rather than a string

packages/core/src/ceramic.ts
```diff
@@ -634,6 +634,13 @@ export class Ceramic implements CeramicApi {
...
+          }
+
           const genesisCID = await this.ipfs.dag.put(query.genesis)
           if (!streamId.cid.equals(genesisCID)) {
             return Promise.reject('Given StreamID CID does not match given genesis content')
```
stbrody, line 649:
```suggestion
            throw new Error('Given StreamID CID does not match given genesis content')
```

packages/3id-did-resolver/src/index.ts
```diff
@@ -52,6 +52,9 @@ export function wrapDocument(content: any, did: string): DIDDocument | null {
     verificationMethod: [],
     authentication: [],
     keyAgreement: [],
+    assertionMethod: [],
```
stbrody, line 55: Did this sneak in somehow?

packages/common/src/ceramic-api.ts
```diff
@@ -135,6 +135,10 @@ export interface CeramicApi extends CeramicSigner {
 }
 
 export interface MultiQuery {
+  /**
+   * The genesis content for the queried stream
```
stbrody, line 139:
```suggestion
   * The genesis content for the queried stream. Useful in cases where the stream might not exist and you want to avoid timing out trying to load the genesis commit from IPFS.
```

packages/common/src/streamopts.ts
```diff
@@ -1,3 +1,4 @@
+import { Stream } from './stream'
```
stbrody, line 1: unnecessary, remove

packages/3id-did-resolver/src/index.ts
haardikk21: was part of a hacky workaround to fix tests when 3id-resolver upstream dep changed, its removed now

packages/core/src/ceramic.ts
haardikk21: I've been able to test this approach using the following:

```
expect(
        ceramic2.multiQuery([
          {
            streamId: stream1.id,
            genesis: genesisCommit,
          },
        ])
      ).rejects.toEqual('Given StreamID CID does not match given genesis content')
      ```
      
I'm not sure how I would test the error throw as the overall function does not care about errors being thrown within `walkNext`

packages/core/src/ceramic.ts
haardikk21: same as below

packages/core/src/ceramic.ts
stbrody: ah I see. You're trying to make failures of this specific type actually cause the whole operation to fail, unlike the existing failures when loading one of the linked streams, which should still allow the rest of the operation to continue.  Instead of using a different error reporting technique to get around the try/catch, why not just move the block of code you added to before the start of the try/catch block?

stbrody (approved):
LGTM!
Thanks for all your work on this!

stbrody:
well I take it back, tests still need to be fixed

haardikk21:
> well I take it back, tests still need to be fixed

o god lol, switching to errors from promise.rejects changed some minor stuff

PR #1756, opened 2021-10-06
feat: Include safe-did resolver

ukstv:


packages/cli/package.json
```diff
@@ -48,6 +48,7 @@
     "cors": "^2.8.5",
     "dag-jose": "^0.3.0",
     "did-resolver": "^3.1.0",
+    "did-safe-resolver": "^0.0.4",
```
oed, line 51: Package should be called `safe-did-resolver`.

oed: Can we contact the author here to get the name? https://www.npmjs.com/package/safe-did-resolver

oed (commented):
Would be nice if we can change the package name. Otherwise looks good.

oed:
yes, it's fine.

PR #1794, opened 2021-10-28
chore(streamid): Don't use exceptions as control flow in StreamRef::from

stbrody:
This allows WebStorm's "Any exception" breakpoint to be useful again without hitting tons of false positives from `StreamRef::from`.

An alternative approach would have been to add `StreamID.isValid` and `CommitID.isValid` static methods that never throw but return true or false based on if their input is a valid representation of a StreamID/CommitID.  That approach though would have resulted in `StreamRef::from` parsing the same input twice - once to determine if it was valid, and a second time to actually parse the input into the appropriate data type.


stbrody:
ping @ukstv @oed

oed (approved):
lgtm

PR #1795, opened 2021-10-29
fix(blockchain-utils-linking): Support eip1193 provider.request

stephhuynh18:
Note: I tried to import the @walletconnect/ethereum-provider package as a dev dependency for the test as this was the library that uncovered the bug but it makes the tests hang. I did manual happy path tests.

### DIFF
--- a/packages/blockchain-utils-linking/src/__tests__/ethereum.test.ts
+++ b/packages/blockchain-utils-linking/src/__tests__/ethereum.test.ts
@@ -162,4 +162,31 @@ describe('EthereumAuthProvider', () => {
     const auth = new ethereum.EthereumAuthProvider(provider, address)
     await expect(auth.withAddress(address).createLink(testDid)).resolves.toMatchSnapshot()
   })
+
+  describe('various providers', () => {
+    test('sendAsync', async () => {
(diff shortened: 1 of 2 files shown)


oed (approved):
lgtm

packages/blockchain-utils-linking/src/__tests__/ethereum.test.ts
```diff
@@ -162,4 +162,31 @@ describe('EthereumAuthProvider', () => {
     const auth = new ethereum.EthereumAuthProvider(provider, address)
     await expect(auth.withAddress(address).createLink(testDid)).resolves.toMatchSnapshot()
   })
+
+  describe('various providers', () => {
+    test('sendAsync', async () => {
```
stbrody, line 167: it would be nice to use `jest.spyOn` to spy on the three send methods (send, sendAsync, request) and then assert in the tests that the right one that you expected gets called and not the other ones

stephhuynh18: 👍

ukstv (approved):

PR #1806, opened 2021-11-03
feat(core): Update running state's pinned commits when adding pins to pin store

stephhuynh18:
# Update running state's pinned commits when adding pins to pin store 
https://github.com/ceramicnetwork/ceramic/issues/118
https://github.com/ceramicnetwork/ceramic/issues/119
https://github.com/ceramicnetwork/ceramic/issues/127

## Description

* `PinStore.add` finishes by calling a `setPinnedState` on `RunningState`, which sets the `pinnedState` field to the given state, and updates the `StateSource` for the `RunningState` to `StateStore`. This will prevent repinning of commits on subsequent `Pinstore.add` calls
* `PinStore.add` is given a second, optional argument `force` which if set to true causes it to pin everything in the entire stream history.  This is useful in case the state store and ipfs pin store get out of sync and there are CIDs that aren't pinned in IPFS even though they're included in the state object stored in the state store.

## How Has This Been Tested?

Unit Tests

## Definition of Done

- [x] The work addresses the description and outcomes in the issue
- [x] I have added relevant tests for new or updated functionality
- [x] My code follows conventions, is well commented, and easy to understand
- [x] My code builds and tests pass without any errors or warnings
- [x] I have tagged the relevant reviewers
- [x] The changes have been communicated to interested parties

### DIFF
--- a/packages/core/src/local-pin-api.ts
+++ b/packages/core/src/local-pin-api.ts
@@ -1,4 +1,4 @@
-import { PinApi, PublishOpts, RunningStateLike } from '@ceramicnetwork/common'
+import { PinApi, PublishOpts, SyncOptions } from '@ceramicnetwork/common'
 import StreamID from '@ceramicnetwork/streamid'
 import { DiagnosticsLogger } from '@ceramicnetwork/common'
 import { Repository } from './state-management/repository'
@@ -13,7 +13,7 @@ export class LocalPinApi implements PinApi {
   ) {}
 
   async add(streamId: StreamID): Promise<void> {
-    const state$ = await this.repository.load(streamId, {})
+    const state$ = await this.repository.load(streamId, { sync: SyncOptions.PREFER_CACHE })
     await this.repository.pin(state$)
     this.logger.verbose(`Pinned stream ${streamId.toString()}`)
   }
--- a/packages/core/src/local-pin-api.ts
+++ b/packages/core/src/local-pin-api.ts
@@ -9,12 +9,11 @@ import { Repository } from './state-management/repository'
 export class LocalPinApi implements PinApi {
   constructor(
     private readonly repository: Repository,
-    private readonly loadStream: (streamId: StreamID) => Promise<RunningStateLike>,
     private readonly logger: DiagnosticsLogger
   ) {}
 
   async add(streamId: StreamID): Promise<void> {
-    const state$ = await this.loadStream(streamId)
+    const state$ = await this.repository.load(streamId, {})
     await this.repository.pin(state$)
     this.logger.verbose(`Pinned stream ${streamId.toString()}`)
   }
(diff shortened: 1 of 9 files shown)


packages/core/src/state-management/repository.ts
```diff
@@ -287,7 +287,7 @@ export class Repository {
   }
 
   pin(streamStateHolder: StreamStateHolder): Promise<void> {
-    return this.#deps.pinStore.add(streamStateHolder)
+    return this.#deps.pinStore.add(new RunningState(streamStateHolder.state, StateSource.NETWORK))
```
stbrody, line 290: this definitely isn't safe.  The StreamStateHolder likely *is* a RunningState, and you're discarding the information from that RunningState about its StateSource, as well as breaking any rxjs pipelines built on the original RunningState.

The thing to do here is probably to update the signature of `pin` to be more specific and specify that its input is a  `RunningState` instance instead of a `StreamStateHolder`

packages/core/src/store/pin-store.ts
```diff
@@ -32,25 +27,28 @@ export class PinStore {
...
+    const newCommits =
+      runningState.pinnedCommits && !force
+        ? commitLog.filter((cid) => !runningState.pinnedCommits.has(cid.toString()))
+        : commitLog
 
     const points = await this.getComponentCIDsOfCommits(newCommits)
     await Promise.all(points.map((point) => this.pinning.pin(point)))
-    await this.stateStore.save(stateHolder)
+    await this.stateStore.save(runningState)
+    runningState.setPinnedState(runningState.state)
```
stbrody, line 51: hmm... if this is the only place `setPinnedState` is ever called outside of tests, then maybe it has the wrong signature.  Maybe it should be `markAsPinned()` and just always take its current state as the newly pinned state

stbrody (changes requested):
mostly looks really good, but there's one issue with constructing a new RunningState in `Repository.pin()`

packages/core/src/store/pin-store.ts
stephhuynh18: I agree

packages/core/src/state-management/repository.ts
stephhuynh18: Yea that was actually my first thought but I went with this tactic so there would be minimal changes. Thank you clarifying the problem with this tactic! I'll try updating the signature again and if it touches too many things i'll let you know

packages/core/src/local-pin-api.ts
```diff
@@ -9,12 +9,11 @@ import { Repository } from './state-management/repository'
 export class LocalPinApi implements PinApi {
   constructor(
     private readonly repository: Repository,
-    private readonly loadStream: (streamId: StreamID) => Promise<RunningStateLike>,
     private readonly logger: DiagnosticsLogger
   ) {}
 
   async add(streamId: StreamID): Promise<void> {
-    const state$ = await this.loadStream(streamId)
+    const state$ = await this.repository.load(streamId, {})
```
stbrody, line 16: let's explicitly pass `{sync: SyncOptions.PREFER_CACHE}` like we used to do in `loaderWithSyncSet` at the higher level

stbrody (approved):
LGTM with one small change.
Nice work!

PR #1808, opened 2021-11-04
Update pull_request_template.md

oed:
PR template was way to verbose. This makes it much simpler so it's not so much a burden on devs.

Removed a bunch of redundant checkboxes.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,8 +1,6 @@
-# [Replace Me With Meaningful Name] - #[Issue]
-
 ## Description
 
-Include a summary of the change and which issue it addresses in the title of the PR. 
+Include a summary of the change or link to the issue it addresses.
 
 Include relevant motivation, context, brief description and impact of the change(s). List follow-up tasks here.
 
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -13,18 +11,13 @@ Describe the tests that you ran to verify your changes. Provide instructions for
...
+## PR checklist
 
 Before submitting this PR, please make sure:
 
-- [ ] The work addresses the description and outcomes in the issue
-- [ ] I have added relevant tests for new or updated functionality
-- [ ] My code follows conventions, is well commented, and easy to understand
-- [ ] My code builds and tests pass without any errors or warnings
-- [ ] I have tagged the relevant reviewers
+- [ ] I have tagged the relevant reviewers and interested parties
```
glcohen, line 18: can we keep one about "adding or updating relevant tests"?

glcohen (approved):
lgtm - main concern is updating this across all our repos is a big hassle / manual process. I know we can do some terraform automation to do it but not sure it's worth investing in that right now.

.github/pull_request_template.md
oed: I think that's sufficiently covered by the "How Has This Been Tested?" section

PR #1819, opened 2021-11-09
chore(core): Add tests for optimized pinning behaviors

stbrody:


packages/core/src/state-management/repository.ts
```diff
@@ -57,7 +57,7 @@ export class Repository {
   /**
    * Various dependencies.
    */
-  #deps: RepositoryDependencies
+  private _deps: RepositoryDependencies
```
ukstv, line 60: IMO, it makes more sense to add a `pinStore` getter.

ukstv (approved):

PR #1822, opened 2021-11-09
chore(core): Remove StateSource enum in favor of pinned bool

stbrody:
builds on https://github.com/ceramicnetwork/js-ceramic/pull/1821

### DIFF
--- a/packages/core/src/state-management/running-state.ts
+++ b/packages/core/src/state-management/running-state.ts
@@ -11,9 +11,9 @@ import CID from 'cids'
 export class RunningState extends StreamStateSubject implements RunningStateLike {
   readonly id: StreamID
   readonly subscriptionSet: SubscriptionSet = new SubscriptionSet()
-  private _pinnedCommits?: Set<string>
+  private _pinnedCommits?: Set<string> | null
 
-  constructor(initial: StreamState, private pinned: boolean) {
+  constructor(initial: StreamState, pinned: boolean) {
     super(initial)
     this.id = new StreamID(initial.type, initial.log[0].cid)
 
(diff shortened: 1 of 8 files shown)


packages/core/src/state-management/running-state.ts
```diff
@@ -8,31 +8,16 @@ import {
...
-   * Indicates that the Stream was loaded from the network.
-   * runningState does not keep track of the commit CID's of the state as they may not have been stored.
-   */
-  NETWORK,
-}
 export class RunningState extends StreamStateSubject implements RunningStateLike {
   readonly id: StreamID
   readonly subscriptionSet: SubscriptionSet = new SubscriptionSet()
-  pinnedCommits?: Set<string>
+  private _pinnedCommits?: Set<string>
```
stbrody, line 14: made this private and added a getter so that external callers can't reassign it without going through a method on RunningState.  FYI @stephhuynh18

packages/core/src/state-management/running-state.ts
```diff
@@ -8,31 +8,16 @@ import {
...
 
-  constructor(initial: StreamState, public stateSource: StateSource) {
+  constructor(initial: StreamState, private pinned: boolean) {
```
ukstv, line 16: @stbrody This `pinned` flag is a side-effect of `_pinnedComments` being empty or not. IMO, it is better to eliminate a possibility of these two being inconsistent.

ukstv (changes requested):
Please, recalculate pinned flag based on `_pinnedCommits`.

packages/core/src/state-management/running-state.ts
stbrody: done

ukstv (approved):

PR #1832, opened 2021-11-15
feat(common): Import types, when only types are imported

ukstv:
Some imports are only used to provide type information. Here we replace vanilla import with `import type`.


stbrody (approved):
if you accidentally use `import type` when you actually do depend on the implementation, not just the type definition, will the compiler yell at you?

ukstv:
> if you accidentally use `import type` when you actually do depend on the implementation, not just the type definition, will the compiler yell at you?

It will.

PR #1841, opened 2021-11-15
chore: Use proper module name

ukstv:
Fix a bug.
1. Use `@ceramicnetwork/streamid` instead of `streamid` for import.
2. Import just `StreamID` type.


oed (approved):
Wow really long time ago we used this pkh name!

PR #1860, opened 2021-11-23
chore: Rename did parameter to controller on TileDocumentHandler::_verifySignature

ukstv:
It is not clear right now what is `did` and why it is here. Came across this while tinkering with CACAO spec.


stbrody (approved):
lgtm

Issue #1836, opened 2021-11-15
TileDocument: Accept optional `useDid` parameter to represent signed comits through a capability


haardikk21:

haardikk21:
Tests:
unhappy path - clearly a wrong DID is used
happy path - use the same DID as normally would be used, but pass it `asDID` instead until CACAO is ready

PR #1864, opened 2021-11-24
add `asDID` option to TileDocumentHandler

haardikk21:
## Description - #1836 

Introduces the ability to make an update with a different DID `asDID` than the one invoking the call. Necessary for OCAP work that is in-progress where updates will be made by the permission holder rather than the permission giver.

## How Has This Been Tested?

Since OCAPs are not fully implemented yet, there are two basic tests. 

The first passes the Ceramic Api's authenticated DID as `asDID` - which works as expected (even though we do switch to using `asDid` instead of `ceramic` for the signatures) 

The second passes a random DID `asDID` and it fails when attempting to sign a Dag JWS. 

More tests will be added as the CACAO library progresses and real use cases become available to test. The changes made should not affect any existing code flows.

## PR checklist

- [-] I have updated the READMEs of affected packages
- [-] I have made corresponding changes to the documentation

### DIFF
--- a/packages/canary-integration/src/__tests__/tile-update.test.ts
+++ b/packages/canary-integration/src/__tests__/tile-update.test.ts
@@ -1,0 +1,44 @@
+import { IpfsApi } from '@ceramicnetwork/common'
+import { createIPFS } from '../create-ipfs'
+import { CeramicApi } from '@ceramicnetwork/common'
+import { createCeramic } from '../create-ceramic'
+import { TileDocument } from '@ceramicnetwork/stream-tile'
+import { DID } from 'dids'
+import { Ed25519Provider } from 'key-did-provider-ed25519'
+import KeyResolver from 'key-did-resolver'
+import crypto from 'crypto'
+
+let ipfs: IpfsApi
+let ceramic: CeramicApi
+
+beforeAll(async () => {
+  ipfs = await createIPFS()
+  ceramic = await createCeramic(ipfs)
+}, 12000)
+
+afterAll(async () => {
+  await ipfs.stop()
+  await ceramic.close()
+})
+
+test('can update a tile document with valid asDID', async () => {
+  const newTile = await TileDocument.create(ceramic, { foo: 'bar' })
+  await newTile.update({ foo: 'baz' }, null, { asDID: ceramic.did })
+
+  expect(newTile.content).toMatchObject({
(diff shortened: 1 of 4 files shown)


packages/stream-tile/src/tile-document.ts
```diff
@@ -279,7 +279,13 @@ export class TileDocument<T = Record<string, any>> extends Stream {
     opts: UpdateOpts = {}
   ): Promise<void> {
     opts = { ...DEFAULT_UPDATE_OPTS, ...opts }
-    const updateCommit = await this.makeCommit(this.api, content, metadata)
+    let signer: CeramicSigner = this.api
```
ukstv, line 282: Suggestion: `const signer: CeramicSigner = opts.asDID ? { signer: opts.asDID } : this.api`

packages/common/src/streamopts.ts
```diff
@@ -1,3 +1,5 @@
+import { DID } from 'dids'
```
ukstv, line 1:
```suggestion
import type { DID } from 'dids'
```

packages/canary-integration/src/__tests__/tile-update.test.ts
```diff
@@ -0,0 +1,44 @@
+import { IpfsApi } from '@ceramicnetwork/common'
+import { createIPFS } from '../create-ipfs'
+import { CeramicApi } from '@ceramicnetwork/common'
+import { createCeramic } from '../create-ceramic'
+import { TileDocument } from '@ceramicnetwork/stream-tile'
+import { DID } from 'dids'
+import { Ed25519Provider } from 'key-did-provider-ed25519'
+import KeyResolver from 'key-did-resolver'
+import crypto from 'crypto'
```
ukstv, line 9: Please, use `@stablelib/random`.

ukstv (changes requested):
minor changes requested: import type DID, use @stablelib/random

ukstv (approved):
🏇

stbrody:
Sorry I'm late to the party here, but why is this necessary?  Why wouldn't we always take the DID that the Ceramic handle is authenticated as?

oed:
Because you might have two different session keys that have write access
to different streams. Also, check out the spec.

Sent via Superhuman iOS ***@***.***>

PR #1862, opened 2021-11-23
Chore: update deps

stephhuynh18:
## Description 

** Most significant changes in IPFS Daemon
Updating the dependencies of all the packages for https://github.com/ceramicnetwork/js-ceramic/issues/1804

Missed: 

## How Has This Been Tested?

-All tests are running
-Manually tested ipfs daemon

## PR checklist

Before submitting this PR, please make sure:

- [x] I have tagged the relevant reviewers and interested parties
- [x] I have updated the READMEs of affected packages
- [x] I have made corresponding changes to the documentation

## References:

https://github.com/ceramicnetwork/js-ceramic/issues/1804

PR #1872, opened 2021-11-26
feat: Do not unpin twice

ukstv:
It leads to the daemon or ipfs hanging.

Slightly more straightforward code.

### DIFF
--- a/packages/core/src/state-management/repository.ts
+++ b/packages/core/src/state-management/repository.ts
@@ -219,21 +219,8 @@ export class Repository {
     opts: CreateOpts | UpdateOpts
   ): Promise<RunningState> {
     const state$ = await this.stateManager.applyCommit(streamId, commit, opts)
-    await this.applyWriteOpts(state$, opts)
-    return state$
-  }
-
-  /**
-   * Apply options relating to authoring a new commit
-   *
-   * @param state$ - Running State
-   * @param opts - Initialization options (request anchor, publish to pubsub, etc.)
-   * @private
-   */
-  async applyWriteOpts(state$: RunningState, opts: CreateOpts | UpdateOpts) {
     this.stateManager.applyWriteOpts(state$, opts)
-
-    await this.handlePinOpts(state$, opts)
+    return state$
   }
 
   async handlePinOpts(state$: RunningState, opts: PinningOpts) {
@@ -251,9 +238,9 @@ export class Repository {
    * @param opts
    */
   async applyCreateOpts(streamId: StreamID, opts: CreateOpts): Promise<RunningState> {
-    const state = await this.load(streamId, opts)
-    await this.applyWriteOpts(state, opts)
-    return state
+    const state$ = await this.load(streamId, opts)
+    this.stateManager.applyWriteOpts(state$, opts)
+    return state$
   }
 
   /**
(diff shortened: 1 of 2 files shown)


stephhuynh18 (approved):

packages/core/src/state-management/repository.ts
```diff
@@ -219,21 +219,8 @@ export class Repository {
...
-   * Apply options relating to authoring a new commit
-   *
-   * @param state$ - Running State
-   * @param opts - Initialization options (request anchor, publish to pubsub, etc.)
-   * @private
-   */
-  async applyWriteOpts(state$: RunningState, opts: CreateOpts | UpdateOpts) {
     this.stateManager.applyWriteOpts(state$, opts)
-
-    await this.handlePinOpts(state$, opts)
```
stbrody, line 236: So where are the pinning opts being handled now?

packages/core/src/state-management/repository.ts
```diff
@@ -251,9 +238,9 @@ export class Repository {
    * @param opts
    */
   async applyCreateOpts(streamId: StreamID, opts: CreateOpts): Promise<RunningState> {
-    const state = await this.load(streamId, opts)
-    await this.applyWriteOpts(state, opts)
-    return state
+    const state$ = await this.load(streamId, opts)
+    this.stateManager.applyWriteOpts(state$, opts)
```
stbrody, line 242: why remove this await?

packages/core/src/state-management/repository.ts
ukstv: Copied from https://github.com/ceramicnetwork/js-ceramic/pull/1862#discussion_r756456702:

@ukstv So i did some investigating and this is what I found:
`stream.update()`  --> `ceramic.applyCommit()` --> `repo.applyCommit()` 
It then calls
`stateManager.applyCommit()` --> `repo.load()` --> `repo.handlePinOpts()` --> `repo.unpin()`
After the above completes the following gets called 
`repo.applyWriteOpts()` --> `repo.handlePinOpts()` --> `repo.unpin()`

So essentially `repo.unpin()` gets called twice. This functions is async and so this double call may be whats confusing jest causing it to hang even after the test successfully completes . If I comment out one of `repo.handlePinOpts` calls the test passes and doesn't hang. If i don't unpin in the test, the test passes and doesn't hang.

I'm not sure if I am right so please help me! The duplicated call to unpin is unexpected 

---

Pinning is handled on `Repository::load` which is called from both `Repository::load` (obviously) and `Repository::applyCommit` (through `StateManager::applyCommit`).

stbrody (changes requested):
sorry I didn't have time to get to this yesterday.  I'll try to give more detailed feedback later today but for now please hold off on merging this, I think we can do more to untangle the layering here

ukstv:
@stbrody I am sure more can be done. I do not think this PR is the right place though to do that.

stbrody (commented):
sorry for the long delay, I wanted to think about this more and try to lay out a concrete plan, but I just haven't had time this week.  I wrote a comment with the rough sketch of my thinking here.

ukstv:
Closing in favour of https://github.com/ceramicnetwork/js-ceramic/pull/1898

Issue #1870, opened 2021-11-25
Add `Access-Control-Max-Age` header to CORS preflight requests


PaulLeCam:
# Description

Browsers make preflight `OPTIONS` requests to cross-origin endpoints to check if they have permission to make requests, which results in doubling the number of requests the client makes to the server in some cases.
By adding the `Access-Control-Max-Age` header to the response, it is possible to have the browser cache the result of the preflight check, therefore avoiding repeating this check.
See [this article](https://blog.bitsrc.io/4-ways-to-reduce-cors-preflight-time-in-web-apps-1f47fe7558) for more details.

# Technical Information

The [`maxAge` option](https://github.com/expressjs/cors#configuration-options) could be set in the [CORS configuration of the daemon](https://github.com/ceramicnetwork/js-ceramic/blob/07cbc9108415c8e71c268e10acf9238f65460a1e/packages/cli/src/ceramic-daemon.ts#L174).

PR #1873, opened 2021-11-28
chore(cli): set "Access-Control-Max-Age"

oed:
## Description

Should address #1870

## How Has This Been Tested?

It has not been tested.

## PR checklist


PaulLeCam (approved):
Thanks for the change, LGTM but any reason for setting the cache to only 10 mins please? Couldn't this be cached for a long time?

oed:
Based on https://stackoverflow.com/questions/52008480/choose-cors-access-control-max-age-value seems like max for chromium is 2h. Updated the value.

PR #1899, opened 2021-12-06
fix(cli,http-client): Properly serialize timeout for multiquery requests through the http client

stbrody:


packages/http-client/src/ceramic-http-client.ts
```diff
@@ -154,6 +154,7 @@ export class CeramicClient implements CeramicApi {
       method: 'post',
       body: {
         queries: queriesJSON,
+        ...{ timeout },
```
stephhuynh18, line 157: Nitpick: I don't think you need to put timeout in an object then spread
```suggestion
        timeout,
```

packages/core/src/ceramic.ts
```diff
@@ -682,7 +682,7 @@ export class Ceramic implements CeramicApi {
    * @param queries - Array of MultiQueries
    * @param timeout - Timeout in milliseconds
    */
-  async multiQuery(queries: Array<MultiQuery>, timeout?: number): Promise<Record<string, Stream>> {
+  async multiQuery(queries: Array<MultiQuery>, timeout = 7000): Promise<Record<string, Stream>> {
```
stephhuynh18, line 685: Nice to have: add default value to the doc comment

stephhuynh18 (approved):
LGTM just a few optional nitpicks

packages/core/src/ceramic.ts
stbrody: the autogenerated docs already include default values

packages/http-client/src/ceramic-http-client.ts
stbrody: I was trying to avoid sending the field with an explicit value of `undefined` over the wire

PR #1901, opened 2021-12-06
fix(core): Allow loading multiple CommitIDs for the same stream in parallel

stbrody:


packages/core/src/state-management/state-manager.ts
```diff
@@ -111,7 +109,7 @@ export class StateManager {
    * @param commitId - Requested commit.
    */
   async atCommit(state$: RunningStateLike, commitId: CommitID): Promise<SnapshotState> {
-    return this.executionQ.forStream(commitId.baseID).run(async () => {
```
ukstv, line 114: and this is _extremely_ dangerous. pls, no.

ukstv (commented):
This is not going to work. I mean, it could for simple cases, but IMO there is an extremely high chance of data corruption.

PR #1902, opened 2021-12-06
chore(core): Cleanup stream load code

stbrody:
Small code cleanup, no functional change

### DIFF
--- a/packages/core/src/state-management/repository.ts
+++ b/packages/core/src/state-management/repository.ts
@@ -137,6 +137,29 @@ export class Repository {
   }
 
   /**
+   * Helper function for loading at least the genesis commit state for a stream.
+   * WARNING: This should only be called from within a thread in the loadingQ!!!
+   *
+   * @param streamId
+   * @returns a tuple whose first element is the state that was loaded, and whose second element
+   *   is a boolean representing whether we believe that state should be the most update-to-date
+   *   state for that stream, or whether it could be behind the current tip and needs to be synced.
+   */
+  async _loadGenesis(streamId: StreamID): Promise<[RunningState, boolean]> {
+    let stream = this.fromMemory(streamId)
+    if (stream) {
+      return [stream, true]
+    }
+
+    stream = await this.fromStateStore(streamId)
+    if (stream) {
+      return [stream, this.stateManager.wasPinnedStreamSynced(streamId)]
+    }
+
+    stream = await this.fromNetwork(streamId)
+    return [stream, false]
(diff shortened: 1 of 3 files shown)


packages/core/src/state-management/repository.ts
```diff
@@ -136,6 +136,29 @@ export class Repository {
     return state$
   }
 
+  /**
+   * Helper function for loading at least the genesis commit state for a stream.
+   * WARNING: This should only be called from within a thread in the loadingQ!!!
```
stbrody, line 141: I hate how easy it would be for someone to accidentally directly run a function that was designed to only be run from the loading/executionQ, and by so doing skip all the rate limiting/serialization that is supposed to be there.

I wonder can use the type system to force correct usage.  I tried to do an experiment of making all tasks given to the task queue require taking a single argument of a special type, so that all functions designed to run within a TaskQueue had to have that argument in their signature, hoping the compiler would catch violators, and that would make it difficult to accidentally run a function designed to run in the queue from outside the queue.  But unfortunately in my little experiment the compiler was too loose and allowed passing in `() => Promise<TaskResultType>`  even if the `Task` type was defined as `export type Task<TaskResultType> = (handle: TaskHandle) => Promise<TaskResultType>`.  I would have thought that the type being defined as taking a single arg of type `TaskHandle` would prevent matching with functions that take no args, but for some reason it was being allowed.  Maybe I did something wrong?

stbrody: Here's a link to a PR of my WIP experiment that doesn't seem to be working: https://github.com/ceramicnetwork/js-ceramic/pull/1903

stbrody: Note - don't feel any need to spend too much time looking at the linked WIP experiment - no more than 5 minutes.  I mostly just wanted to showcase the idea for how to make the API harder to use incorrectly, and see if you have any thoughts on it off the top of your head, or ideas about why it isn't working.  But I can definitely spend more time investigating myself before you look at it closely

packages/core/src/state-management/running-state.ts
```diff
@@ -34,6 +34,10 @@ export class RunningState extends StreamStateSubject implements RunningStateLike
     return this._pinnedCommits
   }
 
+  get isPinned(): boolean {
+    return this.pinnedCommits != null
```
ukstv, line 38: `Boolean(this.pinnedCommits && this.pinnedCommits.size > 0)`

PR #1905, opened 2021-12-07
chore(core): Test that conflictResolution.snapshotAtCommit does not mutate state object

stbrody:
Using deep-freeze unfortunately didn't work as it would throw an error `Cannot freeze array buffer views with elements` when called on a state object.

So instead I made the tests make a deep clone of the state object and compare them before and after loading at a commit to make sure they aren't modified.


stbrody:
closing in favor of https://github.com/ceramicnetwork/js-ceramic/pull/1920

PR #1920, opened 2021-12-08
fix(core): Load commits serially again

stbrody:
This undoes commit 3290a66db7f4063aac1df3781bef2962442740e2


stbrody:
I saw lots of failures loading streams on clay after https://github.com/ceramicnetwork/js-ceramic/commit/3290a66db7f4063aac1df3781bef2962442740e2 was deployed.  I don't know for certain that that commit is the cause, but it made me nervous.  Maybe it's not actually worth it to muck about with this codepath right now.

stbrody:
merging pre-review to unblock release

PR #1925, opened 2021-12-13
feat: Only allow valid CBOR in deterministic streams

ukstv:
Non-deterministic streams are all right.

See https://github.com/ceramicnetwork/ceramic/issues/205 for discussion.

BTW, for all of our tests cases, cbor-encoding takes 0-1ms.

### DIFF
--- a/packages/canary-integration/src/__tests__/deterministic-tile.test.ts
+++ b/packages/canary-integration/src/__tests__/deterministic-tile.test.ts
@@ -28,12 +28,6 @@ test("can create and retreive deterministic tile document", async () => {
   expect(createdTile.content).toMatchObject(retrievedTile.content)
 })
 
-test("can not create invalid deterministic tile document", async () => {
-  await expect(
-    TileDocument.deterministic(ceramic, { family: 'test123', tags: ['foo', undefined, 'blah'] })
-  ).rejects.toThrow(/`undefined` is not supported by the IPLD Data Model and cannot be encoded/)
-})
-
 test("cannot create or retreive deterministic tile document if family or tag not set", async () => {
   await expect(TileDocument.deterministic(ceramic, { forbidControllerChange: false })).rejects.toThrow(
     /Family and\/or tags are required/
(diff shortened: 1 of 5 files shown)


oed (approved):
lgtm

oed (commented):
We should use the latest version. Maybe address in a separate PR?

packages/stream-tile/package.json
```diff
@@ -27,6 +27,7 @@
   "dependencies": {
     "@ceramicnetwork/common": "^2.0.0-alpha.0",
     "@ceramicnetwork/streamid": "^2.0.0-alpha.0",
+    "@ipld/dag-cbor": "^6.0.13",
```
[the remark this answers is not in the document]
ukstv: That's intentional. We have to do another round of dependency updates.

packages/stream-tile/src/tile-document.ts
```diff
@@ -384,8 +385,11 @@ export class TileDocument<T = Record<string, any>> extends Stream {
     }
 
     if (content == null) {
+      const result = { header }
+      // Check if we can encode it in cbor. Should throw an error when invalid payload.
+      dagCbor.encode(result)
```
stbrody, line 390: what happens if we don't do this when we actually go and put it into ipfs?  I would have thought we'd get an error thrown anyway, from ipfs?

ukstv: In the usual scenario, when a request is sent to Ceramic node right away, yes, this change is redundant. We though expose `TileDocument.makeGenesis` as a publicly available method. It can be used for a multiquery with genesis parameter, for example, which leads to a surprising behaviour. Or, most probable case of surprising misuse is a deterministic tile + http-client. Here JSON serialization (opaque, since it is part of `fetch` API) messes with the payload and replaces `undefined` with `null`, which is not right.

stbrody: ah i see, that makes sense.  bummer.

stbrody (approved):
lgtm mod question about the best place for the test

packages/canary-integration/src/__tests__/deterministic-tile.test.ts
```diff
@@ -21,13 +21,19 @@ afterAll(async () => {
...
   await createdTile.update({ foo: 'bar' })
-  
+
   const retrievedTile = await TileDocument.deterministic(ceramic, {family: 'test123'}, { anchor: false, publish: false })
 
   expect(createdTile.id.toString()).toEqual(retrievedTile.id.toString())
   expect(createdTile.content).toMatchObject(retrievedTile.content)
 })
 
+test("can not create invalid deterministic tile document", async () => {
```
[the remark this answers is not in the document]
ukstv: moved.

PR #1933, opened 2021-12-16
feat: First stab at go-ipfs inclusion

ukstv:
Use go-ipfs in canary-integration and core packages.

Ancillary changes:
- slightly more straight forward IPFS resubscribe logic
- **[beware]** make multiquery really swallow an error


packages/canary-integration/package.json
```diff
@@ -21,10 +21,15 @@
...
     "@truffle/hdwallet-provider": "^2.0.0",
-    "dids": "3.0.0-alpha.0",
+    "dids": "^3.0.0-alpha.1",
     "ganache-core": "^2.13.2",
     "key-did-provider-ed25519": "^1.1.0",
-    "key-did-resolver": "^2.0.0-alpha.0"
+    "key-did-resolver": "^2.0.0-alpha.0",
+    "get-port": "^5.1.1",
+    "go-ipfs": "^0.11.0",
+    "ipfs-http-client": "^54.0.3-rc.12",
```
oed, line 30: js-ipfs has been released now.

packages/core/src/ceramic.ts
```diff
@@ -685,12 +685,10 @@ export class Ceramic implements CeramicApi {
...
       queries.map((query) => {
-        try {
-          return this._loadLinkedStreams(query, timeout)
-        } catch (e) {
+        return this._loadLinkedStreams(query, timeout).catch(e => {
           this._logger.warn(`Error during multiQuery: ${e.toString()}`)
-          return Promise.resolve({})
-        }
+          return {}
+        })
```
oed, line 691: Why is this changed?

packages/core/src/dispatcher.ts
```diff
@@ -88,7 +88,10 @@ export class Dispatcher {
         // put the JWS into the ipfs dag
         const cid = await this._ipfs.dag.put(jws, { format: 'dag-jose', hashAlg: 'sha2-256' })
         // put the payload into the ipfs dag
-        await this._ipfs.block.put(linkedBlock, jws.link)
+        const linkCid = jws.link
+        const format = await this._ipfs.codecs.getCodec(linkCid.code).then(f => f.name);
+        const mhtype = await this._ipfs.hashers.getHasher(linkCid.multihash.code).then(mh => mh.name)
```
oed, line 93: oh, that's convenient!

oed (commented):
This seems like a surprisingly easy change :O

packages/core/src/ceramic.ts
ukstv: This is the most controversial change of all here. Original code, it seems, contained a bug.

There is a `try-catch` clause around `_loadLinkedStreams` call, which conveys that an error should be caught. Really, as `_loadLinkedStreams` returns a `Promise`, its rejection is not caught by `try-catch` at all without `await` statement. If adding `await`, `try-catch` clause is too verbose and does not directly tell "just log an error, do nothing" story.

Backstory: if left as it were, the unhandled promise rejection weirdly interfered with Jest and go-ipfs. It prematurely ended the test suite: killed go-ipfs instances (as specified in `afterAll` hook), but did not close ceramic instances, as they are down the control flow of the test. I guess, in general it could cause a multiquery to throw if just one of the queries failed.

oed:
Does things still work with js-ipfs with these changes? Should we care about that?
cc @stbrody

ukstv:
> This seems like a surprisingly easy change :O

Well, yes, after some homework :D

ukstv:
> Does things still work with js-ipfs with these changes? Should we care about that?

IMO, it should not bother us. But, maybe, we could add an environment variable that commands `createIPFS` and `withFleet`  to select go-ipfs or js-ipfs. Then we run the test suite twice - one for each ipfs flavour.

ukstv:
> Maybe that's reasonable to do, at least during the transition period?

Ok. Next PR, maybe, please?

ukstv:
@oed @stbrody For go-ipfs + js-ipfs test https://github.com/ceramicnetwork/js-ceramic/issues/1934

packages/core/src/ceramic.ts
oed: good catch

stbrody: so I think the essence of this change is right, but it has an unfortunate side effect made evident by the changes to ceramic.test.ts.  I think if a user uses the API wrong, they should get an error right away in their app that they are misusing the API.  With this change, now it'll just silently fail.

The problem is that we're treating user-input errors (thrown by `_ensureGenesis`) the same as a failure to load a stream (an error or timeout when `_loadLinkedStreams` loads a stream, which has it's own try-catch).

I think the thing to do would be to move the `_ensureGenesis` call out of `_loadLinkedStreams` and up to `multiQuery`.  So we could then do 2 passes over all the `queries` in `multiQuery` - one pass to `_ensureGenesis` for each, where any failures causes the entire multiquery to throw an error, and then a second pass to `_loadLinkedStreams` where all errors are swallowed.

packages/core/src/__tests__/ceramic.test.ts
```diff
@@ -564,16 +564,14 @@ describe('Ceramic integration', () => {
             genesis: genesisCommit,
           },
         ])
-      ).rejects.toThrowError(
```
stbrody, line 567: I don't like losing this feedback to users who misuse the multiquery api.  All the changes to this test file make me nervous.  Please see my comment on the multiquery code for an idea of how to avoid this while still handling the uncaught exception issue you were hitting.

stbrody (changes requested):
overall this looks really great.  2 main points of concern - one is related to the changes to multiquery, the other is about how we respond to connection errors.  All other comments can be considered optional.

packages/core/src/ceramic.ts
ukstv: @stbrody This changes current semantics of multiquery: individual queries execute separately. Suddenly, there is no way to return back an individual error here.

packages/canary-integration/src/create-ipfs.ts
```diff
@@ -1,40 +1,49 @@
...
-  return new Proxy(instance, {
-    get(target: any, p: PropertyKey): any {
-      if (p === 'stop') {
-        tmpFolder.cleanup()
-      }
-      return target[p]
-    },
+  const ipfsd = await Ctl.createController({
+    ipfsHttpModule: ipfsHttp,
+    ipfsBin: path(),
```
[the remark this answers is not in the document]
ukstv: The instances are fully independent, with own repository, own peer id and so on. Data are wiped after a daemon is stopped. Similar to what we do to js-ipfs.

packages/core/src/pubsub/incoming-channel.ts
```diff
@@ -72,32 +72,27 @@ export class IncomingChannel extends Observable<IPFSPubsubMessage> {
...
+        complete()
+      } else {
+        this.tasks.add(async () => {
+          await new Promise((resolve) => setTimeout(resolve, this.resubscribeEvery))
+          await this.ipfs.pubsub?.unsubscribe(this.topic, handler)
+          await this.subscribeToIpfs(handler, complete)
+        })
       }
-      try {
-        // For some reason ipfs.id() throws an error directly if the
```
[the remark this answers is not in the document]
ukstv: It is not accurate.

packages/core/src/ceramic.ts
ukstv: And, I tend to believe, if we are to handle this erroring behavior, we better approach it holistically. One approach I can think of is to return an error message there in a query entry, but discussion on error message format might take a bit too long.

packages/canary-integration/src/create-ipfs.ts
```diff
@@ -1,40 +1,49 @@
...
  */
-export async function createIPFS(overrideConfig: Record<string, unknown> = {}): Promise<IpfsApi> {
-  const tmpFolder = await tmp.dir({ unsafeCleanup: true })
-
-  const port = await getPort()
+/**
+ * Create an IPFS instance
+ * @param overrideConfig - IFPS config for override
+ */
+export async function createIPFS(overrideConfig: Partial<Options> = {}): Promise<IPFS> {
```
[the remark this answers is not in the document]
ukstv: Going to be done in https://github.com/ceramicnetwork/js-ceramic/issues/1938 Out of scope here.

packages/core/src/pubsub/incoming-channel.ts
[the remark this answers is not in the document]
ukstv: or maybe it is. going to move it back.

packages/canary-integration/src/create-ipfs.ts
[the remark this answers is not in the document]
stbrody: I'm still having a pretty hard time understanding what's happening on this line.  `path` is imported from the `go-ipfs` package, which as best as I can tell is a re-export of the `path` built-in node module: https://github.com/ipfs/npm-go-ipfs/blob/master/src/index.js#L4

but when I `import path` locally and try to run `path()` as a function, I get `Uncaught TypeError: path is not a function`.

What am I missing?  What is actually being returned here?

packages/core/src/ceramic.ts
stbrody: > This changes current semantics of multiquery

I think it's the PR as written that is changing the semantics worse than what I am proposing.  Multiquery currently today will throw an error if the user mis-uses the `genesis` api.  Now this PR removes that error and makes it fail silently instead.  I'm trying to preserve the existing semantics of the application getting an error if they mis-use the API.

It's true that if multiple queries in the list all make the same mistake in how the `genesis` option works, the app will only get an error from the first one.  But this doesn't seem so bad to me, and I agree that figuring out a format for returning and correlating multiple errors from multiple queries is out of scope here.  But chances are that if an application is mis-using the `genesis` api, probably all the queries misuse it in the same way, and just getting an error from the first misuse would cause them to fix their app and likely fix the issue for all the queries in the multiquery.

packages/canary-integration/src/create-ipfs.ts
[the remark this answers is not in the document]
ukstv: `path` in `create-ipfs.ts` is this custom function https://github.com/ipfs/npm-go-ipfs/blob/master/src/index.js#L6 that returns a path to installed `go-ipfs` binary.

`Uncaught TypeError: path is not a function` is possibly caused by how you configure TS+Node and how you import or require the function ;)

If in Node.js REPL, which is CJS compliant, this code should work all right:
```
const goIpfs = require('go-ipfs')
goIpfs.path()
```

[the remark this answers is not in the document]
stbrody: ooph it was right in front of my face and I missed it!  Thanks for clarifying!

ukstv:
Done all the changes to IncomingChannel.

Errors while subscribing, and while being subscribed are treated the same => log, resubscribe in `resubscribeEvery` ms.

oed (approved):
Great! Would like to see a checkmark from @stbrody as well.

packages/core/src/pubsub/incoming-channel.ts
```diff
@@ -16,89 +16,69 @@ export type IPFSPubsubMessage = {
...
-}
-
 /**
- * Incoming IPFS PubSub message stream as Observable.
- * Ensures that ipfs subscription to a topic is live by periodically re-subscribing.
- *
- * One could simplify this by removing periodic re-subscription and serialization via task-queue.
- * Unfortunately, we can end up with concurrency issues: trying to unsubscribe while (or right after) other instance tries to subscribe.
- * So, better keep it all together.
+ * Subscription attempts must be sequential, in FIFO order.
```
stbrody, line 20: why is it important that subscription attempts be serialized?  What happens if they are not?

packages/core/src/pubsub/incoming-channel.ts
```diff
@@ -16,89 +16,69 @@ export type IPFSPubsubMessage = {
...
-        console.warn(`Can not reach ipfs node, resubscribing: ${error}`)
-        // TODO - should probably have some sort of backoff mechanism here.
-        await new Promise((resolve) => setTimeout(resolve, this.resubscribeEvery))
-        this.tasks.add(() => this.subscribeToIpfs(handler, complete))
-      }
-    } else {
-      complete()
-    }
+/**
+ * Incoming IPFS PubSub message stream as Observable.
```
stbrody, line 60:
```suggestion
 * Incoming IPFS PubSub message stream as Observable.  Adds retry logic on top of base PubsubIncoming.
```

stbrody (approved):
lgtm

packages/core/src/pubsub/incoming-channel.ts
ukstv: ".subscribe" call is not idempotent. It throws if already subscribed: https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs-http-client/src/pubsub/subscription-tracker.js#L26

PR #1935, opened 2021-12-17
feat: Test with both go-ipfs and js-ipfs

ukstv:
Make CircleCI test using go-ipfs and js-ipfs based on `IPFS_FLAVOR` env variable. Uses go-ipfs by default, but if `IPFS_FLAVOR` env is set to `js` or `JS`, then `js-ipfs` is used.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,33 +1,41 @@
 version: 2.1
 orbs:
   node: circleci/node@4.7.0
-jobs:
+
+commands:
   build-and-test:
-    machine:
-      image: ubuntu-2004:202107-02
-    environment:
-      NODE_OPTIONS: "--max_old_space_size=4096"
     steps:
       - checkout
       - node/install-npm:
-          version: "6"
+          version: '6'
       - node/install-packages
       - run: npm run lint
       - run: npm run build
       - run:
-          name: Test with go-ipfs
+          name: Test
           command: npm run test
           no_output_timeout: 30m
-          environment:
-            IPFS_FLAVOR: go
-      - run:
-          name: Test with js-ipfs
-          command: npm run test
-          no_output_timeout: 30m
-          environment:
-            IPFS_FLAVOR: js
+jobs:
+  with-go-ipfs:
+    machine:
+      image: ubuntu-2004:202107-02
+    environment:
+      NODE_OPTIONS: '--max_old_space_size=4096'
+      IPFS_FLAVOR: go
+    steps:
+      - build-and-test
       - run: npm run docs
+  with-js-ipfs:
+    machine:
+      image: ubuntu-2004:202107-02
+    environment:
+      NODE_OPTIONS: '--max_old_space_size=4096'
+      IPFS_FLAVOR: js
+    steps:
+      - build-and-test
+
 workflows:
   build-and-test:
     jobs:
-      - build-and-test
+      - with-go-ipfs
+      - with-js-ipfs
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,7 +20,6 @@ jobs:
     machine:
       image: ubuntu-2004:202107-02
     environment:
-      NODE_OPTIONS: '--max_old_space_size=4096'
       IPFS_FLAVOR: go
     steps:
       - build-and-test
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,8 +15,17 @@ jobs:
       - run: npm run lint
       - run: npm run build
       - run:
+          name: Test with go-ipfs
           command: npm run test
           no_output_timeout: 30m
+          environment:
+            IPFS_FLAVOR: go
+      - run:
+          name: Test with js-ipfs
+          command: npm run test
+          no_output_timeout: 30m
+          environment:
+            IPFS_FLAVOR: js
       - run: npm run docs
 workflows:
   build-and-test:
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -15,8 +15,17 @@ jobs:
...
           command: npm run test
           no_output_timeout: 30m
+          environment:
+            IPFS_FLAVOR: go
+      - run:
+          name: Test with js-ipfs
+          command: npm run test
+          no_output_timeout: 30m
+          environment:
+            IPFS_FLAVOR: js
```
oed, line 28: Can we make these two tasks run in parallel?

packages/canary-integration/src/create-ipfs.ts
```diff
@@ -2,21 +2,33 @@ import * as dagJose from 'dag-jose'
 import { path } from 'go-ipfs'
 import * as Ctl from 'ipfsd-ctl'
 import * as ipfsHttp from 'ipfs-http-client'
-import type { IPFS, Options } from 'ipfs-core'
+import type { Options } from 'ipfs-core'
+import { create } from 'ipfs-core'
 import getPort from 'get-port'
 import mergeOpts from 'merge-options'
+import type { IpfsApi } from '@ceramicnetwork/common'
```
oed, line 9: Is this not possible to import directly from js-ipfs yet?

oed (commented):
Would be good if we can run both of these in parallel!

.circleci/config.yml
oed: I guess we could create two different "jobs"?

stbrody (commented):
my only comment is the same as joels - can we make the jobs parallel?  otherwise I think it looks great, though I'm still hoping we can de-dupe the two different paths for launching ipfs to reduce code duplication, but that can happen in a separate pr

ukstv:
> I'm still hoping we can de-dupe the two different paths for launching ipfs to reduce code duplication

Sorry, could you elaborate please? I remember no previous discussion about de-duping anything.

packages/canary-integration/src/create-ipfs.ts
ukstv: It is possible, but for the sake of uniformity, I retained `IpfsApi`, until it is gone everywhere at once.

stbrody:
> > I'm still hoping we can de-dupe the two different paths for launching ipfs to reduce code duplication
> 
> @stbrody Sorry, could you elaborate please? I remember no previous discussion about de-duping anything.

it was a comment on the previous PR.  We have code to create an ipfs node in both the `core` and `canary-integration` pacakges.  Would be good to reuse one version and get rid of the code duplication

.circleci/config.yml
ukstv: it works https://app.circleci.com/pipelines/github/ceramicnetwork/js-ceramic/4772/workflows/a56b2290-5fe5-4386-88bb-abb74552a6f4

packages/canary-integration/src/create-ipfs.ts
ukstv: Relevant: https://github.com/ceramicnetwork/js-ceramic/issues/1937

ukstv:
@stbrody tracked in https://github.com/ceramicnetwork/js-ceramic/issues/1938

Out of scope here, IMO. I know it is not DRY yet.

.circleci/config.yml
oed: Nice!

oed: I wonder if there's a way to not duplicate the clone and build. With doing them in parallel we spend more CPU time but less wall clock time.

ukstv: In theory, you could do a single `build` job, and then run two ('with-go-ipfs', 'with-js-ipfs') jobs in parallel. In practice (prior experience+the internets on CircleCI specifics), with node packages, it gets slow in every dimension. Every CircleCI job executes in own container (or machine, or whatever executor). In order to pass data, like our fully installed and built repo, between jobs, you have to archive that, pass to storage, then download it and dearchive. For thousands of small files in node_modules, the process is far from perfect.

oed: ok, makes sense. Then let's keep as is :)

oed (approved):

.circleci/config.yml
```diff
@@ -1,24 +1,41 @@
...
       - run:
+          name: Test
           command: npm run test
           no_output_timeout: 30m
+jobs:
+  with-go-ipfs:
+    machine:
+      image: ubuntu-2004:202107-02
+    environment:
+      NODE_OPTIONS: '--max_old_space_size=4096'
```
stbrody, line 23: nit: can probably remove this option from go-ipfs

stbrody (approved):

.circleci/config.yml
ukstv: ok, will see, if it works :)

ukstv: it does.

stbrody: yeah, `NODE_OPTIONS` is an env var used by the nodejs runtime, it's not ipfs-specific at all, so I can't imagine it having any affect at all on a go-ipfs process

stbrody:
I just realized that this PR doesn't update https://github.com/ceramicnetwork/js-ceramic/blob/e037f8ea8b6ae0fdd83a8e5d8d435ba89554b7ed/packages/cli/src/__tests__/create-ipfs.ts#L10, which means the ceramic-daemon tests aren't using go-ipfs.  Those are fairly large and extensive tests, so we'll definitely need to get them switched over to go-ipfs as well.  I'll file a ticket

PR #1947, opened 2021-12-21
feat(core): Add env var to configure pubsub qps limit

stbrody:
Tested manually

I thought about adding this fully to the ceramic config and daemon config file, but given that query rate limiting is likely going away soon with tipsync, I figured it might be better just to use an env var as a quick fix that we can easily remove later.

### DIFF
--- a/packages/core/src/__tests__/dispatcher.test.ts
+++ b/packages/core/src/__tests__/dispatcher.test.ts
@@ -1,7 +1,7 @@
 import { Dispatcher } from '../dispatcher'
 import CID from 'cids'
 import StreamID from '@ceramicnetwork/streamid'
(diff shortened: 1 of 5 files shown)


oed (approved):

oed (changes requested):
We should make `maxQueuedQueries` configurable

stbrody:
@oed, are you okay with leaving the max queue size fixed to 10x the rate limit per second, if I also do a follow up to no longer fail the request when the queue is full but just drop the query on the floor?

stbrody:
Follow-up PR to drop queries on the floor once the queue is full: https://github.com/ceramicnetwork/js-ceramic/pull/1955

oed (approved):

PR #1949, opened 2021-12-21
fix(core): Fix ipfs retries when using ipfs http client

stbrody:


oed (approved):

stbrody:
@ukstv This PR seems to be causing the tests to consistently time out in CI, although they pass locally.  Any ideas how to debug?

ukstv (approved):

stbrody:
okay, after a lot of experimentation as best as I can tell it seems like the error is related to using the ipfs http client.  If I change the test to pass the ipfs node directly to the dispatcher, then it doesn't hang.  But if I pass the ipfs-http-client instance then the tests pass but time out after finishing.

I'm not really sure what to do about that.  The issue this PR is fixing only shows up with the ipfs-http-client, so if I don't use it the test is useless.  But I have no idea how to make the tests actually complete without timing out while using it.
I guess we don't actually have any other tests that use the ipfs http client - usually in tests when we set up ceramic we pass the ipfs node instance to ceramic directly.

Any ideas @ukstv?

stbrody:
Okay, I managed to get the tests to finish without timing out at the end.  I had to make two changes.  The first was to split the new test that starts up a real ipfs node and talks to it with the http client out into its own file.  I do not know why it was hanging if this test lived in the same file as the existing `dispatcher.test.ts`, but for some reason it did, and splitting it out into its own `dispatcher-real-ipfs.test.ts` file fixed things.

The second change was to sleep after calling `dispatcher.close()` but before shutting down the ipfs node.  I believe this is because `dispatcher.close()` can return without actually unsubscribing from pubsub, which leaves a long-polling connection open to ipfs in the background, which was the open handle preventing jest from terminating in CI.  I filed https://github.com/ceramicnetwork/js-ceramic/issues/1963 to make `dispatcher.close()` not return until all the cleanup work is complete so we can have safe, clean shutdown for Ceramic.  In the meantime, I just left the 2 second sleep in place with a TODO to remove it once the newly filed ticket is complete.

@ukstv

Issue #1930, opened 2021-12-14
Cache chain_id in ethereum account linking


stbrody:
There are several places in the ethereum implementation of blockchain-utils-linking where we call the `eth_chainId` endpoint of the ethereum provider.  This is resulting in many thousands of calls to this endpoint in our infura account.  We should cache this data.

https://github.com/ceramicnetwork/js-ceramic/blob/e037f8ea8b6ae0fdd83a8e5d8d435ba89554b7ed/packages/blockchain-utils-linking/src/ethereum.ts#L158

https://github.com/ceramicnetwork/js-ceramic/blob/e037f8ea8b6ae0fdd83a8e5d8d435ba89554b7ed/packages/blockchain-utils-linking/src/ethereum.ts#L32

PR #1951, opened 2021-12-21
chore(core): Add transaction cache to EthereumAnchorValidator

stbrody:


oed (commented):
What about a cache for chain_id which is our most used endpoint?

stbrody:
> What about a cache for chain_id which is our most used endpoint?

Separate ticket: https://github.com/ceramicnetwork/js-ceramic/issues/1930

oed (approved):

PR #1956, opened 2021-12-22
fix(core): Allow loading multiple CommitIDs for the same stream in parallel

stbrody:
This is a re-apply of 3290a66db7f4063aac1df3781bef2962442740e2 after it was
reverted in 8c73805991e1f3d960f5451af8fa795fb260fef2.


oed (commented):
Want @ukstv to review this as well.

ukstv (approved):

PR #1969, opened 2022-01-04
fix: return 503 during health-check if connected ipfs is not online

smrz2001:
## Description

Update the Ceramic daemon health-check to return unhealthy if its associated IPFS node in offline.

This ensures that ECS does not mark Ceramic as ready-to-use until IPFS is fully operational.

### DIFF
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -31,6 +31,7 @@ const packageJson = require('../package.json')
 
 const DEFAULT_HOSTNAME = '0.0.0.0'
 const DEFAULT_PORT = 7007
+const HEALTHCHECK_RETRIES = 3
 
 interface MultiQueryWithDocId extends MultiQuery {
   docId?: string
@@ -300,14 +301,15 @@ export class CeramicDaemon {
    * @dev Only checking for IPFS right now but checks for other subsystems can go here in the future
    */
   async healthcheck(req: Request, res: Response): Promise<void> {
-    try {
-      if (await this.ceramic.ipfs.isOnline()) {
-        res.status(200).send('Alive!')
-        return
-      }
-    }
-    catch (e) {
-      this.diagnosticsLogger.err(`Error checking IPFS status: ${e}`)
+    for (let i = 0; i < HEALTHCHECK_RETRIES; i++) {
+        try {
+          if (await this.ceramic.ipfs.isOnline()) {
+            res.status(200).send('Alive!')
+            return
+          }
+        } catch (e) {
+          this.diagnosticsLogger.err(`Error checking IPFS status: ${e}`)
+        }
     }
     res.status(503).send('IPFS unreachable')
   }
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -295,8 +295,13 @@ export class CeramicDaemon {
     }
   }
 
+  // Mark Ceramic unhealthy if associated IPFS is not 'online'
   healthcheck(req: Request, res: Response): void {
-    res.status(200).send('Alive!')
+    if (this.ceramic.ipfs.isOnline()) {
+      res.status(200).send('Alive!')
+    } else {
+      res.status(503).send('IPFS offline')
+    }
   }
 
   /**
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -296,7 +296,7 @@ export class CeramicDaemon {
   }
 
   // Mark Ceramic unhealthy if associated IPFS is not 'online'
-  async healthcheck(req: Request, res: Response): void {
+  async healthcheck(req: Request, res: Response): Promise<void> {
     try {
       if (await this.ceramic.ipfs.isOnline()) {
         res.status(200).send('Alive!')
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -296,12 +296,17 @@ export class CeramicDaemon {
   }
 
   // Mark Ceramic unhealthy if associated IPFS is not 'online'
-  healthcheck(req: Request, res: Response): void {
-    if (this.ceramic.ipfs.isOnline()) {
-      res.status(200).send('Alive!')
-    } else {
-      res.status(503).send('IPFS offline')
+  async healthcheck(req: Request, res: Response): void {
+    try {
+      if (await this.ceramic.ipfs.isOnline()) {
+        res.status(200).send('Alive!')
+        return
+      }
     }
+    catch (e) {
+      this.diagnosticsLogger.err(`Error checking IPFS status: ${e}`)
+    }
+    res.status(503).send('IPFS unreachable')
   }
 
   /**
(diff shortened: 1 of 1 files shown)


packages/cli/src/ceramic-daemon.ts
```diff
@@ -295,8 +295,13 @@ export class CeramicDaemon {
     }
   }
 
+  // Mark Ceramic unhealthy if associated IPFS is not 'online'
   healthcheck(req: Request, res: Response): void {
-    res.status(200).send('Alive!')
+    if (this.ceramic.ipfs.isOnline()) {
```
stbrody, line 300: please make sure to test this with both an in-process and out-of-process (http-client) ipfs node.

smrz2001: Will do 👍🏼

smrz2001: Needed to add some additional code after all for out-of-process IPFS node. Worked correctly if Ceramic came up before IPFS but then it kept returning 200 even after I killed IPFS.

With the latest change, if IPFS dies later on, the health-check will fail and mark Ceramic as unhealthy. I'm wondering if we should do that or not...

Doing so would mean the Ceramic node being marked unhealthy if IPFS crashes, for example, which might cause Ceramic to also be recreated but maybe that's not a bad thing?

stbrody: hmm that's a good point. I don't really know what's best practice around whether or not a service should report itself as healthy if a service it in turn depends on is unhealthy.  Would be a good question for our future Observability Engineer.  In lieu of that person, let's wait for Val to come back and weigh in.

I'm also going to add @oed and @ukstv to the review

oed: > Doing so would mean the Ceramic node being marked unhealthy if IPFS crashes, for example, which might cause Ceramic to also be recreated but maybe that's not a bad thing?

Well that means that streams would be marked as not synced and it would cause a lot more pubsub traffic.

>  Would be a good question for our future Observability Engineer.

What is it that that person could answer but we can't now?

ukstv:
Can we have another call rather than `isOnline` for the liveness check, something more robust? I personally consider Ceramic+IPFS to be the single package. If IPFS is unreachable, Ceramic should be considered dead too.

packages/cli/src/ceramic-daemon.ts
smrz2001: Is there a way to handle IPFS being offline more gracefully without marking Ceramic as unhealthy? I do need to test what would actually happen with the current code - I suspect Ceramic would be restarted if it becomes unhealthy but will double-check.

Assuming that Ceramic also restarts when unhealthy, perhaps it's ok to just postpone "healthy" during init to ensure the right startup seq and then not care if IPFS goes offline after that?

smrz2001:
> Can we have another call rather than `isOnline` for the liveness check, something more robust? I personally consider Ceramic+IPFS to be the single package. If IPFS is unreachable, Ceramic should be considered dead too.

Thanks, @ukstv, that was my thinking too w.r.t. reachability. Maybe it would be better to respond in a certain way to API requests when IPFS is unreachable without marking Ceramic unhealthy and causing it to restart.

packages/cli/src/ceramic-daemon.ts
v-stickykeys: > if IPFS dies later on, the health-check will fail and mark Ceramic as unhealthy
> 
> Doing so would mean the Ceramic node being marked unhealthy if IPFS crashes, for example, which might cause Ceramic to also be recreated but maybe that's not a bad thing?

This is the current behavior. If IPFS dies after Ceramic starts then Ceramic dies. I don't know what part of the code causes this but that's what happens today.

I think this behavior makes sense. As @ukstv said below, they are a package deal.

v-stickykeys: > Is there a way to handle IPFS being offline more gracefully without marking Ceramic as unhealthy?
Why? Can Ceramic operate in a healthy state without IPFS?

v-stickykeys:
> Can we have another call rather than `isOnline` for the liveness check, something more robust? I personally consider Ceramic+IPFS to be the single package. If IPFS is unreachable, Ceramic should be considered dead too.

@ukstv @smrz2001 given how we do our healthcheck for IPFS it seems like `isOnline` should be sufficient. See: https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/ipfs-daemon/src/healthcheck-server.ts

v-stickykeys (approved):
👍

ukstv (approved):
Convinced :)

🏇

packages/cli/src/ceramic-daemon.ts
smrz2001: > Why? Can Ceramic operate in a healthy state without IPFS?

It can't but restarting the Ceramic node does have other undesirable side-effects, e.g. the one @oed mentioned:

> Well that means that streams would be marked as not synced and it would cause a lot more pubsub traffic.

Avoiding this would be a good reason to not restart Ceramic when IPFS restarts, and instead to error for API calls (since Ceramic can't work when IPFS is down) and then wait till IPFS comes back up to allow APIs to work again.

Ceramic and IPFS are definitely a package deal but Ceramic doesn't need to be restarted to ensure that it's only used when IPFS is also online.

I don't want to bikeshed too much 😅 so I'd be fine with merging the code as it is (i.e. Ceramic becomes unhealthy and restarts if IPFS does) and then coming back to it later if y'all are ok with that. Just wanted to point out that there _is_ a tradeoff to consider here. @v-stickykeys @ukstv

v-stickykeys: @smrz2001 thanks for explanation and yeah that makes sense. I don't understand what "streams would be marked as not synced means"

But it seems like there is a way to gracefully shut down--i.e. maybe Ceramic needs to do some clean up before it shuts down if IPFS can not be reached. Will leave it to you to decide if that should be in this PR or come later.

stbrody:
We probably want to add a few retries around isOnline.  We wouldn't want a single missed message due to a flaky network to cause the healthcheck to fail and make Ceramic restart.

smrz2001:
After discussing with @v-stickykeys, I added a separate `isOnline` endpoint to check whether Ceramic is "operational" (currently only checking for IPFS, but could be more subsystems in the future). I wasn't sure if there was a naming convention with upper/lower case REST resource names - I can change the name of the resource if anyone has objections/suggestions.

The `healthcheck` endpoint will continue to be used purely for for Ceramic daemon health while the new endpoint can be used to detect/handle startup sequence-related issues.

smrz2001:
> We probably want to add a few retries around isOnline. We wouldn't want a single missed message due to a flaky network to cause the healthcheck to fail and make Ceramic restart.

Thanks, @stbrody, please see comment above. Shouldn't affect existing behavior after latest update.

stbrody:
What about adding a flag to the existing endpoint rather than adding a new endpoint?

smrz2001:
So @v-stickykeys and I discussed this some more and think reusing the existing `healthcheck` endpoint is the preferable thing to do (I believe @ukstv already agreed with this approach when he approved the PR).

If Ceramic crashes and loses its state because IPFS crashes, so be it. It's not ideal but it's still the right thing to do because:
- We shouldn't be designing around bugs, especially not bugs in 3rd party software.
- Ceramic can't work without IPFS anyway.
- The move to `go-ipfs` is right around the corner and hopefully makes this point moot.

If no one has objections to this, I'll finalize the PR. It's already taken way too long lol.

@oed @stbrody @ukstv @v-stickykeys

PR #1970, opened 2022-01-05
Verify CACAO capability in TileDocument when applying a signed commit 

haardikk21:
## Description

Adds CACAO capability verification in TileDocument when updating a tile document. Includes:
- Pins CACAO blocks on IPFS when pinning streams
- Adds CACAO verification and resource verification in TileDocument when applying commits
- Adds happy path and simple unhappy path integration tests to Canary

## How Has This Been Tested?

Two integration tests currently exist in canary integration

- [x] Test One
- Creates a determinstic TileDocument using a did:pkh
- did:pkh issues a CACAO to a did:key to update that TileDocument
- did:key updates the TileDocument

- [x] Test Two
- Creates a deterministic TileDocument using a did:pkh
- did:pkh issues a CACAO to a did:key, but does not provide the correct streamId in the resources
- did:key attempts to update the TileDocument, and it failed

## PR checklist

resolves NET-843, NET-845


packages/stream-tile-handler/package.json
```diff
@@ -17,7 +17,7 @@
     "lib"
   ],
   "scripts": {
-    "test": "../../node_modules/.bin/jest --silent --coverage",
+    "test": "../../node_modules/.bin/jest --coverage",
```
haardikk21, line 20: need to make this silent again

packages/stream-tile-handler/src/tile-document-handler.ts
```diff
@@ -63,7 +65,8 @@ export class TileDocumentHandler implements StreamHandler<TileDocument> {
       return this._applyAnchor(context, commitData, state)
     }
 
-    return this._applySigned(commitData, state, context)
+    const streamId = StreamUtils.streamIdFromState(state)
```
oed, line 68: Can't we just caclulate this right before we call `_verifySignature` instead?

packages/stream-tile-handler/src/tile-document-handler.ts
```diff
@@ -210,8 +218,31 @@ export class TileDocumentHandler implements StreamHandler<TileDocument> {
...
     commitData: CommitData,
     context: Context,
-    controller: string
+    controller: string,
+    streamId?: StreamID
   ): Promise<void> {
+    if (streamId) {
+      const protectedHeaders = commitData.envelope.signatures.map((s) => s.protected)
+      const decodedProtectedHeaders = protectedHeaders.map((pH) => base64urlToJSON(pH))
+      const capHeaderIndex = decodedProtectedHeaders.findIndex((dph) => !!dph.cap)
```
oed, line 227: Hm, I think we should assume there's only one signature here. I believe we do so in `did.verifyJWS` anyway?

packages/stream-tile-handler/src/tile-document-handler.ts
```diff
@@ -210,8 +218,31 @@ export class TileDocumentHandler implements StreamHandler<TileDocument> {
...
+      if (capHeaderIndex > -1) {
+        const capHeader = decodedProtectedHeaders[capHeaderIndex]
+        const capIPFSUri = capHeader.cap
+        const capCID = CID.parse(capIPFSUri.replace('ipfs://', ''))
+        const cacaoBlock = (await context.ipfs.dag.get(capCID)).value
```
oed, line 232: Does `(await context.ipfs.dag.get(capCID)).value` really return the block and not the value?

packages/stream-tile-handler/src/tile-document-handler.ts
```diff
@@ -210,8 +218,31 @@ export class TileDocumentHandler implements StreamHandler<TileDocument> {
   async _verifySignature(
     commitData: CommitData,
     context: Context,
-    controller: string
+    controller: string,
+    streamId?: StreamID
   ): Promise<void> {
+    if (streamId) {
```
oed, line 224: Why is this conditional on `streamId`? Should it not be conditional on if there is a `cap` prop in the JWS header?

packages/stream-tile-handler/src/tile-document-handler.ts
```diff
@@ -210,8 +218,31 @@ export class TileDocumentHandler implements StreamHandler<TileDocument> {
...
+        const cacao = cacaoBlock.value
+
+        // TODO: Verify resources here
+
+        await context.did.verifyJWS(commitData.envelope, {
+          atTime: commitData.timestamp,
+          issuer: controller,
+          disableTimecheck: commitData.disableTimecheck,
+          capability: cacao,
+        })
```
oed, line 242: Why do we call `verifyJWS` twice?

packages/stream-tile-handler/src/tile-document-handler.ts
haardikk21: Seems like it will inevitably lead to weird errors down the line...

packages/stream-tile-handler/src/tile-document-handler.ts
haardikk21: My bad, this should be the `cacao`, not `cacaoBlock`

packages/stream-tile-handler/src/tile-document-handler.ts
haardikk21: The only time `streamid` is passed is when applying a signed commit. It then only does the decoding for protected headers if we are trying to verify a signed commit

packages/stream-tile-handler/src/tile-document-handler.ts
haardikk21: I missed a `return`, will fix

packages/stream-tile-handler/src/tile-document-handler.ts
haardikk21: nevermind, i see genesis blocks can also be signed. moved it.

packages/stream-tile-handler/src/tile-document-handler.ts
oed: But how will you make sure that the CACAO object you verify is actually for the key that signed the particular protected header?

packages/stream-tile-handler/src/tile-document-handler.ts
```diff
@@ -210,8 +217,41 @@ export class TileDocumentHandler implements StreamHandler<TileDocument> {
   async _verifySignature(
     commitData: CommitData,
     context: Context,
-    controller: string
+    controller: string,
+    streamId?: StreamID
   ): Promise<void> {
+    if (StreamUtils.isSignedCommitData(commitData)) {
```
stbrody, line 223: Can you please pull all of this new functionality out into its own helper function?  Then you'll be able to use early returns to remove indentation layers and improve readability

packages/stream-tile-handler/src/tile-document-handler.ts
```diff
@@ -210,8 +217,41 @@ export class TileDocumentHandler implements StreamHandler<TileDocument> {
   async _verifySignature(
     commitData: CommitData,
     context: Context,
-    controller: string
+    controller: string,
+    streamId?: StreamID
```
stbrody, line 221: why is streamId optional?  When would you want to call `_verifySignature` without the streamid?

packages/stream-tile-handler/src/tile-document-handler.ts
```diff
@@ -210,8 +217,41 @@ export class TileDocumentHandler implements StreamHandler<TileDocument> {
   async _verifySignature(
```
stbrody, line 217: Don't forget to update the doccomments when changing a function signature!  You should be able to get your IDE to give you a warning whenever a function's signature and doccomments don't match

stbrody (commented):
a few small style suggestions

packages/stream-tile-handler/src/tile-document-handler.ts
haardikk21: when its not a signed commit we dont need it, as it cannot have a cacao attached

packages/stream-tile-handler/src/tile-document-handler.ts
haardikk21: will do! thanks :D

packages/stream-tile-handler/src/tile-document-handler.ts
haardikk21: why would it not be? tbh i havent seen a case where theres multiple signatures myself - could you elaborate on how that would happen?

packages/stream-tile-handler/src/tile-document-handler.ts
stbrody: If it's not signed, why would we be calling `_verifySignature` at all?

stbrody: looking closer, it seems like the one place we call this without passing a streamid is for validating signed genesis commits.  So I guess it's about whether and how we support capabilities that grant the ability to create new streams on the user's behalf.  While we could calculate the StreamID in `_applyGenesis`, I doubt we'll have capabilities that grant the right to create a new Stream but only if the new Stream has a specific StreamID, as that would imply knowing fully the data that would be created a-priory.  It seems more likely that we'd have capabilities that allow creating a stream so long as it has specific metadata properties, but it's fine not to worry about that for now.  So I think leaving StreamID optional and not passing it in `_applyGenesis` is fine, but it would be good to explain more about what the StreamID is used for and when it might be missing in the comments for `_verifySignature`

oed: > I doubt we'll have capabilities that grant the right to create a new Stream

We do, but not in this first iteration. I think passing the streamId from genesis should not cause any issues for now either though.


## For context from the spec:

### Format: Capability

For capability representation on IPFS we use [CACAO](https://github.com/ChainAgnostic/CAIPs/pull/74).

Caveats in CACAO are represented as URIs. We are interested in the following caveats:

- write to stream: `ceramic://<stream-id>`
- write to stream payload X, identified by `<payload-cid>` here - this gives a capability owner a permission to update a stream once: `ceramic://<stream-id>?payload=<payload-cid>`
- create a stream: `ceramic://*?payload=<payload-cid>`, where `payload-cid` is actual *payload* for non-deterministic stream genesis commit.
- Make sure that `ceramic://*` (without *payload* query param) is explicitly **not** allowed

packages/stream-tile-handler/src/tile-document-handler.ts
```diff
@@ -212,50 +213,71 @@ export class TileDocumentHandler implements StreamHandler<TileDocument> {
...
+          `Capability does not have appropriate permissions to update this TileDocument`
+        )
+      }
+
+      await context.did.verifyJWS(commitData.envelope, {
+        atTime: commitData.timestamp,
+        issuer: controller,
+        disableTimecheck: commitData.disableTimecheck,
+        capability: cacao,
+      })
```
oed, line 277: We should not need to do this here.
```suggestion
```

packages/stream-tile-handler/src/tile-document-handler.ts
```diff
@@ -212,50 +213,71 @@ export class TileDocumentHandler implements StreamHandler<TileDocument> {
...
+   * @param controller - DID value
+   * @param streamId - Stream ID for the commit
+   * @returns true if capability was present and was verified, false otherwise
+   */
+  async _verifyCapability(
+    commitData: CommitData,
+    context: Context,
+    controller: string,
+    streamId: StreamID
+  ): Promise<boolean> {
```
oed, line 248:
```suggestion
  ): Promise<Cacao | undefined> {
```

packages/stream-tile-handler/src/tile-document-handler.ts
```diff
@@ -212,50 +213,71 @@ export class TileDocumentHandler implements StreamHandler<TileDocument> {
...
-          atTime: commitData.timestamp,
-          issuer: controller,
-          disableTimecheck: commitData.disableTimecheck,
-          capability: cacao,
-        })
-        return
-      }
-    }
+    const verifiedCap = await this._verifyCapability(commitData, context, controller, streamId)
+    if (verifiedCap) return
```
oed, line 226:
```suggestion
    const cacao = await this._verifyCapability(commitData, context, controller, streamId)
```

packages/stream-tile-handler/src/tile-document-handler.ts
```diff
@@ -212,50 +213,71 @@ export class TileDocumentHandler implements StreamHandler<TileDocument> {
...
 
     await context.did.verifyJWS(commitData.envelope, {
       atTime: commitData.timestamp,
       issuer: controller,
       disableTimecheck: commitData.disableTimecheck,
     })
```
oed, line 232:
```suggestion
    await context.did.verifyJWS(commitData.envelope, {
      atTime: commitData.timestamp,
      issuer: controller,
      disableTimecheck: commitData.disableTimecheck,
      capability: cacao,
    })
```

oed (changes requested):
We can simplify this code a little by returning the parsed (and verified) cacao object from the `_verifyCapability` method. This way we only need to include the call to `verifyJWS` once.

packages/stream-tile-handler/src/tile-document-handler.ts
```diff
@@ -205,17 +213,61 @@ export class TileDocumentHandler implements StreamHandler<TileDocument> {
...
+  ): Promise<Cacao | undefined> {
+    const protectedHeader = commitData.envelope.signatures[0].protected
+    const decodedProtectedHeader = base64urlToJSON(protectedHeader)
+
+    if (decodedProtectedHeader.cap) {
+      const capIPFSUri = decodedProtectedHeader.cap
+      const capCID = CID.parse(capIPFSUri.replace('ipfs://', ''))
+      const cacao = (await context.ipfs.dag.get(capCID)).value as Cacao
+      const resources = cacao.p.resources as string[]
+      const payloadCID = commitData.cid.toString() // TODO: does this need to be a specific codec?
```
oed, line 255: Is this the CID of the JWS or the jws.link? (we should verify jws.link)

oed: btw, it does not need to be a specific codec.

packages/stream-tile-handler/src/tile-document-handler.ts
```diff
@@ -205,17 +213,61 @@ export class TileDocumentHandler implements StreamHandler<TileDocument> {
...
+      capability: cacao,
     })
   }
+
+  /**
+   * Verifies capability attached to a signed commit
+   * @param commitData - Commit to be verified
+   * @param context - Ceramic context
+   * @param streamId - Stream ID for the commit
+   * @returns true if capability was present and was verified, false otherwise
```
stbrody, line 240: return type is not boolean, comment needs to be updated

packages/stream-tile-handler/src/tile-document-handler.ts
```diff
@@ -205,17 +213,61 @@ export class TileDocumentHandler implements StreamHandler<TileDocument> {
...
+   */
+  async _verifyCapability(
+    commitData: CommitData,
+    context: Context,
+    streamId: StreamID
+  ): Promise<Cacao | undefined> {
+    const protectedHeader = commitData.envelope.signatures[0].protected
+    const decodedProtectedHeader = base64urlToJSON(protectedHeader)
+
+    if (decodedProtectedHeader.cap) {
```
stbrody, line 250: Make this `if (!decodedProtectedHeader.cap) { return null }`

Always try to early-return when you can.  Removes unnecessary indentation layers and reduced cognitive overhead when reading the code.

packages/stream-tile-handler/src/tile-document-handler.ts
```diff
@@ -205,17 +213,61 @@ export class TileDocumentHandler implements StreamHandler<TileDocument> {
...
+   * @param commitData - Commit to be verified
+   * @param context - Ceramic context
+   * @param streamId - Stream ID for the commit
+   * @returns true if capability was present and was verified, false otherwise
+   */
+  async _verifyCapability(
+    commitData: CommitData,
+    context: Context,
+    streamId: StreamID
+  ): Promise<Cacao | undefined> {
```
stbrody, line 246: nitpick, but might be better the return `null` instead of `undefined` if there's no capability?

stbrody (commented):
a few small nitpicks.  Overall I'm happy once Joel and Sergey are

packages/stream-tile-handler/src/tile-document-handler.ts
oed: Because we can't know the CID of the DagJWS inside of the CACAO, since the hash of the Cacao is inside of the DagJWS header.

packages/stream-tile-handler/src/tile-document-handler.ts
```diff
@@ -205,17 +213,61 @@ export class TileDocumentHandler implements StreamHandler<TileDocument> {
...
   }
+
+  /**
+   * Verifies capability attached to a signed commit
+   * @param commitData - Commit to be verified
+   * @param context - Ceramic context
+   * @param streamId - Stream ID for the commit
+   * @returns Cacao is capability was found and verified, null otherwise
+   */
+  async _verifyCapability(
```
oed, line 242:
```suggestion
  async _verifyCapabilityAuthz(
```
I would suggest renaming to make it clearer that we are only authorizing resources here.

oed (commented):
I don't get why the first test pass since we are not passing the correct controller anywhere.

Also, requesting a review from @smrz2001 to see if we should try to optimise IPFS access using `CommitData`?

oed:
Also, having trouble to make linear see this PR after the fact. Trying again here:
resolves NET-843, NET-845

packages/stream-tile-handler/src/tile-document-handler.ts
```diff
@@ -14,6 +14,8 @@ import {
   StreamUtils,
 } from '@ceramicnetwork/common'
 import { StreamID } from '@ceramicnetwork/streamid'
+import { base64urlToJSON } from './utils'
```
ukstv, line 17: this will break @haardikk21

haardikk21: what a catch!

packages/canary-integration/src/__tests__/capability.test.ts
```diff
@@ -0,0 +1,136 @@
...
+  const capability = Cacao.fromSiweMessage(siweMessage)
+  // Create new did:key with capability attached
+  const didKeyWithCapability = didKey.withCapability(capability)
+  await didKeyWithCapability.authenticate()
+
+  await deterministicDocument.update({ foo: 'bar' }, null, {
+    asDID: didKeyWithCapability,
+    anchor: false,
+    publish: false,
+  })
```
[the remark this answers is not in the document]
haardikk21: SIWE has `address` = `wallet.address`
SIWE to CACAO conversion will convert SIWE's `address` to CACAO's `iss`

packages/canary-integration/src/__tests__/capability.test.ts
```diff
@@ -0,0 +1,136 @@
...
+})
+
+test('verifies capability with signed commit', async () => {
+  // Create a did:pkh for the user
+  const wallet = Wallet.fromMnemonic(
+    'despair voyage estate pizza main slice acquire mesh polar short desk lyrics'
+  )
+  const didPkh = new DID({ resolver: PkhDidResolver.getResolver() })
+  // Create a determinstic tiledocument owned by the user
+  const deterministicDocument = await TileDocument.deterministic(
```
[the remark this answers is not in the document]
haardikk21: will find out as soon as i can get weird jest errors due to the new dids library release out of the way...

packages/canary-integration/src/__tests__/capability.test.ts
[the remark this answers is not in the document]
oed: Right, I know that part. But I don't think we set the controller of the stream to the correct DID PKH anywhere.

packages/canary-integration/src/__tests__/capability.test.ts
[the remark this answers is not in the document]
haardikk21: fixed during 1<>1

packages/canary-integration/src/__tests__/capability.test.ts
```diff
@@ -34,11 +34,13 @@ test('verifies capability with signed commit', async () => {
     {
       deterministic: true,
       family: 'testCapabilities',
+      controllers: [`did:pkh:eip155:1:${wallet.address}`],
     },
     {
       asDID: didPkh,
     }
```
oed, line 41: It should work without this right?

packages/canary-integration/src/__tests__/capability.test.ts
```diff
@@ -78,6 +81,64 @@ test('verifies capability with signed commit', async () => {
...
+  // Create a determinstic tiledocument owned by the user
+  const deterministicDocument = await TileDocument.deterministic(
+    ceramic,
+    {
+      deterministic: true,
+      family: 'testCapabilities',
+    },
+    {
+      asDID: didPkh,
+    }
```
oed, line 99: I don't think this is needed?

packages/canary-integration/src/__tests__/capability.test.ts
```diff
@@ -78,6 +81,64 @@ test('verifies capability with signed commit', async () => {
...
+  const didKeyWithCapability = didKey.withCapability(capability)
+  await didKeyWithCapability.authenticate()
+
+  await expect(
+    deterministicDocument.update({ foo: 'baz' }, null, {
+      asDID: didKeyWithCapability,
+      anchor: false,
+      publish: false,
+    })
+  ).rejects.toThrow()
```
oed, line 139:
```suggestion
  ).rejects.toThrow(/something something/)
```
Can we include simple regex so that we see that it throws where we expect it to throw?

packages/core/src/dispatcher.ts
```diff
@@ -97,27 +97,16 @@ export class Dispatcher {
...
-          const capMhType = await this._ipfs.hashers
-            .getHasher(capCID.multihash.code)
-            .then((mh) => mh.name)
-          await this._ipfs.block.put(cacaoBlock, {
-            format: capFormat,
-            mhtype: capMhType,
-            version: capCID.version,
-          })
+          const putOptions = await Utils.getIPFSBlockPutOptionsFromCID(capIPFSUri, this._ipfs)
+          await this._ipfs.block.put(cacaoBlock, putOptions)
```
oed, line 101: I would just put the `_ipfs.block.put` part inside of the util function as well.

packages/core/src/utils.ts
```diff
@@ -100,6 +100,22 @@ export class Utils {
     if (!commitData.commit.prev) commitData.type = CommitType.GENESIS
     return commitData
   }
+
+  /**
+   * Returns ipfs block PutOptions given a CID instance
+   * @param cid the CID to get the block PutOptions for
+   */
+  static async getIPFSBlockPutOptionsFromCID(cid: CID | string, ipfsApi: IpfsApi) {
```
oed, line 108:
```suggestion
  static async putIPFSBlock(cid: CID | string, block: Uint8Array, ipfsApi: IpfsApi): void {
```

packages/core/src/utils.ts
```diff
@@ -100,6 +100,22 @@ export class Utils {
...
+    if (typeof cid === 'string') cid = CID.parse(cid.replace('ipfs://', ''))
+    const format = await ipfsApi.codecs.getCodec(cid.code).then((f) => f.name)
+    const mhtype = await ipfsApi.hashers.getHasher(cid.multihash.code).then((mh) => mh.name)
+    const version = cid.version
+    return {
+      format,
+      mhtype,
+      version,
+    }
```
oed, line 117:
```suggestion
    await ipfsApi.block.put(block, {
      format,
      mhtype,
      version,
    })
```

oed (changes requested):
Looks mostly good, some minor changes suggested.

packages/canary-integration/src/__tests__/capability.test.ts
```diff
@@ -0,0 +1,180 @@
...
+      anchor: false,
+      publish: false,
+    })
+  ).rejects.toThrow(/invalid_jws/)
+}, 30000)
+
+test('fails to verify capability with invalid resource', async () => {
+  // Create a did:pkh for the user
+  const wallet = Wallet.fromMnemonic(
+    'despair voyage estate pizza main slice acquire mesh polar short desk lyrics'
```
stbrody, line 133: Can we use a different mnemonic for each test?  or use non-deterministic TileDocuments?  I'm worried about the different tests operating on the same deterministic document and creating dependencies between the tests

stbrody (approved):
lgtm mod making sure that test cases are all independent

packages/canary-integration/src/__tests__/capability.test.ts
oed: We can't use non-deterministic Tiles with DID PKH yet. We could however use different metadata families.

packages/canary-integration/src/__tests__/capability.test.ts
```diff
@@ -0,0 +1,180 @@
...
+test('verifies capability with signed commit', async () => {
+  // Create a did:pkh for the user
+  const wallet = Wallet.fromMnemonic(
+    'despair voyage estate pizza main slice acquire mesh polar short desk lyrics'
+  )
+  const didPkh = new DID({ resolver: PkhDidResolver.getResolver() })
+  // Create a determinstic tiledocument owned by the user
+  const deterministicDocument = await TileDocument.deterministic(ceramic, {
+    deterministic: true,
+    family: 'testCapabilities',
```
oed, line 34:
```suggestion
    family: 'testCapabilities1',
```

packages/canary-integration/src/__tests__/capability.test.ts
```diff
@@ -0,0 +1,180 @@
...
+test('does not allow updating if cacao issuer is not document controller', async () => {
+  // Create a did:pkh for the user
+  const wallet = Wallet.fromMnemonic(
+    'despair voyage estate pizza main slice acquire mesh polar short desk lyrics'
+  )
+  const didPkh = new DID({ resolver: PkhDidResolver.getResolver() })
+  // Create a determinstic tiledocument owned by the user
+  const deterministicDocument = await TileDocument.deterministic(ceramic, {
+    deterministic: true,
+    family: 'testCapabilities',
```
oed, line 87:
```suggestion
    family: 'testCapabilities2',
```

packages/canary-integration/src/__tests__/capability.test.ts
```diff
@@ -0,0 +1,180 @@
...
+test('fails to verify capability with invalid resource', async () => {
+  // Create a did:pkh for the user
+  const wallet = Wallet.fromMnemonic(
+    'despair voyage estate pizza main slice acquire mesh polar short desk lyrics'
+  )
+  const didPkh = new DID({ resolver: PkhDidResolver.getResolver() })
+  // Create a determinstic tiledocument owned by the user
+  const deterministicDocument = await TileDocument.deterministic(ceramic, {
+    deterministic: true,
+    family: 'testCapabilities',
```
oed, line 139:
```suggestion
    family: 'testCapabilities3',
```

oed (commented):
To address spencers concern around using the same tile in all tests.

packages/canary-integration/src/__tests__/capability.test.ts
```diff
@@ -136,7 +124,7 @@ test('does not allow updating if cacao issuer is not document controller', async
       anchor: false,
       publish: false,
     })
-  ).rejects.toThrow()
+  ).rejects.toThrow(/invalid_jws/)
```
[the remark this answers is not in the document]
haardikk21: No, it falls back to the same error as https://github.com/ceramicnetwork/js-did/blob/a47bb5a626b5bceed1b9999509f6583c4d5f44ae/src/did.ts#L330

oed:
@haardikk21 can you address my last comment and merge develop into this branch? Then we can merge this!

oed:
The new Utils function used in `packages/core/src/dispatcher.ts` need to be updated to account for the AbortControllers introduced in #2033

oed (approved):
lgtm once tests pass

Issue #1980, opened 2022-01-09
URGENT: Remove @overnightjs/logger or enforce package-lock.json


vaultec81:
**Describe the bug**
Recently, the maintainer of the `colors` npm package has decided to go rogue. Read me [here](https://github.com/Marak/colors.js/commit/074a0f8ed0c31c35d13d28632bd8a049ff136fb6) and [here](https://www.reddit.com/r/javascript/comments/rz56qu/marak_adds_infinite_loop_to_colorsjs/)
The package `@overnightjs/logger` used in Ceramic depends on `colors`

**To Reproduce**

The nodejs code below will demonstrate this issue.  <h3>**NOTE: you will need to do a fresh install with no package-lock.json or already existing node_modules.**.</h6>
```

import CeramicHTTP from '@ceramicnetwork/http-client'

const ceramic = new CeramicHTTP('<ceramic API url>')

```

**Screenshots**
After running the above code

![image](https://user-images.githubusercontent.com/47548474/148665415-179331cd-8377-42c1-9c16-ca74e8fe56a6.png)

And thousands of lines of garbage data being put out.

**Ceramic versions** 
As far as I am aware this affects all ceramic versions. Any version that uses `@overnightjs/logger` or `colors` npm package directly or indirectly.

I was running Ceramic http client v 1.4.0 but it also happens on latest.

A quick work around is having a package-lock.json, however, new installs will not have this available.

v-stickykeys:
Have been able to reproduce the issue as well.

v-stickykeys:
Create a PR and you can use this branch to run js-ceramic without the issue until we create a new release.

oed:
Thank you for reporting this @vaultec81! I believe the issue should now be addressed. Can you confirm?

oed:
Separate note: Seems like `@overnight/logger` has been deprecated: https://github.com/seanpmaxwell/overnight#overnightjs

We should make sure to replace this with something different. @ukstv @stephhuynh18 @stbrody

vaultec81:
> Thank you for reporting this @vaultec81! I believe the issue should now be addressed. Can you confirm?

Yep! Fixed on my end!

ukstv:
> Separate note: Seems like `@overnight/logger` has been deprecated: https://github.com/seanpmaxwell/overnight#overnightjs
> 
> We should make sure to replace this with something different. @ukstv @stephhuynh18 @stbrody

[...]

PR #1981, opened 2022-01-09
hotfix(common): lock colors dependency to stop infinite loop

v-stickykeys:
As reported by Ceramic community member [here](https://github.com/ceramicnetwork/js-ceramic/issues/1980) and in Discord, the maintainer of [colors](https://github.com/Marak/colors.js) pushed an [infinite loop](https://github.com/Marak/colors.js/commit/074a0f8ed0c31c35d13d28632bd8a049ff136fb6) into the code base. The module is used by one of our logging dependencies called [overnight](https://github.com/seanpmaxwell/overnight).

This PR adds a package-lock file to `js-ceramic/packages/common` in order to fix the colors version to an earlier version that is compatible with overnight but does not include this... weirdness.

Tested this locally by doing
```
npm install
npm run build
node packages/cli/bin/ceramic.js daemon
```

### DIFF
--- a/packages/common/package-lock.json
+++ b/packages/common/package-lock.json
@@ -1,0 +1,2165 @@
+{
+  "name": "@ceramicnetwork/common",
+  "version": "1.10.2-rc.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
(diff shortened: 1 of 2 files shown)


smrz2001 (approved):
🚀

PR #1985, opened 2022-01-12
Add Deprecation notice to all functions that have been moved to the Glaze CLI

Sterahi:
## Deprecate all functions that have been transitioned over to the Glaze CLI.

We are now providing an alert after most functions that looks like the following: 
 
Screen Shot 2022-01-12 at 11 24 49 AM

The updated help looks like the following: 
Screen Shot 2022-01-12 at 11 24 05 AM

## How Has This Been Tested?

The deprecation notice has been run multiple times to ensure that it behaves properly & ANSI escapes were used to ensure no issues with colors/different Operating Systems.

## PR checklist

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -60,7 +60,7 @@
     "morgan": "^1.10.0",
     "nft-did-resolver": "^1.0.0",
     "picocolors": "^1.0.0",
-    "pkh-did-resolver": "^0.3.3",
+    "pkh-did-resolver": "^0.3.8-rc.0",
     "reflect-metadata": "^0.1.13",
     "s3leveldown": "^2.2.1",
     "safe-did-resolver": "^0.0.5",
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -59,6 +59,7 @@
     "levelup": "^4.4.0",
     "morgan": "^1.10.0",
     "nft-did-resolver": "^1.0.0",
+    "picocolors": "^1.0.0",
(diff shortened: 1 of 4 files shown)


PaulLeCam (commented):
LGTM but I'm not familiar with how well formatting encoding is supported by different OSes and shells, aren't there differences to take into accounts or at least support checks?

ukstv:
Also, raw colour-chars output could potentially break on a terminal, that does not support colors. So, it maybe is a good idea to use `picocolors` or `kleur` for colors: they both have a decent function to determine if a terminal supports colors.

Sterahi:
@ukstv Implemented picocolors. thanks for pointing it out! :)

packages/cli/package.json
```diff
@@ -59,7 +59,8 @@
     "levelup": "^4.4.0",
     "morgan": "^1.10.0",
     "nft-did-resolver": "^1.0.0",
-    "pkh-did-resolver": "^0.3.8-rc.0",
+    "picocolors": "^1.0.0",
+    "pkh-did-resolver": "^0.3.3",
```
ukstv, line 63: @Sterahi mind I ask you, please, why pkh-did-resolver version has changed?

Sterahi: I have no idea, I've reset it to it's previous iteration though!

ukstv (approved):
🏇

stbrody (approved):
lgtm, thanks!

PR #2003, opened 2022-01-19
feat: Cache chainId requests to providers

ukstv:
From [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap):

> A WeakMap is a collection of key/value pairs whose keys must be objects, with values of any arbitrary JavaScript type, and which does not create strong references to its keys. That is, an object’s presence as a key in a WeakMap does not prevent the object from being garbage collected. Once an object used as a key has been collected, its corresponding values in any WeakMap become candidates for garbage collection as well — as long as they aren't strongly referred to elsewhere.

Here we utilise WeakMap to effectively attach chainId property to an ethereum provider, as long as the provider is available somewhere in JS memory.

A chainId value is set in the weak map the first time chainId is requested. It stays cached as long as provider object is available, i.e. until it is garbage-collected. Similar to LRU Cache, with a difference being expulsion of cached value is based on key being garbage collected.

It seems to be supported widely enough, so we should consider it as safe to use: https://caniuse.com/?search=weakmap

### DIFF
--- a/packages/blockchain-utils-linking/src/__tests__/ethereum.test.ts
+++ b/packages/blockchain-utils-linking/src/__tests__/ethereum.test.ts
@@ -60,7 +60,7 @@ let provider: any
 let addresses: string[]
 let contractAddress: string
 
-beforeAll(async () => {
+beforeEach(async () => {
   provider = ganache.provider(GANACHE_CONF)
   addresses = await send(provider, encodeRpcMessage('eth_accounts'))
   // ganache-core doesn't support personal_sign -.-
@@ -84,7 +84,7 @@ beforeAll(async () => {
   MockDate.set('2018-10-01') // So that the anchors happen at a predictable blockNumber/blockTimestamp
 })
 
-afterAll(() => {
+afterEach(() => {
   MockDate.reset()
   jest.clearAllMocks()
 })
--- a/packages/blockchain-utils-linking/src/__tests__/ethereum.test.ts
+++ b/packages/blockchain-utils-linking/src/__tests__/ethereum.test.ts
@@ -92,20 +92,20 @@ afterAll(() => {
 describe('isEthAddress', () => {
   test('detect eth address correctly', async () => {
     const notEthAddr = '0xabc123'
-    expect(await ethereum.isEthAddress(notEthAddr)).toBeFalsy()
-    expect(await ethereum.isEthAddress(addresses[0])).toBeTruthy()
+    expect(ethereum.isEthAddress(notEthAddr)).toBeFalsy()
+    expect(ethereum.isEthAddress(addresses[0])).toBeTruthy()
   })
 })
 
 describe('isERC1271', () => {
   test('detect erc1271 address', async () => {
     const acc1 = new AccountId({ address: addresses[0], chainId: 'eip155:1' })
-    expect(await ethereum.isERC1271(acc1, provider)).toEqual(false)
+    await expect(ethereum.isERC1271(acc1, provider)).resolves.toEqual(false)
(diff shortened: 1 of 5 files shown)


oed (approved):
Seems like a good solution. lgtm

packages/blockchain-utils-linking/src/__tests__/ethereum.test.ts
```diff
@@ -165,6 +167,29 @@ describe('EthereumAuthProvider', () => {
     const auth = new ethereum.EthereumAuthProvider(provider, address)
     await expect(auth.withAddress(address).createLink(testDid)).resolves.toMatchSnapshot()
   })
+  test('accountId', async () => {
+    const address = addresses[0]
+    const auth = new ethereum.EthereumAuthProvider(provider, address)
+    const providerSpy = jest.spyOn(provider, 'sendAsync')
+    const results = []
+    // No matter how many times `::accountId` is called
+    for (let m = 0; m <= Math.floor(Math.random() * 10) + 1; m++) {
```
stbrody, line 176: I'm always very suspicious of randomness in tests unless there is a very strong rationale for doing so.  Will this test be meaningfully less valuable if we just hard coded this at, say, 10 requests?

Non-determinism in tests makes them very hard to reliably debug and reproduce failures.

packages/blockchain-utils-linking/src/ethereum.ts
```diff
@@ -16,11 +16,23 @@ type EthProviderOpts = {
...
+
 /**
  *  AuthProvider which can be used for Ethereum providers with standard interface
  */
 export class EthereumAuthProvider implements AuthProvider {
   readonly isAuthProvider = true
+  private _accountId: AccountId | undefined
 
   constructor(
     private readonly provider: any,
```
stbrody, line 38: the EthereumAuthProvider maintains a single long-lived instance of the `provider`.  Given that, why keep a map of provider to chainid at all, instead of caching a single chainid in the EthereumAuthProvider instance itself?

stbrody: Follow-up: I see that there is an exported free function `createLink` that can be called without an instance of EthereumAuthProvider.  Do you know why this is?  The other blockchains in blockchain-utils-linking don't seem to have corresponding free functions - they all require creating the link to go through an instance of the provider type.

Can we remove the free-function `createLink` in favor of `EthereumAuthProvider.createLink`, and by so doing remove the need to use WeakMap and instead just cache the single chainid used by the single `provider` instance that `EthereumAuthProvider` uses?

oed: Not sure exactly how these APIs are used upstream, but I think we should try to avoid making breaking changes right now. I think there's an opportunity to revamp the AuthProvider API once we start relying more on CACAO for authentication, but for now I think we should keep apis the same.

stbrody: FWIW this function isn't part of the AuthProvider API, its specific to EthereumAuthProvider.  But anyway your point is still fair.  We can leave it for now, it just makes this caching code more involved than it needs to be but that's fine.

stbrody (approved):
lgtm mod optional-but-encouraged suggestion to remove randomness from unit test

packages/blockchain-utils-linking/src/__tests__/ethereum.test.ts
ukstv: Changed it to be deterministic.

PR #2008, opened 2022-01-20
feat(ipfs-daemon): create ipfs daemon using ipfsd ctl

stephhuynh18:
## Description

Remove all repo creation
repo is created at ipfsPath

## How Has This Been Tested?

Tests pass
IpfsDaemon can be run. Objects can be put and retrieved


packages/ipfs-daemon/src/ipfs-daemon.ts
```diff
@@ -1,32 +1,20 @@
-import { create as createIPFS } from 'ipfs-core'
-import { HttpApi } from 'ipfs-http-server'
-import { HttpGateway } from 'ipfs-http-gateway'
+import { createController } from './create-ipfs'
```
ukstv, line 1:
```suggestion
import { createController } from './create-ipfs.js'
```

PR #2009, opened 2022-01-20
feat(core): Throw error if commit rejected by conflict resolution

stbrody:


packages/core/src/conflict-resolution.ts
```diff
@@ -355,6 +355,12 @@ export class ConflictResolution {
 
     const selectedState = await pickLogToAccept(localState, remoteState)
     if (selectedState === localState) {
+      if (opts.throwOnInvalidCommit) {
```
oed, line 358: I'm confused. Where is this option passed?

stbrody: yeah sorry I should have called that out.

This option was added a few months ago - it's set whenever we're applying a commit sent from the application, but not when applying commits that came in from pubsub (or anchor commits learned from polling the CAS).  It's purpose is basically exactly this - when we're applying a commit for an application client we want errors to get back to the user, but for background operations we just want the commit to fail to apply and move on

smrz2001 (approved):
🚢

packages/core/src/conflict-resolution.ts
oed: Ok. Why did we add the option but not the code to throw? Do we throw in other places, but not here perhaps?

oed (approved):

PR #2010, opened 2022-01-21
feat(ipfs-daemon): remove builds of ipfs daemon

v-stickykeys:
## Description

No longer using js-ipfs so `Dockerfile.ipfs-daemon` is irrelevant. This PR removes it and related workflows that built the image.

Also removed info about running js-ipfs from the README of the ipfs-daemon package.


smrz2001 (approved):
:shipit:

PR #2019, opened 2022-01-27
feat: Add PL-managed list of bootstrap nodes to embedded go-ipfs instance

ukstv:


stbrody (approved):
Fine for now, though I do wish there was a way to offload this to ipfs so we won't have to make a change here ourselves if the PL provided bootstrap list changes in the future

PR #2025, opened 2022-01-31
feat(core,stream-tile,stream-caip10-link): Pin streams by default

stbrody:


PaulLeCam:
@stbrody nice! How does it work for deterministic streams please? Will accessing a stream pin it implicitly or does it still need to be pinned explicitly on first update?

stbrody:
> @stbrody nice! How does it work for deterministic streams please? Will accessing a stream pin it implicitly or does it still need to be pinned explicitly on first update?

`TileDocument.deterministic` uses the default create opts, not the defaault load opts (for better or worse), which means that loading a determinstic stream will pin the stream by default

PR #2033, opened 2022-02-08
chore(core): Interrupt blocked ipfs network i/o at shutdown

stbrody:
If we have outstanding requests to ipfs during shutdown, especially ones that will take along time, like loading a CID that doesn't exist, then shutdown blocks until those outstanding requests terminate.  We should be able to use an AbortController to proactively abort outstanding networking i/o when we begin shutdown.

### DIFF
--- a/packages/common/src/context.ts
+++ b/packages/common/src/context.ts
@@ -5,12 +5,10 @@ import type { IpfsApi, LoggerProvider } from './index.js'
 
 /**
  * Encapsulates Ceramic context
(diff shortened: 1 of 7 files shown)


oed:
Nice. Can we merge this after #1970 which refactors block.put code in the dispatcher.

ukstv (approved):
niice

oed:
@ukstv was hoping this could get merged before #1970 Now we need to take this change into account in that PR.

PR #2035, opened 2022-02-09
chore: use cross-platform rimraf

ukstv:
So that it can work on windows.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,6 +27,7 @@
         "jest-resolver-enhanced": "^1.0.1",
         "lerna": "^4.0.0",
         "prettier": "^2.5.1",
+        "rimraf": "^3.0.2",
         "typedoc": "^0.22.10",
         "typedoc-plugin-pages": "^1.1.0",
         "typescript": "^4.5.4"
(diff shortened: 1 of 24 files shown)


smrz2001:
🚢 🚀

ukstv:
@smrz2001 does it really work for you this way? I am worried, if `rimraf` is not installed globally, the main command would fail right away. There is no `rimraf` executable in this case.

smrz2001:
Ahh, I see. It does not work without global rimraf but is that an issue? Without removing node_modules we won't be able to get a clean install.

smrz2001:
But let's discuss because I do see your point.

ukstv:
Maybe we could `rimraf ./node_modules` the last thing though. It should first load `rimraf` into memory, so no files are required.

smrz2001:
I have another idea but it's a bit ugly. We could replace the top-level rimraf with (rm -rf node_modules || true) && (del node_modules || true).

ukstv:
I like it, really, but I have no idea how to avoid messages like `sh: del: command not found` in a cross-platform fashion.

smrz2001:
Made some updates to my PR @ukstv. Adding `npx` before `rimraf` will download and cache `rimraf` if it is not already installed.

Please check.

PR #2037, opened 2022-02-09
fix(core): Re-enable dispatcher-real-ipfs.test.ts

stbrody:


ukstv:
@stbrody You sure this deserves a whole test file named as pretentiously as `dispatcher-real-ipfs` after these changes? It basically tests (1) if dispatcher can store and retrieve commit, (2) it retries few times. I can not see anything particularly _real_ here. Maybe we could move the tests to the dispatcher test suite?

stbrody:
> @stbrody You sure this deserves a whole test file named as pretentiously as `dispatcher-real-ipfs` after these changes? It basically tests (1) if dispatcher can store and retrieve commit, (2) it retries few times. I can not see anything particularly _real_ here. Maybe we could move the tests to the dispatcher test suite?

haha, I didn't realize a file name could be "pretentious" :stuck_out_tongue: 

The "real" here is referring to the fact that this test uses an actual ipfs instance.  As opposed to dispatcher.test.ts, which uses a mock-ed out ipfs.  I could rename dispatcher.test.ts to dispatcher-mock-ipfs.test.ts to create more symmetry and clarity between the two file names?

If I moved these tests to be in the dispatcher.test.ts file, they would still have to have a completely independent test fixture and setup/teardown, so I don't really see the value in putting them in the same file when they really are completely independent test setups.

ukstv (approved):

ukstv:
> As opposed to dispatcher.test.ts, which uses a mock-ed out ipfs

Now I see. Thanks! Approved.

PR #2048, opened 2022-02-15
chore (common): use jet-logger 

stephhuynh18:
## Description

Change it back to jet-logger from overjet-logger

## How Has This Been Tested?

Tests passing


oed (approved):
lgtm

PR #2049, opened 2022-02-15
chore: Revive tests with js-ipfs

ukstv:
It works. One caveat: had to add `--forceExit` to every jest invocation. Guess, it is because js-ipfs does not clean all the resources after stopped.

### DIFF
--- a/packages/core/src/__tests__/ceramic-api.test.ts
+++ b/packages/core/src/__tests__/ceramic-api.test.ts
@@ -433,7 +433,7 @@ describe('Ceramic API', () => {
       expect(Object.keys(streams).length).toEqual(6)
     })
 
-    it('can load streams for array of multiqueries even if streamid or path throws error', async () => {
+    it.skip('can load streams for array of multiqueries even if streamid or path throws error', async () => {
       const queries = [
         {
           streamId: streamA.id,
--- a/packages/core/src/__tests__/ceramic-api.test.ts
+++ b/packages/core/src/__tests__/ceramic-api.test.ts
@@ -320,7 +320,7 @@ describe('Ceramic API', () => {
     afterAll(async () => {
       await ceramic.close()
       await tmpFolder.cleanup()
-    })
+    }, 10000)
 
     it('can load linked stream path, returns expected form', async () => {
       const streams = await ceramic._loadLinkedStreams(
--- a/packages/core/src/__tests__/ceramic-api.test.ts
+++ b/packages/core/src/__tests__/ceramic-api.test.ts
@@ -1,4 +1,3 @@
-import { jest } from '@jest/globals'
 import tmp from 'tmp-promise'
 import { Ceramic, CeramicConfig } from '../ceramic.js'
 import { Ed25519Provider } from 'key-did-provider-ed25519'
@@ -35,7 +34,6 @@ const generateStringOfSize = (size): string => {
 }
 
 describe('Ceramic API', () => {
-  jest.setTimeout(60000)
   let ipfs: IpfsApi
 
   const stringMapSchema = {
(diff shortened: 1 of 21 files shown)


packages/core/src/__tests__/ceramic-api.test.ts
```diff
@@ -435,7 +433,7 @@ describe('Ceramic API', () => {
       expect(Object.keys(streams).length).toEqual(6)
     })
 
-    it('can load streams for array of multiqueries even if streamid or path throws error', async () => {
+    it.skip('can load streams for array of multiqueries even if streamid or path throws error', async () => {
```
stephhuynh18, line 436: is this skipped purposefully?

packages/core/src/__tests__/dispatcher-real-ipfs.test.ts
```diff
@@ -72,7 +73,7 @@ describe('Dispatcher with real ipfs over http', () => {
     const ipfsGetSpy = jest.spyOn(ipfsClient.dag, 'get')
 
     // try to load a CID that ipfs doesn't know about.  It will timeout
```
stbrody, line 75:
```suggestion
    // try to load a CID that ipfs doesn't know about.  It will timeout.
    // Timeout error message is different depending on if we are talking to a go-ipfs or js-ipfs instance
```

packages/core/src/__tests__/dispatcher-real-ipfs.test.ts
```diff
@@ -86,7 +87,8 @@ describe('Dispatcher with real ipfs over http', () => {
     // shutdownController successfully interrupted waiting on IPFS.
 
     const getPromise = dispatcher.retrieveCommit(FAKE_CID)
+    await delay(300) // Apparently, with js-ipfs, we need some extra time.
```
stbrody, line 90: what happens if this delay isn't there?  It really shouldn't matter even if the shutdownController was triggered entirely before the call to retrieveCommit so I'd like to understand this better

stbrody (commented):
one question about an apparent race condition in a test I'd like to understand more

packages/core/src/__tests__/ceramic-api.test.ts
ukstv: Ooh, good catch. Should not have skipped it.

packages/core/src/__tests__/dispatcher-real-ipfs.test.ts
ukstv: While it should not, in reality, it does. When testing with js-ipfs, signal is triggered before the `dag.get` runs. It seems, that whatever is inside `dag.get` does not respect already triggered abort signal. The delay makes sure that `dag.get` starts first, and only then the signal is triggered.

stbrody: so what happens?  Awaiting the `getPromise` hangs and the test times out?

ukstv: Hey @stbrody could I merge that?

stbrody: Okay, I just checked and even if you reorder the lines of this test to put `shutdownController.abort()` *before* the call to `dispatcher.retrieveCommit`, the test works fine against go-ipfs.  So it seems like this is a js-ipfs specific bug that they don't handle already-activated abortSignals.  Can we check the test configuration and put the sleep under an if-statement so it only runs if we're in js-ipfs mode, but doesn't sleep if we're in go-ipfs mode?

Might also want to increase the sleep to 1 second just to make the test less likely to be flaky on a machine with different performance characteristics.

ukstv: There are two ways to check if the test runs with js-ipfs. One is to look at an environment variable, another is to check functions specific to IPFS node. I believe, using an env variable creates unnecessary connascence. Added a check using ipfs-node-specific `preload` method.

stbrody (approved):
LGTM

PR #2053, opened 2022-02-17
NET-891 update request capability method to use cacao

haardikk21:
## Description

Update the Ethereum Auth Provider `requestCapability` method to use CACAO, instead of the earlier hardcoded solution. 

## How Has This Been Tested?

- [x] Updated pre-existing requestCapability test

## PR checklist


haardikk21:
Resolves NET-891

packages/blockchain-utils-linking/src/ethereum.ts
```diff
@@ -61,25 +61,18 @@ export class EthereumAuthProvider implements AuthProvider {
     return createLink(did, accountId, this.provider, this.opts)
   }
 
-  async requestCapability(params: OcapParams): Promise<Ocap> {
+  async requestCapability(siweMessage: SiweMessage): Promise<Cacao> {
```
oed, line 64:
```suggestion
  async requestCapability(sessionDID: string, streams: Array<StreamId | string>): Promise<Cacao> {
```
I would like to simplify the method a lot here. Really most things inside of the SIWE message should be provided by this method!

oed (changes requested):
We can make the api a lot simpler for developers.

packages/blockchain-utils-linking/src/ethereum.ts
haardikk21: uhh i dont agree.

Nonces should be managed by the dApp, they could use them for linking to other information.
Expiry and not before times should be managed by the dApp
the statement and domain should be managed by the dApp

any defaults we choose here will be opinionated for our use case for 3IDConnectV2 but for devs looking to implement their own capabilities, especially down the road with delegation, it will become a problem

oed: If apps are looking to link additional information / their own capabilities then there is no need for them to use this library. As it stands right now the only thing this method really does is sign the message and output a CACAO. That's doesn't provide much meaningful help to a dev. 

The purpose of this library is really an easy entrypoint for devs. However, we could provide an optional param like so:
```suggestion
  interface CapOpts {} // optional parameters to change our defaults for SIWE
  async requestCapability(sessionDID: string, streams: Array<StreamId | string>, opts: CapOpts): Promise<Cacao> {
```

The defaults I think we should provide:
```js
new SiweMessage({
      domain: window ? window.location.hostname : throw Error(...)
      address: address, // from provider obv
      statement: 'Give the current application access to some of your data on Ceramic', // copy can be updated later
      uri: sessionDID,
      version: '1',
      nonce: random(),
      issuedAt: // now
      expires: // one week, or some other sensible default
      chainId: // from provider
      resources: streams
    })
```

haardikk21: If we want to avoid confusion, we can wrap SiweMessage in a different container which better reflects the use case. How "Sign in with Ethereum" relates to capabilities is a little involved. Why 'uri' is a DID is also a little involved. I like the suggestion of `sessionDID`, but I don't want to enforce defaults that may not work for devs.

haardikk21: I like the options approach much better. Will do.

oed: Ok, great :)

packages/blockchain-utils-linking/src/ethereum.ts
```diff
@@ -61,10 +63,40 @@ export class EthereumAuthProvider implements AuthProvider {
...
+    sessionDID: string,
+    streams: Array<StreamID | string>,
+    opts: CapabilityOpts
+  ): Promise<Cacao> {
     console.warn(
       'WARN: requestCapability os an experimental API, that is subject to change any time.'
     )
+
+    const domain = typeof window !== 'undefined' ? window.location.hostname : opts.domain
+    if (!domain) throw new Error("Missing parameter 'domain'")
```
oed, line 76: 👍

packages/blockchain-utils-linking/src/util.ts
```diff
@@ -22,6 +22,14 @@ export interface ConsentMessage {
   timestamp?: number
 }
 
+export interface CapabilityOpts {
+  domain?: string
+  statement?: string
+  version?: string
+  nonce?: string
+  requestId?: string
+}
```
oed, line 31:
```suggestion
  requestId?: string
  expirationTime? string
  resources? Array<string>
}
```

packages/blockchain-utils-linking/src/ethereum.ts
```diff
@@ -61,10 +63,40 @@ export class EthereumAuthProvider implements AuthProvider {
     return createLink(did, accountId, this.provider, this.opts)
   }
 
-  async requestCapability(siweMessage: SiweMessage): Promise<Cacao> {
+  async requestCapability(
+    sessionDID: string,
+    streams: Array<StreamID | string>,
+    opts: CapabilityOpts
```
oed, line 69:
```suggestion
    opts: CapabilityOpts = {}
```

packages/blockchain-utils-linking/src/ethereum.ts
```diff
@@ -61,10 +63,40 @@ export class EthereumAuthProvider implements AuthProvider {
...
+      statement: opts.statement
+        ? opts.statement
+        : 'Give this application access to some of your data on Ceramic',
+      uri: sessionDID,
+      version: opts.version ? opts.version : '1',
+      nonce: opts.nonce ? opts.nonce : randomString(10),
+      issuedAt: now.toISOString(),
+      expirationTime: oneWeekLater.toISOString(),
+      chainId: (await this.accountId()).chainId.toString(),
+      resources: streams.map((s) => (typeof s === 'string' ? s : s.toUrl())),
```
oed, line 95:
```suggestion
      expirationTime: opts.expirationTime || oneWeekLater.toISOString(),
      chainId: (await this.accountId()).chainId.toString(),
      resources: (opts.resources || []).concat(streams.map((s) => (typeof s === 'string' ? s : s.toUrl()))),
```

packages/blockchain-utils-linking/src/ethereum.ts
```diff
@@ -61,10 +63,40 @@ export class EthereumAuthProvider implements AuthProvider {
...
+
+    const siweMessage = new SiweMessage({
+      domain: domain,
+      address: this.address,
+      statement: opts.statement
+        ? opts.statement
+        : 'Give this application access to some of your data on Ceramic',
+      uri: sessionDID,
+      version: opts.version ? opts.version : '1',
+      nonce: opts.nonce ? opts.nonce : randomString(10),
```
oed, line 91:
```suggestion
      statement: opts.statement || 'Give this application access to some of your data on Ceramic',
      uri: sessionDID,
      version: opts.version || '1',
      nonce: opts.nonce || randomString(10),
```

oed (changes requested):
Looks great now. Some small nitpicks and additional options.

packages/blockchain-utils-linking/src/ethereum.ts
haardikk21: I did not add `expirationTime` to opts because what if someone creates invalid expiry dates? Also, they are now responsible for formatting the string the way we want. Invalid formatting might still work but lead to weird results.

I suggested using a proper datetime library to handle time durations properly instead of exact timestamps of expiry.

packages/blockchain-utils-linking/src/ethereum.ts
haardikk21: reminded me of this - https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx
off the top of my head thought it also applied to `||`
perhaps a stronger approach is to use `??`

packages/blockchain-utils-linking/src/ethereum.ts
oed: I expect people will request this feature if we don't add it now. But we don't need to include it in this PR.

packages/blockchain-utils-linking/src/util.ts
haardikk21: don't want resources directly as a function argument?

packages/blockchain-utils-linking/src/ethereum.ts
oed: Thoughts on this @ukstv ?

packages/blockchain-utils-linking/src/util.ts
oed: Well people might want to include resources that are not streamIds :)

haardikk21: hmmmmmm

packages/blockchain-utils-linking/src/ethereum.ts
ukstv: IMO, it makes sense to leave `||` assuming we can rely on TS type system.

packages/blockchain-utils-linking/src/ethereum.ts
```diff
@@ -61,25 +63,48 @@ export class EthereumAuthProvider implements AuthProvider {
     return createLink(did, accountId, this.provider, this.opts)
   }
 
-  async requestCapability(params: OcapParams): Promise<Ocap> {
+  async requestCapability(
+    sessionDID: string,
+    streams: Array<StreamID | string>,
```
ukstv, line 68: Distant future feature request: we'll need to have a DSL to specify streams as resources here.

ukstv: For now, giving devs freedom to specify all the SIWE payload as Joel suggested would suffice, IMO.

packages/blockchain-utils-linking/src/ethereum.ts
haardikk21: going to remove the ternary but going to use `??`

oed (approved):
lgtm

packages/blockchain-utils-linking/src/__tests__/__snapshots__/ethereum.test.ts.snap
```diff
@@ -18,20 +18,27 @@ Timestamp: 1538352000",
...
-  "signature": "0xd008043db6b85beffb4dc0c8d1371f55ac1d1f31716385f7b5d14457a700453309d2fe9f0f2278140638b000877b7b073f036c34eb0efdd23e6768cdc02e72761c",
+  "h": Object {
+    "t": "eip4361-eip191",
+  },
+  "p": Object {
+    "aud": "did:key:z6MkrBdNdwUPnXDVD1DCxedzVVBpaGi8aSmoXFAeKNgtAer8",
+    "domain": "https://service.org/",
+    "exp": "2018-10-08T00:00:00.000Z",
+    "iat": "2018-10-01T00:00:00.000Z",
+    "iss": "did:pkh:eip155:eip155:1337:0x8fe2c4516e920425e177658aaac451ca0463ed69",
```
oed, line 29:
```suggestion
    "iss": "did:pkh:eip155:1337:0x8fe2c4516e920425e177658aaac451ca0463ed69",
```
Hm, something went wrong in this one.

oed (commented):
Bug in the CACAO library?
If so I guess it can be fixed in the PR you have to update CACAO lib.

packages/blockchain-utils-linking/src/__tests__/__snapshots__/ethereum.test.ts.snap
haardikk21: bug in the cacao library. thanks for catching this. merging for now, will fix in update cacao

oed (approved):
Feel free to merge whenever. As long as cacao library is updated in a later PR.

PR #2058, opened 2022-02-22
chore: Use newer go-ipfs

ukstv:
- Update to go-ipfs v0.12
- It works on Apple Silicon out of the box, no env variables needed.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -51,7 +51,7 @@
     "dids": "^3.0.0-alpha.4",
     "ethr-did-resolver": "^5.0.3",
     "express": "^4.17.2",
-    "go-ipfs": "^0.11.0",
+    "go-ipfs": "^0.12.0",
     "ipfs-http-client": "^55.0.0",
     "ipfsd-ctl": "^10.0.5",
     "key-did-provider-ed25519": "^1.1.0",
--- a/packages/ipfs-daemon/package.json
+++ b/packages/ipfs-daemon/package.json
@@ -40,7 +40,7 @@
     "dag-jose": "^1.0.0",
     "express": "^4.17.2",
     "get-port": "^6.0.0",
-    "go-ipfs": "^0.11.0",
+    "go-ipfs": "^0.12.0",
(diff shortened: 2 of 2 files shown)


stbrody (approved):
ooh awesome.  Great to see this just works out of the box.

FYI @v-stickykeys @smrz2001

oed:
Note that we will likely remain on 0.11 on our infra for a while since 0.12 does a repo migration.

stbrody:
> Note that we will likely remain on 0.11 on our infra for a while since 0.12 does a repo migration.
> 
> We don't use this bundle for our infra anyway.

oh interesting... hmm, we may not want to be running our unit tests with version 12 then if our infra is still on v11

stbrody (changes requested):
revoking my approval - if we're not ready to up our infra to v12 then we should up our testing or what's started by default on the CLI

oed:
Fwiw, our infra uses https://github.com/ceramicnetwork/go-ipfs-daemon/ not this code. 
0.12 doesn't do any changes to networking, only local data store. I think this is a fairly safe change.
https://github.com/ipfs/go-ipfs/releases/tag/v0.12.0

stbrody:
> Fwiw, our infra uses [ceramicnetwork/go-ipfs-daemon](https://github.com/ceramicnetwork/go-ipfs-daemon/) not this code. 0.12 doesn't do any changes to networking, only local data store. I think this is a fairly safe change. [ipfs/go-ipfs@`v0.12.0` (release)](https://github.com/ipfs/go-ipfs/releases/tag/v0.12.0)

even so.  as a general rule we shouldn't be testing with something different than what we're running

oed:
I'm pretty sure our integration tests are using go-ipfs-daemon (or should
be?). Overall though I think let's keep it at 0.11 for now. Once we have
integration test running we can upgrade here more confidently.

*Joel Thorstensson *
Co-founder / CTO

ukstv:
Valid concern. Converting it to draft.

oed:
I guess this one is relevant and should be merged now @stbrody @ukstv

PR #2059, opened 2022-02-22
feat: Use old caip format for proofs

ukstv:
Apparently, found a bug in v2 client communicating with v1 daemon. v2 client should use old CAIP format string in proofs.

### DIFF
--- a/packages/blockchain-utils-linking/src/__tests__/__snapshots__/cosmos.test.ts.snap
+++ b/packages/blockchain-utils-linking/src/__tests__/__snapshots__/cosmos.test.ts.snap
@@ -4,7 +4,7 @@ exports[`Blockchain: Cosmos authenticate create proof for cosmoshub3 1`] = `"0x3
 
 exports[`Blockchain: Cosmos createLink create proof for cosmoshub3 1`] = `
 Object {
-  "account": "cosmos:cosmoshub-3:cosmos1gs6gu0w9xvqcfhm3nlhnmqkf8aphnsxyjlgpqd",
+  "account": "cosmos1gs6gu0w9xvqcfhm3nlhnmqkf8aphnsxyjlgpqd@cosmos:cosmoshub-3",
   "message": "Link this account to your identity
 
 did:3:bafysdfwefwe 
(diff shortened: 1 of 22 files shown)


oed:
Why is this necessary?

Anyway I don't think this matters much as I imagine we will update caip10-links to support CACAOs created by the pkh DID correlating to the caip10 in the *controller*. So new commits will be JWS :)

stbrody (commented):
ooh good catch!  Can we add some testing?

stbrody:
> ooh good catch! Can we add some testing?

looks like there actually was tests already, they just need to be updated

ukstv:
> > ooh good catch! Can we add some testing?
> 
> looks like there actually was tests already, they just need to be updated

yup. and they rightfully fail now.

oed:
> I didn't follow this. Don't we need to preserve the existing format for backwards compatibility? 

Yes ofc! 

My first comment was more regarding that I thought the proof verifier should support both old and new format!

stbrody:
> My first comment was more regarding that I thought the proof verifier should support both old and new format!

Right, so here in blockchain-utils-linking we want to always be *creating* the proofs in the old format for backwards compatibility, at least for the time being.  Then in blockchain-utils-validation we could make it support both old and new formats.  Then after support for both formats has been out long enough that we think most Ceramic nodes have updated and have that support, then we can start producing proofs in the new format.

oed:
> Right, so here in blockchain-utils-linking we want to always be creating the proofs in the old format for backwards compatibility, at least for the time being. Then in blockchain-utils-validation we could make it support both old and new formats. Then after support for both formats has been out long enough that we think most Ceramic nodes have updated and have that support, then we can start producing proofs in the new format.

Makes sense! Good catch indeed. 

My other point was that we likely wouldn't care about changing this to the new caip10 format as we would likely phase in support for new JWS + CACAO proofs.

ukstv:
Changed all of the blockchains to use old caip for proofs :/

PaulLeCam:
cc @zachferland this might affect the changes you made in 3ID Connect

stbrody (approved):
this LGTM but I'd like @zachferland to take a look and confirm that 3id-connect is prepared for this before we merge

zachferland:
yeah ran into this today as well, yeah definitely need this, and looks good

zachferland (approved):

stbrody:
does this need to be backported to the V1 branch and released as a hotfix to the current Ceramic version? @zachferland @ukstv

ukstv:
> does this need to be backported to the V1 branch and released as a hotfix to the current Ceramic version? @zachferland @ukstv 

AFAIK v1 uses old caip everywhere.

PR #2065, opened 2022-02-28
fix: set `keepalive: false` in http(s) agent to ipfs

smrz2001:


packages/core/src/utils.ts
```diff
@@ -115,11 +115,7 @@ export class Utils {
     signal: AbortSignal
   ) {
     if (typeof cid === 'string') cid = CID.parse(cid.replace('ipfs://', ''))
-    let format = await ipfsApi.codecs.getCodec(cid.code).then((f) => f.name)
-    // The try catch behaviour of the js-ipfs-http-client was causing timing issues (socket hangup)
-    // https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs-http-client/src/block/put.js#L34
-    if (format === 'dag-cbor') format = 'cbor'
-    else if (format === 'dag-pb') format = 'protobuf'
+    const format = await ipfsApi.codecs.getCodec(cid.code).then((f) => f.name)
```
v-stickykeys, line 118: Why are we removing the format handling? It seems like if we know the request is going to fail we should avoid it altogether if possible

v-stickykeys (commented):
Just one question

v-stickykeys:
Oh no why are the go-ipfs tests failing now?

packages/core/src/utils.ts
oed: If we don't have any issues with it, it's probably safer to keep using the abstractions that ipfs provides in case they choose to change something under the hood?

smrz2001:
> Oh no why are the go-ipfs tests failing now?

Looks like the Polkadot test again :( @ukstv mentioned that it's known to be flaky?

```
FAIL  src/blockchains/__tests__/polkadot.test.ts
  ● Blockchain: Polkadot › validateLink › validate proof with secp256k
```

packages/core/src/utils.ts
smrz2001: It's because it fails with `js-ipfs` with these changes :(

To reliably not make requests to either `js-ipfs` or `go-ipfs` that are unsuitable would require detecting the type of IPFS node through the `id` calls. That is an alternative though - make the right call to the right node.

We can do that since we make constant `id` calls to IPFS during our `healthcheck`, and `id` returns the node type, but it just seems a little hacky.

v-stickykeys (approved):
🚢

oed (approved):
lgtm

PR #2066, opened 2022-03-02
feat: Allow running embedded IPFS daemon

ukstv:
When running `ceramic daemon`, `ipfsEndpoint` param is a `Multiaddr`, not a `string`.

### DIFF
--- a/packages/cli/src/build-ipfs-connection.util.ts
+++ b/packages/cli/src/build-ipfs-connection.util.ts
@@ -13,12 +13,12 @@ const IPFS_GET_TIMEOUT = 60000 // 1 minute
 
 const mergeOptions = mergeOpts.bind({ ignoreUndefined: true })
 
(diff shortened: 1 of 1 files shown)


smrz2001 (approved):
Oh :oof:, thanks for catching this, @ukstv.

Is this whenever running the daemon or specifically when the daemon is run in bundled mode?

smrz2001:
Regardless, I think it would be good to propagate this fix to CAS as well, where @stephhuynh18 added the change to use a custom http agent yesterday.

smrz2001:
> Yes, it manifests in bundled mode only.

Oh lol, thanks, I see the title of the PR now, was just looking at the description and getting confused.

It doesn't seem impossible that we might in some situations want to use multiaddr instead of URL in CAS, but probably not important right now.

🚢🚀

PR #2070, opened 2022-03-06
feat(stream-tile-handler): Add support for specifying resources by family

oed:
## Description

Currently you have to specify resources using the streamId. This creates a few problems:
#### SelfId SDK can't create streams on the flow within a particular Data Model
Seldom is a Data Model only one stream, rather it's a set of streams which are all referenced from the main Data Model record.

#### You have to know the DID of the user up front
This is problematic in future cases where WalletConnect will allow you to request a SIWE message on connection, basically reducing the flow of authenticating to one interaction. This greatly increases UX, but it means that we can't do the normal request address, then request signature flow.

### Solution
We introduce a new way to specify a Ceramic resource:
```
ceramic://*?family=<family-name>
```
This will allow the developer to specify the family used in a stream as the resource. Since many streams can share the same family this allows us to even create streams using this capability. In IDX we specify the family as the streamId of the definition, which gives us a good way to actually verify what the capability is meant to do (e.g. gives the app access to a particular data model).

## How Has This Been Tested?

Tests have been added to the canary integration:
- [x] Update a stream using a resource which specifies the family
- [x] Update a stream with the wrong family
- [x] Update a stream with null family
- [x] Update someone elses stream using family
- [x] Create a new stream using the family capability
- [x] Create a new stream using the wrong family capability

## PR checklist

### DIFF
--- a/packages/canary-integration/src/__tests__/capability.test.ts
+++ b/packages/canary-integration/src/__tests__/capability.test.ts
@@ -10,12 +10,36 @@ import { randomBytes } from '@stablelib/random'
 import { SiweMessage, Cacao } from 'ceramic-cacao'
 import { createCeramic } from '../create-ceramic.js'
 
-let ipfs: IpfsApi
-let ceramic: CeramicApi
-let wallet: Wallet
-let didKey: DID
+
+const addCapToDid = async (wallet, didKey, resource) => {
+  // Create CACAO with did:key as aud
+  const siweMessage = new SiweMessage({
+    domain: 'service.org',
+    address: wallet.address,
+    chainId: '1',
+    statement: 'I accept the ServiceOrg Terms of Service: https://service.org/tos',
+    uri: didKey.id,
+    version: '1',
+    nonce: '23423423',
+    issuedAt: new Date().toISOString(),
+    resources: [resource],
+  })
+  // Sign CACAO with did:pkh
+  const signature = await wallet.signMessage(siweMessage.toMessage())
+  siweMessage.signature = signature
+  const capability = Cacao.fromSiweMessage(siweMessage)
+  // Create new did:key with capability attached
+  const didKeyWithCapability = didKey.withCapability(capability)
+  await didKeyWithCapability.authenticate()
+  return didKeyWithCapability
+}
 
 describe('CACAO Integration test', () => {
+  let ipfs: IpfsApi
+  let ceramic: CeramicApi
+  let wallet: Wallet
+  let didKey: DID
+
   beforeAll(async () => {
     ipfs = await createIPFS()
     ceramic = await createCeramic(ipfs)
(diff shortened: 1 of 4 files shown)


oed:
EOS testnet servers seem to be down. Maybe we should disable the EOS tests for now?

ukstv:
> EOS testnet servers seem to be down. Maybe we should disable the EOS tests for now?

Yes. https://github.com/ceramicnetwork/js-ceramic/pull/2073

packages/canary-integration/src/__tests__/capability.test.ts
```diff
@@ -65,116 +31,226 @@ test('verifies capability with signed commit', async () => {
...
 
-  // Create did:key for the dApp
-  const seed = new Uint8Array([
-    69, 90, 79, 1, 19, 168, 234, 177, 16, 163, 37, 8, 233, 244, 36, 102, 130, 190, 102, 10, 239, 51,
-    191, 199, 40, 13, 2, 63, 94, 119, 183, 225,
-  ])
-  const didKeyProvider = new Ed25519Provider(seed)
-  const didKey = new DID({ provider: didKeyProvider, resolver: KeyDidResolver.getResolver() })
-  await didKey.authenticate()
+      await deterministicDocument.update({ foo: 'bar' }, null, {
```
stbrody, line 72: include negative test first?  ie try to update just the didKey, without the capability, assert that it fails

packages/canary-integration/src/__tests__/capability.test.ts
```diff
@@ -65,116 +31,226 @@ test('verifies capability with signed commit', async () => {
...
+          asDID: didKeyWithCapability,
+          anchor: false,
+          publish: false,
+        })
+      ).rejects.toThrowError(
+        'Capability does not have appropriate permissions to update this TileDocument'
+      )
+    }, 30000)
+
+    test('can not create new stream without capability', async () => {
```
stbrody, line 240: nit: put this test first - flows better when reading

packages/stream-tile-handler/src/tile-document-handler.ts
```diff
@@ -261,7 +265,8 @@ export class TileDocumentHandler implements StreamHandler<TileDocument> {
 
     if (
```
stbrody, line 266: this conditional is getting kind of complex to read - can we make a helper function to increase clarity?

packages/stream-tile/src/tile-document.ts
```diff
@@ -191,7 +191,8 @@ export class TileDocument<T = Record<string, any>> extends Stream {
       // document as there shouldn't be any existing state for this doc on the network.
       opts.syncTimeoutSeconds = 0
     }
-    const commit = await TileDocument.makeGenesis(ceramic, content, metadata)
+    const signer: CeramicSigner = opts.asDID ? { did: opts.asDID } : ceramic
```
stbrody, line 194: is this related to the core change in this PR?  Or is this a separate bug fix you noticed while in there?

stbrody (approved):
lgtm, all comments optional

packages/stream-tile-handler/src/tile-document-handler.ts
ukstv: @stbrody 2 cents from me. It is complex, and it is ok _for now_. In my opinion, this thing should be simplified after we got additional input from external users/partners, not earlier. IMO, it is still not over the edge, crying "simplify me now, or I will bury you".

packages/stream-tile/src/tile-document.ts
oed: This is needed to add the ability to create a stream with a CACAO enabled DID. This was not supported before.

oed:
Still want @zachferland and @haardikk21 to have a look at this one before merging.

packages/stream-tile-handler/src/tile-document-handler.ts
oed: I was thinking about this as well. The most confusing part to me is that it's negatives + ANDs. Don't feel strongly right now, in the future it would definitely make sense to simplify.

packages/canary-integration/src/__tests__/capability.test.ts
oed: Same thing as with `can not create new stream without capability`? Yep, that would make sense to have.

zachferland (approved):
👍

PR #2073, opened 2022-03-07
chore: Disable tests using EOS Jungle network

ukstv:
Our tests fail after EOS Jungle endpoint in `@smontero/eosio-signing-tools` became unavailable. See https://github.com/sebastianmontero/eosio-signing-tools/issues/1 for details.

This PR removes the tests that use the failed network.


ukstv:
> Can we do `test.skip(` instead?

For a single test only.

PR #2074, opened 2022-03-07
feat: Do not depend on bigint-mod-arith

ukstv:
Unless it can be really used as ES module.

Copied the relevant code inside.

Relevant issue: https://github.com/juanelas/bigint-mod-arith/pull/6

### DIFF
--- a/packages/key-did-resolver/package.json
+++ b/packages/key-did-resolver/package.json
@@ -34,7 +34,6 @@
   },
   "dependencies": {
     "@stablelib/ed25519": "^1.0.2",
-    "bigint-mod-arith": "^3.0.0",
     "multiformats": "^9.5.2",
     "uint8arrays": "^3.0.0",
     "varint": "^6.0.0"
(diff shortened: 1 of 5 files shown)


ukstv:
CC @smrz2001

oed (approved):
ugh, let's monitor upsteam and hope your fix get's merged.

smrz2001:
In case something like this is needed again in the future, wanted to add another option. The followings command will pull in a fork/branch from GitHub directly into the project dependencies. This works for all public repositories and can be an alternative to copying code. Can also use a tag or commit hash in place of the branch (or leave the `#...` part out entirely to use the default branch).
```
npm install "username/repo#branch" --save
```
OR
```
yarn add "username/repo#branch"
```
e.g,
```
npm install "ukstv/bigint-mod-arith#master" --save
```

ukstv:
Thanks! @smrz2001 AFAIK, it would require built artefacts to be stored in the repo, which is not quite nice if the repo uses TypeScript.

smrz2001:
> Thanks! @smrz2001 AFAIK, it would require built artefacts to be stored in the repo, which is not quite nice if the repo uses TypeScript.

Ahh, gotcha, @ukstv 👍🏼

PR #2076, opened 2022-03-10
fix(common): add file extension to local dependency

stephhuynh18:
Its so easy to forget =( especially when everything builds fine


oed (approved):

stbrody:
It's pretty disappointing that this builds fine. @ukstv do you know of anything we could change in our build config to catch errors like this at build time instead of runtime?

PR #2081, opened 2022-03-15
feat(stream-tile-handler): use revocationPhaseOutSec option when verifying signatures

stephhuynh18:
## Description

Use revocation phase out option when verifying jws. Should we make the revocationPhaseOut default more accessible in the docs? 

## How Has This Been Tested?

Added tests

### DIFF
--- a/packages/stream-tile-handler/src/__tests__/tile-document.test.ts
+++ b/packages/stream-tile-handler/src/__tests__/tile-document.test.ts
@@ -17,8 +17,10 @@ import {
   StreamUtils,
   SignedCommitContainer,
   TestUtils,
-  IpfsApi, GenesisCommit
+  IpfsApi,
+  GenesisCommit,
 } from '@ceramicnetwork/common'
+import { parse as parseDidUrl } from 'did-resolver'
 
 jest.unstable_mockModule('did-jwt', () => {
   return {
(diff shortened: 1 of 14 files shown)


stephhuynh18:
Failing tests are the polkadot tests for esm reasons

oed:
Should we wait to merge this before we have the name change of the variable that @stbrody suggested?

packages/stream-tile-handler/src/__tests__/tile-document.test.ts
```diff
@@ -587,6 +592,166 @@ describe('TileDocumentHandler', () => {
     delete state.metadata.unique
     expect(state).toMatchSnapshot()
   })
+
+  it('fails to apply commit if keys revoked', async () => {
```
stbrody, line 596: it's pretty hard to follow what exactly this test is doing.  Can we pull out some pieces into helper functions and add some more comments to make it clearer what we're doing at each phase of this test?  Same for other tests added

stephhuynh18: yes can do

packages/stream-tile-handler/src/tile-document-handler.ts
```diff
@@ -211,7 +213,7 @@ export class TileDocumentHandler implements StreamHandler<TileDocument> {
   }
 
   /**
-   * Verifies commit signature
+   * Verifies commit signature. If a revoked key is used to create the signature, the signature is valid for 24h after the revocation
```
stbrody, line 216:
```suggestion
   * Verifies commit signature. If a revoked key is used to create the signature, the signature is valid for 24h after the revocation.  This is so that if an update made before the key revocation winds up getting anchored after the revocation does, we don't fail the write unnecessarily.
   * TODO: Remove or significantly shorten this grace period once anchors happen far more frequently on the network.
   * ```

stbrody (approved):
LGTM mode one suggestion for additional info in a comment

PR #2085, opened 2022-03-17
chore: fix polkadot tests

ukstv:
Root cause: Jest resolver is buggy. It tries to import CJS by default; some packages do not properly specify conditional exports, which leads to a failure.

The misbehavior is solved in upcoming Jest v28. We can not use it though right now: it changed the way mocks work, which breaks other tests.

Changes: use `jest-resolver-enhanced` for `blockchain-utils-linking` package to overcome Jest resolver limitations.

Also, to make tests fully green again, fixed an issue in TileHandler. Pass `atTime` option to `did` signature verification only if commit has a timestamp. CC @haardikk21

### DIFF
--- a/packages/blockchain-utils-linking/jest.config.json
+++ b/packages/blockchain-utils-linking/jest.config.json
@@ -7,5 +7,6 @@
   },
   "transform": {
     "^.+\\.tsx?$": "babel-jest"
-  }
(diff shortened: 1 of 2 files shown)


stbrody (approved):
🙏

PR #2090, opened 2022-03-21
feat: polyfill AbortController, so that Ceramic node works on Node.js v14

ukstv:
Polyfill AbortController, absent on Node.js v14.

This seems to work. Tests are green on my machine. On CircleCI I see quite cryptic [errors](https://app.circleci.com/pipelines/github/ceramicnetwork/js-ceramic/5341/workflows/324b097b-256f-4305-b8ff-c8f6d6676e34/jobs/5837). This seems to be related to these issues: https://github.com/ipfs/js-ipfs/issues/3998, https://github.com/websockets/ws/issues/1818. That is the reason to drop testing with Node14 on CircleCI.

@PaulLeCam Maybe you could give a shot with using it on AWS, after the release?


stbrody (commented):
core changes seem reasonable.  My main question is what happens when a developer starts using AbortController/AbortSignal in a new file and forgets to call polyfillAbortController() at the top of the file?  If this isn't caught in code review, is there any way we will notice before users complain or deploying to Self.ID fails (which may happen long after the corresponding Ceramic change is released)?

ukstv:
What we do here with AbortController, is an attempt to not re-do the whole deployment scheme of self.id when it migrates to Ceramic v2. We are not going to support Node.js v14 generally. We expect that in few months, we could forget about Node.js v14 completely due to imminent support of Node16 by AWS (ETA 2months from now). It might be the case, that the current hack will not work. In this case, self.id has to stay on Ceramic v1, and we forget about even this attempt of Node14 support.

Future changes is a perfectly valid concern. If I am not mistaken, a generally accepted practice is to polyfill before doing anything in your app (not a library!). I guess, the way to go is for @PaulLeCam to call `polyfillAbortController` function in the app entrypoint (`_app.ts` for Next.js, if I am not mistaken). Then `AbortController` would be available to the whole process.

PaulLeCam:
Yes shouldn't be a problem to add the AbortController polyfill to the app itself 👍

stbrody (approved):
Okay so sounds like the plan is, if we accidentally slip in new uses of AbortController without polyfilling them in js-ceramic, and then some user complains that they can no longer run Ceramic with Node 14, we'll just tell them to manually pollyfill AbortController themselves, and that should get them working again.  Seems reasonable.

PR #2091, opened 2022-03-22
[NET-1195] add ceramic version to heartbeat pubsub, send it at least once a day

haardikk21:
## Description

Updates the Pubsub `KeepAliveMessage` to include the current Ceramic core version being used. Also ensures that a KeepaliveMessage is sent once on startup, and at least once per day, regardless of other messages being sent.

It also introduces a `genversion` dev dependency on core, that is needed to create an importable `version` file during the buildstep. This is needed because we shouldn't read directly from `package.json` as it will then ship with our compiled code, and we cannot read from `process.env.npm_package_version` as it only works if ceramic is started using `npm` i.e. the binary will not work.

## How Has This Been Tested?

- [ ] Test B

## PR checklist

## References:


packages/core/src/pubsub/pubsub-keepalive.ts
```diff
@@ -60,12 +66,18 @@ export class PubsubKeepalive
    */
   publishPubsubKeepaliveIfNeeded(): void {
     const now = Date.now()
-    if (now - this.lastPublishedMessageDate < this.maxPubsubPublishInterval / 2) {
-      // We've published a message recently enough, no need to publish another
+    const needToPublishKeepaliveOnceADay =
```
stbrody, line 69: please pull the interval into a constant `MIN_KEEPALIVE_INTERVAL`

stbrody (changes requested):
mostly looks good, but needs test coverage.  You should be able to expand the existing pubsub-keepalive.test.ts.

I'd also like @ukstv to take a look at how we're generating the package version and make sure it looks good

ukstv:
@stbrody We discussed the way version got injected with @haardikk21 . It seems the most appropriate way, as it does not introduce a run-time dependency on `package.json`. Looking forward, IMO, we should migrate to this in other packages as well.

And, I agree with @stbrody, that there should be a test.

ukstv (changes requested):
test please

packages/core/src/pubsub/pubsub-keepalive.ts
haardikk21: pulled it out, but named it `ONE_DAY_MS` for clearer naming

stbrody (commented):
the test changes just added a check that the new version field was added.  I think the more interesting, complex, and likely to include a bug part is the scheduling logic, which still isn't tested.

haardikk21:
> the test changes just added a check that the new version field was added. I think the more interesting, complex, and likely to include a bug part is the scheduling logic, which still isn't tested.

hmm. so basically overwriting the scheduling intervals in the tests through some config object and maybe wanting a message every 30 seconds regardless of other messages?

stbrody:
> > the test changes just added a check that the new version field was added. I think the more interesting, complex, and likely to include a bug part is the scheduling logic, which still isn't tested.
> 
> hmm. so basically overwriting the scheduling intervals in the tests through some config object and maybe wanting a message every 30 seconds regardless of other messages?

yeah, if you make `PubsubKeepalive` take the max time without a keepalive message as an argument in the constructor, you should be able to do something very similar to the existing pubsub keepalive test by setting that value down to something small

packages/core/src/pubsub/__tests__/pubsub-keepalive.test.ts
```diff
@@ -30,16 +31,22 @@ test('publish keepalive', async () => {
...
+    pubsub,
+    keepaliveInterval,
+    maxIntervalWithoutKeepalive
+  )
   const subscription = pubsubWithKeepalive.subscribe()
 
   await delay(keepaliveInterval * 5)
 
-  expect(ipfs.pubsub.publish.mock.calls.length).toBeGreaterThan(4)
+  expect(ipfs.pubsub.publish.mock.calls.length).toBeGreaterThanOrEqual(8)
```
stbrody, line 44: this test isn't really testing the nuances of the logic here. It's also conflating testing the max time without any message with testing the max time without a keepalive.  This should be broken up into a bunch of smaller tests testing specific pieces of the logic.  i.e. We've had *no* messages over a period of time, we've had some messages but not keepalive messages, we've had keepalive messages, etc.

stbrody (changes requested):

packages/core/src/pubsub/__tests__/pubsub-keepalive.test.ts
haardikk21: hmm makes sense, will do

packages/core/src/pubsub/__tests__/pubsub-keepalive.test.ts
```diff
@@ -12,35 +13,106 @@ const diagnosticsLogger = loggerProvider.getDiagnosticsLogger()
...
-  const pubsub = new Pubsub(
-    ipfs as unknown as IpfsApi,
-    TOPIC,
-    3000,
-    pubsubLogger,
-    diagnosticsLogger
-  )
-  const pubsubWithKeepalive = new PubsubKeepalive(pubsub, keepaliveInterval)
-  const subscription = pubsubWithKeepalive.subscribe()
+describe('pubsub keepalive', () => {
```
stbrody, line 16: one case that's still missing is making sure that a user publishing a random non-keepalive message prevents the need to send keepalives in the maxPubsubIntervalTime, but we'll still need to send one if there's no keepalive messages specifically within the maxKeepaliveIntervalTime

stbrody: that said this PR has gone on long enough, I think the coverage you added thus far is good enough for now.  So consider that an optional suggestion or future improvement.

stbrody (approved):

PR #2096, opened 2022-03-28
include cacao in CommitData when fetching log

haardikk21:
## Description

Fetches and includes Cacao capability in `CommitData` when fetching log (`fetchLog`). Allows us to use the dispatcher to retrieve the capability object, which is more resilient than just calling `ipfs.dag.get(...)` as before.

## How Has This Been Tested?

Existing tests pass, including integration tests

## PR checklist

## References:

### DIFF
--- a/packages/core/src/utils.ts
+++ b/packages/core/src/utils.ts
@@ -92,6 +92,17 @@ export class Utils {
       if (!linkedCommit) throw new Error(`No commit found for CID ${commit.link.toString()}`)
       commitData.commit = linkedCommit
       commitData.envelope = commit
+
+      if (commit.signatures && commit.signatures.length > 0) {
+        const protectedHeader = commit.signatures[0].protected
+        const decodedProtectedHeader = base64urlToJSON(protectedHeader)
+        if (decodedProtectedHeader.cap) {
+          const capIPFSUri = decodedProtectedHeader.cap
+          const capCID = CID.parse(capIPFSUri.replace('ipfs://', ''))
+          const cacao = await dispatcher.retrieveFromIPFS(capCID)
+          commitData.capability = cacao
+        }
+      }
     } else if (StreamUtils.isAnchorCommit(commit)) {
       commitData.type = CommitType.ANCHOR
       commitData.proof = await dispatcher.retrieveFromIPFS(commit.proof)
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -46,7 +46,8 @@
     "logfmt": "^1.3.2",
     "multiformats": "^9.5.8",
     "rxjs": "^7.5.2",
-    "uint8arrays": "^3.0.0"
+    "uint8arrays": "^3.0.0",
+    "ceramic-cacao": "^0.0.16"
   },
   "devDependencies": {
     "@types/lodash.clonedeep": "^4.5.6",
--- a/packages/common/src/stream.ts
+++ b/packages/common/src/stream.ts
@@ -8,6 +8,7 @@ import { Observable } from 'rxjs'
 import type { RunningStateLike } from './running-state-like.js'
 import type { CeramicApi } from './ceramic-api.js'
 import { LoadOpts, SyncOptions } from './streamopts.js'
+import type { Cacao } from 'ceramic-cacao'
 
 /**
  * Describes signature status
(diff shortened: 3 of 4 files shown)


packages/core/src/utils.ts
```diff
@@ -92,6 +92,17 @@ export class Utils {
       if (!linkedCommit) throw new Error(`No commit found for CID ${commit.link.toString()}`)
       commitData.commit = linkedCommit
       commitData.envelope = commit
+
+      if (commit.signatures && commit.signatures.length > 0) {
+        const protectedHeader = commit.signatures[0].protected
```
stbrody, line 97: nit: can you pull this out into an `extractCapability` helper function please?

stbrody: Would also be good if that function could catch an error loading the capability from IPFS and re-throw a more descriptive error message indicating that the CID that failed to load was that of the capability

stbrody (commented):
very nice cleanup.  A few small suggestions to cleanup even further inline

packages/core/src/utils.ts
```diff
@@ -92,17 +92,8 @@ export class Utils {
...
-        const decodedProtectedHeader = base64urlToJSON(protectedHeader)
-        if (decodedProtectedHeader.cap) {
-          const capIPFSUri = decodedProtectedHeader.cap
-          const capCID = CID.parse(capIPFSUri.replace('ipfs://', ''))
-          const cacao = await dispatcher.retrieveFromIPFS(capCID)
-          commitData.capability = cacao
-        }
-      }
+      const cacao = await this.extractCapability(commit, dispatcher)
+      if (cacao) commitData.capability = cacao
```
oed, line 96:
```suggestion
      commitData.capability = await this.extractCapability(commit, dispatcher)
```

packages/core/src/utils.ts
```diff
@@ -131,6 +122,30 @@ export class Utils {
...
     await ipfsApi.block.put(block, { format, mhtype, version, signal })
   }
+
+  /**
+   * Attempts to load CACAO capability from IPFS if present in protected header of commit
+   * @param commit the commit to load the capability for
+   * @param dispatcher instance of dispatcher to load from IPFS
+   * @returns a Cacao capability object if found, else null
+   */
+  static async extractCapability(commit: any, dispatcher: Dispatcher): Promise<any | null> {
```
oed, line 132:
```suggestion
  static async extractCapability(commit: any, dispatcher: Dispatcher): Promise<CACAO | undefined> {
```

We should convert the type to CACAO before we return. If nothing is returned the function will return undefined.

oed (commented):
Looks mostly good 👍

packages/core/package.json
```diff
@@ -50,6 +50,7 @@
     "ajv": "^8.8.2",
     "ajv-formats": "^2.1.1",
     "await-semaphore": "^0.1.3",
+    "ceramic-cacao": "^0.0.17",
```
oed, line 53: This can be a dev dep since we only import the type

oed (approved):
lgtm, would just change how we import the package.

packages/core/src/utils.ts
```diff
@@ -120,6 +122,30 @@ export class Utils {
...
   }
+
+  /**
+   * Attempts to load CACAO capability from IPFS if present in protected header of commit
+   * @param commit the commit to load the capability for
+   * @param dispatcher instance of dispatcher to load from IPFS
+   * @returns a Cacao capability object if found, else null
+   */
+  static async extractCapability(commit: any, dispatcher: Dispatcher): Promise<Cacao | undefined> {
+    if (commit.signatures && commit.signatures.length > 0) {
```
stbrody, line 133: Can do an early return here using the inverse of this condition, and avoid an extra indentation layer.  Early returns can do a lot to improve readability and reduce cognitive overhead when reading long functions like this

packages/core/src/utils.ts
```diff
@@ -120,6 +122,30 @@ export class Utils {
...
+      const protectedHeader = commit.signatures[0].protected
+      const decodedProtectedHeader = base64urlToJSON(protectedHeader)
+      if (decodedProtectedHeader.cap) {
```
stbrody, line 136: ditto re: early return.

```
if (!decodedProtectedHeader.cap) {
  return null
}
```

stbrody (commented):
a few more small cleanups but overall this is looking really nice

PR #2115, opened 2022-04-13
feat(cli): Add stream cache size to config file

stbrody:


oed (approved):
lgtm, I assume this opt will find it's way to the right place.

PR #2122, opened 2022-04-18
feat: Make IPFS instance on CLI non-disposable

ukstv:
When run in CLI mode, IPFS instance is non-disposable.

Test procedure:
- run ceramic daemon,
- create a stream,
- stop ceramic daemon,
- run ceramic daemon,
- query state of the stream.

Reports the state just all right.


stbrody (approved):
🚀

packages/cli/src/build-ipfs-connection.util.ts
```diff
@@ -58,6 +47,7 @@ async function createGoIPFS(overrideConfig: Partial<ipfsClient.Options> = {}): P
   const gatewayPort = 9011
   const defaultConfig = {
     ipld: { codecs: [dagJose] },
+    repo: '~/.ipfs',
```
stephhuynh18, line 50: Is this the default repo we want to use? I don't mind either way but before it was `.jsipfs` and in `ipfs-daemon` its `.goipfs`

stephhuynh18 (approved):
One question tho

packages/cli/src/build-ipfs-connection.util.ts
ukstv: true. thanks!

PR #2138, opened 2022-04-27
feat(stream-tile): use dids capability iss as controller when capabil…

zachferland:
~~DRAFT - waits on DIDs package~~

Reference [SIWE+ spec](https://www.notion.so/threebox/CACAO-DID-PKH-SIWE-Client-Libraries-5ce1f1f6e8c7417683c2b63e07f91cef#741885cca9d94532afee11a46621ab7d) for motivation. Allows devs to use DIDs w/ capability and ceramic without having to know an implementation detail of CACAO (ie controller is cap issuer, and did w/cap is the signer)

### DIFF
--- a/packages/canary-integration/src/__tests__/capability.test.ts
+++ b/packages/canary-integration/src/__tests__/capability.test.ts
@@ -246,13 +246,12 @@ describe('CACAO Integration test', () => {
       test('can not create new stream with wrong family', async () => {
         const family = 'testFamily1'
         const didKeyWithCapability = await addCapToDid(wallet, didKey, `ceramic://*?family=${family}`)
+        ceramic.did = didKeyWithCapability
 
         await expect(
           TileDocument.create(ceramic, { foo: 'bar' }, {
             family: `${family}-wrong`,
-            controllers: [`did:pkh:eip155:1:${wallet.address}`],
           }, {
-            asDID: didKeyWithCapability,
             anchor: false,
             publish: false,
           })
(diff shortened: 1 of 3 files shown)


stbrody (commented):
makes sense to me, but I'll defer to @ukstv for final approval

ukstv (changes requested):
In the current iteration, the PR _replaces_ existing tests while it _adds_ new functionality.

IMO, this PR should add few new tests rather than changing old ones.

Otherwise, LGTM. The core change improves DX a lot.

zachferland:
yeah considered, but agree, restored existing tests, added new now

packages/canary-integration/src/__tests__/capability.test.ts
```diff
@@ -315,4 +318,56 @@ describe('CACAO Integration test', () => {
...
+      const deterministicDocument = await TileDocument.deterministic(ceramic, {
+        deterministic: true,
+        family: 'testCapabilities1',
+      })
+      const streamId = deterministicDocument.id
+      const didKeyWithCapability = await addCapToDid(wallet, didKey, `ceramic://${streamId.toString()}`)
+      ceramic.did = didKeyWithCapability
+
+      await deterministicDocument.update({ foo: 'bar' }, null, {
+        asDID: didKeyWithCapability,
```
ukstv, line 335: @zachferland Do I miss something or `asDID` is not necessary here?

ukstv (approved):
🏇

packages/canary-integration/src/__tests__/capability.test.ts
zachferland: yup! removed

PR #2144, opened 2022-05-04
feat(core): replace cas-dev for dev-unstable with cas-qa

v-stickykeys:
Now that we have the qa environment up and running, dev-unstable network should anchor to it by default

### DIFF
--- a/packages/core/src/ceramic.ts
+++ b/packages/core/src/ceramic.ts
@@ -58,7 +58,7 @@ const DEFAULT_ANCHOR_SERVICE_URLS = {
   [Networks.MAINNET]: 'https://cas.3boxlabs.com',
   [Networks.ELP]: 'https://cas.3boxlabs.com',
   [Networks.TESTNET_CLAY]: 'https://cas-clay.3boxlabs.com',
-  [Networks.DEV_UNSTABLE]: 'https://cas-dev.3boxlabs.com',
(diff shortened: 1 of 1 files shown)


packages/core/src/ceramic.ts
```diff
@@ -58,7 +58,7 @@ const DEFAULT_ANCHOR_SERVICE_URLS = {
   [Networks.MAINNET]: 'https://cas.3boxlabs.com',
   [Networks.ELP]: 'https://cas.3boxlabs.com',
   [Networks.TESTNET_CLAY]: 'https://cas-clay.3boxlabs.com',
-  [Networks.DEV_UNSTABLE]: 'https://cas-dev.3boxlabs.com',
+  [Networks.DEV_UNSTABLE]: 'https://cas-qa.3boxlabs.com',
```
smrz2001, line 61: Should this be injected from Terraform instead?

smrz2001: Ok to do it this way for now, though 👍🏼

smrz2001 (approved):
🚢

packages/core/src/ceramic.ts
v-stickykeys: not really. these should be hardcoded. shouldn't rely on ppl running ceramic daemon to import these via environment variable

smrz2001: Ok, makes sense 👍🏼

stbrody (approved):

PR #2145, opened 2022-05-04
fix(canary-integration): Increase test fixture cleanup timeout

stbrody:
I don't really understand why @zachferland's recent commit would have made this teardown take longer.
I suspect that wasn't actually the root cause, and the issue was some upstream package getting updated and slowing down, although it is suspicious that the test that keeps failing is exactly the test file that zach's recent commit touched.  Something fishy is definitely going on here, but hopefully increasing the timeout on the teardown here will help


ukstv:
LGTM lol

PR #2146, opened 2022-05-05
chore(core,stream-tile-handler): Move schema validation out of core and into streamtype-specific code

stbrody:


packages/stream-tile-handler/src/tile-document-handler.ts
```diff
@@ -159,12 +170,24 @@ export class TileDocumentHandler implements StreamHandler<TileDocument> {
...
-    const content = state.next?.content ?? state.content
-    const metadata = state.next?.metadata ?? state.metadata
+    const oldContent = state.next?.content ?? state.content
+    const oldMetadata = state.next?.metadata ?? state.metadata
+
+    const newContent = jsonpatch.applyPatch(oldContent, payload.data).newDocument
+    const newMetadata = { ...oldMetadata, ...payload.header }
+
+    if (newMetadata.schema) {
+      // TODO: SchemaValidation.validateSchema does i/o to load a Stream.  We should pre-load
```
stbrody, line 180: so, this is in many ways a step backward.  We've been working to get network i/o out of the commit application path by pre-loading the data required into `CommitData`, but here I'm pushing network i/o down into a function that's already been CommitData-ified.  I'm disappointed to do this, but I looked into trying to proactively load the schema contents into the CommitData at a higher level, but there are so many circular dependencies it quickly spiraled out of control.  In the interest of unblocking the Model and ModelInstanceDocument streamtype asap, I think doing this for now is worth it.

I do think, however, that the need for a larger refactor of stream loading and commit application is becoming more and more urgent by the day.

oed (commented):
What about pinning? Does the schema get pinned if you pin the tile?
This is something we need to think about with Model and MID as well.

stbrody:
> What about pinning? Does the schema get pinned if you pin the tile? This is something we need to think about with Model and MID as well.

Ooh good question.  It does not look like we pin the schema stream currently.  This PR doesn't change that one way or the other.  It does seem like it would be good to make stream pinning be handled by the StreamTypeHandler so it can have streamtype-specific logic.  That would probably be a non-trivial refactor though...

oed:
Let's add a debt ticket to address this. We should at least make sure it happens for model / MID.

ukstv (approved):

PR #2159, opened 2022-05-13
fix(stream-tile, stream-tile-handler): don't allow updating controllers to invalid values

willex:
## Description

The current input sanity check in `makeCommit()` didn't account for controllers being set to `null` or `undefined`. Setting a stream controller to `null` prevents future updates on the same stream to succeed or propagate and goes against the spec that a _singular_ valid controller must be present.

## How Has This Been Tested?

- [ ] Manual test to update stream with incorrect `controller metaData` to verify fix
- [ ] Added unit test to cover multiple undesired controller update values: `null`, `undefined` & `''`
- [ ] Ran full test via `npm run test` (or you confirm via tile controller test only: `cd packages/core/src/__tests__/; npm run test -- tile-controller.test.ts`

## References:

Documentation: https://developers.ceramic.network/learn/glossary/#controllers

Refers to controllers: `A given stream may have one or more controllers.` which seems to be incorrect given the enforcement of one controller in the `makeCommit()` function and needs to be updated to: `A given stream may have one controller only.`


packages/stream-tile/src/tile-document.ts
```diff
@@ -344,8 +344,13 @@ export class TileDocument<T = Record<string, any>> extends Stream {
       newContent = this.content
     }
 
-    if (header.controllers && header.controllers?.length !== 1) {
-      throw new Error('Exactly one controller must be specified')
+    if (header.controllers) {
+      if (header.controllers?.length !== 1) {
+        throw new Error('Exactly one controller must be specified')
+      }
+      if (header.controllers[0] === null || header.controllers[0] === undefined || header.controllers[0] == '') {
```
oed, line 351:
```suggestion
      if (!Boolean(header.controllers[0])) {
```
[Boolean](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean) should suffice for this.

oed (approved):
lgtm, one suggestion.

packages/cli/src/daemon-config.ts
```diff
@@ -90,7 +89,7 @@ export class DaemonHTTPApiConfig {
    */
   @jsonMember(AnyT, {serializer: inPort => {
       const validPort = Number(inPort)
-      if (inPort == undefined || inPort == null) {
+      if (inPort == null) {
```
ukstv, line 92: @willex how is it relevant to stream controllers?

packages/stream-tile/src/tile-document.ts
```diff
@@ -344,8 +344,13 @@ export class TileDocument<T = Record<string, any>> extends Stream {
       newContent = this.content
     }
 
-    if (header.controllers && header.controllers?.length !== 1) {
-      throw new Error('Exactly one controller must be specified')
+    if (header.controllers) {
+      if (header.controllers?.length !== 1) {
```
ukstv, line 348:
```suggestion
      if (header.controllers.length !== 1) {
```

We (and TS compiler) already know, that `header.controllers` is defined.

packages/cli/src/daemon-config.ts
willex: It is not related. It was a left over suggestion from a previous PR that I overlooked and put it in here as a chore vs. opening a new PR for a small change.

stbrody (changes requested):

packages/stream-tile/src/tile-document.ts
```diff
@@ -344,8 +344,13 @@ export class TileDocument<T = Record<string, any>> extends Stream {
       newContent = this.content
     }
 
-    if (header.controllers && header.controllers?.length !== 1) {
-      throw new Error('Exactly one controller must be specified')
+    if (header.controllers) {
```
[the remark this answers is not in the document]
willex: Updated with latest fixes and added unit test.

packages/stream-tile/src/tile-document.ts
```diff
@@ -307,7 +293,7 @@ export class TileDocument<T = Record<string, any>> extends Stream {
       prev: this.tip,
       id: this.state$.id.cid,
     }
-    const commit = await _signDagJWS(this.api, rawCommit)
+    const commit = await TileDocument._signDagJWS(this.api, rawCommit)
```
ukstv, line 296: why make it a static method?

packages/stream-tile/src/tile-document.ts
```diff
@@ -355,7 +361,7 @@ export class TileDocument<T = Record<string, any>> extends Stream {
       prev: this.tip,
       id: this.state.log[0].cid,
     }
-    return _signDagJWS(signer, commit)
```
ukstv, line 358: @willex what is a rationale behind this?

willex: Creation of the helper function `_makeRawCommit` with breakout from `makeCommit` was necessary as to write a test case. It would otherwise not able to update the `metaData` after the content has been signed to verify to side step the `tile-controller` input verification and be able to trigger the `tile-controller-handler` check instead.

packages/stream-tile/src/tile-document.ts
willex: To indicate that it is a helper function and to mimic the prior behavior outside the class, but can be invoked now as part of the object during the test-case without the need of importing it separately (hence the move inside the class).

packages/stream-tile/src/tile-document.ts
ukstv: IMO, it is better to use old fashioned functions.

packages/stream-tile-handler/src/__tests__/tile-document.test.ts
```diff
@@ -615,6 +615,82 @@ describe('TileDocumentHandler', () => {
...
+    // apply genesis
+    const genesisCommitData = {
+      cid: FAKE_CID_1,
+      type: CommitType.GENESIS,
+      commit: payload,
+      envelope: genesisCommit.jws,
+    }
+    const genesisState = await tileDocumentHandler.applyCommit(genesisCommitData, context)
+
+    // make a first update
```
stbrody, line 637: do we need this first update?  Can we remove it to simplify and focus the test?

packages/stream-tile-handler/src/__tests__/tile-document.test.ts
```diff
@@ -615,6 +615,82 @@ describe('TileDocumentHandler', () => {
...
+        cid: FAKE_CID_3,
+        type: CommitType.SIGNED,
+        commit: sPayload2,
+        envelope: signedCommit2.jws,
+      }
+      await expect(tileDocumentHandler.applyCommit(
+        signedCommitData_2,
+        context,
+        deepCopy(state1)
+      )).rejects.toThrow(/undefined/)
```
stbrody, line 689: can we check for a more meaningful error message here?

packages/stream-tile/src/tile-document.ts
```diff
@@ -338,14 +324,34 @@ export class TileDocument<T = Record<string, any>> extends Stream {
     newContent: T | null | undefined,
     newMetadata?: TileMetadataArgs
   ): Promise<CeramicCommit> {
+    const commit = await this._makeRawCommit(signer, newContent, newMetadata)
+    return TileDocument._signDagJWS(signer, commit)
+  }
+
+  /**
+   * Helper function for makeCommit() to check and
```
stbrody, line 332: incomplete sentence

stbrody (approved):
lgtm with a few small cleanups suggested

packages/stream-tile/src/tile-document.ts
stbrody: This _signDagJWS helper is an implementation detail of `makeCommit()`, it shouldn't be considered part of the public API.  The only reason we need it exposed at all is for a unit test.  I don't think exporting it as its own top-level function makes sense for something that should never be called directly outside of this file or its corresponding unit test

packages/stream-tile-handler/src/__tests__/tile-document.test.ts
willex: Done. Also updated the `tile-controller.test.ts` for symmetry.

packages/stream-tile-handler/src/__tests__/tile-document.test.ts
willex: You are correct - first update is not necessary. Removed.

packages/stream-tile/src/tile-document.ts
willex: Rectified.

PR #2160, opened 2022-05-14
Origin/spec dispatchers outstanding queries set should time out entries after a minute

Jovonni:
## Description
Adding support for event-driven management of a Node's outstanding queries to the network.

This PR abstracts outstanding queries from a map, defined below, to a class: 

```typescript 
readonly outstandingQueries: Map<string, StreamID> = new Map()
```

```OutstandingQueries``` defines properties to be used in conditional logic supporting multiple responses being considered by a single node in response to the nodes ```Query``` to the network. 

In essence, this creates a 1-to-many relationship between a ```Query ID```, and a ```Query Response```. Before, this was a **strictly** 1-to-1.

Currently, upon a node receiving a response from the network, inside:
```javascript
async _handleResponseMessage(message: ResponseMessage): Promise<void> {
```

The node will remove the value in the outstanding queries map corresponding with the key, queryId:
https://github.com/ceramicnetwork/js-ceramic/blob/942a7c1cd1985ebc3d36070aa5dee8e25211fcc9/packages/core/src/dispatcher.ts#L362

This change allows for a node to continue to consider more than 1 response from the network, in response to its own query. 

## How Has This Been Tested?
This has been tested with a single node connected to the network. Manual monitoring of node behavior while querying the network, and receiving responses. Unit tests are also integrated for outstanding queries. 

- [X] Units Tests for Outstanding Queries Passing
- [X] Behavioral Test on a running node performing the intended actions.

## PR checklist

## References:


package.json
```diff
@@ -37,6 +37,7 @@
   },
   "prettier": "eslint-config-3box/prettier.config",
   "dependencies": {
+    "@types/js-priority-queue": "^0.0.6",
```
ukstv, line 40: It seems, this package is not needed.

packages/core/src/dispatcher.ts
```diff
@@ -345,21 +345,34 @@ export class Dispatcher {
...
-      if (!newTip) {
-        throw new Error(
-          "Response to query with ID '" +
-            queryId +
-            "' is missing expected new tip for StreamID '" +
-            expectedStreamID +
-            "'"
-        )
+    const outstandingQuery = this.messageBus.outstandingQueries.queryMap.get(queryId)
+    if(outstandingQuery == undefined){
```
ukstv, line 349: pls, use prettier, Cmd+Option+Shift+P in WebStorm to auto-format the code

packages/core/src/dispatcher.ts
```diff
@@ -345,21 +345,34 @@ export class Dispatcher {
...
+      console.warn(`Response to query with ID ${queryId} is not a valid outstanding query`)
```
ukstv, line 350: IMO, it is better to do
```
if (!outstandingQuery) {
  console.warn(`...`)
  return
}
// .. and here goes your else-block
```

ukstv: or, alternatively, there is a question mark operator in TS: https://www.delftstack.com/howto/typescript/question-mark-in-typescript/

Could work like `const expectedStreamID = outstandingQuery?.streamID`.

packages/core/src/dispatcher.ts
```diff
@@ -345,21 +345,34 @@ export class Dispatcher {
...
-      this.messageBus.outstandingQueries.delete(queryId)
-      // TODO Iterate over all streams in 'tips' object and process the new tip for each
+    }
+
+    //either way, cleanup outstanding queries
+    try{
+      this.messageBus.outstandingQueries.cleanUpExpiredQueries();
+    }catch(e){
+      const errorMessage = `Error in _handle while trying ot clean up outstanding queries, ${e.message}`
+      console.error(errorMessage)
```
ukstv, line 374: Consider using logger

packages/core/src/pubsub/message-bus.ts
```diff
@@ -26,11 +26,16 @@ function betweenTimeout<T>(betweenMs: number): UnaryFunction<Observable<T>, Obse
   )
 }
 
+/*
+  @active-branch
```
ukstv, line 30: ?!

packages/core/src/pubsub/message-bus.ts
```diff
@@ -68,7 +73,16 @@ export class MessageBus extends Observable<PubsubMessage> implements Subscriptio
...
     const queryMessage = buildQueryMessage(streamId)
     this.next(queryMessage)
-    this.outstandingQueries.set(queryMessage.id, streamId)
+    const timeNow: number = Date.now();
+    const query = new Query(timeNow, streamId, queryMessage.id);
+    
+    try{
+      //add query to outstanding query set
+      this.outstandingQueries.add(queryMessage.id, query)
+    }catch(e){
```
ukstv, line 82: definitely, prettier

packages/core/src/pubsub/message-bus.ts
```diff
@@ -68,7 +73,16 @@ export class MessageBus extends Observable<PubsubMessage> implements Subscriptio
   queryNetwork(streamId: StreamID): Observable<CID> {
     const queryMessage = buildQueryMessage(streamId)
     this.next(queryMessage)
-    this.outstandingQueries.set(queryMessage.id, streamId)
+    const timeNow: number = Date.now();
+    const query = new Query(timeNow, streamId, queryMessage.id);
+    
+    try{
+      //add query to outstanding query set
+      this.outstandingQueries.add(queryMessage.id, query)
```
ukstv, line 81: Does it even makes sense for `outstandingQueries.add` to throw an error?

packages/core/src/pubsub/outstanding-queries.ts
```diff
@@ -0,0 +1,129 @@
...
+}
+
+/**
+ * Query abstraction with a timestamp, streamID, and queryID
+ */
+export class Query implements IQuery {
+    timestamp: number;
+    streamID: StreamID;
+    queryID: string;
+    constructor(t: number, sid: StreamID, qid: string){
```
ukstv, line 25:
```suggestion
    constructor(readonly timestamp: number, readonly streamID: StreamID, readonly queryID: string){
```

Works as a shortcut for lines 22-24, 26-28. They are not needed anymore.

packages/core/src/pubsub/outstanding-queries.ts
```diff
@@ -0,0 +1,129 @@
...
+
+    //otherwise they are equal, return RHS
+    return -1;
+
+};
+
+/**
+ * Outstanding Queries Abstraction holding the query map, and queue
+ */
+export class OutstandingQueries {
```
ukstv, line 58: Could you add a comment, telling what it is and what it does, please?

packages/core/src/pubsub/outstanding-queries.ts
```diff
@@ -0,0 +1,129 @@
...
+        this.queryQueue.enqueue(query);
+        return true;
+      }catch(e){
+        const errorMessage = `Error in OutstandingQueries.add(), ${e.message}`
+        console.error(errorMessage)
+        throw new Error(errorMessage)
+      }
+    }
+
+    remove(topQuery: Query) : Boolean {
```
ukstv, line 81:
```suggestion
    remove(topQuery: Query) : boolean {
```

Beware the dragons: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean

packages/core/src/pubsub/outstanding-queries.ts
```diff
@@ -0,0 +1,129 @@
...
+          
+          const differenceInMinutes = Math.round(((diffMs % 86400000) % 3600000) / 60000); // minutes
+          
+          if(differenceInMinutes > this._minutesThreshold){
+              try{
+                this.remove(topQuery);
+                //recursively check if the most prioritized query can be cleaned
+                this.cleanUpExpiredQueries()
+              }catch(e){
+                const errorMessage = `Error in OutstandingQueries.cleanUpExpiredQueries(), ${e.message}`
```
ukstv, line 117: Smells not good. 1) I fail to see what benefit it provides while hiding an original error. 2) Please, use logger. Or do not log here, instead relying on error messages in a hope that a caller would log the error.

ukstv:
Note on your dev env: We use Node.js v16 and npm v8. Even if you use yarn, please, do not commit yarn artefacts like `yarn.lock`.

packages/core/src/dispatcher.ts
```diff
@@ -345,21 +345,27 @@ export class Dispatcher {
...
-            queryId +
-            "' is missing expected new tip for StreamID '" +
-            expectedStreamID +
-            "'"
-        )
-      }
-      this.repository.stateManager.update(expectedStreamID, newTip)
-      this.messageBus.outstandingQueries.delete(queryId)
-      // TODO Iterate over all streams in 'tips' object and process the new tip for each
+    const outstandingQuery = this.messageBus.outstandingQueries.queryMap.get(queryId)
```
stbrody, line 348: what if this response message isn't for a query we sent?  Then the outstanding query won't be in the set.  Looks like as written we would wind up throwing an exception on line 352, but it's totally expected behavior that pubsub might contain RESPONSE messages that have nothing to do with this node.  We shouldn't error in that case.  We should probably check if there's no outstanding query for this queryId and early return without error from _handleResponseMessage if so

packages/core/src/dispatcher.ts
```diff
@@ -345,21 +345,27 @@ export class Dispatcher {
...
+          expectedStreamID +
+          "'"
+      )
+    }
+    this.repository.stateManager.update(expectedStreamID, newTip)
+    // TODO Iterate over all streams in 'tips' object and process the new tip for each
+
+    //either way, cleanup outstanding queries
+    try {
+      this.messageBus.outstandingQueries.cleanUpExpiredQueries()
```
stbrody, line 365: I think this should happen automatically inside the outstandingQueries object every time a new query is added.  That way the garbage collection can be entirely an implementation detail of the OutstandingQueries class, and calling code doesn't need to know anything about it, or have to remember to call anything to make sure it happens.

packages/core/src/pubsub/__tests__/outstanding-queries.test.ts
```diff
@@ -0,0 +1,143 @@
...
+    outstandingQueries.add(testQueryID, q1)
+    assert(outstandingQueries.queryMap.size === 1)
+    assert(outstandingQueries.queryQueue.size() === 1)
+    outstandingQueries.remove(q1)
+    expect(outstandingQueries.queryMap.size).toEqual(0)
+
+    expect(outstandingQueries.queryQueue.size()).toEqual(0)
+  })
+
+  test('Oustanding Queries, cleanUpExpiredQueries', async () => {
```
stbrody, line 124: would be good to have a test where some of the queries are expired and others aren't, to make sure it only cleans up the ones that are actually expired.

packages/core/src/pubsub/outstanding-queries.ts
```diff
@@ -0,0 +1,102 @@
...
+  if (a.timestamp < b.timestamp) {
+    return -1
+  }
+
+  //otherwise they are equal, return RHS
+  return -1
+}
+
+/**
+ * Outstanding Queries Abstraction holding the query map, and queue.
```
stbrody, line 42: this comment has more information about implementation details (holds a map and queue) than about what this class represents and how it's used.  How about something like:

"OutstandingQueries tracks a set of all query messages that have been sent to pubsub, for which we are still waiting on a response pubsub message.  It also takes care of garbage collecting old queries that have been outstanding for more than 1 minute."

packages/core/src/pubsub/outstanding-queries.ts
```diff
@@ -0,0 +1,102 @@
...
+ * This class contains the Queue & Map representing the outstanding queries being expected from the network
+ */
+export class OutstandingQueries {
+  readonly queryQueue: PriorityQueue<IQuery> = new PriorityQueue<IQuery>(compareQueryTimestamps)
+  readonly queryMap: Map<string, Query> = new Map()
+  readonly _logger: DiagnosticsLogger = new DiagnosticsLogger(LogLevel.important, false)
```
stbrody, line 48: there should be a single logger instance for an entire Ceramic node.  We shouldn't be constructing new loggers on the fly, but rather any class that needs access to the logger should take it as an argument in its constructor, and should be provided the Ceramic-wide logger instance during process initialization.

In this case though, OutstandingQueries should be a simple container object and I don't think should be doing any logging anyway

packages/core/src/pubsub/outstanding-queries.ts
```diff
@@ -0,0 +1,102 @@
...
+  //set the time in minutes we want to allow outstanding requests to be considered
+  private _minutesThreshold = 1
+
+  add(id: string, query: Query): boolean {
+    //enforce no duplicate outstanding queries
+    if (this.queryMap.get(id) != undefined) {
```
stbrody, line 54: query ids are deterministically generated from the query contents, so it's actually totally normal to send multiple queries with the same queryid.

I think if a query with a duplicate queryid is added it should replace the existing query (i.e. update its expiration time)

packages/core/src/pubsub/outstanding-queries.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  remove(topQuery: Query): boolean {
+    try {
+      //remove queryId key for top query
+      this.queryMap.delete(topQuery.queryID)
+      //dequeue top query
+      this.queryQueue.dequeue()
+      return true
+    } catch (e) {
+      const errorMessage = `Error in OutstandingQueries.remove(), ${e.message}`
```
stbrody, line 73: none of this code should throw.  This is a simple storage type, adding/removing from a map or priority queue should never fail.  There should be no need for try/catches, logging errors, or throwing any exceptions within this class

packages/core/src/pubsub/outstanding-queries.ts
```diff
@@ -0,0 +1,102 @@
...
+    } else {
+      // add to map
+      this.queryMap.set(id, query)
+      // add to queue
+      this.queryQueue.enqueue(query)
+      return true
+    }
+  }
+
+  remove(topQuery: Query): boolean {
```
stbrody, line 65: no need for a boolean return type when it always returns true.  The mutation functions here can all be `void` methods

packages/core/src/pubsub/outstanding-queries.ts
```diff
@@ -0,0 +1,102 @@
...
+    }
+  }
+
+  /**
+   * Event-driven method to clean up outdates outstanding queries
+   * @param
+   * @public
+   */
+  cleanUpExpiredQueries() {
+    if (this.queryQueue.size() > 0) {
```
stbrody, line 85:
```suggestion
    if (!this.queryQueue.size()) {
      return
    }
```

stbrody: Early returns allow you to avoid an extra indentation level and help keep code much more readable IMO

packages/core/src/pubsub/outstanding-queries.ts
```diff
@@ -0,0 +1,102 @@
...
+      const topQuery: Query = this.queryQueue.front()
+      const diffMs = Date.now() - topQuery.timestamp // milliseconds
+      const differenceInMinutes = Math.round(((diffMs % 86400000) % 3600000) / 60000) // minutes
+      if (differenceInMinutes > this._minutesThreshold) {
+        try {
+          this.remove(topQuery)
+          //recursively check if the most prioritized query can be cleaned
```
stbrody, line 92: I think this code would be much simpler and easier to read if it was written iteratively instead of recursively.  Then you also don't need to worry about exhausting the stack if there are a lot of expired queries.

packages/stream-tile-handler/src/tile-document-handler.ts
```diff
@@ -273,3 +273,4 @@ export class TileDocumentHandler implements StreamHandler<TileDocument> {
     return cacao
   }
 }
+
```
stbrody, line 276: rm extraneous change to this file, helps keep the git history clean when files are only updated by real changes

stbrody (changes requested):
good foundation, some small comments throughout

packages/core/src/pubsub/outstanding-queries.ts
oed: @Jovonni can you please mark comments you addressed as resolved?

Makefile
```diff
@@ -0,0 +1,35 @@
+build:
```
stbrody, line 1: please don't check in a Makefile into the repo.  If you want one locally that's fine if it's helpful for your workflow but we shouldn't put this into the codebase.

ukstv (approved):

packages/core/src/pubsub/__tests__/outstanding-queries.test.ts
stbrody: sorry where? I'm not seeing it

packages/core/src/pubsub/outstanding-queries.ts
```diff
@@ -0,0 +1,93 @@
...
+export class OutstandingQueries {
+  readonly queryQueue: PriorityQueue<IQuery> = new PriorityQueue<IQuery>(compareQueryTimestamps)
+  readonly queryMap: Map<string, Query> = new Map()
+  //set the time in minutes we want to allow outstanding requests to be considered
+  private _minutesThreshold = 1
+
+  add(id: string, query: Query): void {
+    //enforce no duplicate outstanding queries
+    this.cleanUpExpiredQueries()
+    if (this.queryMap.get(id) == undefined) {
```
stbrody, line 55: I think if a duplicate query id gets added it should replace the existing entry, updating its expiration time, rather than keeping the older expiration time around.  Otherwise you might age out a query from the set that's less than a minute old because there was a previous query for the same id that was older

packages/core/src/pubsub/__tests__/outstanding-queries.test.ts
```diff
@@ -0,0 +1,161 @@
...
+    expect(outstandingQueries.queryMap.size).toEqual(0)
+    expect(outstandingQueries.queryQueue.size()).toEqual(0)
+  })
+
+  test('Oustanding Queries, restrict duplicate outstanding queries', async () => {
+    const t1 = new Date('2022-05-12T10:20:30Z').getTime()
+    const testQueryID = 'testID'
+    const q1 = new Query(t1, FAKE_STREAM_ID, testQueryID)
+    outstandingQueries.add(testQueryID, q1)
+    expect(!outstandingQueries.add(testQueryID, q1))
```
stbrody, line 159: the `add` method is a void method, so this `expect` check isn't really checking anything since `add` has no return value

stbrody: I think this test should be updated to test that when a duplicate query id is added, the timestamp in the set is updated to that of the newer query

packages/core/src/pubsub/outstanding-queries.ts
```diff
@@ -0,0 +1,93 @@
...
+      // add to map
+      this.queryMap.set(id, query)
+      // add to queue
+      this.queryQueue.enqueue(query)
+    }
+  }
+
+  remove(topQuery: Query): void {
```
stbrody, line 63: do we need a remove method?  Is it used anywhere?  Better to keep the API smaller if we don't need it

packages/core/src/pubsub/outstanding-queries.ts
```diff
@@ -0,0 +1,93 @@
...
+    //remove queryId key for top query
+    this.queryMap.delete(topQuery.queryID)
+    //dequeue top query
+    this.queryQueue.dequeue()
+  }
+
+  /**
+   * Event-driven method to clean up outdates outstanding queries
```
stbrody, line 71:
```suggestion
   * Event-driven method to clean up outdated outstanding queries
```

packages/core/src/pubsub/outstanding-queries.ts
```diff
@@ -0,0 +1,93 @@
...
+   * @param
+   * @public
+   */
+  cleanUpExpiredQueries(): void {
```
stbrody, line 75: Let's make this private.  There's no reason for consuming code to ever call this directly since it's called as part of the `add()` method.  Always good to keep the surface area of public APIs as small as possible

stbrody, line 75:
```suggestion
  private _cleanUpExpiredQueries(): void {
```

packages/core/src/pubsub/outstanding-queries.ts
```diff
@@ -0,0 +1,93 @@
...
+    for (const _ of this.queryQueue.toArray()) {
```
stbrody, line 76:
```suggestion
    while (isExpired(this.queryQueue.front())
```

stbrody: If you take the logic of lines 78-80 and put them into a `isExpired` helper method that can take a Query object and tell you whether or not it is expired, then you can have the logic of the loop here be checking the exact condition you want, rather than having it set to iterate over the entire set, but have a `break` inside the loop body

packages/core/src/pubsub/outstanding-queries.ts
stbrody: I see now we use it in `cleanUpExpiredQueries`.  Let's just make this private then

packages/core/src/pubsub/outstanding-queries.ts
```diff
@@ -0,0 +1,93 @@
...
+      const topQuery: Query = this.queryQueue.front()
+      const diffMs = Date.now() - topQuery.timestamp // milliseconds
+      const differenceInMinutes = Math.floor(diffMs / 1000 / 60)
+      if (differenceInMinutes > this._minutesThreshold) {
+        try {
+          this.remove(topQuery)
```
stbrody, line 82: why would remove() ever throw?

stbrody (changes requested):
getting close, a few final suggestions

packages/core/src/pubsub/__tests__/outstanding-queries.test.ts
```diff
@@ -139,23 +139,27 @@ describe('Outstanding Queries', () => {
...
     const t1 = new Date('2022-05-12T10:20:30Z').getTime()
     const testQueryID = 'testID'
     const q1 = new Query(t1, FAKE_STREAM_ID, testQueryID)
     outstandingQueries.add(testQueryID, q1)
-    expect(!outstandingQueries.add(testQueryID, q1))
+    const t2 = new Date('2022-05-13T10:20:30Z').getTime()
+    const q2 = new Query(t2, FAKE_STREAM_ID, testQueryID)
+    outstandingQueries.add(testQueryID, q2)
+    expect(outstandingQueries.queryMap.size == 1)
+    expect(outstandingQueries.queryQueue.size() == 1)
```
stbrody, line 163: would be good to also test that the query in the map has the new timestamp not the old one

packages/core/src/pubsub/outstanding-queries.ts
```diff
@@ -67,27 +72,24 @@ export class OutstandingQueries {
...
+    } else {
+      return false
+    }
+  }
+
   /**
-   * Event-driven method to clean up outdates outstanding queries
+   * Event-driven method to clean up outdated outstanding queries
    * @param
    * @public
```
stbrody, line 88: private

packages/core/src/pubsub/outstanding-queries.ts
```diff
@@ -67,27 +72,24 @@ export class OutstandingQueries {
     this.queryQueue.dequeue()
   }
 
+  private _isExpired(query: Query): boolean {
+    const diffMs = Date.now() - query?.timestamp // milliseconds
+    const differenceInMinutes = Math.floor(diffMs / 1000 / 60)
+    if (differenceInMinutes > this._minutesThreshold) {
```
stbrody, line 78: could just be `return differenceInMinutes > this._minutesThreshold`

packages/core/src/pubsub/outstanding-queries.ts
```diff
@@ -67,27 +72,24 @@ export class OutstandingQueries {
...
-        } catch (e) {
-          const errorMessage = `Error in OutstandingQueries.cleanUpExpiredQueries(), ${e.message}`
-          throw new Error(errorMessage)
-        }
-      } else {
-        // front queue is not expired, stop iteration over queue
-        break
-      }
+  private _cleanUpExpiredQueries(): void {
+    while (this._isExpired(this.queryQueue.front())) {
```
stbrody, line 91: nice, very clean now

packages/core/src/pubsub/outstanding-queries.ts
```diff
@@ -0,0 +1,95 @@
...
+      this.queryQueue.enqueue(query)
+    } else {
+      //replace query
+      this.remove(query)
+      this.queryMap.set(id, query)
+      this.queryQueue.enqueue(query)
+    }
+  }
+
+  remove(topQuery: Query): void {
```
stbrody, line 68:
```suggestion
  private _remove(topQuery: Query): void {
```

stbrody (approved):
LGTM!  A few small cleanups suggested, but feel free to go ahead and merge after addressing them.

Nice work!

stbrody (approved):
LGTM, feel free to merge!

PR #2164, opened 2022-05-17
chore(core, model-instance-document, model-instance-document-handler): Add basic packages for Model and ModelInstaceDocument copied from TileDocument

stbrody:
Main actual pieces worth considering/reviewing are:
* the package names - they're long and verbose, but clear.  I am however a little worried about the length for commit messages using semver, eg: `feat(stream-model-instance-document-handler): add a new feature`
* the streamtype identifier.  I went with `instance`


oed (commented):
Naming suggestions.

Also some small refactor suggestions, but might be out of scope.

stbrody:
@oed what about `@ceramicnetwork/stream-model-instance` and `@ceramicnetwork/stream-model-instance-handler` as the package names?  So dropping the "document" for the sake of brevity, but still leaving the name long enough to have clarify for what it represents.

oed (approved):
lgtm

PR #2165, opened 2022-05-17
temporarily log the full headers from dapps in the wild - PLAT-535

gvelez17:
## Description

In order to determine what metrics we are able to gather about partners and dapps using Ceramic, we need to examine the full headers produced by dapps in the wild.  The only way to reliably gather this is by temporarily logging full headers in the wild, ie at least on clay.

## How Has This Been Tested?

This is only modifying a well exercised log pathway, the smoke tests should be sufficient.

## PR checklist

- [x ] I have tagged the relevant reviewers and interested parties

## References:

See https://www.notion.so/threebox/js-ceramic-instrumentation-8e1baf720b764b7ab0cf1e25d8a50da1

### DIFF
--- a/packages/cli/src/daemon/log-requests.ts
+++ b/packages/cli/src/daemon/log-requests.ts
@@ -3,7 +3,7 @@ import { Request, Response } from 'express'
 import morgan from 'morgan'
 
 const ACCESS_LOG_FMT =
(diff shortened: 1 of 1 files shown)


smrz2001 (approved):
🚢

PR #2166, opened 2022-05-18
Indexing configuration

ukstv:
- Pass `--config` option from CLI to actual code
- Read connection string for a database
- Specify models to index in the config file

Also: replace `path.resolve` with `URL`-based resolution.

### DIFF
--- a/packages/cli/src/daemon-config.ts
+++ b/packages/cli/src/daemon-config.ts
@@ -88,7 +88,8 @@ export class DaemonHTTPApiConfig {
   /**
    * Port to listen on.
    */
-  @jsonMember(AnyT, {serializer: inPort => {
+  @jsonMember(AnyT, {
+    serializer: (inPort) => {
       const validPort = Number(inPort)
       if (inPort == undefined || inPort == null) {
         return inPort
@@ -97,7 +98,7 @@ export class DaemonHTTPApiConfig {
         process.exit(1)
       }
       return validPort
-    }
+    },
   })
   port?: number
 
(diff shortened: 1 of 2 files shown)


packages/cli/src/daemon-config.ts
```diff
@@ -166,6 +168,26 @@ export class DaemonAnchorConfig {
...
+    deserializer: (arr?: Array<string>) => {
+      if (!arr) return arr
+      return arr.map(StreamID.fromString)
+    },
+    serializer: (arr?: Array<StreamID>) => {
+      if (!arr) return arr
+      return arr.map((s) => s.toString())
+    },
+  })
+  models: string[]
```
stbrody, line 188: this seems backwards to me.  Shouldn't the `models` field be of type `StreamID[]`, and the serializer translate a string array into a streamid array?

That would match the way the serializer for `port` works above - the serializer turns the input from the user as a string into the type we want to use internally

stbrody (commented):
one question about the options serializer, otherwise looks good

packages/cli/src/daemon-config.ts
ukstv: Other way around. Serialisers were always present in the correct form. Types were a bit off.

ukstv: I guess, `port` one is broken. According to https://github.com/JohnWeisz/TypedJSON#serializer-and-deserializer `deserializer` turns input from JSON text to a runtime object, and `serializer` turns it back to JSON text.

stbrody: Huh.  I wonder then why we haven't seen issues with the `port` serializer.  That was a pretty new change - maybe its broken but no one's noticed yet?  @willex can you double-check this?

stbrody (approved):
lgtm.  Presumably if there's any issues with serialization/deserialization of the config file options, they'll turn up later in testing

packages/cli/src/daemon-config.ts
willex: Yes, looking at it now.

PR #2167, opened 2022-05-19
feat(core): Setup database connection for indexing, SQLite only

ukstv:
- add indexing module to Ceramic core
- parse connection string and initialise an appropriate instance of `DatabaseIndexAPI`

There is a slight divergence from the spec: added `init` method to `DatabaseIndexAPI`.

Note: custom base branch for PR to reduce perceived diff size. Going to change to `develop` once https://github.com/ceramicnetwork/js-ceramic/pull/2166 is merged.


packages/core/src/indexing/types.ts
```diff
@@ -0,0 +1,66 @@
+import type { StreamID } from '@ceramicnetwork/streamid'
+import type { DID } from 'dids'
+
+export interface IndexStreamArgs {
+  streamID: StreamID
+  model: StreamID
+  controller: DID
+  lastAnchor: Date | null
+  content: Record<string, any>
```
stbrody, line 9: I'd probably leave out `content` and `relationFields` entirely for now, we can add them in the future when we actually support doing anything with them

packages/core/src/indexing/types.ts
```diff
@@ -0,0 +1,66 @@
...
+  relationFields?: Map<string, StreamID | DID>
+}
+
+export enum Ordering {
```
stbrody, line 13: I thought the conclusion on the spec was to leave Ordering out entirely for now

packages/core/src/indexing/types.ts
```diff
@@ -0,0 +1,66 @@
...
+  CHRONOLOGICAL = 'chronological', // last_anchored_at DESC NULLS FIRST, created_at DESC
+  INSERTION = 'insertion', // created_at DESC = when an entry was added to the index
+}
+
+export interface BaseQuery {
+  model: StreamID | string
+  account?: string
```
stbrody, line 20: `DID | string`?

stbrody (commented):
looks good, a couple of small questions about leaving out a few things that we had discussed on the spec leaving out until a later version

packages/core/src/indexing/types.ts
oed: No, a DID identifier is always a string. the DID class contains functionality for actions using a DID.

packages/core/src/indexing/types.ts
ukstv: We better have it here.

If we go with a single ordering method while implementing the indexer, there is a high chance for it to be omitted in some internal API. The consequence is at the later stage, when we really need multiple methods, we would need to refactor some of the code. That will happen at an unfortunate moment when most of the code is solidified. To me it makes more sense to approach indexing code now having in mind and at hand variability of ordering methods, even if only one of the options is finally exposed to a user.

stbrody: okay, so long as it doesn't wind up exposed to any user-facing APIs, and we can change the names used later if we want to do before exposing them, then that's fine

ukstv:
So, @stbrody @oed @willex do you mind marking the PR approved? Any more comments?

stbrody (approved):
LGTM

BTW, the reason I hadn't come back to this PR is because it hadn't been sent back to me for additional review.  In the future, if you click the "re-request review" button next to my name in the reviewers list, that will make sure I circle back to any PRs that have been updated since I last looked

PR #2169, opened 2022-05-19
chore(stream-tile-handler, stream-model-handler, stream-model-instance-handler): Unify implementation of signature verification across streamtypes

stbrody:
FYI @Jovonni as this moves the code you need to update for https://linear.app/3boxlabs/issue/NET-1437/replace-cacao-support-for-family-with-model-instead


packages/common/src/utils/signature_utils.ts
```diff
@@ -0,0 +1,71 @@
+import type { Cacao } from 'ceramic-cacao'
+import type { DID } from 'dids'
+import { CommitData } from '../index.js'
+import { StreamID } from '@ceramicnetwork/streamid'
```
oed, line 4:
```suggestion
import type { CommitData } from '../index.js'
import type { StreamID } from '@ceramicnetwork/streamid'
```
I believe we only need type here?

oed (approved):
lgtm

PR #2170, opened 2022-05-19
chore(stream-model, stream-model-handler): Rm unneeded schema and deterministic functionality from Model streamtype

stbrody:


oed (approved):
lgtm

PR #2172, opened 2022-05-19
Origin/net 1393 add family to pubsub update messages

Jovonni:
## Description

[Draft]
Refactoring to add 'model' to the UPDATE pubsub messages

## How Has This Been Tested?

- [ ] Test B

## PR checklist

## References:

### DIFF
--- a/packages/core/src/pubsub/pubsub-message.ts
+++ b/packages/core/src/pubsub/pubsub-message.ts
@@ -102,7 +102,7 @@ export function serialize(message: PubsubMessage): Uint8Array {
         doc: message.stream.toString(),
         stream: message.stream.toString(),
         tip: message.tip.toString(),
-        model: message.model?.toString()
+        ...(message.model && { model: message.model.toString() })
       }
       return textEncoder.encode(JSON.stringify(payload))
     }


packages/core/src/__tests__/dispatcher-mock-ipfs.test.ts
```diff
@@ -225,10 +230,17 @@ describe('Dispatcher with mock ipfs', () => {
...
     const queryID = queryMessageSent.id
 
-    // Handle UPDATE message
+    // Handle UPDATE message without model
     dispatcher.repository.stateManager.update = jest.fn()
     await dispatcher.handleMessage({ typ: MsgType.UPDATE, stream: FAKE_STREAM_ID, tip: FAKE_CID })
-    expect(dispatcher.repository.stateManager.update).toBeCalledWith(state$.id, FAKE_CID)
+
+    // Handle UPDATE message with model (to ensure optional use of model parameter)
+    dispatcher.repository.stateManager.update = jest.fn()
```
stbrody, line 238: the update function is already mocked on line 234, doing it again here is redundant.

packages/core/src/dispatcher.ts
```diff
@@ -310,10 +310,14 @@ export class Dispatcher {
   async _handleUpdateMessage(message: UpdateMessage): Promise<void> {
     // TODO Add validation the message adheres to the proper format.
 
-    const { stream: streamId, tip } = message
+    /*
+      @active-branch
```
stbrody, line 314: ?

packages/core/src/dispatcher.ts
```diff
@@ -310,10 +310,14 @@ export class Dispatcher {
...
+    */
+    console.log("_handleUpdateMessage")
```
stbrody, line 316: don't forget to clean up extra logs used during development before sending code out for review

packages/core/src/pubsub/pubsub-message.ts
```diff
@@ -20,7 +20,8 @@ export enum MsgType {
 export type UpdateMessage = {
   typ: MsgType.UPDATE
   stream: StreamID
-  tip: CID
+  tip: CID,
+  model?: string
```
stbrody, line 24: model should be a StreamID

packages/core/src/pubsub/pubsub-message.ts
```diff
@@ -96,6 +97,11 @@ export function serialize(message: PubsubMessage): Uint8Array {
         doc: message.stream.toString(),
         stream: message.stream.toString(),
         tip: message.tip.toString(),
+        /*
+          @active-branch
```
stbrody, line 101: if your editor is putting these comments in please reconfigure it so it stops doing that.  Or at least make sure to take them out before putting code up for review

packages/core/src/pubsub/pubsub-message.ts
```diff
@@ -96,6 +97,11 @@ export function serialize(message: PubsubMessage): Uint8Array {
...
+          placeholder, optional type
+        */
+        model: message.model?.toString()
```
stbrody, line 104: what happens if model is missing?  Then the `payload` object will have `model: undefined`.  How will `textEncoder.encode()` handle that?  Will it strip out the `model` field (desired behavior), or will it still include the `model` field as an explicit `undefined` (not desired behavior as it increases message size unnecessarily)?

stbrody: instead of trying to figure out what `textEncoder` does here, it may be better to just proactively only add the `model` field to the payload if you know it's non-null

packages/core/src/pubsub/pubsub-message.ts
```diff
@@ -124,6 +130,11 @@ export function deserialize(message: any): PubsubMessage {
         typ: MsgType.UPDATE,
         stream,
         tip: toCID(parsed.tip),
+        /*
+          @active-branch
+          placeholder, optional type
+        */
+        model: parsed.model
```
stbrody, line 137: you'll need to parse the `model` into a StreamID like is done for the `stream` field above

packages/core/src/state-management/state-manager.ts
```diff
@@ -202,7 +202,12 @@ export class StateManager {
    * @param tip - Stream Tip CID
    * @private
    */
-  update(streamId: StreamID, tip: CID): void {
+  update(streamId: StreamID, tip: CID, model?: string): void {
```
stbrody, line 205: There's no need for this function to take a `model` argument.  This function is for applying the new update to the StreamState for the updated stream - the model property doesn't affect that behavior at all.  For now, nodes that receive UPDATE messages can just ignore the model field entirely.  It'll be used later by the indexer but that isn't hooked up yet.

stbrody (changes requested):
great start, some comments throughout.  There will of course need to be a follow-up commit to actually include the model when sending the update messages, but that can happen after this goes in.

packages/core/src/state-management/state-manager.ts
```diff
@@ -192,6 +192,7 @@ export class StateManager {
   }
 
   publishTip(state$: RunningState): void {
+    // NOTE: pass model to publishTip
     this.dispatcher.publishTip(state$.id, state$.tip)
```
oed, line 196: Something like this, naming is likely off.
```suggestion
    // NOTE: pass model to publishTip
    this.dispatcher.publishTip(state$.id, state$.tip, state$.state.metadata.model)
```

packages/core/src/pubsub/pubsub-message.ts
```diff
@@ -96,6 +97,7 @@ export function serialize(message: PubsubMessage): Uint8Array {
         doc: message.stream.toString(),
         stream: message.stream.toString(),
         tip: message.tip.toString(),
+        ...(message.model && { model: message.model.toString() }), // note: conditionally add the model key if message.model is not undefined        
       }
```
oed, line 101:
```suggestion
      }
      if (message.model) payload.model = message.model.toString()
```
I think we can make this a little easier to read.

packages/core/src/pubsub/pubsub-message.ts
```diff
@@ -124,6 +126,7 @@ export function deserialize(message: any): PubsubMessage {
         typ: MsgType.UPDATE,
         stream,
         tip: toCID(parsed.tip),
+        model: StreamID.fromString(parsed.model),
```
oed, line 129: Does `StreamID.fromString` return undefined if `parsed.model` is not defined?

packages/core/src/state-management/state-manager.ts
```diff
@@ -192,7 +192,9 @@ export class StateManager {
   }
 
   publishTip(state$: RunningState): void {
-    this.dispatcher.publishTip(state$.id, state$.tip)
+    // NOTE: pass model to publishTip
+    console.log(state$)
+    this.dispatcher.publishTip(state$.id, state$.tip, state$.state.metadata.model)
```
oed, line 197: I guess you are still working this out?

packages/core/src/state-management/state-manager.ts
```diff
@@ -192,7 +192,8 @@ export class StateManager {
   }
 
   publishTip(state$: RunningState): void {
-    this.dispatcher.publishTip(state$.id, state$.tip)
+    // NOTE: pass model to publishTip
```
oed, line 195:
```suggestion
```
This comment is no longer relevant.

packages/core/src/pubsub/pubsub-message.ts
oed: Still curious about this.

oed (approved):
lgtm, one Q and one small update suggested.

oed:
I guess also make sure tests pass :)

oed (approved):
I think this can be merged when the tests pass

packages/common/src/stream.ts
```diff
@@ -91,7 +91,7 @@ export interface StreamMetadata {
   schema?: string
   tags?: Array<string>
   forbidControllerChange?: boolean
-
+  model?: StreamID
```
stbrody, line 94: this already exists in the `develop` branch, please merge in develop branch and tag me to take another look after you do and the tests are passing

packages/core/src/__tests__/dispatcher-mock-ipfs.test.ts
```diff
@@ -253,4 +256,67 @@ describe('Dispatcher with mock ipfs', () => {
...
+
+    // Store the query ID sent when the stream is registered so we can use it as the response ID later
+    const publishArgs = ipfs.pubsub.publish.mock.calls[0]
+    expect(publishArgs[0]).toEqual(TOPIC)
+    const queryMessageSent = JSON.parse(
+      new TextDecoder().decode(publishArgs[1] as unknown as Uint8Array)
+    )
+    const queryID = queryMessageSent.id
+
+    // Handle UPDATE message without model
```
stbrody, line 289:
```suggestion
    // Handle UPDATE message with model
```

packages/core/src/__tests__/dispatcher-mock-ipfs.test.ts
```diff
@@ -253,4 +256,67 @@ describe('Dispatcher with mock ipfs', () => {
...
+      log: [
+        {
+          cid: FAKE_STREAM_ID.cid,
+          type: CommitType.GENESIS,
+        },
+      ],
+    } as unknown as StreamState
+    const state$ = await register(initialState)
+
+    // Store the query ID sent when the stream is registered so we can use it as the response ID later
```
stbrody, line 281: no need for testing QUERY or RESPONSE messages in this test, they are already tested in the previous test you copied this one from, that would just be redundant duplicate testing.  I think this test case can be made way smaller and only focused on the UPDATE message

packages/core/src/dispatcher.ts
```diff
@@ -309,8 +309,8 @@ export class Dispatcher {
    */
   async _handleUpdateMessage(message: UpdateMessage): Promise<void> {
     // TODO Add validation the message adheres to the proper format.
-
-    const { stream: streamId, tip } = message
+    // TODO model isn't used in update yet, will be in later versions
```
stbrody, line 312:
```suggestion
    // TODO(NET-1527) model isn't used in update yet, will be in later versions
```

stbrody (commented):
basically looks good.  A few small comments, and needs to be merged with develop and the tests made to pass

ukstv (approved):
LGTM mod @stbrody 's comments.

oed:
@smrz2001 Why does the new ci pipeline fail? Can it be ignored for now or should we be concerned?

smrz2001:
> @smrz2001 Why does the new ci pipeline fail? Can it be ignored for now or should we be concerned?

@oed, it fails because of known flaky unit tests. @stbrody mentioned adding a ticket this coming cycle to clean such tests up.

My recommendation would be to ignore these till the end of next cycle, after which we can both deprecate CircleCI and enforce that the new pipeline must pass before a PR can be merged.

packages/core/src/pubsub/pubsub-message.ts
stbrody: looks like this is back to always adding `model` instead of only adding it if it's not undefined

packages/core/src/pubsub/pubsub-message.ts
```diff
@@ -101,7 +102,9 @@ export function serialize(message: PubsubMessage): Uint8Array {
         doc: message.stream.toString(),
         stream: message.stream.toString(),
         tip: message.tip.toString(),
+        model: message.model?.toString()
```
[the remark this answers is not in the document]
stbrody: then you could also delete line 107 and keep all the construction of this object together in one place

packages/core/src/pubsub/pubsub-message.ts
oed: I don't think this matters since `JSON.stringify` (on line 108) removes all properties that are not defined.

oed: `undefined` is not a thing in JSON.

stbrody: Okay cool, in that case you can probably just delete line 107 and call it a day

PR #2173, opened 2022-05-19
chore(common): Add 'model' property to StreamState and RawCommit

stbrody:


stbrody:
FYI @Jovonni, your follow-up for putting model into pubsub update messages can build on this to check the stream metadata to see if it has a `model` property

oed (approved):
This doesn't touch IPLD yet right? 
Just want to make sure we use `Uint8Array` on the IPLD block encoding.

PR #2179, opened 2022-05-23
feat: Create table per indexed model

ukstv:
Also, add a stream to the index.
If a stream exists, updates `last_anchor_at` and `updated_at` fields to (now).


packages/core/src/indexing/types.ts
```diff
@@ -4,7 +4,7 @@ import type { DID } from 'dids'
 export interface IndexStreamArgs {
   streamID: StreamID
   model: StreamID
-  controller: DID
+  controller: DID | string
```
stbrody, line 7: as Joel pointed out to me on a spec doc, the controller should probably always be a `string`, not a `DID` instance.  DID instances are only useful when you actually want access to the methods for signature creation/verification or encryption/decryption, if you just want a DID identifier it's best to always just use a string.

stbrody: I do kind of which there was a different type for a DID identifier though that enforced the string structure and provided accessors for the different parts of the DID string, but that doesn't exist atm as far as I know and isn't important enough to add now

stbrody (approved):
this looks great, very clean.

stbrody:
I wonder if it's worth spending time now to detect the scenario where the table already exists, but doesn't match the structure we expect.  This seems like this could become important in future releases after we change the expected database schema.  It might be good to some time now to make sure that older versions will detect if they get pointed at a database in a format they don't understand (potentially because the database was upgraded to a new schema by a newer version of Ceramic, and then Ceramic was downgraded without downgrading the database, for example).  Could help future-proof us against human error from future node operators as managing and upgrading the index database becomes more complicated.

Fine not to include in this PR, but worth considering adding a ticket to the project for this.
CC @oed @v-stickykeys @smrz2001 @PaulLeCam

packages/core/src/indexing/sqlite/migrations/1-create-model-table.ts
```diff
@@ -0,0 +1,79 @@
...
+      {
+        name: 'stream_id',
+        type: 'VARCHAR(1024)',
+        isPrimary: true,
+        isNullable: false,
+        isUnique: true,
+      },
+      {
+        name: 'controller_did',
+        type: 'VARCHAR(1024)',
```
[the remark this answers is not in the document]
oed: There is no max length for a DID according to the did-core spec. This limit seems fine practically though

PaulLeCam:
@stbrody I agree it would be good even if not implementing the full logic, at least having an idea of how this could work and if anything needs to be defined now to anticipate the implementation of this logic.
Maybe it could simply be adding a version number to the table names so that different versions of Ceramic can know how to handle them?

ukstv:
> I wonder if it's worth spending time now...

I believe it is not.

For dev purposes we can safely assume _now_ a developer is free to wipe their database and start from scratch. For a production mode, a dedicated tool is required, as per the specification, which has own set of considerations. We have a plan for it already!

Until we engage in the design of such a tool though, there is no clarity and/or agreement on constraints of the problem space and the solution worth replicating for dev-mode. Current file/code layout have in mind future changes to accommodate for the table shape migrations. A real mechanism is to be designed later, with the production-mode tool.

I encourage you @stbrody @PaulLeCam to have a little more faith in the plan we have all agreed upon.

stbrody:
> I encourage you @stbrody @PaulLeCam to have a little more faith in the plan we have all agreed upon.

I think of the tool for configuring the database that we have in the spec as complementary to what I was proposing here.  The tool will be useful for node operators to configure their database properly, but what I was proposing was guardrails within the Ceramic node itself to detect misconfigurations, operators forgetting or failing to use the tool properly, etc.  One provides the tools to do the right thing, the other prevents you from shooting yourself in the foot if you do the wrong thing.

In any case, I don't feel so strongly that we have to do this now, so fine with leaving out of scope for now and revisiting in the future if need be

PR #2182, opened 2022-05-25
feat(stream-model, stream-model-handler): Core APIs for Model streamtype

stbrody:


packages/stream-model/src/model.ts
```diff
@@ -158,63 +68,146 @@ async function throwReadOnlyError(): Promise<void> {
...
   static STREAM_TYPE_ID = 2
 
+  // The hardcoded "model" StreamID that all Model streams have in their metadata. This provides
+  // a "model" StreamID that can be indexed to query the set of all published Models.
+  private static readonly MODEL: StreamID = (function () {
+    const data = encode('model-v1')
+    const multihash = multihashes.encode(data, 'identity')
+    const digest = create(code, multihash)
+    const cid = CID.createV1(code, digest)
+    return new StreamID(Model.STREAM_TYPE_ID, cid)
```
stbrody, line 107: @oed not sure if I did this right or how you had been thinking, please let me know if there's a better/different way to construct the meta-model StreamID

packages/streamid/src/stream-id.ts
```diff
@@ -92,8 +92,8 @@ const TAG = Symbol.for('@ceramicnetwork/streamid/StreamID')
 export class StreamID implements StreamRef {
   protected readonly _tag = TAG
 
-  readonly #type: number
-  readonly #cid: CID
+  private readonly _type: number
```
stbrody, line 95: Unfortunately using proper private variables with the hashsign prefix causes `lodash.cloneDeep` to be unable to properly clone StreamIDs.  This caused all kinds of errors as we're now adding StreamIDs into StreamState and StreamMetadata for the first time, both of which have `cloneDeep()` called on them in several places throughout the codebase.

packages/stream-model-handler/src/__tests__/model-handler.test.ts
```diff
@@ -191,6 +131,7 @@ const setDidToNotRotatedState = (did: DID) => {
   did.createJWS = async () => jwsForVersion0
 }
 
+// TODO: De-dupe this with similar code from tile-document-handler.test.ts and model-instance-document.test.ts
```
stbrody, line 134: I'm pretty disappointed about this.  There is quite a lot of somewhat nasty test code here for managing key rotations that is completely duplicated between this test and the equivalent tests for TileDocumentHandler and ModelInstanceDocumentHandler.  I tried to pull this code into `TestUtils`, but unfortunately moving this into the `common` package caused a bunch of circular dependencies that broke compilation.  I think sometime soon we're going to need to rethink our package structure and create some more shared packages for helper functionality and test functionality that can be shared across the majority of our packages.  But for now it was simplest just to leave this code duplicated.  I don't love it, but at least it's only test code.

packages/common/src/stream.ts
```diff
@@ -33,7 +33,7 @@ export enum AnchorStatus {
 export interface CommitHeader {
   controllers: Array<string>
   family?: string
-  model?: StreamID
+  model?: Uint8Array // StreamID encoded as byte array
```
oed, line 36: 🙌🏻

packages/stream-model/src/model.ts
oed: This is what I had in mind yes 👍

packages/streamid/src/stream-id.ts
oed: :(

oed (commented):
Looks good from a quick scan.

packages/stream-model/src/model.ts
```diff
@@ -12,119 +10,31 @@ import {
...
   GenesisCommit,
   RawCommit,
   CeramicApi,
   SignedCommitContainer,
-  StreamMetadata,
   CeramicSigner,
   GenesisHeader,
 } from '@ceramicnetwork/common'
 import { CommitID, StreamID, StreamRef } from '@ceramicnetwork/streamid'
+import type { JSONSchema } from 'json-schema-typed/draft-07'
```
pawartur, line 21:
```suggestion
import type { JSONSchema } from 'json-schema-typed/draft-2020-12';
```

packages/stream-model/src/model.ts
```diff
@@ -158,63 +68,146 @@ async function throwReadOnlyError(): Promise<void> {
   )
 }
 
+/**
+ * Represents the relationship between an instance of this model and the controller account.
+ * 'list' means there can be many instances of this model for a single account. 'link' means
+ * there can be only one instance of this model per account (if a new instance is created it
+ * overrides the old one).
+ */
+export enum ModelAccountRelation {
```
pawartur, line 77: In `@glazed/types` we currently have `export declare type ModelAccountRelation = 'list' | 'set' | 'link' | 'none';`

Would it make sense to put this in `common` and use the same types everywhere?

packages/stream-model/src/model.ts
```diff
@@ -158,63 +68,146 @@ async function throwReadOnlyError(): Promise<void> {
...
+  LIST = 'list',
+  LINK = 'link',
+}
+
+/**
+ * Contents of a Model Stream.
+ */
+export interface ModelDefinition {
```
pawartur, line 85: In `@glazed/types` we have:
```
export declare type ModelDefinition = {
    name: string;
    description?: string;
    schema: JSONSchema.Object;
    accountRelation: ModelAccountRelation;
    relations?: ModelRelationsDefinition;
    views?: ModelViewsDefinition;
};
```

I think `relations` and `views` will not be implemented for the MVP, but just highlighting in case we want to use the same types from `commons` in `glazed` and `ceramic`.

oed: Good callout. Would be great if we can import types from Ceramic in glaze.

stbrody: yeah I think once this PR gets merged it would be great if glaze switched to importing and using these types.  I don't think anything needs to change with this PR for that to be possible

stbrody: I'd rather leave them out for now so it's clear to users that those fields can't be set yet.  I don't want people trying to make models with those fields now and have them ignored or be invalid when we eventually add those fields for real

pawartur: Sounds good, thanks :) !

ukstv (approved):

PR #2184, opened 2022-05-27
feat: Chronological order for indexing, SQLite-only

ukstv:
Supports forward and backward pagination.

Few things to have in mind.
1. CC @PaulLeCam. In the current implementation one could traverse along the selected order.
For the forward pagination, after retrieving page `N` (=`first M after CURSOR`), you could only request page `N + 1`: `hasPrevPage` is set to `false`.
For the backward pagination, after retrieving page `N` (=`last M before CURSOR`), you could only request page `N - 1`: `hasNextPage` is set to `false`.

It conforms to the [GraphQL Cursor Connections Specification](https://relay.dev/graphql/connections.htm#HasNextPage()). See "If the server can efficiently determine that elements exist prior to after" note there.

Benefits:
- Just one SQL query per `page` request. If we add an indication of a previous page, that would add one more SQL query of comparable complexity.
- No surprising behaviour. Requesting previous pages for a forward (from the most recent to the oldest) chronological order might be surprising, because of out-of-order updates.

2. SQL query language via Knext. TypeORM do not have good support for building queries. So, I used https://knexjs.org/ to build them. It is less obvious than vanilla SQL, but removes lots of duplication by making queries composable.

3. Testing. There is a CSV file with entries in chronological order. We test the query results against the file.


packages/core/src/indexing/sqlite/chronological-order.ts
```diff
@@ -0,0 +1,184 @@
...
+        throw new UnsupportedOrderingError(paginationKind)
+    }
+  }
+
+  private chronologicalForwardQuery(
+    query: BaseQuery,
+    pagination: ForwardPaginationQuery
+  ): Knex.QueryBuilder {
+    const tableName = asTableName(query.model)
+    const base = this.knexConnection
```
stbrody, line 118: I'm surprised you couldn't do this with TypeOrm.  TypeOrm does have a querybuilder API that can express some fairly complex queries.  See for example this query from the CAS: https://github.com/ceramicnetwork/ceramic-anchor-service/blob/f938eee1960591160126481fb580afe1706c1fbe/src/repositories/request-repository.ts#L122-L146

packages/core/src/indexing/sqlite/chronological-order.ts
```diff
@@ -0,0 +1,184 @@
...
+      .from(tableName)
+      .select('stream_id', 'last_anchored_at', 'created_at')
+      .orderBy(CHRONOLOGICAL_ORDER)
+      .limit(pagination.first + 1)
```
stbrody, line 122: comment for why the +1 is there please

packages/core/src/indexing/sqlite/chronological-order.ts
```diff
@@ -0,0 +1,184 @@
...
+            startCursor: Cursor.stringify(asChronologicalCursor(firstEntry)),
+          },
+        }
+      }
+      default:
+        throw new UnsupportedOrderingError(paginationKind)
+    }
+  }
+
+  private chronologicalForwardQuery(
```
stbrody, line 113: would be nice to have some comments spelling out exactly what "chronological" order represents (and talking about how nulls are handled)

packages/core/src/indexing/sqlite/chronological-order.ts
```diff
@@ -0,0 +1,184 @@
...
+    const tableName = asTableName(query.model)
+    const base = this.knexConnection
+      .from(tableName)
+      .select('stream_id', 'last_anchored_at', 'created_at')
+      .orderBy(CHRONOLOGICAL_ORDER)
+      .limit(pagination.first + 1)
+    if (pagination.after) {
+      const after = Cursor.parse(pagination.after)
+      if (after.last_anchored_at) {
+        return base.where('last_anchored_at', '<', after.last_anchored_at)
```
stbrody, line 126: I'm having a hard time making sense of the sorting logic here.  Will revisit after comments are added explaining which of ForwardQuery or BackwardQuery is supposed to get the most recent entries or the least recent entries

packages/core/src/indexing/sqlite/chronological-order.ts
```diff
@@ -0,0 +1,184 @@
...
+    return JSON.parse(uint8arrays.toString(uint8arrays.fromString(cursor, 'base64url')))
+  }
+
+  /**
+   * base64url-encode cursor from +input+ object.
+   * Return `undefined` if +input+ object is `undefined` or `null`.
+   */
+  static stringify(input: any): string | undefined {
+    if (input) {
+      return uint8arrays.toString(uint8arrays.fromString(JSON.stringify(input)), 'base64url')
```
stbrody, line 31: nice

packages/core/src/indexing/sqlite/chronological-order.ts
stbrody: Okay, reading more I think I mostly get it now.  I think ForwardQuery is supposed to get the most recent entries and return them in decreasing recency order, while BackwardQuery gets the oldest entries but the results get newer as you iterate.  I was a bit thrown by ForwardCursor using `after` for subsequent batches given that it's actually requesting older entries, but I guess that's because the Page API is general and doesn't assume the query is in chronological order, so it's "after" the previous batch, but in this case that corresponds to entries created "before" on the network.

Assuming I have that all right, my only remaining question is why `CHRONOLOGICAL_ORDER` says `last_anchored_at` should sort `nulls: 'last'`.  If you're trying to get the most recent entries, the unanchored entries should come first, so then shouldn't it be `nulls: 'first'` in `CHRONOLOGICAL_ORDER`?

I see your unit test CSV file has the order I expect, so I'm sure it actually works, but I'm just confused as to how it's working.

packages/core/src/indexing/sqlite/chronological-order.ts
stbrody: please also include clarification of which of `chronologicalForwardQuery` vs `chronologicalBackwardQuery` return the most recent vs least recent entries

stbrody (commented):
looks really good overall.  Some questions and requests for comments throughout

packages/core/src/indexing/sqlite/__tests__/sqlite-index-api.test.ts
```diff
@@ -126,3 +141,101 @@ describe('indexStream', () => {
...
+  let ALL_ENTRIES_IN_CHRONOLOGICAL_ORDER: Array<string>
+
+  beforeEach(async () => {
+    indexAPI = new SqliteIndexApi(dataSource, knexConnection, MODELS_TO_INDEX)
+    await indexAPI.init()
+    const rows = await readFixture(new URL('./chronological-order.fixture.csv', import.meta.url))
+    for (const row of rows) {
+      await indexAPI.indexStream(row)
+    }
+    ALL_ENTRIES_IN_CHRONOLOGICAL_ORDER = rows.map((r) => r.streamID.toString())
```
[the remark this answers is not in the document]
ukstv: From the PR description:

> 3. Testing. There is a CSV file with entries in chronological order. We test the query results against the file.

Simple sort is not enough.

packages/core/src/indexing/sqlite/chronological-order.ts
ukstv: The link shows that you could do a query if you create an entity class per table at development time. It is not our case here. TypeORM can not really create a query if you do not have a corresponding entity class. That's the reason to introduce here a dedicated query builder.

I am eagerly waiting for @pawartur and @willex spike results regarding this issue for the dedicated migration tool. If they can make TypeORM handle dynamic set of tables in dev-friendly way, knex is to be ditched. If not, this thing is here to stay.

@stephhuynh18 might have some additional input really soon about using TypeORM in CAS. As surprising, as it is, there is a chance, we might have to replace TypeORM with knex.

packages/core/src/indexing/sqlite/__tests__/sqlite-index-api.test.ts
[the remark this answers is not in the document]
ukstv: Added a comment about that in the test code.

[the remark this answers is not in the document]
ukstv: https://github.com/ceramicnetwork/js-ceramic/pull/2184/commits/079df988e1785dd7aa19db52be16bf6825410348#diff-bb93f3297376873d46cba28caa3b8b705d717661b3ef77f16e839d3fa4f1f247R180

packages/core/src/indexing/sqlite/chronological-order.ts
```diff
@@ -0,0 +1,184 @@
...
+ */
+function asChronologicalCursor(
+  input: { created_at: number; last_anchored_at: number } | undefined
+) {
+  if (!input) return undefined
+  return { created_at: input.created_at, last_anchored_at: input.last_anchored_at }
+}
+
+const CHRONOLOGICAL_ORDER = [
+  { column: 'last_anchored_at', order: 'DESC', nulls: 'last' }, // (`last_anchored_at` is null) DESC
```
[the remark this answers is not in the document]
ukstv: Changed to `// (last_anchored_at is null) DESC, sorts nulls first`

packages/core/src/indexing/sqlite/chronological-order.ts
ukstv: `// To know if we have more entries to query`.

packages/core/src/indexing/sqlite/chronological-order.ts
```diff
@@ -0,0 +1,184 @@
...
+  ForwardPaginationQuery,
+  PaginationKind,
+  parsePagination,
+} from '../parse-pagination.js'
+import { UnsupportedOrderingError } from '../ordering.js'
+import { asTableName } from '../as-table-name.util.js'
+
+type Selected = { stream_id: string; last_anchored_at: number; created_at: number }
+
+abstract class Cursor {
```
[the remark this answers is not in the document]
ukstv:
```
/**
 * Contains functions to handle GraphQL cursors
 * as per [GraphQL Cursor Connections Spec](https://relay.dev/graphql/connections.htm).
 */
```

packages/core/src/indexing/sqlite/chronological-order.ts
ukstv: Added the comments there.

packages/core/src/indexing/sqlite/chronological-order.ts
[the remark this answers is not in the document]
stbrody: yep we did, that's why making it clear in the comments that these *aren't* database cursors is valuable.  Future readers of this code might not have been here for our discussions about database cursors.

Added comment looks good.

packages/core/src/indexing/sqlite/chronological-order.ts
stbrody: Ah got it, that makes sense.

packages/core/src/indexing/sqlite/chronological-order.ts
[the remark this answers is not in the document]
ukstv: > Future readers of this code might not have been here for our discussions about database cursors

Indeed.

packages/core/src/indexing/sqlite/chronological-order.ts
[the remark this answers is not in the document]
stbrody: woah yeah that was super not clear to me, and I didn't really understand what that comment was trying to convey.

Why does setting `nulls: 'last'` mean the nulls come first?  That seems so backwards.  Is it because the order is `DESC` and that inverts the meaning of `nuls: 'last'` for some reason?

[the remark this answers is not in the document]
ukstv: Hope, `sorts nulls first` makes it clearer.

This `nulls` thing is quite peculiar. Vanilla `{column: 'foo', order: 'DESC'}` translates to an obvious `ORDER BY foo DESC`. If you add `nulls: first`, the corresponding SQL is `ORDER BY (foo is not null) DESC`, and `nulls: last` turn it into `ORDER BY (foo is null) DESC`. A database interprets `foo is null` or `foo is not null` as a boolean condition. So, how a database sorts booleans?

SQLite stores boolean as an integer. False = 0, True = 1. Sorting booleans DESC results in [True, False] order.

This means rows with `(last_anchored_at is null) DESC` are sorted in [last_anchored_at is null, last_anchored_at is not null] order, which means "nulls first".

Just in case, this behaviour is specific to SQLite, because of the way it treats booleans. Postgres, and I suspect most other RDBMs, behave properly.

packages/core/src/indexing/sqlite/chronological-order.ts
```diff
@@ -0,0 +1,193 @@
...
+ */
+function asChronologicalCursor(
+  input: { created_at: number; last_anchored_at: number } | undefined
+) {
+  if (!input) return undefined
+  return { created_at: input.created_at, last_anchored_at: input.last_anchored_at }
+}
+
+const CHRONOLOGICAL_ORDER = [
+  { column: 'last_anchored_at', order: 'DESC', nulls: 'last' }, // (`last_anchored_at` is null) DESC, sorts nulls first on **SQLite only**
```
stbrody, line 55:
```suggestion
  { column: 'last_anchored_at', order: 'DESC', nulls: 'last' }, // This translates to the SQL "(`last_anchored_at` is null) DESC", which results in sorting nulls first.  **SQLite only**
```

PR #2189, opened 2022-05-30
Throw during model stream validation, if any of the objects within mo…

pawartur:
…del's schema doesn't have additional properties disabled

## Description

We shouldn't allow models to use schemas that allow additional properties.  The Model streamtype should error if a user tries to create a schema that doesn't explicitly set additionalProperties:false

## How Has This Been Tested?

- [x] Built and ran ceramic daemon locally
- [x] Added additional test cased to the suite testing `packages/stream-model-handler/schema-utils/SchemaValidation`

## PR checklist

### DIFF
--- a/packages/stream-model-handler/src/__tests__/schema-utils.test.ts
+++ b/packages/stream-model-handler/src/__tests__/schema-utils.test.ts
@@ -16,12 +16,89 @@ describe('SchemaValidation', () => {
           type: 'string',
           maxLength: 80,
         },
+        objectPropName: {
+          $ref: '#/$defs/EmbeddedObject'
+        },
       },
-      additionallyProperties: false,
+      $defs: {
+        EmbeddedObject : {
+          type: 'object',
+          properties: {
+            stringPropName: {
+              type: 'string',
+              maxLength: 80
+            }
+          },
+          additionalProperties: false,
+        }
+      },
+      additionalProperties: false,
       required: ['stringPropName'],
     })).resolves.not.toThrow()
   })
-  it('returns false for an incorrect 2020-12 schema', async () => {
+
+  it('throws for correct 2020-12 schema with `additionalProperties === true` enabled on top-level', async () => {
+    expect(schemaValidator.validateSchema({
+      $schema: "https://json-schema.org/draft/2020-12/schema",
+      type: 'object',
+      props: {
+        stringPropName: {
+          type: 'string',
+          maxLength: 80,
+        },
+      },
+      additionalProperties: true,
+      required: ['stringPropName'],
+    }))
+    .rejects
+    .toThrow("All objects in schema need to have additional properties disabled")
+  })
+
+  it('throws for correct 2020-12 schema with `additionalProperties === <allowed_property_type>` enabled on top-level', async () => {
+    expect(schemaValidator.validateSchema({
+      $schema: "https://json-schema.org/draft/2020-12/schema",
+      type: 'object',
+      props: {
+        stringPropName: {
+          type: 'string',
+          maxLength: 80,
+        },
+      },
+      additionalProperties: {type: 'string'},
+      required: ['stringPropName'],
+    }))
+    .rejects
+    .toThrow("All objects in schema need to have additional properties disabled")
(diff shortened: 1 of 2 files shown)


packages/stream-model-handler/src/schema-utils.ts
```diff
@@ -1,6 +1,33 @@
 
 import ajv, { AnySchema } from 'ajv/dist/2020'
 
+type JSONSchemaObjectType = {
+  type: 'object',
+  additionalProperties?: boolean
+}
+
+function traverseObjectRecursivelyAndApplyFunction(o, fn) {
```
ukstv, line 9: types for the input parameters, please

packages/stream-model-handler/src/schema-utils.ts
```diff
@@ -13,12 +40,13 @@ export class SchemaValidation {
   })
 
   public async validateSchema(
```
oed, line 42: Where is this method called? If it's used in TileDocuments then this might break old schemas?

packages/stream-model-handler/src/__tests__/schema-utils.test.ts
```diff
@@ -16,12 +16,89 @@ describe('SchemaValidation', () => {
...
-      additionallyProperties: false,
+      additionalProperties: false,
       required: ['stringPropName'],
     })).resolves.not.toThrow()
   })
-  it('returns false for an incorrect 2020-12 schema', async () => {
+
+  it('throws for correct 2020-12 schema with `additionalProperties === true` enabled on top-level', async () => {
+    expect(schemaValidator.validateSchema({
+      $schema: "https://json-schema.org/draft/2020-12/schema",
```
stbrody, line 42: it might cleanup this test file and make it a bit easier to read if you defined a valid schema object in a constant at the top of the file, then for the tests that you want to fail you can copy the valid object and then make a targeted change to the copy to make it invalid in the way you want.

packages/stream-model-handler/src/schema-utils.ts
```diff
@@ -1,5 +1,32 @@
...
+
+type JSONSchemaObjectType = {
+  type: 'object',
+  additionalProperties?: boolean
+}
+
+function traverseSchemaRecursivelyAndApplyFunction(o: SchemaObject, fn: (o: object) => void) {
+  fn(o)
+  Object.getOwnPropertyNames(o).forEach( (name) => {
+    fn.apply(this,[o[name]]);
```
stbrody, line 12: isn't this redundant given the recursive call will cause the fn call on line 10 to run on the subobj anyway?

packages/stream-model-handler/src/schema-utils.ts
stbrody: this file is specific to the `stream-model-handler` package.  TileDocument has its own schema validation code that is separate

stbrody (commented):
lgtm with a few small suggestions

packages/stream-model-handler/src/schema-utils.ts
oed: ok, ty!

packages/stream-model-handler/src/schema-utils.ts
pawartur: Awesome catch, thanks @stbrody !

packages/stream-model-handler/src/schema-utils.ts
```diff
@@ -1,5 +1,31 @@
 
-import ajv, { AnySchema } from 'ajv/dist/2020.js'
+import ajv, { SchemaObject } from 'ajv/dist/2020.js'
+
+type JSONSchemaObjectType = {
+  type: 'object',
+  additionalProperties?: boolean
+}
+
+function applyFuntionToSchemaAndItsObjectPropsRecursively(schema: SchemaObject, fn: (schemaProperty: object) => void) {
```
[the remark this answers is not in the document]
ukstv: Also, `funtion`.

stbrody (approved):
I'm happy when Sergey is

PR #2192, opened 2022-05-30
chore(stream-model-instance, stream-model-instance-handler):  metadata removal; family, tags, schema, deterministic

willex:
## Description

code/metadata removal for MID preparation. Combined NET-1443 & NET-1441 into this one PR. One question around `SignatureUtils.verifyCommitSignature()`.

## How Has This Been Tested?

- [x] Updated local tests & ran full test suite

## PR checklist

## References:


packages/stream-model-instance-handler/src/model-instance-document-handler.ts
```diff
@@ -84,12 +84,12 @@ export class ModelInstanceDocumentHandler implements StreamHandler<ModelInstance
...
     if (isSigned) {
       const streamId = await StreamID.fromGenesis('MID', commitData.commit)
-      const { controllers, family } = payload.header
+      const { controllers } = payload.header
       await SignatureUtils.verifyCommitSignature(
         commitData,
         context.did,
         controllers[0],
-        family,
+        null,
```
willex, line 92: @stbrody I've added a null value right now to side-step the `family` requirement for `SignatureUtils.verifyCommitSignature()` but would like to follow up on it if this needs a breakout/duplication of this class for MID.

packages/stream-model-instance-handler/src/__tests__/model-instance-document-handler.test.ts
```diff
@@ -303,7 +303,7 @@ describe('ModelInstanceDocumentHandler', () => {
     expect(modelInstanceDocumentHandler.name).toEqual('MID')
   })
 
-  it('makes genesis commits correctly', async () => {
+  /*it('makes genesis commits correctly', async () => {
```
ukstv, line 306: Please, add a comment on top like `TODO NET-XXXX Description`, where `XXXX` is a reference to a corresponding Linear ticket, and `Description` is a short rationale for commenting this out.

ukstv (approved):

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
oed: We should change this method signature to accept `model` as `StreamID` instead of family. We should also make this param optional.

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
```diff
@@ -109,10 +109,6 @@ export class ModelInstanceDocumentHandler implements StreamHandler<ModelInstance
       log: [{ cid: commitData.cid, type: CommitType.GENESIS }],
     }
 
-    if (state.metadata.schema) {
-      await this._schemaValidator.validateSchema(context.api, state.content, state.metadata.schema)
-    }
```
oed, line 114: Maybe just comment out this code so that we can put it back once we add the model schema validation?

packages/stream-model-instance/src/model-instance-document.ts
```diff
@@ -32,30 +32,6 @@ export interface ModelInstanceDocumentMetadataArgs {
...
-   * If true, then two calls to ModelInstanceDocument.create() with the same content and the same metadata
-   * will only create a single document with the same StreamID. If false, then otherwise
-   * identical documents will generate unique StreamIDs and be able to be updated independently.
-   * @deprecated use deterministic function instead
-   */
-  deterministic?: boolean
-
   /**
    * If true, all changes to the 'controllers' array are disallowed.  This guarantees that the
    * Stream will always have the same controller. Especially useful for Streams controlled by
```
oed, line 37: The `forbidControllerChange` should be removed imo. It should never be possible to change controller of a MID.

packages/stream-model-instance/src/model-instance-document.ts
```diff
@@ -32,30 +32,6 @@ export interface ModelInstanceDocumentMetadataArgs {
    */
   controllers?: Array<string>
```
oed, line 34:
```suggestion
  controllers?: Array<string>

  model: StreamID
```

A MID always need a model.

packages/stream-model-instance/src/model-instance-document.ts
```diff
@@ -85,37 +61,20 @@ function headerFromMetadata(
...
     controllers: metadata?.controllers,
-    family: metadata?.family,
-    schema: metadata?.schema?.toString(),
-    tags: metadata?.tags,
   }
 
   // Handle properties that can only be set on the genesis commit.
   if (genesis) {
-    if (!metadata?.deterministic) {
-      header.unique = uint8arrays.toString(randomBytes(12), 'base64')
```
oed, line 106: I think we still need to include the `unique` random bytes by default, e.g. there should be no way to disable it.

packages/stream-model-instance/src/model-instance-document.ts
```diff
@@ -85,37 +61,20 @@ function headerFromMetadata(
...
-  if (typeof metadata?.schema === 'string') {
-    try {
-      CommitID.fromString(metadata.schema)
-    } catch {
-      throw new Error('Schema must be a CommitID')
-    }
-  }
 
   const header: CommitHeader = {
     controllers: metadata?.controllers,
```
oed, line 66:
```suggestion
  const header: CommitHeader = {
    controllers: metadata?.controllers,
    mode: metadata.model,
    unique: uint8arrays.toString(randomBytes(12), 'base64')
```
Actually, let's just do this.

Here I'm assuming that it's only possible to set metadata in the genesis record. Otherwise we need to put the `unique` property inside of the `genesis` condition.

oed (commented):
Few comments. I realize that some of them might be out of scope for this issue though.

packages/stream-model-instance/src/model-instance-document.ts
stbrody: I think the goal of this PR was just to remove no-longer-needed functionality.  Adding in the model to the APIs will happen in https://linear.app/3boxlabs/issue/NET-1446/bring-modelinstancedocument-public-api-into-alignment-with-the-spec

packages/stream-model-instance/src/model-instance-document.ts
willex: Gotcha, yeah missed that. Added back in.

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
willex: Done.

packages/stream-model-instance/src/model-instance-document.ts
willex: done.

packages/stream-model-instance-handler/src/__tests__/model-instance-document-handler.test.ts
willex: Fixed an issue on my end and tests work as expected (added back in).

packages/stream-model-instance/src/model-instance-document.ts
willex: Checked and the `unique` property has to be set only for the genesis record, so moved it into the `genesis` condition block where it was originally.

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
willex: Do we want that as part of this PR (removal) or with the follow up one as @stbrody you mentioned in a different comment as well.

packages/stream-model-instance-handler/src/__tests__/model-instance-document-handler.test.ts
```diff
@@ -329,13 +329,7 @@ describe('ModelInstanceDocumentHandler', () => {
...
   })
 
-  it('throws error for deterministic genesis commit with data', async () => {
-    await expect(
-      ModelInstanceDocument.makeGenesis(context.api, COMMITS.genesis.data, { deterministic: true })
-    ).rejects.toThrow(/Initial content must be null/)
-  })
-
-  it('Takes controller from authenticated DID if controller not specified', async () => {
+  it('Does not sign commit if no content', async () => {
```
stbrody, line 332: now two tests with the same name?  See the next test on line 337.  Is this supposed to be `Takes controller from authenticated DID if controller not specified`?

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
stbrody: fine to leave for a follow-up IMO

stbrody (approved):
lgtm

PR #2194, opened 2022-05-31
feat: warn if indexing is not configured

ukstv:
If using an old ceramic daemon config with no `indexing` section, display a warning.


stbrody (approved):
🚀

PR #2195, opened 2022-05-31
implement simple collection metrics

gvelez17:
## Description

In this PR we implement two simple metrics on js-ceramic, in order that we can start gathering data in clay testnet and create end-to-end processing and display of metrics collected.  

## How Has This Been Tested?

Wrote a unit test for the metrics

Tested locally on dev-unstable and verified that metrics appear at the new endpoint

## PR checklist

## References:

See https://www.notion.so/threebox/Metrics-Collection-b1afc5e5cb6f4952a8b794932a397571


gvelez17:
This PR optionally uses variables defined in https://github.com/3box/ceramic-infra/pull/565 but defaults to disabling the exporter if the METRICS_ENABLED variable is not defined, and to the default port if the METRICS_EXPORTER_PORT is not defined

packages/cli/src/daemon/instrument-requests.ts
```diff
@@ -0,0 +1,13 @@
+import {Count, REQUEST_METRIC} from '@ceramicnetwork/common'
+
+export function instrumentRequest(req, res, next) {
```
oed, line 3: Is this method used anywhere?

oed (commented):
I'd like @ukstv to have a look at this.

packages/common/src/metrics-setup.ts
```diff
@@ -0,0 +1,57 @@
...
+export const PUBLISHED_METRIC = 'published'
+export const PINNED_METRIC = 'pinned'
+
+const VALID_METRIC_NAMES = [REQUEST_METRIC, RESPONSE_METRIC, RECEIVED_METRIC, PUBLISHED_METRIC, PINNED_METRIC]
+
+// could have subclasses or specific functions with set params, but we want to
+// easily and quickly change what is recorded, there are no code dependencies on it
+
+export function Count(name:string, value:number, params?:any) {
+  if (! VALID_METRIC_NAMES.includes(name)) {
```
v-stickykeys, line 46: Small suggestion here: make the metric names a map instead of an array

```
METRIC_NAMES = {
  STREAM_PINNED: "stream_pinned"
  ...
```

then you can do

```
if !METRIC_NAMES[name]
```

and when importing the names in another file it would be easy to see all names and use any

PR #2196, opened 2022-05-31
feat(stream-model-instance, stream-model-instance-handler): ModelInstanceDocument API

stbrody:


packages/stream-model-instance/src/model-instance-document.ts
```diff
@@ -116,48 +78,18 @@ export class ModelInstanceDocument<T = Record<string, any>> extends Stream {
...
-   * @param metadata - Genesis metadata
+   * @param metadata - Genesis metadata, including the model that this document belongs to
    * @param opts - Additional options
    */
   static async create<T>(
     ceramic: CeramicApi,
-    content: T | null | undefined,
-    metadata?: ModelInstanceDocumentMetadataArgs,
+    content: T | null,
+    metadata: ModelInstanceDocumentMetadata,
```
stbrody, line 87: metadata is now required, and must have a `model` field within it.  I debated pulling model up out of `metadata` into a top-level required positional arg, but then we'd either need an optional `controller` field or an optional `metadata` field with only a single `controller` property.  Keeping all the metadata in a `metadata` field makes the API a bit more extensible so that's what I went with.  But I could easily be convinced to pull model out to the top level

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
```diff
@@ -213,33 +182,32 @@ export class ModelInstanceDocumentHandler implements StreamHandler<ModelInstance
...
       cid: commitData.cid,
       type: CommitType.ANCHOR,
       timestamp: proof.blockTimestamp,
     })
     let content = state.content
-    let metadata = state.metadata
 
     if (state.next?.content) {
       content = state.next.content
       delete state.next.content
```
oed, line 202:
```suggestion
```
Looks like we delete `state.next` anyway below.

packages/stream-model-instance/src/model-instance-document.ts
```diff
@@ -279,38 +196,48 @@ export class ModelInstanceDocument<T = Record<string, any>> extends Stream {
...
-    if (content == null) {
-      const result = { header }
-      // Check if we can encode it in cbor. Should throw an error when invalid payload.
-      dagCbor.encode(result)
-      // No signature needed if no genesis content
-      return result
+    // TODO(NET-1464): enable GenesisHeader to receive 'controller' field directly
+    const header: GenesisHeader = {
+      controllers: [metadata.controller],
+      unique: uint8arrays.toString(randomBytes(12), 'base64'),
```
oed, line 222: Maybe we can make this raw bytes as well to save space?
```suggestion
      unique: randomBytes(12),
```
Just a  thought, not super important.

oed (approved):
lgtm

PR #2200, opened 2022-06-01
feat: add dummy implementation of IndexClientApi to core and http-client

ukstv:
- Move shared types to common
- Add dummy impl to core and http-client


packages/common/src/index-client-api.ts
```diff
@@ -0,0 +1,73 @@
+import type { StreamID } from '@ceramicnetwork/streamid'
+import type { StreamState } from './stream.js'
+
+/**
+ * Allowed ordering for an indexing query.
+ */
+export enum Ordering {
```
stbrody, line 7: I don't think this should be exposed in the client API.  As we discussed earlier, we want to always use chronological ordering for now, and avoid exposing any ordering options until we've settled on user-facing names for them.

packages/core/src/ceramic.ts
```diff
@@ -175,11 +177,12 @@ export class Ceramic implements CeramicApi {
   public readonly dispatcher: Dispatcher
   public readonly loggerProvider: LoggerProvider
   public readonly pin: PinApi
+  public readonly index: IndexClientApi
```
stbrody, line 180: `IndexClientApi` sounds to me like what `RemoteIndexApi` actually is.  Can we rename `IndexClientApi` to `IndexApi` and `IndexApi` to `ServerIndexApi` or `LocalIndexApi`?

packages/core/src/ceramic.ts
```diff
@@ -175,11 +177,12 @@ export class Ceramic implements CeramicApi {
...
   readonly repository: Repository
 
   readonly _streamHandlers: HandlersMap
   private readonly _anchorValidator: AnchorValidator
-  private readonly _indexing: DatabaseIndexAPI | undefined
+  private readonly _indexing: DatabaseIndexApi | undefined
```
stbrody, line 185: Can we hide the `DatabaseIndexApi` within the `IndexApi`?  Does the `Ceramic` class need both or can one abstract away the other?

stbrody (changes requested):

packages/common/src/index-client-api.ts
ukstv: Again, let me remind you about tradeoffs here. It is easier to remove the option than to add it at a later time.

packages/core/src/ceramic.ts
ukstv: good idea.

packages/core/src/ceramic.ts
stbrody: sorry, can you explain the concern a bit more?  Couldn't `LocalIndexApi` have an `init()` method which in turn calls `DatabaseIndexApi::init()`, and Ceramic's `_init()` can just call the `LocalIndexApi::init()`?

ukstv: That was the first version of the initialisation logic haha :) Maybe it is not a big deal to make `init` available after all.

ukstv: Okay, I feel it is better now.

stbrody (approved):
:shipit:

PR #2201, opened 2022-06-01
feat: Make IndexAPI call the database

ukstv:
Potential controversies:
- We might have indexing database backend not configured. It makes the `queryIndex` method return an empty page. We also log a warning for every query.
- Use `Repository::streamState` to load StreamState for StreamID. We assume that a state store always contains StreamState for an indexed stream.

Base branch is not `development` to reduce mental overhead for reviewing.

### DIFF
--- a/packages/core/src/ceramic.ts
+++ b/packages/core/src/ceramic.ts
@@ -203,7 +203,6 @@ export class Ceramic implements CeramicApi {
     this._gateway = params.gateway
     this._networkOptions = params.networkOptions
     this._loadOptsOverride = params.loadOptsOverride
-    this._index = new LocalIndexApi(modules.indexing)
 
     this.context = {
       api: this,
@@ -234,6 +233,7 @@ export class Ceramic implements CeramicApi {
       anchorService: modules.anchorService,
       conflictResolution: conflictResolution,
     })
+    this._index = new LocalIndexApi(modules.indexing, this.repository, this._logger)
   }
 
   get index(): IndexApi {
--- a/packages/core/src/indexing/local-index-api.ts
+++ b/packages/core/src/indexing/local-index-api.ts
@@ -1,15 +1,51 @@
-import type { BaseQuery, IndexApi, Page, Pagination, StreamState } from '@ceramicnetwork/common'
+import type {
+  BaseQuery,
+  IndexApi,
+  Page,
+  Pagination,
+  StreamState,
+  DiagnosticsLogger,
+} from '@ceramicnetwork/common'
+import { SyncOptions } from '@ceramicnetwork/common'
 import type { DatabaseIndexApi } from './database-index-api.js'
-import { NotImplementedError } from './not-implemented-error.js'
+import type { Repository } from '../state-management/repository.js'
 
+/**
+ * API to query an index.
+ */
 export class LocalIndexApi implements IndexApi {
-  constructor(private readonly databaseIndexApi: DatabaseIndexApi | undefined) {}
+  constructor(
+    private readonly databaseIndexApi: DatabaseIndexApi | undefined,
+    private readonly repository: Repository,
+    private readonly logger: DiagnosticsLogger
+  ) {}
 
-  queryIndex(query: BaseQuery & Pagination): Promise<Page<StreamState>> {
-    throw new NotImplementedError('IndexApi::queryIndex')
-  }
-
-  async init(): Promise<void> {
-    await this.databaseIndexApi.init()
+  /**
+   * Query the index. Ask an indexing database for a list of StreamIDs,
+   * and convert them to corresponding StreamState instances via `Repository::load`.
+   */
+  async queryIndex(query: BaseQuery & Pagination): Promise<Page<StreamState>> {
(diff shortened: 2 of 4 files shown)


packages/core/src/ceramic.ts
```diff
@@ -205,7 +205,6 @@ export class Ceramic implements CeramicApi {
     this._networkOptions = params.networkOptions
     this._loadOptsOverride = params.loadOptsOverride
     this._indexing = modules.indexing
```
stbrody, line 207: Related to my comment on the previous PR: now that we pass the indexing into the IndexApi, do we need to store a reference here in the Ceramic class, or can we just keep the reference to `this.index` and have it be responsible for managing the `DatabaseIndexApi`?

packages/core/src/indexing/index-api.ts
```diff
@@ -4,14 +4,47 @@ import type {
...
+   * Query the index. Ask an indexing database for a list of StreamIDs,
+   * and convert them to corresponding StreamState instances via `Repository::streamState`.
+   *
+   * We assume that a state store always contains StreamState for an indexed stream.
+   */
+  async queryIndex(query: BaseQuery & Pagination): Promise<Page<StreamState>> {
+    if (this.databaseIndexApi) {
+      const page = await this.databaseIndexApi.page(query)
+      const streamStates = await Promise.all(
+        page.entries.map((streamId) => this.repository.streamState(streamId))
```
stbrody, line 33: `Repository.streamState()` doesn't use the loading queue or add the state to the cache.  Should we be using `Repository.get()` instead?  https://github.com/ceramicnetwork/js-ceramic/blob/e8947c9ae2d7b050e34fdbd69cd33e0c80173050/packages/core/src/state-management/repository.ts#L268

packages/core/src/indexing/index-api.ts
```diff
@@ -4,14 +4,47 @@ import type {
...
+  /**
+   * Query the index. Ask an indexing database for a list of StreamIDs,
+   * and convert them to corresponding StreamState instances via `Repository::streamState`.
+   *
+   * We assume that a state store always contains StreamState for an indexed stream.
+   */
+  async queryIndex(query: BaseQuery & Pagination): Promise<Page<StreamState>> {
+    if (this.databaseIndexApi) {
+      const page = await this.databaseIndexApi.page(query)
+      const streamStates = await Promise.all(
```
stbrody, line 32: I'm a little worried about us hammering S3 with hundreds of simultaneous requests.  I think we might want to break this down into smaller batches and/or rate limit how fast we hammer the state store with these requests

packages/core/src/indexing/index-api.ts
```diff
@@ -4,14 +4,47 @@ import type {
...
 
-  queryIndex(query: BaseQuery & Pagination): Promise<Page<StreamState>> {
-    throw new NotImplementedError('IndexApi::queryIndex')
+  /**
+   * Query the index. Ask an indexing database for a list of StreamIDs,
+   * and convert them to corresponding StreamState instances via `Repository::streamState`.
+   *
+   * We assume that a state store always contains StreamState for an indexed stream.
+   */
+  async queryIndex(query: BaseQuery & Pagination): Promise<Page<StreamState>> {
```
stbrody, line 29: I'm also worried that we don't have any protection against a client sending a query requesting a huge number of results in a single batch.  I think we should have some server-enforced maximum batch size. I'll file a ticket for this.

packages/core/src/indexing/index-api.ts
ukstv: I believe it is a responsibility of S3StateStore then, and not this thing. Created an issue for that.

packages/core/src/indexing/local-index-api.ts
```diff
@@ -4,14 +4,47 @@ import type {
...
+   * Query the index. Ask an indexing database for a list of StreamIDs,
+   * and convert them to corresponding StreamState instances via `Repository::streamState`.
+   *
+   * We assume that a state store always contains StreamState for an indexed stream.
+   */
+  async queryIndex(query: BaseQuery & Pagination): Promise<Page<StreamState>> {
+    if (this.databaseIndexApi) {
+      const page = await this.databaseIndexApi.page(query)
+      const streamStates = await Promise.all(
+        page.entries.map((streamId) => this.repository.streamState(streamId))
```
stbrody, line 33:
```suggestion
        // For database queries we bypass the stream cache and repository loading queue
        page.entries.map((streamId) => this.repository.streamState(streamId))
```

packages/core/src/indexing/index-api.ts
stbrody: I more buy the argument that we should skip the loading queue since we know we won't be generating pubsub queries so the need to rate limit is less.  I'm a little less convinced that we should skip caching streams that are loaded via queries.  I do see the concern though that one large query could blow away your entire cache in one fell swoop.  So I'm okay with going with this for now and we can revisit in the future if this causes issues.

stbrody (approved):
lgtm

PR #2202, opened 2022-06-01
fix: Filter by account

ukstv:
SQLite, chronological order. I forgot to filter by account. This fixes the mistake.


stbrody (approved):
lgtm

PR #2205, opened 2022-06-03
feat: gitgnore generated version.ts file

ukstv:
Ignore auto-generated version.ts file.

As a meta-rule, we should avoid checking generated files in.


stbrody (approved):
🚀

PR #2211, opened 2022-06-06
feat: Implement IndexAPI on http client side

ukstv:
Core part:
- Make RemoteIndexApi call HTTP endpoint

Housekeeping:
- Use native JS URL extensively. We do not need `query-string` package anymore.

### DIFF
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -453,8 +453,7 @@ export class CeramicDaemon {
 
   async getCollection(req: Request, res: Response): Promise<void> {
     const httpQuery = parseQueryObject(req.query)
-    const defaultPagination = { first: 100 }
-    const query = collectionQuery(httpQuery, defaultPagination)
+    const query = collectionQuery(httpQuery)
     const indexResponse = await this.ceramic.index.queryIndex(query)
     res.json({
       entries: indexResponse.entries.map(StreamUtils.serializeState),
--- a/packages/http-client/src/__tests__/remote-index-api.test.ts
+++ b/packages/http-client/src/__tests__/remote-index-api.test.ts
@@ -1,0 +1,51 @@
+import { jest } from '@jest/globals'
+import { RemoteIndexApi } from '../remote-index-api.js'
+import { CommitType, fetchJson, StreamState, TestUtils } from '@ceramicnetwork/common'
+import { StreamID } from '@ceramicnetwork/streamid'
+
+const FAUX_ENDPOINT = new URL('https://example.com')
+const MODEL = new StreamID(1, TestUtils.randomCID())
+
+const EMPTY_RESPONSE = {
+  entries: [],
+  pageInfo: {},
+}
+const FAUX_STREAM_STATE = {
+  type: 0,
+  log: [
+    {
+      type: CommitType.GENESIS,
+      cid: TestUtils.randomCID(),
+    },
+  ],
+} as unknown as StreamState
+
+test('model in query', async () => {
+  const fauxFetch = jest.fn(async () => EMPTY_RESPONSE) as typeof fetchJson
+  const indexApi = new RemoteIndexApi(FAUX_ENDPOINT, fauxFetch)
+  const result = await indexApi.queryIndex({ model: MODEL, first: 5 })
+  expect(result).toEqual(EMPTY_RESPONSE)
+  expect(fauxFetch).toBeCalledWith(new URL(`https://example.com/collection?model=${MODEL}&first=5`))
+})
+
+test('model, account in query', async () => {
+  const fauxFetch = jest.fn(async () => EMPTY_RESPONSE) as typeof fetchJson
+  const indexApi = new RemoteIndexApi(FAUX_ENDPOINT, fauxFetch)
+  const result = await indexApi.queryIndex({ model: MODEL, account: 'did:key:foo', first: 5 })
+  expect(result).toEqual(EMPTY_RESPONSE)
+  expect(fauxFetch).toBeCalledWith(
+    new URL(
+      `https://example.com/collection?model=${MODEL}&account=${encodeURIComponent(
+        'did:key:foo'
+      )}&first=5`
+    )
+  )
+})
+
+test('serialize stream stte', async () => {
+  const response = { ...EMPTY_RESPONSE, entries: [FAUX_STREAM_STATE] }
+  const fauxFetch = jest.fn(async () => response) as typeof fetchJson
+  const indexApi = new RemoteIndexApi(FAUX_ENDPOINT, fauxFetch)
+  const result = await indexApi.queryIndex({ model: MODEL, account: 'did:key:foo', first: 5 })
+  expect(result.entries[0]).toEqual(FAUX_STREAM_STATE)
+})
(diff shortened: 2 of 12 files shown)


packages/http-client/src/__tests__/remote-index-api.test.ts
```diff
@@ -0,0 +1,51 @@
...
+  expect(fauxFetch).toBeCalledWith(
+    new URL(
+      `https://example.com/collection?model=${MODEL}&account=${encodeURIComponent(
+        'did:key:foo'
+      )}&first=5`
+    )
+  )
+})
+
+test('serialize stream stte', async () => {
```
stbrody, line 45:
```suggestion
test('serialize stream state', async () => {
```

packages/http-client/src/remote-index-api.ts
```diff
@@ -1,7 +1,29 @@
...
 
+/**
+ * IndexAPI implementation on top of HTTP endpoint.
+ */
 export class RemoteIndexApi implements IndexApi {
-  queryIndex(query: BaseQuery & Pagination): Promise<Page<StreamState>> {
-    throw new Error(`NotImplemented: RemoteIndexApi::queryIndex`)
+  private readonly _collectionURL: URL
+
+  constructor(apiUrl: URL, private readonly fetchJsonFn: typeof fetchJson = fetchJson) {
```
stbrody, line 10: why have `fetchJsonFn` as an argument?  It's easy enough to mock out the function in unit tests without this, and seems to just complicate the API.

packages/http-client/src/remote-index-api.ts
```diff
@@ -1,7 +1,29 @@
...
+    this._collectionURL = new URL('./collection', apiUrl)
+  }
+
+  /**
+   * Issue a query to `/collection` endpoint.
+   */
+  async queryIndex(query: BaseQuery & Pagination): Promise<Page<StreamState>> {
```
stbrody, line 17: I just realized this function (and therefore also the signature definition in `IndexApi` and the implementation in `LocalIndexApi`) shouldn't be returning `StreamState` objects but actually should be returning Stream instances of the appropriate streamtype, same as how `CeramicApi.load()` transforms the resulting StreamStates into Stream instances

stbrody: Unless you want to make the IndexApi private in `Ceramic` and `CeramicClient` and have a `query()` method on `CeramicApi` that wraps `IndexApi.queryIndex` and transforms the StreamStates to stream instances.  But whatever the end-user application dev consumes in their application code should be returning Stream instances of the proper StreamType.

stbrody (changes requested):
mostly looking good, but one small comment about removing test-specific code from the RemoteIndexApi constructor, and one bigger comment around making sure that the code that app devs consume give them Stream instances rather than raw StreamStates

packages/http-client/src/remote-index-api.ts
ukstv: It _was_ easy. Jest has very minimal support for mocking of ES modules. The tradeoff is this non-fancy argument, or quite a few `await import` clauses, along with mocking fetch output per test. I tried that. Mocking fetch here is ugly.

packages/http-client/src/remote-index-api.ts
ukstv: > should be returning Stream instances of the proper StreamType

1. I really thought we agreed on returning StreamState here.
2. Returning actual Streams would be a huge waste of compute resources.

packages/http-client/src/remote-index-api.ts
stbrody: Okay, what about if we take it out of the constructor, but still have it as a private member variable that just always gets assigned to fetchJson at construction?  Then the unit tests can still override the function member, but we haven't expanded the public API

ukstv: It's a less obvious mess, but I am ok with it.

packages/http-client/src/remote-index-api.ts
stbrody: @ukstv are you okay with going with what we discussed in discord?
`IndexApi.queryIndex` stays as-is, but `IndexApi` becomes private within the `CeramicApi` implementations.  Then we add a new `query` method to `CeramicApi` that is more user-friendly: returns an AsyncIterable, handles requesting subsequent batches internally, and returns stream instances one at a time

stbrody: decision from standup: this will basically stay as-is in this PR.  `IndexApi.queryIndex` is the primary way we expose querying functionality.  It is designed to be consumed by the GraphQL runtime tooling which will handle paging and translating StreamStates to Stream instances.

In the future we can add an API that wraps `IndexApi.queryIndex` into something more user-friendly, if there is demand for it

packages/http-client/src/remote-index-api.ts
stbrody: I think this is the only change now I'm waiting on to re-review and approve this PR

packages/cli/src/ceramic-daemon.ts
```diff
@@ -448,6 +444,17 @@ export class CeramicDaemon {
     })
   }
 
+  async getCollection(req: Request, res: Response): Promise<void> {
+    const httpQuery = parseQueryObject(req.query)
+    const defaultPagination = { first: 100 }
```
PaulLeCam, line 449: I'm wondering if we should even have this default, rather than simply making it an invalid request to no provide either `first` or `last`?

PaulLeCam (approved):
Looks great, thanks!
I'm just wondering about setting a default value for pagination, but we can discuss and eventually change this later so don't hesitate to keep it for now.

packages/http-client/src/remote-index-api.ts
ukstv: @stbrody Changed to private _fetchJson. Please, have a look!

packages/cli/src/ceramic-daemon.ts
ukstv: @PaulLeCam I can make no strong argument here. I basically borrowed this from TheGraph default behaviour. Removing it now in favour of an error. It is based on an incoming payload only, so is fine wrt security.

packages/http-client/src/remote-index-api.ts
```diff
@@ -1,7 +1,30 @@
 import type { BaseQuery, IndexApi, Page, Pagination, StreamState } from '@ceramicnetwork/common'
+import { StreamUtils, fetchJson } from '@ceramicnetwork/common'
 
+/**
+ * IndexAPI implementation on top of HTTP endpoint.
+ */
 export class RemoteIndexApi implements IndexApi {
-  queryIndex(query: BaseQuery & Pagination): Promise<Page<StreamState>> {
-    throw new Error(`NotImplemented: RemoteIndexApi::queryIndex`)
+  private readonly _fetchJson: typeof fetchJson = fetchJson
```
stbrody, line 8:
```suggestion
  // Stored as a member to make it easier to inject a mock in unit tests
  private readonly _fetchJson: typeof fetchJson = fetchJson
```

stbrody (approved):
LGTM!

PR #2212, opened 2022-06-06
feat(model, model-instance-document): Encode 'unique' as Uint8Array

stbrody:


packages/common/src/utils/stream-utils.ts
```diff
@@ -128,6 +128,9 @@ export class StreamUtils {
     if (state.next?.metadata?.model) {
       cloned.next.metadata.model = state.next.metadata.model.toString()
     }
+    if (state.metadata?.unique && state.metadata.unique instanceof Uint8Array) {
+      cloned.metadata.unique = Array.from(state.metadata.unique)
```
oed, line 132: Maybe use `uint8arrays` to encode it as base64 to be more compact?

oed: This is what we do with CACAO and `linkedBlock` when we serialize them

oed (approved):
lgtm, one suggestion

packages/common/src/utils/stream-utils.ts
stbrody: Okay done, but heads up that required me to change behavior based on whether the stream is a TileDocument or not, to avoid changing the type of 'unique' in TileDocument as well, which required me to duplicate the type id number for tile in StreamUtils since I can't import TileDocument from the `common` package.  New PR up if you want to take a look

oed (approved):
Ok, unfortunate but seems alright.

PR #2213, opened 2022-06-06
feat(model-handler): Enforce that Model streams cannot have extra fields in their content

stbrody:


pawartur (approved):
Looks great, thanks!

The only thought that came up when reading the changes was that maybe it would be useful to have a single interface for validating the whole Model Definition payload, including allowed keys and the schema.

But I think that this approach is consistent with how we validate other stream types, so it's great for now and maybe later we can figure out some improvements, if needed.

PR #2214, opened 2022-06-06
Replace cacao support for 'family' with 'model' instead

zachferland:
BREAKING for anyone that uses cacao with family already, but from issue looks like we decided that there is little to no impact there 

This removes cacao family support from tiles and adds cacao model support for model instance.


zachferland:
pushed this up to see failure in CI, any idea why logging here causes the snapshot to change and tests to fail? (even if try/catch)  cc @ukstv  @stbrody

stbrody:
Woah, weird.  Seems like it might have to do with the fact that StreamID's Memoize their toString() representation: https://github.com/ceramicnetwork/js-ceramic/blob/cfbb0dd890b08a8ce99959daa4a9421ad2e77a19/packages/streamid/src/stream-id.ts#L208

zachferland:
oh interesting! yeah does seem to be that

packages/canary-integration/src/__tests__/capability.test.ts
```diff
@@ -169,159 +191,160 @@ describe('CACAO Integration test', () => {
...
-        const didKeyWithCapability = await addCapToDid(
-          wallet,
-          didKey,
-          `ceramic://*?family=${family}`
-        )
+    test('can create stream with model resource', async () => {
+      const didKeyWithCapability = await addCapToDid(
+        wallet,
+        didKey,
+        `ceramic://*?model=${METADATA.model.toString()}`
```
stbrody, line 313: can you either use `FAKE_STREAM_ID` here or use `METADATA.model` below when making the stream?  It's confusing to see two different things in the test and have to realize that they are actually the same.

Ditto for the other tests

packages/stream-model-handler/src/model-handler.ts
```diff
@@ -90,13 +90,13 @@ export class ModelHandler implements StreamHandler<Model> {
     }
 
     const streamId = await StreamID.fromGenesis('model', commitData.commit)
-    // TODO(NET-1437): replace family with model
-    const { controllers, family } = payload.header
+    const { controllers, model } = payload.header
```
stbrody, line 93: don't you have to do `StreamID.fromBytes` like you do in model-instance-handler?

We should probably have some tests of capabilities with `Model` Streams

packages/common/src/utils/signature_utils.ts
```diff
@@ -23,10 +23,10 @@ export class SignatureUtils {
     commitData: CommitData,
     did: DID,
     controller: string,
-    family: string,
+    model: StreamID,
```
stbrody, line 26:
```suggestion
    model: StreamID | null,
```

stbrody: since TileDoc passes in null

stbrody (commented):
main comment is testing for Model streams

packages/common/src/utils/signature_utils.ts
ukstv: also, there is a `@param` JSDoc above, that has to be changed

ukstv (approved):
lgtm mod @stbrody 's comments

packages/canary-integration/src/__tests__/capability.test.ts
zachferland: yeah agree, added more consistent naming

packages/stream-model-handler/src/model-handler.ts
zachferland: correct types here, but was wondering why we are using bytes in one place (use string or streamid in tile handler before), any reason? 

I can look closer again to maybe see why

stbrody: `model` is definitely a byte array in the header: https://github.com/ceramicnetwork/js-ceramic/blob/7fa990c1a7185381743ab6f038e510ce2b28b3c9/packages/stream-model/src/model.ts#L310.

@oed  had the idea to use a byte array for model to speed up serialization/deserialization.  TileDocument doesn't use model so isn't directly relevant, and for other places that serialized StreamIDs as strings we mostly left alone to avoid breaking anything

zachferland: yup! could have sworn i saw it differ before, but must have confused it with the strings in tile stream, updated 

pr should be all set to merge now

packages/common/src/utils/signature_utils.ts
```diff
@@ -42,13 +43,13 @@ export class SignatureUtils {
...
    * @param streamId - Stream ID for the commit
-   * @param family - family of the stream being updated
+   * @param model - model of the stream being updated 
    * @returns Cacao is capability was found and verified, null otherwise
    */
   private static async _verifyCapabilityAuthz(
     commitData: CommitData,
     streamId: StreamID,
-    family: string
+    model: StreamID
```
stbrody, line 52:
```suggestion
    model: StreamID | null
```

stbrody (approved):

packages/stream-model-handler/src/model-handler.ts
zachferland: last week I believe they did, now they do throw, just checked (maybe model validation be added)

zachferland:
cleaned up those last bits, merging

PR #2215, opened 2022-06-06
bug(cli): fix port coercion and parsing for config file additions

willex:
## Description

Updates port validation and coercion to be properly invoked with config file CLI switch and fixes the improper use of serializer vs. deserializer. See ticket for details.

## How Has This Been Tested?

- [x] Manual local daemon testing via `--port`, `--config` switches & default config file
- [x] Full local `npm run test`

## PR checklist

## References:


packages/cli/src/daemon-config.ts
```diff
@@ -88,18 +88,7 @@ export class DaemonHTTPApiConfig {
   /**
    * Port to listen on.
    */
-  @jsonMember(AnyT, {
```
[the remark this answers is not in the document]
willex: Yes; the checks happen now explicitly during the config assignment (newly added) and daemon creation - I actually had it at the CLI port evaluation as well but that seemed redundant. This follows the initial pattern before I added my first fix.

Prior to @ukstv updates there was no easy/separate way of coercing the port for CLI switches besides putting it into the daemon config parser itself, but it turned out it was a bit hacky. Also, doing it during the config parser it is actually fired twice - once for instantiation and the second time for the actual launch of the daemon. Having it now explicitly check on the entry points feels a bit cleaner and establishes a pattern to do additional sanity checks on potential future values down the road too (e.g. db URL enforcement via ENV).

packages/cli/src/ceramic-cli-utils.ts
```diff
@@ -601,6 +601,21 @@ export class CeramicCliUtils {
   }
 }
 
+/**
+ * Helper function: Parse provided port and verify validity or exit process
+ * @param inPort
+ */
+export function _validatePort(inPort) {
```
ukstv, line 608: @stbrody @willex Are you ok with using usual names, with no underscores?

ukstv (approved):
Approve in general.

I am not quite sure about the leading underscore here though.

Consider it fully approved when we settle with @stbrody on the leading underscore here.

packages/cli/src/ceramic-cli-utils.ts
stbrody: Agreed, this is an exported function, so no need for an underscore.  I usually use underscores exclusively for private methods within a class to make clear the public vs private API

packages/cli/src/ceramic-cli-utils.ts
```diff
@@ -130,7 +130,7 @@ export class CeramicCliUtils {
         config.httpApi.hostname = hostname
       }
       if (port) {
-        config.httpApi.port = port
+        config.httpApi.port = _validatePort(port)
```
stbrody, line 133: do we even need this here if we're going to check it again in the CeramicDaemon constructor?

packages/cli/src/daemon-config.ts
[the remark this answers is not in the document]
stbrody: It feels to me like it would make sense to check it in the two entry points for the two ways it can be set - config file and CLI flags.

The alternative would be to do it in a single place further down, in a path that all entry points call into.  This version of the PR seems to mix both - checking at the edge for CLI flags, but also further down in the CeramicDaemon constructor that all paths call into

packages/cli/src/bin/ceramic.ts
```diff
@@ -1,7 +1,7 @@
 import { program } from 'commander'
 import pc from 'picocolors'
 
-import { CeramicCliUtils } from '../ceramic-cli-utils.js'
+import {CeramicCliUtils} from '../ceramic-cli-utils.js'
```
stbrody, line 4: make sure you're running prettier, the original formatting is what it should be

packages/cli/src/ceramic-cli-utils.ts
```diff
@@ -601,6 +601,21 @@ export class CeramicCliUtils {
   }
 }
 
+/**
+ * Helper function: Parse provided port and verify validity or exit process
+ * @param inPort
+ */
+export function validatePort(inPort) {
```
stbrody, line 608: now that you only use this function in ceramic-daemon.ts, you can just put it there and not need to export it.

packages/cli/src/daemon-config.ts
```diff
@@ -88,18 +88,7 @@ export class DaemonHTTPApiConfig {
...
-      if (inPort == null) {
-        return inPort
-      } else if (isNaN(validPort) || validPort > 65535) {
-        console.error('Invalid port number passed.')
-        process.exit(1)
-      }
-      return validPort
-    },
-  })
+  @jsonMember(AnyT)
```
stbrody, line 91: Number
```suggestion
  @jsonMember(Number)
```

willex: Need to verify that this indeed works to switch it back into `Number`. I had an issue with that in the past because of the instantiation with the value in memory (hence why I moved the coercion initially into the member serializer).

stbrody: oh right, that was where the original error we were seeing came from, the whole reason we started this line of work.  If using a serializer here stopped that error from happening, then it seems like the serializer was indeed being used?

willex: yeah, going full circle with this one. It was being used but in a "hacky" form (using serializer vs. deserializer etc.) that didn't fully work for the config  so pushing it into the `CeramicDaemon` class addresses now all use cases and ensures that either the port is correctly parsed or the daemon exits.

stbrody: I'm so confused.  How could it work but hacky?  When a config file is loaded and parsed, either it calls the serializer or it calls the deserializer.  It doesn't call both.  Or am I just totally misunderstanding how the option parser works?

willex: Added comment as per our discussion to explain the type requirement. Once you give your final OK going to merge it in.

stbrody (approved):
lgtm

PR #2218, opened 2022-06-09
feat: Add InsertionOrder and remove ChronologicalOrder

ukstv:


packages/core/src/indexing/sqlite/__tests__/read-fixture.util.ts
```diff
@@ -0,0 +1,35 @@
+import { IndexStreamArgs } from '../../database-index-api.js'
```
stbrody, line 1: `read-csv-fixture.util.ts` as the filename, to be more specific and clear what it contains?

packages/core/src/indexing/sqlite/__tests__/sqlite-index-api.test.ts
```diff
@@ -111,8 +106,8 @@ describe('indexStream', () => {
     expect(raw.stream_id).toEqual(STREAM_ID_B)
     expect(raw.controller_did).toEqual(CONTROLLER)
     expect(raw.last_anchored_at).toBeNull()
-    const createdAt = new Date(raw.created_at * 1000)
```
stbrody, line 114: why did this change?

packages/core/src/indexing/sqlite/insertion-order.ts
```diff
@@ -115,39 +110,24 @@ export class ChronologicalOrder {
     }
   }
 
-  /**
-   * Forward query: traverse from the most recent to the last.
```
stbrody, line 119: why lose this comment?

packages/core/src/indexing/sqlite/insertion-order.ts
```diff
@@ -115,39 +110,24 @@ export class ChronologicalOrder {
...
-            builder.whereNull('last_anchored_at').andWhere('created_at', '<', after.created_at)
-          )
-          .orWhereNotNull('last_anchored_at')
-      }
-    } else {
-      return base
+      return base.where('created_at', '<', after.created_at)
     }
+    return base
+    // throw new NotImplementedError('forwardQuery')
```
stbrody, line 128: rm

packages/core/src/indexing/sqlite/insertion-order.ts
```diff
@@ -115,39 +110,24 @@ export class ChronologicalOrder {
...
   }
 
-  /**
-   * Backward query: traverse from the last to the most recent.
```
stbrody, line 148: ditto, why lose the comment?

stbrody (approved):
lgtm with a few nitpick-y optional suggestions

packages/core/src/indexing/sqlite/__tests__/insertion-order.test.ts
```diff
@@ -0,0 +1,111 @@
...
+let EXPECTED: Array<string>
+
+let tmpFolder: tmp.DirectoryResult
+let dataSource: DataSource
+let order: InsertionOrder
+
+beforeEach(async () => {
+  tmpFolder = await tmp.dir({ unsafeCleanup: true })
+  dataSource = new DataSource({
+    type: 'sqlite',
```
willex, line 23: You could just replace this with:

`dataSource = knex({
  client: 'sqlite3', // or 'better-sqlite3'
  connection: {
    filename: "${tmpFolder.path}/tmp-ceramic.sqlite"
  }
});`

and remove typeorm then.

willex (approved):
lgtm - see comments about mixing knex with typeorm and potentially just moving to knex all together

packages/core/src/indexing/sqlite/__tests__/read-fixture.util.ts
ukstv: Ack. Also renamed `readFixture` to `readCsvFixture`.

packages/core/src/indexing/sqlite/__tests__/sqlite-index-api.test.ts
ukstv: SQLite does not _really_ handle timestamps. So, the decision how to store timestamps is up to a developer. Before this change we assumed second precision. That can lead to problems for a developer, if she does multiple insertions during a single second. A situation easy to imagine. Insertion ordering in this case would be effectively broken for her, as order is undefined for the entries having the same timestamp.

Just in case, PG does not suffer, as it has microseconds precision.

packages/core/src/indexing/sqlite/insertion-order.ts
ukstv: No reason. A mistake here. Thanks!

packages/core/src/indexing/sqlite/insertion-order.ts
ukstv: Same response :D

packages/core/src/indexing/sqlite/insertion-order.ts
ukstv: ack

PR #2219, opened 2022-06-09
feat: Attempt to limit concurrent S3 reads

ukstv:
Using Semaphore, nothing fancy. Semaphore limit is set based on the S3 minimum 5500 limit for GET/HEAD requests.

Although reality is a bit more complex, this might be a good first approach. According to the [docs](https://aws.amazon.com/premiumsupport/knowledge-center/s3-503-within-request-rate-prefix/) S3 is really flexible, but it requires the exponential backoff in the ideal case. IMO, it is a responsibility of s3leveldown ([issue](https://github.com/loune/s3leveldown/issues/13) created) to handle that.

I am not certain that the semaphore approach here makes sense, so feel free to say no to it.

### DIFF
--- a/packages/cli/src/s3-state-store.ts
+++ b/packages/cli/src/s3-state-store.ts
@@ -4,7 +4,6 @@ import { StateStore } from '@ceramicnetwork/core'
 import LevelUp from 'levelup'
 import S3LevelDOWN from 's3leveldown'
 import toArray from 'stream-to-array'
-import { Semaphore } from 'await-semaphore'
 
 const MAX_CONCURRENT_READS = 4000
 
@@ -13,7 +12,6 @@ const MAX_CONCURRENT_READS = 4000
  */
 export class S3StateStore implements StateStore {
   readonly #bucketName: string
-  readonly #loadSemaphore: Semaphore = new Semaphore(MAX_CONCURRENT_READS)
   #store
 
   constructor(bucketName: string) {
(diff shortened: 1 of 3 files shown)


willex (approved):
thanks for the semaphore background just now - lgtm

packages/cli/src/s3-state-store.ts
```diff
@@ -4,12 +4,16 @@ import { StateStore } from '@ceramicnetwork/core'
 import LevelUp from 'levelup'
 import S3LevelDOWN from 's3leveldown'
 import toArray from 'stream-to-array'
+import { Semaphore } from 'await-semaphore'
+
+const MAX_CONCURRENT_READS = 4000
```
stbrody, line 9: The 5500 limit from S3 is for requests *per second* I assume?  Allowing 4k simultaneous requests could still easily go over the per second limit.  We could do something like we do in pubsub-ratelimit to actually enforce a request per second limit.  I'd also be fine just turning this number down a bit to something like 1000.  

I don't feel like I have a strong instinct on how likely this is to actually bite us in practice.  I'd be interested in @oed and @smrz2001's opinions here

oed: I wonder if there's a way to batch requests into one S3 call?

smrz2001: @oed, `s3leveldown` does have a batch API but seems like it just makes individual API calls behind-the-scenes. Making direct S3 calls when using this plugin might be complicated.

smrz2001: Yeah, @stbrody, I agree, this does seem likely to bust the 5.5K limit. I think it really depends on how S3 measures per-second rates.

Would probably be good to run a quick test with standalone TS (i.e. not from inside Ceramic) calling `s3leveldown` with the semaphore to see when we get throttled. Perhaps a worst case test where we instantly fill up the semaphore and keep it full for a little while.

Alternatively, how hard would it be to use prefixes?

oed: The hard part about using prefixes would be the migration.

ukstv:
All right, so the best solution I am aware of is using p-queue. It allows you to control actions per second. Now the limit is honest 4000 _requests per second_.

smrz2001 (approved):
:shipit:

stbrody (approved):
lgtm

PR #2226, opened 2022-06-14
build: cacao patch v1.1.1 clockskew

zachferland:
- cacao patch v1.1.1 clockskew
- dids v3.2.0 with cacao phaseout
- abort controller lib change

### DIFF
--- a/packages/key-did-resolver/src/nist_weierstrass_common.ts
+++ b/packages/key-did-resolver/src/nist_weierstrass_common.ts
@@ -8,8 +8,8 @@ import { base64url } from 'multiformats/bases/base64'
   * For BigInt see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt
   */
 interface BigIntPoint {
-   x: BigInt,
-   y : BigInt
+   x: bigint,
+   y : bigint
 }
 
 /**
--- a/package-lock.json
+++ b/package-lock.json
@@ -31,12 +31,24 @@
         "typescript": "^4.5.4"
       }
     },
+    "node_modules/@ampproject/remapping": {
+      "version": "2.2.0",
+      "resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.2.0.tgz",
+      "integrity": "sha512-qRmjj8nj9qmLTQXXmaR1cck3UXSRMPrbsLJAasZpF+t3riI71BXed5ebIOYwQntykeZuhjsdweEc9BxH5Jc26w==",
+      "dev": true,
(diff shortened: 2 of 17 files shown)


zachferland:
hmm not sure about ci build issue yet, do not see this locally

packages/key-did-resolver/src/nist_weierstrass_common.ts
```diff
@@ -8,8 +8,8 @@ import { base64url } from 'multiformats/bases/base64'
   * For BigInt see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt
   */
 interface BigIntPoint {
-   x: BigInt,
-   y : BigInt
+   x: bigint,
+   y : bigint
```
oed, line 12: Why this change?

oed (commented):
one question otherwise lgtm

stbrody:
@zachferland can you explain why this change made the abort controller issue show up?  I don't see you using abort controller in your code anywhere, is it used in one of our dependencies?  If it's one of the dependencies we control, should we fix it there?

packages/key-did-resolver/src/nist_weierstrass_common.ts
oed: Why is your linter different from the rest of the teams?

oed: I don't understand why this change is needed based on the changes you did in the code.

zachferland:
@stbrody yeah its only due to upgrading dependencies (ref issue linked above, also seen in earlier CI fail), likely due to node types or typescript dependencies upgrading (not sure exactly unfortunately). But the current abort-controller library wont align on types now until they change it. 

even if we tried to only up dependencies needed here in package lock, would likely run into same issue soon 

we use abort-controller in common

zachferland:
basically node type and abort-controller pollyfill/lib dont match types now

stbrody:
I will defer to @ukstv on this review

stbrody:
FYI I'd like to do a release tomorrow with fixes for gitcoin and ideally this would be included in it if we can get to consensus on the changes in time

ukstv:
I am not sure what happened there, but apparently it was something weird, considering changes in abort-controller package and package-lock file. Maybe simpler https://github.com/ceramicnetwork/js-ceramic/pull/2237 could work so that @zachferland does not have to redo the work in this PR?

ukstv:
I do not know, maybe different version of npm contributed to the confusion?

ukstv:
Can we close it now @zachferland ?

PR #2231, opened 2022-06-14
chore(core): Always log streamId on errors loading commits from IPFS

stbrody:
Makes the `streamId` argument to `Dispatcher.retrieveCommit` mandatory, which means that anytime there is an error loading a commit from IPFS we will have the StreamID it belongs to in the log message


willex (approved):
lgtm

PR #2232, opened 2022-06-14
chore(core): Warn if no ethereumRpcUrl provided on mainnet

stbrody:


packages/core/src/ceramic.ts
```diff
@@ -49,6 +49,7 @@ import * as path from 'path'
 import { buildIndexing } from './indexing/build-indexing.js'
 import type { DatabaseIndexApi } from './indexing/database-index-api.js'
 import { LocalIndexApi } from './indexing/local-index-api.js'
+import * as net from 'net'
```
oed, line 52: What is this? I don't see it being used anywhere.

oed (commented):
Great to add this.

packages/core/src/ceramic.ts
stbrody: gah, must have been done my by IDE, I'll remove

oed (approved):

PR #2233, opened 2022-06-14
feat(ipfs-topology): Do not connect to peerlist nodes at startup

stbrody:
FYI @ukstv

### DIFF
--- a/packages/ipfs-topology/src/ipfs-topology.ts
+++ b/packages/ipfs-topology/src/ipfs-topology.ts
@@ -1,29 +1,11 @@
-import fetch from 'cross-fetch'
 import { Networks } from '@ceramicnetwork/common'
 import type { DiagnosticsLogger, IpfsApi } from '@ceramicnetwork/common'
 import { Multiaddr } from 'multiaddr'
 
-const PEER_FILE_URLS = (ceramicNetwork: Networks): string | null => {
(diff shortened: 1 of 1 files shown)


packages/ipfs-topology/src/ipfs-topology.ts
```diff
@@ -118,41 +81,32 @@ export class IpfsTopology {
...
-    if (connectedPeers.length > 0) {
-      const peerAddresses = connectedPeers.map((obj) => obj.addr)
-      this.logger.imp(`Connected to peers: ${peerAddresses.join(', ')}`)
+    for (const node of filteredBootstrapList) {
+      try {
+        await ipfs.swarm.connect(node)
+      } catch (error) {
+        this.logger.warn(`Can not connect to ${node}`)
+        this.logger.warn(error)
+      }
```
oed, line 109: Why is this code changed from an async map to a sync for loop?

oed (approved):
lgtm, question though around one of the changes

packages/ipfs-topology/src/ipfs-topology.ts
stbrody: so it happens one at a time instead of hitting the IPFS node with a bunch of requests all at once.  Trying to minimize load on IPFS

PR #2235, opened 2022-06-16
chore: Make memoization slightly faster and more reliable

ukstv:
_This is not related to any of the ongoing work. Feel free to consider that the lowest priority._

@stbrody's [hypothesis](https://github.com/ceramicnetwork/js-ceramic/pull/2214#issuecomment-1148909870) about memoization changing shape of a `StreamID` instance caught my attention. That _should not_ affect Jest snapshots, but technically memoization modifies the instance.

Apparently, there is a better way to handle memoization. It is [20%-40% faster](https://github.com/ukstv/mapmoize/runs/6923693346?check_suite_focus=true#step:6:16), smaller([802B](https://bundlephobia.com/package/mapmoize@1.0.0) vs [1.4kB](https://bundlephobia.com/package/typescript-memoize@1.1.0) now). It does not pollute the instances thanks to [`WeakMap`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap) usage.

Also, it could support [LRU](https://github.com/ukstv/mapmoize#custom-arguments-cache) for memoization, so we could use it more widely.


stbrody:
the change to js-ceramic isn't really signficant, the real thing that needs review is the implementation of the mapmoize package: https://github.com/ukstv/mapmoize/blob/main/src/index.ts.  (note @ukstv the NPM repo doesn't link to the source code, I had to find it on your github directly)

I don't really feel equipped to review the mapmoize library implementation.  Maybe @zachferland or @oed do?

stbrody:
> Folks @stbrody @willex @stephhuynh18 any news on approval?
> 
> The merge is obviously conditional on conflict resolution.

Can you open a pr for the implementation of mapMemoize? That's actually the important and complex part.

stephhuynh18 (approved):

ukstv:
@stephhuynh18 Thank you!
@stbrody Good to go?

PR #2236, opened 2022-06-16
fix(core): Creating a stream via a multiquery should pin it

stbrody:
I don't really *love* this approach.  I think it's weird to use a query as a backdoor way to create a stream.  But that's what we did when we introduced https://github.com/ceramicnetwork/js-ceramic/issues/1647 so this seems like the only way to make the behavior actually correct.


oed (commented):
Using `createStreamFromGenesis` doesn't really seem like the right approach. Couldn't we call `this.pin.add(streamId.baseID)` in the `_loadLinkedStreams` method?

oed (approved):
Actually, looks like `pin.add` also loads the stream a second time which was my main concern. So this lgtm.

PR #2238, opened 2022-06-17
Added support for ModelDefinitions.views field

pawartur:
## Description

This will unblock us to enable devs to mark some fields as referencing document controller their GraphQL SDL definitions of models.

The `views` field should be:
```
interface ModelDefinition {
  ...
  views?: ModelViewsDefinition
}
```
, where
```
type ModelViewsDefinition = Record<string, ModelViewDefinition>
```
and
```
ModelViewDefinition = { type: 'documentAccount' }
```

## How Has This Been Tested?

- [X] Added unit tests for the new `ViewsValidation` interface
- [X] Updated tests for `ModelStreamHandler`

## PR checklist


packages/stream-model-handler/src/__tests__/model-handler.test.ts
```diff
@@ -94,6 +94,52 @@ const FINAL_CONTENT = {
...
 }
 
+const FINAL_CONTENT_WITH_ACCOUNT_DOCUMENT_VIEW: ModelDefinition = {
+  name: 'MyModel',
+  accountRelation: ModelAccountRelation.LIST,
+  schema: {
+    $schema: 'https://json-schema.org/draft/2020-12/schema',
+    type: 'object',
+    properties: {
+      owner: {
```
stbrody, line 104: I thought these view properties wouldn't show up in the schema itself, but it just metadata for the client library to expose the stream metadata with a given name?  @PaulLeCam

packages/stream-model-handler/src/__tests__/model-handler.test.ts
```diff
@@ -290,6 +336,18 @@ describe('ModelHandler', () => {
     await checkSignedCommitMatchesExpectations(did, commit, expectedGenesis)
   })
 
+  it('supports view properties in genesis comment', async () => {
```
stbrody, line 339:
```suggestion
  it('supports view properties in genesis commit', async () => {
```

packages/stream-model-handler/src/__tests__/model-handler.test.ts
pawartur: Yes! That's right, sorry for the mistake.

PaulLeCam (approved):
Looks great, thanks!
I'm not the most familiar with this codebase so better wait for approval from other reviewers, but it looks all good to me.

packages/stream-model-handler/src/__tests__/model-handler.test.ts
```diff
@@ -94,6 +94,46 @@ const FINAL_CONTENT = {
   },
 }
 
+const FINAL_CONTENT_WITH_ACCOUNT_DOCUMENT_VIEW: ModelDefinition = {
```
ukstv, line 97: Based on what I see, you could do something like
```typescript
{
  ...FINAL_CONTENT,
  views: {
    owner: {
      type: 'documentAccount'
    }
  }
}
```

packages/stream-model-handler/src/model-handler.ts
```diff
@@ -125,6 +128,9 @@ export class ModelHandler implements StreamHandler<Model> {
 
     if (state.content.schema !== undefined) {
```
ukstv, line 129:
```suggestion
    if (state.content.schema) {
```

IMO this should work too, and is a bit more idiomatic.

packages/stream-model-handler/src/model-handler.ts
```diff
@@ -125,6 +128,9 @@ export class ModelHandler implements StreamHandler<Model> {
...
       await this._schemaValidator.validateSchema(state.content.schema)
+      if (state.content.views !== undefined) {
```
ukstv, line 131:
```suggestion
      if (state.content.views) {
```

packages/stream-model-handler/src/views-utils.ts
```diff
@@ -0,0 +1,22 @@
...
+
+const SUPPORTED_VIEW_TYPES = [
+  'documentAccount'
+]
+
+export class ViewsValidation {
+  public validateViews(
+    views: ModelViewsDefinition,
+    schema: JSONSchema.Object
+  ) {
```
ukstv, line 12:
```suggestion
  ): void {
```

packages/stream-model-handler/src/views-utils.ts
```diff
@@ -0,0 +1,22 @@
...
+    Object.entries(views).map(([key, value]) => {
```
ukstv, line 13:
```suggestion
    Object.entries(views).forEach(([key, value]) => {
```

you do not really return anything here

packages/stream-model-handler/src/views-utils.ts
```diff
@@ -0,0 +1,22 @@
...
+      if (!SUPPORTED_VIEW_TYPES.includes(value.type)) {
```
ukstv, line 14: Suggested micro-optimisation. Replace `SUPPORTED_VIEW_TYPES` array with an object like
```
{
documentAccount: true
}
```

Then this line would be O(1): `SUPPORTED_VIEW_TYPES[value.type]`.

ukstv (approved):
few beautifications here and there are possible, but otherwise LGTM

🏇

stbrody (approved):
two small nitpicks, but overall looks good

PR #2243, opened 2022-06-17
feat(stream-model-instance): Enforce that model field must refer to a Model StreamID

stbrody:


packages/stream-model-instance-handler/src/model-instance-document-handler.ts
```diff
@@ -1,5 +1,6 @@
 import jsonpatch from 'fast-json-patch'
 import cloneDeep from 'lodash.clonedeep'
+import { Model } from '@ceramicnetwork/stream-model'
```
oed, line 3: Please use the [StreamID](https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/streamid/src/stream-type.ts) package instead!

oed (changes requested):

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
stbrody: that doesn't give a way to directly get the Model streamtype code.  I'd either need to hardcode the type number here or hardcode the string description (`"model"`) and look up the code from that.  In either case this package would need to be making an assumption about the identifier of the Model streamtype, so at that point I might as well just hardcode the type number.  I can do that if you prefer.

oed: Ok, hardcoding seems better imo. Type int should never change anyway.

stbrody: updated

PR #2249, opened 2022-06-20
chore(stream-tile-handler, stream-caip10-link-handler): Add 'prev' connection check to TileDocumentHandler and Caip10LinkHandler

stbrody:
We added this check to ModelHandler and ModelInstanceDocumentHandler, but left TileDocumentHandler alone to minimize the changes there.  But a lot of our test coverage of conflict resolution is still done with TileDocument, so we should keep the logic here consistent across our stream types to make sure we have good test coverage.  The change should be harmless anyway


packages/stream-tile-handler/src/tile-document-handler.ts
```diff
@@ -149,6 +148,15 @@ export class TileDocumentHandler implements StreamHandler<TileDocument> {
       throw new Error(`Invalid streamId ${payload.id}, expected ${state.log[0].cid}`)
     }
 
+    const expectedPrev = state.log[state.log.length - 1].cid
+    if (!payload.prev.equals(expectedPrev)) {
+      // This should never happen and would indicate a programming error if it did
```
oed, line 153: I guess the same is true for `payload.id` that is checked above.

oed (approved):

stbrody:
Took this PR a little bit farther and made a standard helpful function for doing this validation and made all 4 streamtypes use it in the same way, for both signed commits and anchor commits.  So now all the behavior should be totally consistent.  PTAL

stbrody:
oh boy, adding this validation to Caip10Link uncovered that we've been setting the `id` field in update commits for Caip10Links for a very long time.  I'll put together a patch to fix that, but it means we won't be able to do this validation in Caip10Link since there are likely many Caip10Link streams out there with the wrong value for the 'id' field in their commits that are out there already

stbrody:
> oh boy, adding this validation to Caip10Link uncovered that we've been setting the `id` field in update commits for Caip10Links incorrectly for a very long time (setting it to the previous tip instead of the genesis commit). I'll put together a patch to fix that, but it means we won't be able to do this validation in Caip10Link since there are likely many Caip10Link streams out there with the wrong value for the 'id' field in their commits that are out there already

Also applies to TileDocument commits made with `patch()` instead of `update()`.  More details in the PR to fix: https://github.com/ceramicnetwork/js-ceramic/pull/2256

stbrody:
Okay, updated and ready for re-review @oed.  Also adding @ukstv to review

packages/stream-caip10-link-handler/src/caip10-link-handler.ts
```diff
@@ -151,7 +154,8 @@ export class Caip10LinkHandler implements StreamHandler<Caip10Link> {
     commitData: CommitData,
     state: StreamState
   ): Promise<StreamState> {
-    // TODO: Assert that the 'prev' of the commit being applied is the end of the log in 'state'
+    StreamUtils.assertCommitLinksToState(state, commitData.commit)
```
oed, line 157: Wasn't there a bug (or still is) that the anchor service construction of the anchor commit doesn't include the `id` ?
Please investigate before merging.

oed (commented):
Lgtm, mod comment. 
Also, should we add tests for this, or is it already covered enough?

packages/stream-caip10-link-handler/src/caip10-link-handler.ts
stbrody: ooph yeah good call.  It's fixed in current CAS but did exist in the past so would affect some existing streams.  I'll have the check only apply if the `id` field actually exists

oed (approved):

PR #2251, opened 2022-06-20
feat(stream-model, stream-model-instance): Use 'controller' instead of 'controllers' in metadata

stbrody:


stbrody:
> What's the benefit of this change? To shave off 3 bytes or for DX? If it's for DX, can't we just change the APIs rather than the commit structure? I worry that if we keep adding changes to the commit structure we end up having a lot of logic just to deal with that.

yeah it's mostly to avoid confusion by people who see it and assume that means we support multiple controllers.  I guess I was hoping that at some point we might stop supporting TileDocument and Caip10Link entirely (e.g. when we launch the incentivized network maybe it only supports newer streamtypes) and then we'd be able to standardize on the single new way of doing things.  If you think we're stuck supporting TileDocument and Caip10Link for all of time, then I'd agree we should minimize the number of different formats we need to support.

oed:
Even if we drop support for Tile and Caip10Link, there might be a need to have a streamtype that supports multiple controllers (3ID v2 would need this), so we would always need to keep support for `controllers` regardless I assume.

stbrody:
> Even if we drop support for Tile and Caip10Link, there might be a need to have a streamtype that supports multiple controllers (3ID v2 would need this), so we would always need to keep support for `controllers` regardless I assume.

Okay, I'll update this to change what shows up in the StreamState that gets shown to clients, but leave the field in the actual commit headers the same (basically will do what I did to TileDocument to Model and MID)

packages/canary-integration/src/__tests__/capability.test.ts
```diff
@@ -114,7 +115,7 @@ describe('CACAO Integration test', () => {
       const deterministicDocument = await TileDocument.deterministic(ceramic, {
         deterministic: true,
         family: 'testCapabilities1',
-        controllers: [`did:pkh:eip155:1:${wallet.address}`],
+        controller: [`did:pkh:eip155:1:${wallet.address}`],
```
oed, line 118: This doesn't seem right. Passing an array to the `controller` prop?

packages/canary-integration/src/__tests__/model-instance-document.test.ts
```diff
@@ -199,7 +175,6 @@ describe('ModelInstanceDocument API http-client tests', () => {
     const doc2 = await ModelInstanceDocument.create(ceramic, CONTENT0, midMetadata)
 
     expect(doc1.id.toString()).not.toEqual(doc2.id.toString())
-    expect(doc1.metadata.unique.toString()).not.toEqual(doc2.metadata.unique.toString())
```
oed, line 202: Why was this line removed?

packages/canary-integration/src/__tests__/model.test.ts
```diff
@@ -47,7 +47,6 @@ describe('Model API http-client tests', () => {
     expect(model.state.log.length).toEqual(1)
     expect(model.state.log[0].type).toEqual(CommitType.GENESIS)
     expect(model.state.anchorStatus).toEqual(AnchorStatus.PENDING)
-    expect(model.metadata.unique instanceof Uint8Array).toBeTruthy()
```
oed, line 50: Removed?

packages/canary-integration/src/__tests__/model.test.ts
```diff
@@ -98,7 +97,6 @@ describe('Model API http-client tests', () => {
     const model2 = await Model.create(ceramic, FINAL_CONTENT)
 
     expect(model1.id.toString()).not.toEqual(model2.id.toString())
-    expect(model1.metadata.unique.toString()).not.toEqual(model2.metadata.unique.toString())
```
oed, line 101: Same here?

packages/canary-integration/src/__tests__/model.test.ts
oed, line 50: Removed?

packages/canary-integration/src/__tests__/model.test.ts
oed, line 101: Same here?

packages/core/src/__tests__/__snapshots__/caip10-link.test.ts.snap
```diff
@@ -264,6 +264,7 @@ Object {
     },
   ],
   "metadata": Object {
+    "controller": "0x8fe2c4516e920425e177658aaac451ca0463ed69@eip155:1337",
```
oed, line 267: Why is this exposed on existing stream types? I thought we wanted only to do it on new ones. Seems like this adds a bunch of duplication?

packages/stream-caip10-link-handler/src/caip10-link-handler.ts
```diff
@@ -59,23 +59,25 @@ export class Caip10LinkHandler implements StreamHandler<Caip10Link> {
       throw new Error('Caip10Link genesis commit cannot have data')
     }
 
+    const metadata = commit.header
+    if (!(metadata.controllers && metadata.controllers.length === 1)) {
+      throw new Error('Exactly one controller must be specified')
+    }
+    metadata.controller = metadata.controllers[0]
```
oed, line 66: I don't understand why we want to duplicate data in existing stream types.

packages/stream-model-handler/src/__tests__/model-handler.test.ts
```diff
@@ -805,7 +810,8 @@ describe('ModelHandler', () => {
     const state$ = TestUtils.runningState(state)
     const doc = new Model(state$, context)
     const rawCommit = doc._makeRawCommit(FINAL_CONTENT)
-    rawCommit.header = { controllers: [did.id, did.id] }
+    const newDid = 'did:3:k2t6wyfsu4pg0t2n4j8ms3s33xsgqjhtto04mvq8w5a2v5xo48idyz38l7zzzz'
+    rawCommit.header = { controller: newDid }
```
oed, line 814: I thought we said not to change the commit headers?

oed (commented):
I don't feel very confident about these changes. I'm afraid it might cause more confusion for a developer looking at StreamState than it clarifies things.

Seems like we are trying to improve DX by allowing passing of a single controller string instead of a controllers array. This is great. Unfortunately it seems like this API abstraction leaks into our internal representation of the StreamState, but we limit the leakage such that commits are not affected.

It seems weird to me that our APIs for streams are so tightly coupled to the StreamState.

packages/canary-integration/src/__tests__/capability.test.ts
stbrody: oops, typo.  Fixed

packages/core/src/__tests__/__snapshots__/caip10-link.test.ts.snap
stbrody: I thought it would be good to have the `controller` field show up in all streamtypes, so applications that use multiple streamtypes (like cerscan for example) doesn't have to look for the controller in a different place depending on streamtype.  The 2 new streamtypes now only have `controller`, but the old streamtypes have both `controller` and `controllers` - `controllers` for backwards compatibility, `controller` for consistency with other streamtypes.

If you'd prefer, I could leave TileDoc and Caip10Link with only having `controllers` and have Model and MID only have `controller`

packages/stream-caip10-link-handler/src/caip10-link-handler.ts
stbrody: maintaining consistency between streamtypes without breaking backwards compatibility for existing apps

packages/stream-model-handler/src/__tests__/model-handler.test.ts
stbrody: left over from old version.  Doesn't really matter to the test as it will error if the header is set at all.  Anyway fixed.

stbrody:
New version up that leaves the internal StreamState representation unchanged, and leaves the user-facing APIs of TileDocument and Caip10Link unchanged, while making the metadata returned for Model and MID streams have just a single `controller` field.  This involved making the `metadata` method in the `Stream` base class purely abstract so streamtypes are forced to implement it themselves and can make their own decisions on how the metadata in the StreamState maps to their stream-specific metadata format that should be shown to users.  PTAL

packages/stream-model/src/model.ts
```diff
@@ -118,6 +118,10 @@ export class Model extends Stream {
     return super.content
   }
 
+  get metadata(): ModelMetadata {
+    return { controller: this.state$.value.metadata.controllers[0] }
```
oed, line 122: Should we also return the model property here?

oed (approved):
lgtm, mod one question

packages/stream-model/src/model.ts
stbrody: I'm worried that if we expose it, users will try to use the streamid to load a stream and then be confused when it doesn't work.

oed:
But isn't indexing based on this property? Doesn't the index api use this
accessor?

*Joel Thorstensson *
Co-founder / CTO

stbrody:
> But isn't indexing based on this property? Doesn't the index api use this accessor?

Gah yeah that's a good point.  I'll put it in.  I still kind of wish we could figure make an actual Model for the Models so that loading that StreamID would actually do something.

PR #2252, opened 2022-06-21
feat(core, canary-integration): add stream to index api http

willex:
## Description

Added ability to add stream to index db (api) via extension of the core ceramic service. The stream will be automatically added to the relevant table, but currently does not check for table existence.

## How Has This Been Tested?

- [ ] Tested locally manually by creating/updating streams against the local node

## PR checklist

## References:


packages/canary-integration/src/__tests__/model-instance-document.test.ts
```diff
@@ -2,20 +2,81 @@ import { jest } from '@jest/globals'
...
+    },
+  })
+})
+
+afterEach(async () => {
+  await dbConnection.destroy()
+  await tmpFolder.cleanup()
+})
+
+describe('init', () => {
```
ukstv, line 57: Does it really belong here? It looks like an existing test suite for sqlite-index-api.

packages/canary-integration/src/__tests__/model-instance-document.test.ts
```diff
@@ -54,11 +113,28 @@ describe('ModelInstanceDocument API http-client tests', () => {
...
 
     const port = await getPort()
     const apiUrl = 'http://localhost:' + port
-    daemon = new CeramicDaemon(core, DaemonConfig.fromObject({ 'http-api': { port } }))
+    daemon = new CeramicDaemon(
+      core,
+      DaemonConfig.fromObject({
+        'http-api': { port },
+        indexing: {
+          db: 'sqlite:///Users/Alex/Documents/GitHub/indexing.sqlite',
```
ukstv, line 133: Your local path?

ukstv: BTW, `indexing` clause is ignored in `CeramicDaemon` constructor.

packages/canary-integration/src/__tests__/model-instance-document.test.ts
```diff
@@ -112,27 +180,34 @@ describe('ModelInstanceDocument API http-client tests', () => {
...
   test('Create and update doc', async () => {
-    const doc = await ModelInstanceDocument.create(
-      ceramic,
-      CONTENT0,
-      midMetadata
-    )
+    const doc = await ModelInstanceDocument.create(ceramic, CONTENT0, midMetadata)
     expect(doc.content).toEqual(CONTENT0)
 
+    // confirm stream was added to index
```
ukstv, line 186: Dev artefact to remove, maybe?

packages/core/src/indexing/local-index-api.ts
```diff
@@ -20,6 +21,14 @@ export class LocalIndexApi implements IndexApi {
     private readonly logger: DiagnosticsLogger
   ) {}
 
+  /**
+   * Add stream to index in appropriate model table
+   * @param args
+   */
+  async indexStream(args: IndexStreamArgs): Promise<void> {
```
ukstv, line 28: IMO this method should handle the check https://github.com/ceramicnetwork/js-ceramic/pull/2252/files#diff-0625065ede662c738d45fe3f49a8a1afaba805c6208872cdbf82859722562e5cR47

Let's leave the SqliteIndexAPI to deal with the databases only.

IMO, when we add an intermediate table for StreamID -> table, this `LocalIndexAPI` should be a place to maintain an inmemory cache of it.

ukstv (changes requested):
It is not clear if the commented code in the test file is a remnant of a dev time, or by design.

Also, here we are mainly concerned with checking if `Ceramic` works as expected. This leads me to a thought, that the test should be in `core` package as well.

packages/core/src/indexing/local-index-api.ts
stbrody: Hmm... I agree about moving the check of whether or not the stream should be indexed up here, that way it only needs to live in one place even when we have multiple index database implementations.

As for the cache of the mapping from Model streamid to table name, that part I think should live at the lower level in the database-specific index implementation.  This is because we might use a different system to maintain the streamid->table mapping for different database implementations.

packages/core/src/indexing/sqlite/sqlite-index-api.ts
```diff
@@ -26,22 +26,39 @@ export class SqliteIndexApi implements DatabaseIndexApi {
     this.insertionOrder = new InsertionOrder(dbConnection)
   }
 
+  public getActiveModelsToIndex(): Array<StreamID> {
+    // Helper function to get array of active models that are currently being indexed by node
+    // TODO: update to runtime check once adminAPI unlocks to load models on the fly
```
stbrody, line 31: Even when we have the adminAPI and the table tracking the set of indexed models, I don't think we should be doing a database query here every time a stream is updated.  Instead I think the this class should maintain an in-memory cache of which streams are indexed and what their corresponding tables are.  This cache can get populated from the database at startup, and updated anytime there is a call to the admin API endpoint to add or remove a model from the set of models to index.

packages/core/src/indexing/sqlite/sqlite-index-api.ts
```diff
@@ -26,22 +26,39 @@ export class SqliteIndexApi implements DatabaseIndexApi {
...
+    return this.modelsToIndex
+  }
+
+  private _isIndexStream(args: StreamID): Boolean {
```
stbrody, line 35: `_isIndexedStream`?  `_shouldIndexStream`?

packages/core/src/indexing/sqlite/sqlite-index-api.ts
```diff
@@ -26,22 +26,39 @@ export class SqliteIndexApi implements DatabaseIndexApi {
...
+    // TODO: find more efficient way
```
stbrody, line 36: can use a Set instead of an array for `modelsToIndex`

packages/core/src/indexing/sqlite/sqlite-index-api.ts
```diff
@@ -26,22 +26,39 @@ export class SqliteIndexApi implements DatabaseIndexApi {
     this.insertionOrder = new InsertionOrder(dbConnection)
   }
 
+  public getActiveModelsToIndex(): Array<StreamID> {
```
stbrody, line 29: using a Set as the return value will make it more efficient to check for inclusion of a stream.  It will have to be a Set<string> though not a Set<StreamID> as the latter would only consider the StreamID as being included if you provided it the exact same StreamID instance at the same memory address

packages/core/src/indexing/sqlite/sqlite-index-api.ts
```diff
@@ -26,22 +26,39 @@ export class SqliteIndexApi implements DatabaseIndexApi {
...
-        updated_at: asTimestamp(args.updatedAt) || now,
-      })
-      .onConflict('stream_id')
-      .merge({
-        last_anchored_at: asTimestamp(args.lastAnchor),
-        updated_at: asTimestamp(args.updatedAt) || now,
-      })
+
+    // only index streams with active models in config
+    if (this._isIndexStream(args.model)) {
```
stbrody, line 47: use early return to reduce indentation and clean up the code.

```
if (!this._shouldIndexStream(args.model)) {
  return
}```

File: packages/canary-integration/src/__tests__/model-instance-document.test.ts [hunk 10]
```diff
@@ -2,20 +2,81 @@ import { jest } from '@jest/globals'
...
 import { CeramicClient } from '@ceramicnetwork/http-client'
 import { StreamID } from '@ceramicnetwork/streamid'
 import first from 'it-first'
 import { Model, ModelAccountRelation, ModelDefinition } from '@ceramicnetwork/stream-model'
+import { SqliteIndexApi } from '@ceramicnetwork/core/lib/indexing/sqlite/sqlite-index-api'
+import { listMidTables } from '@ceramicnetwork/core/lib/indexing/sqlite/init-tables'
+import knex, { Knex } from 'knex'
+import tmp from 'tmp-promise'
+
+const INDEXING_MODELS = [
```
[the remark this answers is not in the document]
stbrody: I think the one test worth adding is a simple one that spies on the ceramic node's `indexApi.indexStream` function and just asserts that it is called whenever a stream is created or updated.  We should also have an equivalent test in `model.test.ts`

stbrody (commented):
overall structure looks right, some small comments throughout.  Also think it needs some simple tests for the changes to the sqllite index api and the LocalIndexApi

packages/core/src/indexing/sqlite/sqlite-index-api.ts
ukstv: You could be surprised (I was), but Set is significantly slower than an array for <10000 entries. It is more idiomatic, but slow.

stbrody: yeah, that actually does make sense and is a fair point.  At my last job we used to have a type that implemented a `Set` API but used an array under-the-hood as the implementation, and we used that for small sets for exactly this reason.  I guess for now it's fine to just stick with Array.  But then let's remove that `TODO` comment too.

packages/core/src/indexing/local-index-api.ts
willex: Sounds good. I'll move the check as suggested by Sergey for the time being. For the cache that makes sense going forward.

packages/core/src/indexing/sqlite/sqlite-index-api.ts
willex: yeah, I like that approach. the idea was to have a starting point here as I think we will start out still with in-memory cache of models as per defined in the config file. I'd change the name then to `getActiveModelsToIndexFromCache` as the helper function and the subsequent cache updates as you described it then are done via the future extension.

packages/core/src/indexing/sqlite/sqlite-index-api.ts
willex: I like it. will update.

packages/core/src/indexing/sqlite/sqlite-index-api.ts
stbrody: nitpick, but IMO should leave the name as `getActiveModelsToIndex`, the fact that the data is coming from a cache is an implementation detail.  I think for now the `getActiveModelsToIndex` function can just return the models as listed in the config file, and then later can be from the cache that's build off the database table

packages/core/src/ceramic.ts
```diff
@@ -621,12 +621,29 @@ export class Ceramic implements CeramicApi {
...
+    const stream = streamFromState<T>(
       this.context,
       this._streamHandlers,
       state$.value,
       this.repository.updates$
     )
+
+    // add stream to MID indexing if model is present
+    if (stream.metadata.model) {
+      const last_anchor_ts = state$.state.anchorProof
```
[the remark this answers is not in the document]
willex: Looking at this, does it make sense to move the whole block into the helper function - that way it can be called externally to add streams directly if necessary.

packages/canary-integration/src/__tests__/model-instance-document.test.ts
willex: reverted back to original state as discussed for future canary test.

stbrody (approved):
LGTM!  Nice work

PR #2254, opened 2022-06-21
feat: warn at startup if runs SQLite in production

ukstv:


packages/core/src/initialization/make-index-api.ts
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Make IndexAPI instance. Call `buildIndexing` inside.
+ */
+export function makeIndexApi(
+  indexingConfig: IndexingConfig | undefined,
+  network: Networks,
+  logger: DiagnosticsLogger
+): DatabaseIndexApi | undefined {
+  if (!indexingConfig) {
+    logger.warn(`Indexing is not configured. Please, add the indexing settings to your config file`)
```
stbrody, line 32:
```suggestion
    logger.warn(`Indexing is not configured. Please add the indexing settings to your config file`)
```

packages/core/src/initialization/make-index-api.ts
```diff
@@ -0,0 +1,42 @@
...
+    return undefined
+  }
+  const indexApi = buildIndexing(indexingConfig)
+  if (isNonProduction(indexApi, network, indexingConfig.models)) {
+    logger.warn(
+      'For production deployments we recommend using Postgres not SQLite as it will scale better'
```
stbrody, line 38:
```suggestion
      'SQLite configuration detected for indexing database.  For production deployments we recommend using Postgres not SQLite as it will scale better'
```

stbrody (approved):
lgtm. Included two small suggestions for rewording some of the warning messages

packages/core/src/initialization/make-index-api.ts
willex: Potentially for an ever stronger message: `SQLite configuration detected for indexing database.  For production deployments we strongly recommend using Postgres, NOT SQLite, to avoid scalability and data integrity issues`

PR #2256, opened 2022-06-22
fix(core): StreamID comes from genesis commit CID, not tip

stbrody:
I discovered this when working on https://github.com/ceramicnetwork/js-ceramic/pull/2249 when I saw some Caip10Link tests failing the new validation that was added

Thankfully this bug only affects `StateLink`, not `RunningState` or all `RunningStateLike` objects.  Since I **believe** that `StateLink` is only used withing actual stream instances, but not within any of our internal machinery, I think the damage this bug did is at least somewhat limited.  The main places I believe that this can manifest is as an incorrect `id` field in all [Caip10Link update commits](https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/stream-caip10-link/src/caip10-link.ts#L195), and in TileDocument update commits that were done with the [patch()](https://github.com/ceramicnetwork/js-ceramic/blob/1346d88dd3ad91e23f91792b28730937cefb1416/packages/stream-tile/src/tile-document.ts#L296) method (but not the ones done with [update()](https://github.com/ceramicnetwork/js-ceramic/blob/1346d88dd3ad91e23f91792b28730937cefb1416/packages/stream-tile/src/tile-document.ts#L365))

This would affect indexers trying to index off anchor commits, especially from before we added the StreamID array to the merkle metadata as then the `id` from the commit would have been the best way to know what stream the commit belonged to, but now you'd have to walk the `prev` links all the way back to genesis to actually know the proper streamid.

### DIFF
--- a/packages/stream-caip10-link/src/caip10-link.ts
+++ b/packages/stream-caip10-link/src/caip10-link.ts
@@ -192,7 +192,7 @@ export class Caip10Link extends Stream {
    * @param proof
    */
   makeCommit(proof: LinkProof): RawCommit {
-    return { data: proof, prev: this.tip, id: this.state$.id.cid }
+    return { data: proof, prev: this.tip, id: this.id.cid }
   }
 
   /**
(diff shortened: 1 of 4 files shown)


ukstv (approved):
That is rather curious. LGTM of course.

stbrody:
Notes from standup: 
- change all places in streams to use `this.id` instead of `state$.id`.
- put back validation for TileDocument and Caip10Link and test with an existing Caip10Link from the network

PR #2262, opened 2022-06-24
Add `documentVersion` view to model type definitions

PaulLeCam:
Following our call yesterday, we need to add support for the `documentVersion` view to expose the version to set in mutations.


stbrody (approved):
lgtm

PR #2267, opened 2022-06-24
fix(stream-tile-handler,stream-model-handler,stream-model-instance-handler): Don't cache schemas

stbrody:
@ukstv could you double-check that this fixes the memory growth when associated with validating schemas (both for streams that share a schema and for streams with different schemas)?

### DIFF
--- a/packages/stream-model-handler/src/schema-utils.ts
+++ b/packages/stream-model-handler/src/schema-utils.ts
@@ -1,58 +1,57 @@
-
 import ajv, { SchemaObject } from 'ajv/dist/2020.js'
 
 /**
  * A type that we use to check if object properties within JSON Schema schemas
  * allow for additional properties.
- * 
+ *
  * According to JSON Schema standard, object properties within a valid schema have
  * type field set to 'object' and an optional additionalProperties field that can have a boolean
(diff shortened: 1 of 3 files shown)


oed (approved):
lgtm, why is ajv caching schemas, and why are the cached schemas not being reused if the same schema is validated again?

stbrody:
> lgtm, why is ajv caching schemas, and why are the cached schemas not being reused if the same schema is validated again?

Not sure what the reasoning for caching schemas is, but the reason they're not being reused is because the cache is keyed by the schema object, so only the exact same instance at the same memory address will register as the same schema and hit the cache.  Two separate instances with identical contents will be considered different and will fill up the cache separately

oed:
Could we boost performance by actually using this cache?

*Joel Thorstensson *
Co-founder / CTO

ukstv:
> Could we boost performance by actually using this cache?

AJV does it by design, in a very peculiar way. It leads to unbounded memory growth if we rely solely on AJV. See https://linear.app/3boxlabs/issue/CDB-1907#comment-900d849e

Best approach IMO is to use yet another LRUMap for AJV instances. Yes, for AJV instances, not for schemas.

PR #2269, opened 2022-06-27
feat: Heapdump on SIGUSR2

ukstv:
Make a heapdump when `SIGUSR2` signal is received. The heapdump is written to `<config-folder>/ceramic-<timestamp>.heapsnapshot`. 

It takes time to make a heapdump. We log when the heapdumping starts and finishes in `<config-folder>/heapdump-progress` text file.

The heapdump later can be analysed by Chrome DevTools.

### DIFF
--- a/packages/cli/src/ceramic-cli-utils.ts
+++ b/packages/cli/src/ceramic-cli-utils.ts
@@ -11,13 +11,14 @@ import { CeramicApi, LogLevel, Networks, StreamUtils } from '@ceramicnetwork/com
 import { StreamID, CommitID } from '@ceramicnetwork/streamid'
 
 import { CeramicDaemon } from './ceramic-daemon.js'
-import { DaemonConfig, IpfsMode, StateStoreMode, DaemonMetricsConfig } from './daemon-config.js'
+import { DaemonConfig, IpfsMode, StateStoreMode } from './daemon-config.js'
 import { TileDocument, TileMetadataArgs } from '@ceramicnetwork/stream-tile'
 
 import * as ThreeIdResolver from '@ceramicnetwork/3id-did-resolver'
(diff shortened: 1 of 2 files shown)


packages/cli/src/ceramic-cli-utils.ts
```diff
@@ -184,6 +188,7 @@ export class CeramicCliUtils {
       }
     }
 
+    handleHeapdumpSignal(new URL('./', configFilepath))
```
stbrody, line 191: isn't the configFilepath already an absolute path at this point?  Why are we adding `./` to the front of it here but not in `_loadDaemonConfig`?

stbrody (commented):
I think it'd be really valuable to have a log message in the normal ceramic log when a heapdump begins and ends.

packages/cli/src/ceramic-cli-utils.ts
ukstv: `configFilepath` is an absolute file path _to a config file_. We are interested in its parent folder, thus `./`.

stbrody: Ah got it, that makes sense.  Thanks

ukstv: Just in case, for `absolute = new URL('file:///hello/world.txt')`, `new URL('./', absolute)` would return `new URL('file:///hello/')`.

`./` means "get me the current folder". If  a URL references a file, i.e. no trailing slash, it means the parent folder. If the  URL references a folder, i.e. there is a trailing slash, it returns the same folder.

ukstv:
> I think it'd be really valuable to have a log message in the normal ceramic log when a heapdump begins and ends.

Makes sense. Done.

stbrody (approved):
LGTM

PR #2273, opened 2022-06-27
remove cross-fetch import

lsmod:
`cross-fetch` package creates issues while trying to use `@ceramicnetwork/http-client` in a chrome-browser extension (within a background script).
```
error while trying to authenticate:  ReferenceError: XMLHttpRequest is not defined
    at browser-ponyfill.js:462:1
    at new Promise (<anonymous>)
    at fetch (browser-ponyfill.js:455:1)
    at fetchJson (http-utils.js:16:1)
    at Function.createFromGenesis (document.js:42:1)
    at CeramicClient.createStreamFromGenesis (ceramic-http-client.js:40:1)
    at Function.createFromGenesis (tile-document.js:97:1)
    at TileLoader.deterministic (index.js:112:1)
    at async ThreeIDX.loadAuthLink (three-idx.js:122:1)
    at async ThreeIDX.loadIDX (three-idx.js:164:1)
```

The issue is that chrome extension doesn't provide `XMLHttpRequest` access within background script. Only **fetch** API is available

PS:
Any way to make `cross-fetch`optional ?

### DIFF
--- a/packages/common/src/utils/http-utils.ts
+++ b/packages/common/src/utils/http-utils.ts
@@ -1,4 +1,3 @@
-import fetch from 'cross-fetch'
 import { mergeAbortSignals, TimedAbortSignal } from './abort-signal-utils.js'
 
 const DEFAULT_FETCH_TIMEOUT = 60 * 1000 * 3 // 3 minutes


ukstv:
Thank you for the report. Will handle that as soon as possible.

Directly relying on native fetch is not an option right now, unfortunately: Node.js v16 LTS does not have it. Nevertheless, we shall take care of this.

lsmod:
What about optional import: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import#importing_different_modules_based_on_environment

Something like:
```
let fetch;

if (process.env.CROSS_FETCH_DISABLED !== "true") {
  fetch = await import("cross-fetch");
}
```

ukstv:
Not going to merge that anyway until we finally move to Node 18 with native fetch.

PR #2274, opened 2022-06-27
feat: Add path expansion to config file handling

ukstv:
Handle:
- `~/` is replaced with `os.homedir()`,
- `./`, `../` are resolved against the config file path,
for:
- `config.logger.logDirectory`,
- `config.stateStore.localDirectory`.

### DIFF
--- a/packages/cli/src/__tests__/daemon-config.test.ts
+++ b/packages/cli/src/__tests__/daemon-config.test.ts
@@ -2,6 +2,7 @@ import tmp from 'tmp-promise'
 import { writeFile } from 'node:fs/promises'
 import { DaemonConfig } from '../daemon-config'
 import { homedir } from 'node:os'
+import * as process from 'process'
 
 describe('reading from file', () => {
   let folder: tmp.DirectoryResult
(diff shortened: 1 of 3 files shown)


packages/cli/src/__tests__/daemon-config.test.ts
```diff
@@ -0,0 +1,52 @@
...
+        'local-directory': '~/statestore',
+      },
+    }
+    await writeFile(configFilepath, JSON.stringify(config))
+    const read = await DaemonConfig.fromFile(configFilepath)
+    const home = new URL(`file://${homedir()}/`)
+    expect(read.logger.logDirectory).toEqual(new URL('./log-dir', home).pathname)
+    expect(read.stateStore.localDirectory).toEqual(new URL('./statestore', home).pathname)
+  })
+})
```
stbrody, line 52: Can we also have a test that absolute paths work as well, for a path outside of the config directory?

packages/cli/src/daemon-config.ts
```diff
@@ -1,6 +1,46 @@
...
+import { readFile } from 'node:fs/promises'
+import { homedir } from 'os'
+
+const HOME_DIR = `${homedir()}/`.replace(/\/\/$/, '')
+
+/**
+ * Replace `~/` with `<homedir>/` absolute path.
+ * @param input Relative path.
+ */
+function expandHomedir(input: string): string {
```
stbrody, line 13: I'm really surprised we have to write this ourselves?  Is there really not a library that would handle this for us, in a platform agnostic way?

packages/cli/src/__tests__/daemon-config.test.ts
```diff
@@ -0,0 +1,52 @@
...
+import { writeFile } from 'node:fs/promises'
+import { DaemonConfig } from '../daemon-config'
+import { homedir } from 'node:os'
+
+describe('reading from file', () => {
+  let folder: tmp.DirectoryResult
+  let configFilepath: URL
+  beforeEach(async () => {
+    folder = await tmp.dir({ unsafeCleanup: true })
+    const base = new URL(`file://${folder.path}/`)
```
gvelez17, line 11: I thought file: urls had three slashes ///

https://en.wikipedia.org/wiki/File_URI_scheme#How_many_slashes?

packages/cli/src/daemon-config.ts
gvelez17: if they do it bare on the command line the shell should do it for us, but if they quote it prob not - normally I think cli's don't expect to get '~' literals, the shell expands those?  but users might do it anyway.

packages/cli/src/__tests__/daemon-config.test.ts
ukstv: It is three slashes indeed. Two are here in `file://` string, and another one is provided by `folder.path`, as it is an absolute path starting with `/`.

packages/cli/src/__tests__/daemon-config.test.ts
ukstv: added.

PR #2276, opened 2022-06-28
chore: Bump Jest to v28

ukstv:
- Bump Jest to v28
- Move anchorUpdate helper from Core to common/TestUtils: it is needed for canary-tests
- Add zondax stuff to moduleNameMapper: handle long-not-updated filecoin packages.


packages/common/src/utils/test-utils.ts
```diff
@@ -65,4 +67,15 @@ export class TestUtils {
     ])
     return CID.create(1, SHA256_CODE, decodeMultiHash(body))
   }
+
+  static async anchorUpdate(ceramic: CeramicApi, stream: Stream): Promise<void> {
```
stbrody, line 71:
```suggestion
  static async anchorUpdate(ceramic: Ceramic, stream: Stream): Promise<void> {
```

stbrody: This code actually won't work if it's given a `CeramicClient` it has to a be a Ceramic 'core' instance to be able to access the anchorService

packages/common/package.json
```diff
@@ -28,7 +28,7 @@
     "./lib/loggers.js": "./lib/loggers-browser.js"
   },
   "scripts": {
-    "test": "NODE_OPTIONS=--experimental-vm-modules npx jest --silent --coverage --passWithNoTests --forceExit",
+    "test": "NODE_OPTIONS=--experimental-vm-modules npx jest --silent --coverage --passWithNoTests",
     "build": "npx tsc --project tsconfig.json",
```
stbrody, line 32: This PR adds a lot of calls to Ceramic core code into TestUtils in the `common` package.  Do we need to add devDependencies on the 'core' package?

stbrody (approved):
lgtm mod my question about dependencies for the `common` package.

packages/common/src/utils/test-utils.ts
ukstv: Better not create a dependency on Core, as it would create a circular dependency. Added a doc-comment with a warning.

PR #2278, opened 2022-06-28
feat: Remove AbortController polyfill

ukstv:
- Remove AbortController polyfill. It is not needed anymore: Vercel supports Node v16 with a native AbortController. CC @PaulLeCam 
- Eliminate source of memory leak related to retained event handlers on AbortSignal's inside ipfs codebase.
- Update TypeDoc and TypeScript packages.

One of the peculiar things that I noticed while exploring GitCoin heapdump, was immense amount of AbortController and AbortSignal instances. Digging deeper into that, I discovered, that ipfs-http-client sets an event listener to an AbortSignal but never deletes it. This creates an obvious memory leak.

Screenshot 2022-06-28 at 16 13 37

This PR introduces a wrapper around a function (referenced as `fn` below) that uses AbortSignal (by attaching an event handler to it) and returns Promise. It is called `abortable`. I have put it to `common` package. `abortable` creates an intermediate AbortController. When a promise returned from `fn` is resolved, the event handler is cleared, eliminating the memory leak.

To test locally, I loaded five streams and created a heapdump.

Before (i.e. the current upstream):
Screenshot 2022-06-28 at 16 10 25

After (on the PR branch):
Screenshot 2022-06-28 at 16 10 08


ukstv:
There is still a MaxListenersExceededWarning you could witness. It is due to excessive concurrent handlers attached to a shutdownAbortController. Not a big deal, as we clear all the handlers eventually. Nevertheless, it is annoying, so I am considering replacing the global shutdownAbortController with an observable.

packages/common/src/utils/abort-signal-utils.ts
```diff
@@ -46,3 +43,26 @@ export class TimedAbortSignal {
     this._subscription?.unsubscribe()
   }
 }
+
+/**
+ * Call a function with abort signal and clear the memory.
+ *
+ * Some functions do not clear a signal listener after successful execution. Here we make sure
```
stbrody, line 50:
```suggestion
 * Some functions do not clear a signal listener after successful execution. By wrapping the original AbortSignal (which may be long-lived) in a temporary AbortSignal that we can throw out when `fn` completes, we make sure
```

stbrody (approved):
lgtm with one comment suggestion

PR #2280, opened 2022-06-28
feat(core,model-handler,model-instance-handler): Disable indexing and query features by default until they are ready

stbrody:
This PR puts all indexing and query functionality behind an environment variable flag guard temporarily.  If users discover the new functionality and start creating models or MIDs on mainnet, then any breaking changes we make to those streamtypes as part of testing and finalizing the release would break any existing streams of those streamtypes on the network.  Worse, even if we don't make breaking changes, any MID or Model streams created on mainnet before the necessary changes are in place on the CAS to enable indexing from anchors would never get picked up by indexers performing a historical sync.  We shouldn't have any Model or MID streams created on mainnet until the anchors being performed on mainnet are indexable.

As soon as we're ready to release indexing for real, we should remove this environment variable entirely.

### DIFF
--- a/packages/core/src/indexing/local-index-api.ts
+++ b/packages/core/src/indexing/local-index-api.ts
@@ -22,6 +22,10 @@ export class LocalIndexApi implements IndexApi {
   ) {}
 
   private _shouldIndexStream(args: StreamID): Boolean {
+    if (!this.databaseIndexApi) {
+      return false
+    }
+
     return this.databaseIndexApi.getActiveModelsToIndex().some(function (streamId) {
       return String(streamId) === String(args)
     })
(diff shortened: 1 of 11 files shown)


stbrody:
Assuming this looks good, please merge while I'm OOO so it will get picked up in the next release.  I'm already a bit worried about overzealous community members discovering the indexing functionality in the release that's already out there and trying to use it before its ready

packages/core/src/indexing/local-index-api.ts
```diff
@@ -37,7 +41,7 @@ export class LocalIndexApi implements IndexApi {
     if (!this._shouldIndexStream(args.model)) {
       return
     }
-    this.databaseIndexApi.indexStream(args)
+    await this.databaseIndexApi.indexStream(args)
```
willex, line 44: good catch

ukstv (approved):

PR #2286, opened 2022-06-29
feat: Add a method to CeramicAPI that transforms raw StreamState to an instance of Streamtype

ukstv:
`ceramic.buildStreamFromState(streamState)` builds an instance of a corresponding Streamtype based on the StreamState.
Works on Core and Client.

@pawartur @PaulLeCam


PaulLeCam (approved):
Thanks!

PR #2287, opened 2022-06-30
feat: Store first anchored time in the indexing database

ukstv:
It is a breaking change, but we are free _now_ to do this.


willex (approved):
lgtm - as you mentioned this introduces breaking changes to the current postgres integration for immediate follow up.

PR #2288, opened 2022-06-30
feat(core): Postgres MID table creation and indexing

willex:
## Description

Integration of Postgres MID table creation, indexStream functionality & self contained unit tests.

**This PR also addresses**:
NET-1625: Add contained Postgres unit test capability
NET-1518: Implement IndexStream for Postgres
Add edge cursors and use expected order (same as sqlite)

## How Has This Been Tested?

- [x] Ran create stream locally against Node with locally hosted Postgres instance
- [x] Tested unit tests locally which defaults to `DATABASE_URL` and can used by the CI/CD pipeline

NOTE: following up with Mohsin to see why CircleCLI fails

## PR checklist

## References:

New npm test package (requires Docker): https://www.atdatabases.org/docs/pg-test


packages/core/src/indexing/build-indexing.ts
```diff
@@ -53,6 +54,14 @@ export function buildIndexing(indexingConfig: IndexingConfig): DatabaseIndexApi
       })
       return new SqliteIndexApi(dbConnection, indexingConfig.models)
     }
+    case 'postgres': {
+      const dataSource = knex({
+        client: 'pg',
+        connection: connectionString.toString(),
+        searchPath: ['ceramic', 'public'],
```
ukstv, line 61: @willex you sure we need this flexibility?

willex: yeah, this can be removed.

willex:
@ukstv I'm looking into updating the PR with the ordering changes right now too

packages/core/src/indexing/postgres/__tests__/postgres-index-api.test.ts
```diff
@@ -42,21 +42,22 @@ export async function listMidTables(dbConnection: Knex) {
...
-
 describe('init', () => {
   describe('create tables', () => {
     test('create new table from scratch', async () => {
       const modelsToIndex = [StreamID.fromString(STREAM_ID_A)]
       const indexApi = new PostgresIndexApi(dbConnection, modelsToIndex)
       await indexApi.init()
-      const created = await listMidTables(dbConnection)
+      const created = await listMidTables()
       const tableNames = modelsToIndex.map((m) => `${m.toString()}`)
```
ukstv, line 61:
```suggestion
      const tableNames = modelsToIndex.map(String)
```

ukstv, line 61:
```suggestion
      const tableNames = modelsToIndex.map(String)
```

packages/core/src/indexing/postgres/__tests__/postgres-index-api.test.ts
```diff
@@ -0,0 +1,158 @@
...
+  await dbConnection.destroy()
+})
+
+export async function listMidTables() {
+  const dataArr = []
+  await dbConnection
+    .select('tablename')
+    .from('pg_tables')
+    .whereRaw("schemaname='public' AND tablename LIKE ('kjz%')")
+    .then(function (result) {
```
ukstv, line 37: @willex Now it is all right, but going forward, please, lean towards either async/await or `.then` style. async/await is preferred.

packages/core/src/indexing/postgres/postgres-index-api.ts
```diff
@@ -0,0 +1,54 @@
...
+  constructor(private readonly dbConnection: Knex, readonly modelsToIndex: Array<StreamID>) {
+    this.insertionOrder = new InsertionOrder(dbConnection)
+  }
+
+  public getActiveModelsToIndex(): Array<StreamID> {
+    /**
+     * Helper function to return array of active models that are currently being indexed by node
+     * as defined in the config file.
+     * TODO: extend to runtime check once adminAPI unlocks to add and load models on the fly
+     */
```
ukstv, line 20: When you do `TODO` like this, there has to be a corresponding issue on Linear.

`TODO(NET-XXX)` would then allow us to recover context and act on it.

packages/core/src/indexing/postgres/postgres-index-api.ts
```diff
@@ -0,0 +1,54 @@
...
+  async indexStream(args: IndexStreamArgs & { createdAt?: Date; updatedAt?: Date }): Promise<void> {
+    const tableName = asTableName(args.model)
+
+    // created_at and last_updated_at set by default default value
+    await this.dbConnection(tableName)
+      .insert({
+        stream_id: String(args.streamID),
+        controller_did: String(args.controller),
+        last_anchored_at: args.lastAnchor,
+        created_at: args.createdAt || new Date(),
```
ukstv, line 33: It is postgres. Do you know if we can use [`NOW()`](https://www.postgresql.org/docs/current/functions-datetime.html) here?

packages/core/src/indexing/postgres/__tests__/postgres-index-api.test.ts
ukstv:
```suggestion
      const tableNames = modelsToIndex.map(asTableName)
```

Better be consistent.

ukstv (approved):
LGTM for now to unblock us.

Few actions before the merge:
1. Create an issue for the TODO comment. This one should go to "Indexing Post-MVP" project.
2. Create an issue in the Query Execution project to DRY the code. SQLite and Postgres are very similar, and beg for deduplication.

And after the merge:
1. Please, could you figure out if we could use native pg `now()`? IMO this should be a separate issue on Query Execution Project.

packages/core/src/indexing/postgres/__tests__/postgres-index-api.test.ts
willex: done!

packages/core/src/indexing/postgres/postgres-index-api.ts
willex: yes. Updated with `this.dbConnection.fn.now()`. This was already used in the mid table creation logic as default value.

PR #2297, opened 2022-07-04
chore: Beautify index api test

ukstv:
- Use Ceramic Client instead of fetchJSON
- Cover http-to-indexAPI flow.


willex (approved):
lgtm - awesome cleanup!

PR #2306, opened 2022-07-07
feat(stream-model-handler,stream-model-instance-handler): Remove state.next from Model and MID StreamState

stbrody:


oed (commented):
There are no implications for conflict resolution? We don't use `.next` at all there?

stbrody:
> There are no implications for conflict resolution? We don't use `.next` at all there?

Nope!  Conflict resolution is all based on the commit log, we don't use any other part of the streamstate

ukstv (approved):
🏇

PR #2307, opened 2022-07-07
fix(core): Don't unpin anchor proof, merkle tree, or CACAO when unpinning streams

stbrody:
This does technically leave some leaked CIDs pinned in IPFS after unpinning a stream, but that seems better to me than accidentally unpinning data that might be in use by other streams.

This PR also dedupes some code and improves the test coverage of the pin store


packages/core/src/store/__tests__/pin-store.test.ts
```diff
@@ -229,13 +273,96 @@ describe('#add', () => {
...
+    expect(pinning.unpin).toBeCalledTimes(4)
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    expect(pinning.unpin.mock.calls[0][0].toString()).toEqual(
+      stateWithSignedCommit.log[0].cid.toString()
+    )
+    expect(pinning.unpin.mock.calls[1][0].toString()).toEqual(
+      stateWithSignedCommit.log[1].cid.toString()
+    )
+    expect(pinning.unpin.mock.calls[2][0].toString()).toEqual(capCID.toString())
```
ukstv, line 322: So, we unpin a capability, even if it can be used by some other stream?

ukstv (changes requested):
LGTM for the change in the description.

Not quite relevant: Do you think we can leave a capability pinned? There is no guarantee a capability is issued just for a particular stream. Most often it is reused across many streams.

stbrody:
> Do you think we can leave a capability pinned? There is no guarantee a capability is issued just for a particular stream. Most often it is reused across many streams.

Ooh really good catch.  Yeah I'll update this to leave the CACAO alone during unpin, just like I do with the anchor proof and merkle tree

ukstv (approved):
🏇

PR #2316, opened 2022-07-11
chore: Add missing packages

ukstv:


ukstv:
@stbrody The change originates from @PaulLeCam 's finding of missing dependencies when using PNPM for js-composedb.

PNPM does not hoist dependencies, i.e. it does not flatten the dependency graph into a single node_modules folder. The folder layout it makes directly reflects the dependency graph.

Let's say package `root` depends on package `b`, which depends on package `c`. NPM and Lerna would create a single `node_modules` folder with `b` and `c`. PNPM instead would create `node_modules/b` and `node_modules/b/node_modules/c` with `b` and `c` packages correspondingly.

The PR essentially is a side-effect of temporarily using PNPM instead of NPM+Lerna for the repository. I made PNPM install all the dependencies, and then tried to build the packages. If a dependency is missing, the build fails. With the changes here, `js-ceramic` builds successfully with PNPM-managed dependency.

PR #2317, opened 2022-07-11
feat(core): Add stream from pubsub for UPDATE msg types

willex:
## Description

This adds that UPDATE messages with model property will be added via `IndexAPI::indexStream` with appropriate properties. This includes the discussed refactoring of the original implementation in `ceramic.ts` as well covering:

- moved indexStreamIfNeeded to repository.ts
- Refactored `repository.ts` and `ceramic.ts` to move the `indexStreamIfNeeded` function & switched over from `stream` to `state` as input value
- updated state-manager.ts and added `pinned==true` check as part of `state-manager::_updateStateIfPinnedOrIndexed`

**NOTE**: This still containst some console.log messages for debugging. Overall, I would like to have some eyes on it if this makes sense - especially around the refactor and the dependency inheritance between `repository`, `ceramic` & `state-manager` (@stbrody / @ukstv)

## How Has This Been Tested?

- [ ] Test B

## PR checklist

## References:


packages/core/src/state-management/repository.ts
```diff
@@ -97,7 +103,9 @@ export class Repository {
       deps.conflictResolution,
       this.logger,
       (streamId) => this.fromMemoryOrStore(streamId),
-      (streamId, opts) => this.load(streamId, opts)
+      (streamId, opts) => this.load(streamId, opts),
+      this.indexStreamIfNeeded,
```
stbrody, line 107: I want to moving out of the habit of the state manager calling back up into the Repository.  That is a circular dependency nightmare IMO.  This seems to be making the problem worse, not better

packages/core/src/state-management/repository.ts
```diff
@@ -330,6 +338,36 @@ export class Repository {
     return this.#deps.pinStore.stateStore.load(StreamID.fromString(streamID))
   }
 
+  /**
+   * Helper function to add stream to db index if it has a 'model' in its metadata.
+   * @param state
+   * @public
+   */
+  public async indexStreamIfNeeded(state: RunningState): Promise<void> {
```
stbrody, line 346: convention is to use `state$` for `RunningState`, and just `state` for `StreamState`

packages/core/src/state-management/state-manager.ts
```diff
@@ -54,7 +56,9 @@ export class StateManager {
     private readonly load: (
       streamId: StreamID,
       opts?: LoadOpts | CreateOpts
-    ) => Promise<RunningState>
+    ) => Promise<RunningState>,
+    private readonly indexStreamIfNeeded,
```
stbrody, line 60: this would need a type specifier, except that I want to get it out of here entirely

stbrody (changes requested):
man, the circular relationship between Repository and StateManager really makes me sad.  I think the simplest way to break out of it for now is to push the main implementation of `indexStreamIfNeeded` down into the StateManager.  That would mean giving the StateManager a handle to the IndexApi, which I'd rather not do as it seems like that really does belong to the Repository, but that seems the lesser evil to me then having a callback back into the Repository.

All external calls from other parts of the system (like the dispatcher) should still go through the Repository though, so that eventually the StateManager can be an entirely hidden implementation detail of the Repository.  The repository can have a simple wrapper around whatever StateManager functionality needs to be exposed.

packages/core/src/ceramic.ts
```diff
@@ -224,15 +224,17 @@ export class Ceramic implements CeramicApi {
...
       dispatcher: this.dispatcher,
       pinStore: pinStore,
       context: this.context,
       handlers: this._streamHandlers,
       anchorService: modules.anchorService,
       conflictResolution: conflictResolution,
+      indexing: localIndex,
     })
-    this._index = new LocalIndexApi(modules.indexing, this.repository, this._logger)
+    this._index = localIndex //new LocalIndexApi(modules.indexing, this.repository, this._logger)
```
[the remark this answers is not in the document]
willex: Looking at it, It should be able to go through the repository now.. but I think needs to be initialized there then if I'm not mistaken.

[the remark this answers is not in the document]
stbrody: That seems fine.  We can add an init() method to the Repository if we need

packages/core/src/state-management/repository.ts
willex: done.

packages/core/src/ceramic.ts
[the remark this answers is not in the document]
willex: `IndexApi` removed completely now from` Ceramic`.

packages/core/src/state-management/state-manager.ts
willex: done - removed.

packages/core/src/ceramic.ts
```diff
@@ -224,15 +224,17 @@ export class Ceramic implements CeramicApi {
...
       dispatcher: this.dispatcher,
       pinStore: pinStore,
       context: this.context,
       handlers: this._streamHandlers,
       anchorService: modules.anchorService,
       conflictResolution: conflictResolution,
+      indexing: localIndex,
     })
-    this._index = new LocalIndexApi(modules.indexing, this.repository, this._logger)
+    this._index = localIndex
```
stbrody, line 237: are you going to get rid of this in a follow-up PR?  If so can you add a TODO with the ticket number?

packages/core/src/state-management/repository.ts
stbrody: So I guess the plan is to fix this in the follow-up PR?  Can you add a TODO comment?

packages/core/src/state-management/repository.ts
stbrody: ping

packages/core/src/state-management/state-manager.ts
```diff
@@ -172,22 +176,27 @@ export class StateManager {
...
     } else {
       return false
     }
   }
 
-  private async _updateStateIfPinned(state$: RunningState): Promise<void> {
+  private async _updateStateIfPinnedOrIndexed(state$: RunningState): Promise<void> {
     const isPinned = Boolean(await this.pinStore.stateStore.load(state$.id))
+    console.log('>>>>>>>> _updateStateIfPinnedOrIndexed')
     if (isPinned) {
```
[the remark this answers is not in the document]
stbrody: ping.  We need to make sure that streams that are in a model that is configured to be indexed wind up getting pinned, even if they aren't currently explicitly pinned already

packages/core/src/state-management/state-manager.ts
```diff
@@ -200,16 +205,25 @@ export class StateManager {
...
    * @param streamId
    * @param tip - Stream Tip CID
-   * @private
+   * @param model - Model Stream ID
    */
-  handlePubsubUpdate(streamId: StreamID, tip: CID): void {
-    this.fromMemoryOrStore(streamId).then((state$) => {
+  async handlePubsubUpdate(streamId: StreamID, tip: CID, model?: StreamID): Promise<void> {
+    if (model && this._index.shouldIndexStream(model)) {
+      const inPinStore = await this.pinStore.ls(streamId)
```
stbrody, line 212: this results in going to the state store twice, since it also happens below in `fromMemoryOrStore`.  You should be able to combine the logic

stbrody: I think you should be able to do something like:
```
let state$ = await this.fromMemoryOrStore(streamId)
const shouldIndex = model && this._index.shouldIndexStream(model)
if (!shouldIndex && !state$) {
  // stream isn't pinned or indexed, nothing to do
  return
}

if (!state$) {
  state$ = await this.load(streamId)
}
this.executionQ.forStream(streamId).add(async() => {
  await this._handleTip(state$, tip)
}
this.indexStreamIfNeeded(state$)
```

stbrody (changes requested):

packages/core/src/state-management/repository.ts
willex: Yes, adding the TODOs right now to close this one out.

packages/core/src/state-management/repository.ts
willex: corrected.

packages/core/src/ceramic.ts
willex: correct, adding TODO right now.

packages/core/src/state-management/state-manager.ts
willex: Oh, I see what you mean and your example should work to address it the initial issue of it not being found. If possible, I'd love to have a follow up chat about this though to dive a bit deeper into the execution queue. I was under the impression that we still have to add it first to the state store for it then to be properly processed.

packages/core/src/state-management/state-manager.ts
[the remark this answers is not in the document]
willex: Added an additional check but will need to clean that one up further (TODO added).

stbrody (approved):

PR #2319, opened 2022-07-12
feat(core): Pinning a ModelInstanceDocument should also pin its Model

stbrody:
I had really been hoping on using this to make stream pinning behavior streamtype-specific so each StreamHandler could specify its own behavior for when streams of that type get pinned. That way the model-specific pinning logic wouldn't have to live in a generic codepath that applies to all streamtypes. But doing that would have been a much bigger refactor and so in the interest of time and simplicity this seemed like the more straightforward change that gets us what we need.

### DIFF
--- a/packages/core/src/store/__tests__/state-store-integration.test.ts
+++ b/packages/core/src/store/__tests__/state-store-integration.test.ts
@@ -16,6 +16,7 @@ import { CID } from 'multiformats/cid'
 import { createIPFS } from '@ceramicnetwork/ipfs-daemon'
 import { createCeramic } from '../../__tests__/create-ceramic.js'
 import { RunningState } from '../../state-management/running-state.js'
+import { Repository } from '../../state-management/repository.js'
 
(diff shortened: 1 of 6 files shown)


packages/core/src/store/__tests__/state-store-integration.test.ts
```diff
@@ -16,6 +16,7 @@ import { CID } from 'multiformats/cid'
 import { createIPFS } from '@ceramicnetwork/ipfs-daemon'
 import { createCeramic } from '../../__tests__/create-ceramic.js'
 import { RunningState } from '../../state-management/running-state.js'
+import { Repository } from '../../state-management/repository'
```
ukstv, line 19:
```suggestion
import { Repository } from '../../state-management/repository.js'
```

ukstv (approved):

packages/core/src/store/__tests__/state-store-integration.test.ts
stbrody: ugh good catch.  I wonder why the docker build CI step didn't catch this this time.

I also wonder if there's a way I can configure Webstorm to always add the `.js` by default...

PR #2322, opened 2022-07-13
feat(stream-model-instance-handler): Enable soft-delete

pawartur:
## Description

Enable soft-delete of MIDs by setting their content to `null` or `{}` and skipping attached model's schema validation in such case.

## How Has This Been Tested?

- [X] Added relevant tests

## PR checklist


packages/stream-model-instance-handler/src/model-instance-document-handler.ts
```diff
@@ -204,6 +204,10 @@ export class ModelInstanceDocumentHandler implements StreamHandler<ModelInstance
    * @private
    */
   async _validateContent(context: Context, modelStreamId: StreamID, content: any): Promise<void> {
+    if (isEmpty(content)) {
+      // We leave the possibility to set content to null/{} as a way of 'soft' deletion of an MID
+      return Promise.resolve()
```
ukstv, line 209:
```suggestion
      return
```

ukstv: @pawartur does it mean that you could do "soft-delete", and then add content? Is it intended?

ukstv (changes requested):

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
pawartur: Good question! I assumed that this is what we meant when we were discussing it.

Is there a way that we can soft delete and prevent any following content changes from that point?

willex (approved):

packages/stream-model-instance-handler/src/__tests__/model-instance-document-handler.test.ts
```diff
@@ -298,6 +298,18 @@ describe('ModelInstanceDocumentHandler', () => {
     await checkSignedCommitMatchesExpectations(did, commit, expectedGenesis)
   })
 
+  it('null content in genesis is valid', async () => {
```
stbrody, line 301: what's the use case for allowing null content in genesis?

packages/stream-model-instance-handler/src/schema-utils.ts
```diff
@@ -1,6 +1,12 @@
 import ajv, { SchemaObject } from 'ajv/dist/2020.js'
 import addFormats from 'ajv-formats'
 
+export function isEmpty(obj: any): boolean {
```
stbrody, line 4: this would return `false` if given `null` - is that the desired behavior?

stbrody: I actually feel like we **shouldn't** allow setting to an empty document `{}` but only allow setting to an explicit null.  That is a more clear signal that the goal is to delete the stream.  Allowing an empty document could let through an application programming bug where the content object is built programmatically but winds up with no fields set.  Making the app explicitly set null means its less likely to happen by accident.

stbrody (changes requested):
If the contents of a MID get set to null, we should probably take that as a sign to remove the document from the index as well so it doesn't show up in queries.

Also, rather than allowing setting content to `null` directly using the `replace()` method, I wonder if we should have a separate function in the MID API for doing these soft-deletes.  Would make things more clear and explicit.  The hard part of course is figuring out what we would name it.  `advisoryDelete()`?  `unIndex()`?

CC @oed @PaulLeCam

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
stbrody: We certainly could change the commit handler to not allow setting contents if the current contents are null, but I'm not sure that we actually want to do that?

pawartur:
Closing this PR. We've decided with @stbrody and @PaulLeCam that this needs to be properly designed and discussed before implementation, specifically with regard to how removing a soft-deleted MID from index dbs should work. We'll come back to this when we have time to properly design it.

PR #2324, opened 2022-07-14
feat(cli): use env vars to override config file

v-stickykeys:
## Description

- Uses environment variables to override config file defaults (note: the hierarchy here is cli flags > env vars > config file > defaults)
- Adds env var for indexing db uri

## How Has This Been Tested?

- [x] ran js-ceramic locally and configured indexing with env var


v-stickykeys:
cc @gvelez17 note environment variable name change for metrics. reason: we need to namespace Ceramic specific env vars to avoid conflicts with other processes running in the environment

packages/cli/src/ceramic-cli-utils.ts
```diff
@@ -117,8 +117,13 @@ export class CeramicCliUtils {
       : DEFAULT_DAEMON_CONFIG_FILENAME
     const config = await this._loadDaemonConfig(configFilepath)
 
+    // Environment variables override values from config file
+    if (envVarExists(process.env.CERAMIC_INDEXING_DB_URI)) config.indexing.db = process.env.CERAMIC_INDEXING_DB_URI
```
ukstv, line 121: This could be simplified to just
```suggestion
    if (process.env.CERAMIC_INDEXING_DB_URI) config.indexing.db = process.env.CERAMIC_INDEXING_DB_URI
```

An empty string "", and `undefined` are both coerced to a boolean `false`.

ukstv (changes requested):
IMO, `envVarExists` is redundant.

ukstv (approved):
🏇

PR #2333, opened 2022-07-21
feat (cli, metrics): be able to see at a glance which metrics are from which sources - #PLAT-774

gvelez17:
## Description

Since our prometheus will scrape metrics from multiple sources it is desirable to namespace the sources, both for context of where the metrics are coming from and to prevent stepping on each other. 

This PR is to accept an optional parameter for caller name and prepend it to metric names 

## How Has This Been Tested?

- [ x] tests pass, changes are exercised

## PR checklist

## References:

(will be documenting once we have approved and tested the metrics, want to finalize the methods first)


stbrody (approved):
LGTM except the commit message should include the package names that are updated, and should not reference linear tickets as those are private.  So: `feat(cli,metrics): be able to see at a glance which metrics are from which sources`

gvelez17:
hey @stbrody thx for review, I forgot actually to increment the package version it depends on - since this both updates the metrics package and the code depending on it, is there any order we have to do when merging & deploying?  Should we do the new npm package release first?

packages/metrics/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@ceramicnetwork/metrics",
-  "version": "0.0.3-rc.1",
```
stbrody, line 3: this should probably be reverted also

gvelez17: yep done

packages/metrics/src/metrics-setup.ts
```diff
@@ -66,7 +72,7 @@ class _Metrics {
     }
     // Create this counter if we have not already
     if (!(name in this.counters)) {
-      this.counters[name] = this.meter.createCounter(name)
+      this.counters[name] = this.meter.createCounter('${caller}:${name}')
```
stbrody, line 75: this.caller?

packages/metrics/src/metrics-setup.ts
```diff
@@ -79,7 +85,7 @@ class _Metrics {
     }
     // Create this Histogram if we have not already
     if (!(name in this.histograms)) {
-      this.histograms[name] = this.meter.createHistogram(name)
+      this.histograms[name] = this.meter.createHistogram('${caller}:${name}')
```
stbrody, line 88: this.caller?

packages/metrics/src/metrics-setup.ts
gvelez17: ah!  yes. fixed

stbrody (approved):
lgtm

PR #2339, opened 2022-07-26
feat(cli): Enable ceramic --version flag

ukstv:


v-stickykeys (approved):
woohoo! 🚀

packages/cli/src/bin/ceramic.ts
```diff
@@ -2,8 +2,10 @@ import { program } from 'commander'
 import pc from 'picocolors'
 
 import { CeramicCliUtils } from '../ceramic-cli-utils.js'
+import { version } from '../version.js'
 
 program
+  .version(version, '-v, -V, --version', 'output the version number')
```
stbrody, line 8: don't we need code in this file to actually print the version when the flag is passed?

ukstv: We don't. It is handled by commander for us.

stbrody (approved):
neat!

PR #2340, opened 2022-07-26
fix(core): Detect model model index table and don't recreate

stbrody:


packages/core/src/indexing/postgres/__tests__/postgres-index-api.test.ts
```diff
@@ -51,6 +53,16 @@ describe('init', () => {
       expect(created).toEqual(tableNames)
     })
 
+    test('table creation is idempotent', async () => {
+      const modelsToIndex = [Model.MODEL, StreamID.fromString(STREAM_ID_A)]
+      const indexApi = new PostgresIndexApi(dbConnection, modelsToIndex, true)
+      await indexApi.init()
+      await indexApi.init()
```
willex, line 60: duplication of init.

stbrody: that's intentional, it's checking that you can init twice without an error.  The double-init is the main part of the test

stbrody: added a comment

PR #2342, opened 2022-07-29
fix(core): Add undefined type to did setter

plesastapevka:
I was trying to run the code inside the MetaMask Snaps system, and it didn't want to compile. After adding an optional undefined type, as stated in the diff, the code runs.

## Description

There has not been any open issue regarding this. It seems to happen in MetaMask Snaps only, which run in a separate environment from MetaMask extension and browser. It works as expected when running the code in the browser or node.js environment.

The bug shows already when connecting to the Clay Ceramic Testnet.

## How Has This Been Tested?

- [x] Tried to compile the MetaMask Snap code after applying the change

To reproduce:
Implement the connection in MetaMask Snaps as in the [tutorial here](https://developers.ceramic.network/tools/self-id/write/#using-the-selfid-class). When compiling the MetaMask Snap, the error fires.

## PR checklist

### DIFF
--- a/packages/http-client/src/ceramic-http-client.ts
+++ b/packages/http-client/src/ceramic-http-client.ts
@@ -107,7 +107,7 @@ export class CeramicClient implements CeramicApi {
    * Sets the DID instance that will be used to author commits to streams.
    * @param did
    */
-  set did(did: DID) {
+  set did(did: DID | undefined) {
     this.context.did = did
   }
 


stbrody:
What's the use case for setting the DID instance on a CeramicClient to `undefined`?

stbrody:
I guess if this was failing to compile with MetaMask Snaps, that probably means that MetaMask Snaps is trying to set the DID to undefined for some reason?  Do you know why it is doing that?

plesastapevka:
I have no use case for setting the DID instance to `undefined`, and however, as stated above, when trying to use the library inside of MetaMask Snap environment, it doesn't compile. What exactly happens and why the issue persists, I have no clue.

stbrody:
> I have no use case for setting the DID instance to `undefined`, and however, as stated above, when trying to use the library inside of MetaMask Snap environment, it doesn't compile. What exactly happens and why the issue persists, I have no clue.

I feel like rather than merge this PR and loosen the API, losing some of the advantage of the type system, I think instead we should try to understand why MetamaskSnap seems to be trying to set this to undefined in the first place.  This seems like a bug with MetaMask Snap abusing our API, not a problem with our API itself

PR #2347, opened 2022-08-01
feat(cli): Remove support for did:safe and did:nft

stbrody:
They've always been marked as experimental, have never been safe to use in production, we aren't planning on investing more work into supporting them, and even if we wanted to we aren't actually sure what the path towards making them safe would even be.

CC @ukstv @3avi @LFeld

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -58,13 +58,11 @@
     "levelup": "^5.1.1",
     "merge-options": "^3.0.4",
     "morgan": "^1.10.0",
-    "nft-did-resolver": "^2.0.0",
     "p-queue": "^7.2.0",
     "picocolors": "^1.0.0",
     "pkh-did-resolver": "^1.0.6-rc.0",
     "reflect-metadata": "^0.1.13",
     "s3leveldown": "^2.2.2",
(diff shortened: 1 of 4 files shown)


ukstv:
Sad, but makes sense.

oed:
Not sure it really make sense to remove them from the code base? Can't we keep them as long as we are clear about them being experimental?

stbrody:
@oed as long as we have them around we will continue to get questions about them, and people will continue to want to use them in production.  This will cause a constant low-level drag on our team to be always having to answer these questions and explain why these methods are not supported in Ceramic.
I really don't think it makes sense to keep around a feature that isn't safe to use and that we have no plans now or ever to fix.  It just increases the complexity and surface area of Ceramic for something that we don't even want to support.

qbig:
@stbrody @oed Could you elaborate how they are not secure?

3avi:
> @stbrody @oed Could you elaborate how they are not secure?

hey Liang, can you ask this question on the forum https://forum.ceramic.network/ -- I'll look for your question there

stbrody:
closing in favor of https://github.com/ceramicnetwork/js-ceramic/pull/3021

PR #2349, opened 2022-08-02
fix(stream-model-instance-handler,streamid): Fix test snapshots with new Uint8Array format

stbrody:
It seems that the tests that explicitly call `expect().toMatchSnapshot` on a Uint8Array directly had the format change in the snapshot.  There are plenty of other test snapshots, however, that include Uint8Arrays in them but indirectly as part of nested object ([for example](https://github.com/ceramicnetwork/js-ceramic/blob/df527aba62e9205f8e8bf153b44d0a322e82f069/packages/stream-tile-handler/src/__tests__/__snapshots__/tile-document-handler.test.ts.snap#L19-L54)).  Not sure why those one's didn't need to be updated also...

Changes seems to have been introduced by the recent release of the `uint8arrays` package: https://github.com/achingbrain/uint8arrays/commit/56329d16d6ca575c9638f3abb9601b8a034783b8


stbrody:
> It seems that the tests that explicitly call expect().toMatchSnapshot on a Uint8Array directly had the format change in the snapshot. There are plenty of other test snapshots, however, that include Uint8Arrays in them but indirectly as part of nested object ([for example](https://github.com/ceramicnetwork/js-ceramic/blob/df527aba62e9205f8e8bf153b44d0a322e82f069/packages/stream-tile-handler/src/__tests__/__snapshots__/tile-document-handler.test.ts.snap#L19-L54)). Not sure why those one's didn't need to be updated also...

Any idea about why this is @ukstv?

ukstv:
> > It seems that the tests that explicitly call expect().toMatchSnapshot on a Uint8Array directly had the format change in the snapshot. There are plenty of other test snapshots, however, that include Uint8Arrays in them but indirectly as part of nested object ([for example](https://github.com/ceramicnetwork/js-ceramic/blob/df527aba62e9205f8e8bf153b44d0a322e82f069/packages/stream-tile-handler/src/__tests__/__snapshots__/tile-document-handler.test.ts.snap#L19-L54)). Not sure why those one's didn't need to be updated also...
> 
> Any idea about why this is @ukstv?

`uint8arrays` now tries to be smart about the underlying implementation, be it `Uint8Arrays` or nodejs `Buffer`. See https://github.com/achingbrain/uint8arrays/blob/master/src/to-string.js#L25 Whenever _we_ manipulate the bytes, on Node.js we should expect now an instance of `Buffer`. Hence the difference. 

The example you mention uses bytes from inside CID package, which is unaffected by `uint8arrays` smartness.

PR #2350, opened 2022-08-03
feature (core): add first_anchored add to Postgres Index

willex:
## Description

Adds additional column `first_anchored_at` to index Postgres integration for feature parity with SQLite. This updates the schema validation as well on startup and informs the developer that tables are being created in case new models have been added to the config.

NOTE: if you have an existing Postgres instance with a prior table structure, this will break node startup as the column does not exist in your database and schema verification will fail.

To remedy the issue stop the node and either:
1. DROP the specific indexing tables - the following SQL command will create the drop commands for all applicable tables

`SELECT 'DROP TABLE ' || TABLE_NAME || ' CASCADE;'
FROM INFORMATION_SCHEMA.TABLES
WHERE TABLE_NAME LIKE 'kjz%';`

2. DROP the whole database and re-create it from scratch

then restart the node and the new correct tables will be created.

## How Has This Been Tested?

- [x] verified locally during DB startup
- [x] verified via unit tests

## PR checklist

## References:


packages/core/src/indexing/postgres/migrations/1-create-model-table.ts
```diff
@@ -1,13 +1,15 @@
 import type { Knex } from 'knex'
 
 export async function createModelTable(dataSource: Knex, tableName: string) {
+  console.log(`Creating ComposeDB Indexing table for model: ${tableName}`)
```
ukstv, line 4: Use logger, or let it remain silent.

packages/core/src/indexing/sqlite/migrations/1-create-model-table.ts
```diff
@@ -1,6 +1,7 @@
 import type { Knex } from 'knex'
 
 export async function createModelTable(dbConnection: Knex, tableName: string) {
+  console.log(`Creating ComposeDB Indexing table for model: ${tableName}`)
```
ukstv, line 4: Use logger, or let it remain silent.

packages/core/src/indexing/postgres/__tests__/postgres-index-api.test.ts
```diff
@@ -166,6 +167,7 @@ describe('indexStream', () => {
       ...STREAM_CONTENT,
       updatedAt: updateTime,
       lastAnchor: updateTime,
+      firstAnchor: null,
```
ukstv, line 170: Do you know if there is a follow up ticket that actually sets the `firstAnchor` value?

ukstv (changes requested):
console.log has no place in the code

packages/core/src/indexing/postgres/__tests__/postgres-index-api.test.ts
willex: Just checked and there is not - I can create one and do a follow up to this one or just add it to the current PR. I saw that it's currently also not set for sqlite.

ukstv: > Just checked and there is not - I can create one and do a follow up to this one or just add it to the current PR. I saw that it's currently also not set for sqlite.

Indeed. IMO, it should be a different slice.

packages/core/src/indexing/postgres/postgres-index-api.ts
```diff
@@ -13,7 +13,8 @@ export class PostgresIndexApi implements DatabaseIndexApi {
   constructor(
     private readonly dbConnection: Knex,
     readonly modelsToIndex: Array<StreamID>,
-    private readonly allowQueriesBeforeHistoricalSync: boolean
+    private readonly allowQueriesBeforeHistoricalSync: boolean,
+    private logger: DiagnosticsLogger
```
stbrody, line 17:
```suggestion
    private readonly logger: DiagnosticsLogger
```

stbrody (approved):
LGTM

packages/core/src/indexing/build-indexing.ts
```diff
@@ -45,11 +46,13 @@ function parseURL(input: string) {
  * Build DatabaseIndexAPI instance based on passed indexing configuration.
  */
 export function buildIndexing(indexingConfig: IndexingConfig): DatabaseIndexApi {
+  const diagnosticsLogger = new LoggerProvider().getDiagnosticsLogger()
```
ukstv, line 49: IMO, better to pass the `diagnosticsLogger` from a caller.

packages/core/src/indexing/build-indexing.ts
```diff
@@ -45,11 +46,13 @@ function parseURL(input: string) {
...
   const connectionString = parseURL(indexingConfig.db)
   const protocol = connectionString.protocol.replace(/:$/, '')
   switch (protocol) {
     case 'sqlite':
     case 'sqlite3': {
+      diagnosticsLogger.imp('Initializing SQLite connection')
```
ukstv, line 55: nice

ukstv (approved):
🏇 mod logger like you do in https://github.com/ceramicnetwork/js-ceramic/pull/2350/files#diff-4f310bc7966d14e6e2a9f3b95a833505580949d821a1e30fa3253134e098e679R26

packages/core/src/indexing/build-indexing.ts
willex: good catch, added it since it was already available from within the caller.

PR #2355, opened 2022-08-04
feat(key-did-resolver): Use upstream bigint-mod-arith package

ukstv:
There was a problem with bigint-mod-arith package in the past. It is [solved now](https://github.com/juanelas/bigint-mod-arith/pull/6#issuecomment-1200601111), so no need to use own copy of its code.


willex (approved):
🌊🐴

PR #2366, opened 2022-08-09
fix(core): Add default endpoint for gnosis

stephhuynh18:
ethers js does not have a default provider for gnosis so we are providing a default endpoint

### DIFF
--- a/packages/core/src/anchor/ethereum/ethereum-anchor-validator.ts
+++ b/packages/core/src/anchor/ethereum/ethereum-anchor-validator.ts
@@ -16,6 +16,7 @@ interface EthNetwork {
   chainId: number
   networkId: number
   type: string
+  endpoint?: string
 }
 
 /**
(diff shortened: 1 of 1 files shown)


v-stickykeys (commented):
LGTM 🚀

packages/core/src/anchor/ethereum/ethereum-anchor-validator.ts
```diff
@@ -26,7 +27,14 @@ const ETH_CHAIN_ID_MAPPINGS: Record<string, EthNetwork> = {
...
   'eip155:4': { network: 'rinkeby', chain: 'ETH', chainId: 4, networkId: 4, type: 'Test' },
   'eip155:5': { network: 'goerli', chain: 'ETH', chainId: 5, networkId: 5, type: 'Test' },
-  'eip155:100': { network: 'mainnet', chain: 'Gnosis', chainId: 100, networkId: 100, type: 'Test' },
+  'eip155:100': {
+    network: 'mainnet',
+    chain: 'Gnosis',
+    chainId: 100,
+    networkId: 100,
+    type: 'Test',
+    endpoint: 'https://rpc.ankr.com/gnosis',
```
stbrody, line 36: where did we find this endpoint URL?  Are we confident it's a good and stable one?

stbrody (approved):
code changes LGTM.  Pinging @oed and @haardikk21 to confirm the endpoint

packages/core/src/anchor/ethereum/ethereum-anchor-validator.ts
v-stickykeys: I have used it before and it appears to be reliable re: https://chainlist.org/

oed: This was also suggested by Stefan of Gnosis

stbrody: Sounds good!

PR #2367, opened 2022-08-09
feat(core): Add tests and validation for anchor author

stephhuynh18:
Instead of hardcoding the contracts I could make an API endpoint for the CAS that returns the anchor author. Then I can call that endpoint and use the returned value for the anchor proof validation. But that requires a release, and may not be desirable. The contract address should not change and if so it would be a big deal. So I have hardcoded them into the validator. I am also not married to the name anchor author if you dislike it.


packages/core/src/anchor/__tests__/ethereum/ethereum-anchor-validator.test.ts
```diff
@@ -0,0 +1,92 @@
...
+}
+
+const MockJsonRpcProvider = {
+  getTransaction: null,
+  getNetwork: null,
+  getBlock: null,
+
+  reset() {
+    this.getTransaction = jest.fn(() => {
+      return Promise.resolve(TEST_TRANSACTION)
```
stbrody, line 28: would be good if this checked that the txHash requested matches what we expect (the txHash from the ANCHOR_PROOF)

packages/core/src/anchor/ethereum/ethereum-anchor-validator.ts
```diff
@@ -43,8 +43,16 @@ const BLOCK_THRESHHOLDS = {
   'eip155:1': 1000000000, //mainnet
   'eip155:3': 1000000000, //ropsten
   'eip155:5': 1000000000, //goerli
+  'eip155:100': 1000000000, //gnosis
   'eip155:1337': 1, //ganache
 }
+const ANCHOR_AUTHORS = {
```
stbrody, line 49: are these the addresses of the smart contracts?  Can we call this ANCHOR_CONTRACT_ADDRESSES?

stbrody (approved):
lgtm with two small suggestions

packages/core/src/anchor/ethereum/ethereum-anchor-validator.ts
oed: +1 to @stbrody 

Also, are they really going to have the same address on all networks?
I assume we will update these addresses when we redeploy the contract with the new function signature!

oed (commented):
Open question unrelated to the changes in this PR:
Maybe we should use blockTimestamps instead of block numbers and the `BLOCK_THRESHHOLDS` map? This might simplify things as we wouldn't need to figure out the correct block number for each network.

packages/core/src/anchor/ethereum/ethereum-anchor-validator.ts
stephhuynh18: no they won't. I will update the issues+comment for the block numbers to also include adding the correct contract addresses

PR #2368, opened 2022-08-09
chore(core,stream-tests): Basic end-to-end indexing test

stbrody:


packages/core/src/indexing/sqlite/__tests__/insertion-order.test.ts
```diff
@@ -103,7 +104,7 @@ describe('backward pagination', () => {
         last: PAGE_SIZE,
         before: beforeCursor,
       })
-      beforeCursor = result.pageInfo.startCursor
+      beforeCursor = result.pageInfo.endCursor
```
stbrody, line 107: it's really tough to reason about this stuff, but I'm pretty sure we want to always use `endCursor`, regardless of the iteration direction.  If you look at how the startCursor and endCursor are set [here](https://github.com/ceramicnetwork/js-ceramic/blob/7db84e01f47bb54f811db346bcdf672f29ae7bfc/packages/core/src/indexing/sqlite/insertion-order.ts#L91) and [here](https://github.com/ceramicnetwork/js-ceramic/blob/7db84e01f47bb54f811db346bcdf672f29ae7bfc/packages/core/src/indexing/sqlite/insertion-order.ts#L112) you see that the endCursor is always the 'last' entry in the batch, in whatever order you requested.  So in a forward (newest to oldest) query the `endCursor` will be the oldest entry in the batch, and in a reverse (oldest to newest) query the `endCursor` will be the newest entry in the batch.  In both cases, you want to continue iterating in the same order starting after the `endCursor`

PaulLeCam: For context if that might be useful, I think this is the relevant section of the spec: https://relay.dev/graphql/connections.htm#sec-Edge-order

> [...] ordering must be consistent from page to page, and importantly, The ordering of edges should be the same when using first/after as when using last/before, all other arguments being equal. It should not be reversed when using last/before.

stbrody: > It should not be reversed when using last/before.

wait what?  That seems very in contrast to how we are using them.  @PaulLeCam can you take a look at the `basic-indexing.test.ts` file and see if the way iteration works matches your expectations and understanding of the spec?

packages/stream-tests/src/__tests__/basic-indexing.test.ts
```diff
@@ -0,0 +1,252 @@
...
+    const doc3 = await ModelInstanceDocument.create(ceramic, CONTENT3, midMetadata)
+    const doc4 = await ModelInstanceDocument.create(ceramic, CONTENT4, midMetadata)
+    const doc5 = await ModelInstanceDocument.create(ceramic, CONTENT5, midMetadata)
+
+    const resultObj0 = await ceramic.index.queryIndex({ model: model.id, last: 2 })
+    expect(resultObj0.pageInfo.hasPreviousPage).toBeTruthy()
+    const resultObj1 = await ceramic.index.queryIndex({
+      model: model.id,
+      last: 2,
+      before: resultObj0.pageInfo.endCursor,
```
PaulLeCam, line 222: Should be before the `startCursor` here I think, otherwise you're querying the same items from the end, except off by one.

packages/stream-tests/src/__tests__/basic-indexing.test.ts
```diff
@@ -0,0 +1,252 @@
...
+    })
+    expect(resultObj1.pageInfo.hasPreviousPage).toBeTruthy()
+    const resultObj2 = await ceramic.index.queryIndex({
+      model: model.id,
+      last: 2,
+      before: resultObj1.pageInfo.endCursor,
```
PaulLeCam, line 228:
```suggestion
      before: resultObj1.pageInfo.startCursor,
```

packages/stream-tests/src/__tests__/basic-indexing.test.ts
```diff
@@ -0,0 +1,252 @@
...
+    expect(results[0].id.toString()).toEqual(doc5.id.toString())
+    expect(results[0].content).toEqual(CONTENT5)
+    expect(results[1].id.toString()).toEqual(doc4.id.toString())
+    expect(results[1].content).toEqual(CONTENT4)
+    expect(results[2].id.toString()).toEqual(doc3.id.toString())
+    expect(results[2].content).toEqual(CONTENT3)
+    expect(results[3].id.toString()).toEqual(doc2.id.toString())
+    expect(results[3].content).toEqual(CONTENT2)
+    expect(results[4].id.toString()).toEqual(doc1.id.toString())
+    expect(results[4].content).toEqual(CONTENT1)
```
PaulLeCam, line 250: Looks like the order of the results is reversed here.
My understanding from the connection spec is that if we have `[A, B, C, D, E, F]` and we query `last: 2`, we should get `[E, F]` with `{ hasNextPage: false, hasPreviousPage: true, startCursor: E, endCursor: F }`, then `last: 3, before: E` should return `[B, C, D]`.
Essentially when using `last` we iterate backwards but that shouldn't change how the slices of results are returned.

packages/core/src/indexing/sqlite/__tests__/insertion-order.test.ts
PaulLeCam: I left some comments there, would be good to get input from @ukstv as well as we discussed this before and we have the current behavior working well for the clients, so I think the current implementation is correct?

packages/stream-tests/src/__tests__/basic-indexing.test.ts
stbrody: woah.  That makes no sense to me at all.  What if you just want to iterate in literal reverse order?  Is there no way to do that in GraphQL?

stbrody: isn't that also going to be way less efficient in the database as you can't just iterate the index in reverse order?

stbrody: It looks like GraphQL is supposed to also support an `order` argument: https://dgraph.io/docs/graphql/queries/order-page/.  We don't currently seem to support that - should we?  Being able to query in reverse order seems way more useful to me than being able to fetch the `last` N items in forward order would be.

stbrody: I guess we removed that when we stopped supporting multiple orderings.  I didn't realize at the time that by removing support for ordering on different groups of fields we also removed support for querying in reverse order :disappointed: 

Probably too late to add to the upcoming developer preview release now, but seems like it should be added as part of one of the follow-ups?  Either the one with relations or the one with unions and interfaces.  What do you think @PaulLeCam?

PaulLeCam: > It looks like GraphQL is supposed to also support an `order` argument: https://dgraph.io/docs/graphql/queries/order-page/. We don't currently seem to support that - should we? Being able to query in reverse order seems way more useful to me than being able to fetch the `last` N items in forward order would be.

This is not the official GraphQL spec, seems more like a custom implementation on top with its own additions, no?

The official spec isn't opinionated about custom ordering, we can do what we want for this, but what matters is that the pagination slices are consistent, cf https://relay.dev/graphql/connections.htm#sec-Edge-order

> When before: cursor is used, the edge closest to cursor must come last in the result edges.
> When after: cursor is used, the edge closest to cursor must come first in the result edges.
 
That seems pretty logical to me, no? It's mainly a convention so the GraphQL servers and clients don't have to worry about different pagination mechanisms I guess.



> I guess we removed that when we stopped supporting multiple orderings. I didn't realize at the time that by removing support for ordering on different groups of fields we also removed support for querying in reverse order 😞

Not sure I follow, what is the problem with querying in reverse order? GraphQL connections support using either `first` with optional `after` or `last` with optional `before` arguments, so the ordering can be applied as requested by the client, no?

stbrody: >  what is the problem with querying in reverse order?

So let's say you create streams A, B, C, D, and E in order.  A normal forward query will return ABCDE.
If you use `last` with a page size of 2 you get back DEBCA (or if you use `last` with a large enough page size, you just get back ABCDE again).  That doesn't seem like a very useful ordering to me.  It seems like we should have a way for users to get back EDCBA - actual reverse order

PaulLeCam: You don't actually get DEBCA though, you first get [DE], then [BC] that you need to prepend to the results list, so now you have [BCDE] and then you get [A] that you prepend again, and get [ABCDE].
In the client you can reverse the result list if needed, or simply iterate from the end to the start of the list, that's not really a spec problem at this point.

I guess one thing to remember here is that this is designed for cursor-based access, the ordering of queries is a "just in time" concern.
Say you want to query messages in a live chat, you start by querying the last 2 messages for example, you get [EF]. Now you also start polling for new messages, start getting [G], then [H] that you just append, so you have [EFGH]. The user scrolls up the message lists, so we want to load the last 2 messages messages before E, we get back [CD] that needs to be prepended to the list, which becomes [CDEFGH] and is likely what the app wants, no?

packages/stream-tests/src/__tests__/basic-indexing.test.ts
```diff
@@ -0,0 +1,253 @@
...
+      last: 2,
+      before: resultObj1.pageInfo.startCursor,
+    })
+    expect(resultObj2.pageInfo.hasPreviousPage).toBeFalsy()
+
+    const results = (
+      await Promise.all([
+        extractDocuments(ceramic, resultObj0),
+        extractDocuments(ceramic, resultObj1),
+        extractDocuments(ceramic, resultObj2),
```
PaulLeCam, line 235: I think this is where the test doesn't match how GraphQL clients handle the results, if instead it was
```suggestion
        extractDocuments(ceramic, resultObj2),
        extractDocuments(ceramic, resultObj1),
        extractDocuments(ceramic, resultObj0),
```
then the results order would be consistent.

packages/stream-tests/src/__tests__/basic-indexing.test.ts
stbrody: This sounds like the app is managing all these pages manually.  I had thought the client would present a simple iterator interface for the client where they could just get results one by one with a `next()/hasNext()` API, and all the pagination would be handled automatically by the client.  Is the client really expected to understand all the nuances around startCursor vs endCursor, before vs after, last vs first etc?  I only just got my head around it - is it really the expectation that every app dev will need to do the same?

PaulLeCam: Depends how high-level the GraphQL client goes, but for example with Relay this is handled as you describe: https://relay.dev/docs/guided-tour/list-data/pagination/

PaulLeCam (approved):

PR #2371, opened 2022-08-12
feature (core): add stream content to composedb

willex:
## Description

Stores the the stream content via an additional field in the indexing tables to unlock DB side querying and/or filtering on the content directly efficiently through Posgres' JSONB implementation (https://www.postgresql.org/docs/current/functions-json.html). SQLite implementation is there as well, but only as a base `string` so usability is fairly limited and only supplied for feature parity reasons. This will need further addressing to make a decision on how we want to treat different DB capabilities going forward adding e.g. MongoDB.

**_NOTE_**: if you have an existing Postgres instance with a prior table structure, this will break node startup as the column does not exist in your database and schema verification will fail.

To remedy the issue stop the node and:

DROP the specific indexing tables - the following SQL command will create the drop commands for all applicable tables:
`SELECT 'DROP TABLE ' || TABLE_NAME || ' CASCADE;' FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_NAME LIKE 'kjz%';`

alternatively, DROP the whole database and re-create it from scratch then restart the node and the new correct tables will be created.

## How Has This Been Tested?

- [x] Tested locally on
- [x] Extended unit tests including test sample data

## PR checklist

## References:


packages/core/src/indexing/postgres/postgres-index-api.ts
```diff
@@ -36,6 +36,7 @@ export class PostgresIndexApi implements DatabaseIndexApi {
       .insert({
         stream_id: String(args.streamID),
         controller_did: String(args.controller),
+        stream_content: JSON.stringify(args.streamContent),
```
willex, line 39: Unsure about this; if we just want to use `String` instead since from what've seen stream content is already coming in as a string and Postgres does the validation itself on the field too.

packages/core/src/indexing/database-index-api.ts
```diff
@@ -5,6 +5,7 @@ export interface IndexStreamArgs {
   readonly streamID: StreamID
   readonly model: StreamID
   readonly controller: string
+  readonly streamContent: string
```
stbrody, line 8:
```suggestion
  readonly streamContent: Record<string, any>
```

stbrody: I think the content should probably get passed down as an object as far down the stack as possible, and then individual database API implementations can decide how to translate the content into a format the database understands.  If we had a MongoDB database backend for example, you wouldn't want to send it a string, the mongodb Node.js client has support for receiving a javascript object directly.

packages/core/src/indexing/postgres/__tests__/postgres-index-api.test.ts
```diff
@@ -134,6 +135,7 @@ describe('indexStream', () => {
   const STREAM_CONTENT = {
     model: MODELS_TO_INDEX[0],
     streamID: StreamID.fromString(STREAM_ID_B),
+    streamContent: STREAM_TEST_DATA_PROFILE,
     controller: CONTROLLER,
     lastAnchor: null,
   }
```
stbrody, line 141: I think we should also have a test that builds an index on one of the fields in the content JSON and does a query against the postgres database to make sure it works.  That way we know that we've definitely provided the data in a way that Postgres JSONB indexing can actually work.

packages/core/src/indexing/postgres/migrations/1-create-model-table.ts
```diff
@@ -7,6 +7,7 @@ export async function createModelTable(dataSource: Knex, tableName: string) {
 
     table.string('stream_id').primary(`idx_${indexName}_pkey`).unique(`constr_${indexName}_unique`)
     table.string('controller_did', 1024).notNullable()
+    table.jsonb('stream_content').notNullable().defaultTo('{}')
```
stbrody, line 10: why give this a default?  Don't we expect it to be always provided?

packages/core/src/indexing/postgres/postgres-index-api.ts
stbrody: sorry, I don't really understand what the two options are?

packages/core/src/indexing/postgres/__tests__/postgres-index-api.fixture.ts
```diff
@@ -0,0 +1,805 @@
+/**
+ * Stream content test data-set
+ */
+export const STREAM_TEST_DATA_PROFILE = JSON.stringify([
```
stbrody, line 4: why such a large and complex test content?  I feel like it makes it seem like we're using all this for something.  If a simple dictionary with a 2-3 fields would test just as well, let's just use that and keep it simple.

Also, I don't think MIDs support arrays at the top level, the content hast to be a dictionary object

stbrody (changes requested):

packages/core/src/indexing/postgres/postgres-index-api.ts
willex: It's either to have cast a straightforward `String(args.streamContent)` or the current active conversion via `JSON.stringify(args.streamContent)`.

packages/core/src/indexing/postgres/__tests__/postgres-index-api.test.ts
willex: Yeah I agree. I was actually thinking about that after I opened the PR and will add that as well.

packages/core/src/state-management/repository.ts
```diff
@@ -364,6 +364,7 @@ export class Repository {
       model: state$.value.metadata.model,
       streamID: state$.id,
       controller: state$.value.metadata.controller,
+      streamContent: state$.value.content,
```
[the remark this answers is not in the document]
ukstv: I have a question to both you @willex and @stbrody. Do you feel like `streamContent` without any information about CID of the log entry makes sense?

ukstv (changes requested):
pls, "store" an object  instead of a string

packages/core/src/state-management/repository.ts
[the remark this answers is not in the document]
stbrody: > I have a question to both you @willex and @stbrody. Do you feel like `streamContent` without any information about CID of the log entry makes sense?

Hmm...
It probably would be nice to include the CID of the current tip in the database.  That way apps/devs that are querying the database for stream content would have a way to see what version of the data they are getting, which could be helpful for debugging issues with stale data from out-of-date indexes.  I'd be open to adding a `tip` column to the database for this.  The downside is that later on when we support streams with multiple tips we may have to do a database migration to pull tips out to their own table with a foreign key relation.  But maybe that's far enough away that we can cross that bridge when we come to it.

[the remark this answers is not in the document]
willex: Yeah, thats a good point that I haven't considered. Definitely makes sense to add `tip` column to it for debugging reasons. For the multiple tips support - do we have a timeline on that? I'm sure there will be other migrations as well along that road... so if its more than two seasons out (or post full prod rollout) I think its OK to cross that bridge then.

packages/core/src/indexing/postgres/migrations/1-create-model-table.ts
willex: we do currently, but wanted to standardize around the default value thats also provided here: https://github.com/ceramicnetwork/js-ceramic/blob/413b64490cfeb1a8430ecedaaeb55f106e103e2a/packages/stream-model-instance-handler/src/model-instance-document-handler.ts#L124

'{}' in the db table is defaulting to '{}::jsonb' so already an object and not a string. Fairly agnostic to keeping the default value or removing it.

packages/core/src/indexing/postgres/postgres-index-api.ts
willex: this is resolved with the object comment.

packages/core/src/indexing/postgres/__tests__/postgres-index-api.test.ts
```diff
@@ -147,17 +161,52 @@ describe('indexStream', () => {
...
     expect(raw.last_anchored_at).toBeNull()
     expect(raw.first_anchored_at).toBeNull()
     const createdAt = new Date(raw.created_at)
     const updatedAt = new Date(raw.updated_at)
     expect(closeDates(createdAt, now)).toBeTruthy()
     expect(closeDates(updatedAt, now)).toBeTruthy()
+    expect(raw.dark_mode).toEqual(STREAM_TEST_DATA_PROFILE.settings.dark_mode)
+    expect(raw.id).toEqual(STREAM_TEST_DATA_PROFILE.id)
+
+    // create index on jsonb content
```
stbrody, line 186: can we break this into two tests?  I feel like we've conflated two test cases into one - first the basic test that the index gets maintained at all, then the test that the content is indexable and queryable with jsonb

packages/core/src/indexing/postgres/__tests__/postgres-index-api.test.ts
```diff
@@ -147,17 +161,52 @@ describe('indexStream', () => {
...
+    result = await dbConnection.raw(
```
stbrody, line 187: seems like creating the index should happen in the test fixture setup

packages/core/src/indexing/postgres/__tests__/postgres-index-api.test.ts
```diff
@@ -147,17 +161,52 @@ describe('indexStream', () => {
...
+      `CREATE INDEX idx_postgres_jsonb ON ${MODELS_TO_INDEX[0]}(stream_id, (stream_content->'settings'->'dark_mode'))`
+    )
+    expect(result.command).toEqual('CREATE')
+
+    // verify index usage
+    await dbConnection.raw(`SET enable_seqscan = off;`)
+    result = await dbConnection.raw(
+      `EXPLAIN SELECT *
+       FROM ${MODELS_TO_INDEX[0]}
+       WHERE stream_id='${STREAM_ID_B}'
```
stbrody, line 197: querying on the streamid seems to negate querying on the content since there will only ever be one match on streamid anyway.  Can we query *just* on the content?

packages/core/src/indexing/postgres/__tests__/postgres-index-api.test.ts
```diff
@@ -147,17 +161,52 @@ describe('indexStream', () => {
...
+         AND (stream_content->'settings'->>'dark_mode')::BOOLEAN IS true;`
```
stbrody, line 198: Would be valuable in the test of content indexing to have one stream where the value is true and one where it's false so we make sure it's actually matching the values properly

packages/core/src/indexing/postgres/migrations/1-create-model-table.ts
stbrody: That default there is happening at a very different level.  It's saying that if the genesis **commit** doesn't have a `data` field, the resulting **stream** still comes back with a default `{}` in its content.  The stream itself should never have empty content, and I don't think it makes sense for the database layer to apply a default content separate from the stream layer.  If we ever had code that tried to add a stream to the index without providing content, that code would be a bug.  We'd want to notice that bug rather than silently ignoring it and losing the content from the database.  I recommend taking this default out.

packages/core/src/indexing/postgres/migrations/1-create-model-table.ts
```diff
@@ -34,5 +36,8 @@ export async function createModelTable(dataSource: Knex, tableName: string) {
         storageEngineIndexType: 'hash',
       }
     )
+    table.index(['stream_id', 'tip'], `idx_${indexName}_stream_id_tip`, {
```
stbrody, line 39: do we need this index?  how would it be used?

packages/core/src/indexing/postgres/migrations/mid-schema-verification.ts
```diff
@@ -16,6 +16,18 @@ export const validTableStructure = {
...
+    type: 'jsonb',
+    maxLength: null,
+    nullable: false,
+    defaultValue: "'{}'::jsonb",
+  },
+  tip: {
+    type: 'character varying',
+    maxLength: 255,
+    nullable: false,
+    defaultValue: null,
```
stbrody, line 29: above this field is set as `notNullable()`.  How can it be not nullable and also have a default value of null?

packages/core/src/indexing/postgres/postgres-index-api.ts
```diff
@@ -36,6 +36,8 @@ export class PostgresIndexApi implements DatabaseIndexApi {
       .insert({
         stream_id: String(args.streamID),
         controller_did: String(args.controller),
+        stream_content: args.streamContent,
+        tip: String(args.tip),
```
stbrody, line 40: is casting to a String like this the same as calling `toString()` on the CID object?  I've always used `toString()`, is there an advantage to doing the cast this way instead?

stbrody (commented):
overall looking good, left comments with some little nitpicks to clean up

packages/core/src/indexing/postgres/migrations/1-create-model-table.ts
willex: Ok that makes sense. I'll adjust it and drop the default then.

packages/core/src/indexing/postgres/__tests__/postgres-index-api.test.ts
willex: ooh, yes I thought wrongly inserts vs. updates (override) here. I'll adjust and do just the content.

packages/core/src/indexing/postgres/migrations/1-create-model-table.ts
willex: ooh, yes I thought wrongly inserts vs. updates (override) here. I'll remove the index.

packages/core/src/indexing/postgres/migrations/mid-schema-verification.ts
willex: This is a very good question that I don't have a concrete answer to... it is what `knex` returns me when I dump the schema. Looking at the table structure itself it is correct though - my assumption is it is just a default value returned to ensure the structure stays the same between tables when dumped.

packages/core/src/indexing/postgres/postgres-index-api.ts
willex: I've honestly used it only because we used it initially during the sqlite implementation. It's similar with a major difference though:

`var value = null;
String(value);     // "null" -> String
value.toString(); // TypeError`

I need to look into that `"null"` or `"undefined"` is interpreted by knex correctly as null value vs. doing the check ourselves if we switch to `toString` for nullable fields. 

Overall thats not enough of an advantage though IMHO and I'm OK to switch it to `toString()` instead with a simple in-line `null/undefined` check.

packages/core/src/indexing/postgres/__tests__/postgres-index-api.test.ts
willex: My thought process was if it would not match it correctly here that it would return 0 instead. I'll add the second stream as you suggested though to make it explicit.

packages/core/src/indexing/postgres/__tests__/postgres-index-api.test.ts
willex: I can push it up/split - I mainly kept it together here since we already are populating and confirming the data but yeah I can push that up into the fixture.

packages/core/src/indexing/postgres/__tests__/postgres-index-api.test.ts
willex: Yeah, I can push it up/split it into to and parse out the JSONB parts of it explicitly but I also feel we could leave it in one.

Since querying the content is about confirming that it was stored indeed as an object vs. a string, and therefor traversable, fits for me as part of the new stream creation/data insert test.

For the maintained index I see it's less about checking that the index is indeed maintained and more so as an extension of object storage and subsequent query-ability. I could see this only really fail if something would be severely wrong with the Postgres instance since we are just confirming base DB functionality via its built in operators at that point (which kinda makes the index test overkill in the first place thinking about it).

if we break it out I would rather do it as the db specific verifications then (namely jsonb / object related in the case of Postgres):
1. storage: is it stored correctly as an object confirmed via db operators
2. filtering: can I query against it with typescript data types being honored (string, bool etc.) without the need of addt. casting

and extend that then with e.g. MongoDB specific tests in the future.

stbrody: I don't feel that strongly about this.  IMO it'd be nice to split up, but we don't have to if you'd prefer not to.

packages/core/src/indexing/postgres/__tests__/read-csv-fixture.util.ts
```diff
@@ -2,8 +2,10 @@ import { IndexStreamArgs } from '../../database-index-api.js'
 import fs from 'node:fs'
 import csv from 'csv-parser'
 import { StreamID } from '@ceramicnetwork/streamid'
+import { CID } from 'multiformats/cid'
 
 const MODEL_ID = 'kjzl6cwe1jw145m7jxh4jpa6iw1ps3jcjordpo81e0w04krcpz8knxvg5ygiabd'
+const FAKE_CID = CID.parse('bafybeig6xv5nwphfmvcnektpnojts33jqcuam7bmye2pb54adnrtccjlsu')
```
ukstv, line 8: This CID is used in other places, while having no explicit connection between different (yet same) constants.
I see two options to fix that:
1) Export this thing and reuse in the tests.
2) Add another optional parameter to `readCsvFixture` function, so that a test can decide what to pass there.

I would encourage you to go with (2).

ukstv (approved):
🏇 

Mod optional (but highly encouraged) CID constant deduplication.

packages/core/src/indexing/postgres/__tests__/read-csv-fixture.util.ts
willex: in that case, wouldn't we want to do the same thing for the `MODEL` as well? So extend `readCsvFixture` to accept both `MODEL` & `CID` but not make it optional but force the tests to pass the values and make it a pure util function as a result?

packages/core/src/indexing/postgres/__tests__/postgres-index-api.test.ts
willex: I looked into the fixture creation but given how it is structured right now with tables being dropped before each run I made it a separate test instead. 

As I wrote in the other comment it's mainly a confirmation that the index is indeed invoked when queried BUT I'm honestly not sure if this test overall adds any value given that it only tests a base db function and not really anything indexing feature related.

packages/core/src/indexing/postgres/__tests__/postgres-index-api.test.ts
```diff
@@ -186,6 +221,51 @@ describe('indexStream', () => {
...
+       FROM ${MODELS_TO_INDEX[0]}
+       WHERE stream_content->'settings'->'dark_mode' = ?;`,
+      true
+    )
+    expect(result.command).toEqual('EXPLAIN')
+    expect(result.rows.length).toBeGreaterThan(1)
+    expect(result.rows[0]['QUERY PLAN'].includes('idx_postgres_jsonb')).toBeTruthy()
+  })
+
+  test('query and filter jsonb stream content', async () => {
```
stbrody, line 247: does this test pass if you run it on its own, without running the previous test that creates the JSONB index?

packages/core/src/state-management/repository.ts
```diff
@@ -363,7 +363,9 @@ export class Repository {
     const STREAM_CONTENT = {
       model: state$.value.metadata.model,
       streamID: state$.id,
-      controller: state$.value.metadata.controller,
+      controller: state$.value.metadata.controllers[0],
```
stbrody, line 366: Confirmed that this is right and makes sense.  `state$` is a [StreamState](https://github.com/ceramicnetwork/js-ceramic/blob/5fc4e6157f52ef513abda2845005673f1955ae48/packages/common/src/stream.ts#L137), `metadata` is a [StreamMetadata](https://github.com/ceramicnetwork/js-ceramic/blob/5fc4e6157f52ef513abda2845005673f1955ae48/packages/common/src/stream.ts#L90) and it has a `controllers` array

stbrody (approved):
LGTM!

packages/core/src/indexing/postgres/__tests__/postgres-index-api.test.ts
willex: No, it would fail as it relies on the INDEX to be created because it checks for the `idx_postgres_jsonb` to be present and properly invoked when data is being queried.

stbrody: Right that was my concern.  In that case the index should be created in `beforeAll` or `beforeEach` setup of the test fixture.  Tests shouldn't have side affects that affect other tests.

PR #2373, opened 2022-08-12
chore(stream-tests): Test index maintenance between nodes

stbrody:


packages/stream-tests/src/__tests__/cross-node-indexing.test.ts
```diff
@@ -0,0 +1,172 @@
...
+    await TestUtils.delay(5 * 1000)
+
+    resultObj = await ceramic2.index.queryIndex({ model: model.id, first: 100 })
+    results = extractDocuments(ceramic2, resultObj)
+
+    expect(results.length).toEqual(2)
+    expect(results[0].id.toString()).toEqual(doc1.id.toString())
+    expect(results[0].content).toEqual(CONTENT1)
+    // TODO: Why do I need to serialize state for this to work?
+    expect(StreamUtils.serializeState(results[0].state)).toEqual(
```
stbrody, line 140: I don't understand why this is necessary in this one case but not the other times I compare states in this test.  I think it has something to do with how a StreamID inside the state object is Memoized.  @ukstv I know you were looking into some weirdness around memoization before.

PR #2383, opened 2022-08-19
feat(stream-model): Only allow genesis commit

oed:
This PR changes the model stream type to not have a controller and be completely immutable. 

Tests need to be updated.


PaulLeCam:
@pawartur I think that drives a bunch of changes in the tooling, we wouldn't need the Ceramic instance to be authenticated to create models anymore. Hopefully that's not a problem if we provide a controller, but we might want to update the docs accordingly before the release.
I think that will also require recreating the composite used in https://composedb.js.org/docs/0.2.x/first-composite#importing-a-model cc @v-stickykeys

packages/stream-model-handler/src/model-handler.ts
```diff
@@ -97,27 +97,10 @@ export class ModelHandler implements StreamHandler<Model> {
   async _applyGenesis(commitData: CommitData, context: Context): Promise<StreamState> {
     const payload = commitData.commit
     const isSigned = StreamUtils.isSignedCommitData(commitData)
-    if (!isSigned) {
-      throw Error('Model genesis commit must be signed')
+    if (isSigned) {
```
PaulLeCam, line 100: Does that mean attempting to deploy existing signed models would fail?

PaulLeCam (approved):
LGTM, just wondering about the changes we need to make in the tools to support it?

pawartur:
Sounds good, @PaulLeCam , thanks for the context! I'll test this with composedb cli on monday and create a draft PR with changes to tooling, cli and docs.

packages/stream-model-handler/src/model-handler.ts
oed: Yes, I think it would be weird if our example model is signed when all other models are not?

oed:
It shouldn't be a problem if your ceramic instance is authenticated. So I don't think you actually need to remove the auth step from the cli?

packages/stream-model-handler/src/model-handler.ts
PaulLeCam: Yes should be easy to make the change, just need to make sure we update the composite used in the docs before release.

oed: Ok, thanks :)

oed:
Added signatures back to the genesis object after discussion with @stbrody. Still need to update some of the tests.

pawartur:
So just to make sure, @oed , we don't want to remove authentication from the cli for Developer Preview? Or should I try removing it and see, if there are any problems with that?

Now that the genesis commits are signed, do model streams still have controllers? If not, I should remove the model:controller command from the cli.

oed:
That's correct @pawartur. We are not removing it, and yes models will still have controllers.

pawartur:
Ok, thanks! Than no changes in the cli needed.

ukstv (approved):
🏇

packages/stream-model-handler/src/__tests__/model-handler.test.ts
```diff
@@ -788,137 +458,10 @@ describe('ModelHandler', () => {
...
-      type: CommitType.SIGNED,
-      commit: sPayload,
-      envelope: signedCommit.jws,
-    }
-    await expect(handler.applyCommit(signedCommitData, context, state)).rejects.toThrow(
-      /Updating metadata for Model Streams is not allowed/
-    )
-  })
-
-  it('fails to apply commit with invalid prev link', async () => {
```
stbrody, line 834: some of these tests could in principle be updated to work with anchor commits and maintain the core of what the test was trying to check.  Up to you whether it's worth the effort or not

packages/stream-model-handler/src/model-handler.ts
```diff
@@ -137,70 +136,14 @@ export class ModelHandler implements StreamHandler<Model> {
       log: [{ cid: commitData.cid, type: CommitType.GENESIS }],
     }
```
stbrody, line 138: Recommend adding `Model.assertComplete(payload.data)` before the call to `assertNoExtraKeys` on line 120

packages/stream-model/src/model.ts
```diff
@@ -269,36 +229,6 @@ export class Model extends Stream {
     return model
```
stbrody, line 229: I think the checks from line 263-268 here can be removed.  `load()` no longer needs to check for incomplete models as it should be impossible to create incomplete models in the first place.

stbrody (commented):
a few small suggestions, otherwise looks good

packages/stream-model/src/model.ts
oed: good call

packages/stream-model-handler/src/__tests__/model-handler.test.ts
oed: Isn't this assured by other parts of the codebase anyway?

stbrody (approved):
LGTM

PR #2393, opened 2022-08-23
fix(core): Use package, not relative path to metrics

ukstv:
Apparently, we forgot to replace one of the relative imports with `@ceramicnetwork/metrics` import. The PR fixes that.

I hope it is the last occurrence of the relative imports.

CC @gvelez17


stbrody (approved):
lgtm

PR #2396, opened 2022-08-24
feat(stream-model): Remove 'unique' from Model metadata

stbrody:


oed (approved):
lgtm, this doesn't break already created models with `unique` I assume.

stbrody:
> lgtm, this doesn't break already created models with `unique` I assume.

I haven't explicitly tested it, but it shouldn't

PR #2397, opened 2022-08-24
feat(core): Reject client-initiated updates that build on stale state

stbrody:
Builds on https://github.com/ceramicnetwork/js-ceramic/pull/2395

The goal of this change is to better handle the case where the client application authors a new update to a stream based on stale state.  Imagine that an application loads a stream and the stream handle stays open in the user's browser tab for a while, possibly hours or days.  In the meantime, new updates are made to the stream, possibly via a different Ceramic node powering a different app, or even possibly via the same Ceramic node and app, just in a different session.  Then the user tries to do an update via the original stream handle, which has an old, out of date version of the stream state in it's client-side view.  In this case, one of two things will happen: either the write will be rejected by conflict resolution, or it will succeed by winning the conflict resolution against the existing state on the node, and thus overriding an existing write.

With this change, if the user tries to initiate a write via the http-client but that write is built off of state the Ceramic node *already knows* is stale, the Ceramic node will proactively reject that write to preserve the data that it already knows about that has been around longer, over the new write that was authored based on a stale view of the data.

Note that this change has no affect to the conflict resolution rules applied when learning about a commit via pubsub.  The arbitrary-but-deterministic tie-break based on CID lexographic ordering still applies for commits learned via the network.  This is a special case to prefer existing data over new data when we know the new data is stale, as a way to minimize the number of writes that successfully get applied to a node and only to be later rejected.

### DIFF
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -193,9 +193,9 @@ function validatePort(inPort) {
  * @param opts
  */
 function markShouldRejectConflicts(opts) {
-  if (opts.throwOnInvalidCommit === undefined) {
-    opts.throwOnInvalidCommit = true
-  }
+  opts.throwOnInvalidCommit = opts.throwOnInvalidCommit ?? true
+  opts.throwOnConflict = opts.throwOnConflict ?? true
+  opts.throwIfStale = opts.throwIfStale ?? true
 }
 
 /**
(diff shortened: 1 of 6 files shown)


stbrody:
Not yet ready for review, needs cleanups to the tests that depend on https://github.com/ceramicnetwork/js-ceramic/pull/2358

stbrody:
Closing in favor of https://github.com/ceramicnetwork/js-ceramic/pull/2579

PR #2398, opened 2022-08-25
fix(core): Allow fast-forward of a stream state if newer commit is anchored

ukstv:
CC @v-stickykeys

### DIFF
--- a/packages/common/src/utils/stream-utils.ts
+++ b/packages/common/src/utils/stream-utils.ts
@@ -196,11 +196,7 @@ export class StreamUtils {
       }
     }
 
-    if (state.anchorStatus === AnchorStatus.ANCHORED) {
-      return true
-    }
-
-    if (state.log.length == base.log.length && state.anchorStatus != base.anchorStatus) {
+    if (state.log.length === base.log.length && state.anchorStatus != base.anchorStatus) {
       // Re-creating a state object from the exact same set of commits can still lose information,
       // such as whether or not an anchor has been requested.
       return false
--- a/packages/common/src/utils/stream-utils.ts
+++ b/packages/common/src/utils/stream-utils.ts
@@ -200,7 +200,7 @@ export class StreamUtils {
       return true
     }
 
-    if (state.anchorStatus != base.anchorStatus) {
+    if (state.log.length == base.log.length && state.anchorStatus != base.anchorStatus) {
       // Re-creating a state object from the exact same set of commits can still lose information,
       // such as whether or not an anchor has been requested.
       return false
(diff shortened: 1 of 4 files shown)


ukstv:
Converted to Draft, as it needs tests.

packages/common/src/utils/stream-utils.ts
```diff
@@ -197,6 +196,10 @@ export class StreamUtils {
       }
     }
 
+    if (state.anchorStatus === AnchorStatus.ANCHORED) {
+      return true
+    }
+
     if (state.anchorStatus != base.anchorStatus) {
```
stbrody, line 203:
```suggestion
    if (state.log.length == base.log.length && state.anchorStatus != base.anchorStatus) {
```

stbrody: So I think the thing this code was trying to deal with is two states that are otherwise identical, but where one has information about the anchor status and the other doesn't.  In that case you want to prefer the log with the anchor information.

The oversight with how it was written is that `state` could be **ahead** of `base` and legitimately have a different anchor status but still be a superset of `base`.  So I think if you make the change I propose here you can remove the `if(state.anchorStatus === AnchorStatus.ANCHORED)` check you added above.

stbrody (changes requested):

packages/common/src/utils/stream-utils.ts
ukstv: The suggestion makes sense. Thanks.

packages/core/src/__tests__/state-manager.test.ts
```diff
@@ -348,16 +348,41 @@ describe('atCommit', () => {
...
+  test('return read-only snapshot: do not lose own anchor status', async () => {
+    // Prepare commits to play
+    const tile1 = await TileDocument.create<any>(ceramic, INITIAL_CONTENT, null, {
+      anchor: false,
+      syncTimeoutSeconds: 0,
+    })
+    await tile1.update({ abc: 321, def: 456, gh: 987 })
+    await TestUtils.anchorUpdate(ceramic, tile1)
+
+    // Let's pretend we have a stream in PENDING state
```
stbrody, line 372: it would be weird to have a stream in anchorStatus PENDING where the most recent commit is an anchor commit.  Just to make the test slightly more realistic, I'd recommend dropping the `anchorUpdate` on line 370 and having one of the states have anchorStatus PENDING and the other have NOT_REQUESTED

stbrody (approved):
LGTM

packages/core/src/__tests__/state-manager.test.ts
ukstv: make sense, thanks

PR #2406, opened 2022-08-31
WIP: Check for expired CACAOs when applying new commits

stbrody:
Still a WIP.  There are some issues to iron out if the genesis commit itself has an expired CACAO, including removing the stream from the index in that case.  But this PR is getting complex enough that before I go too much further with it I wanted to get some more eyes on it and see if the general direction seems right.

### DIFF
--- a/packages/common/src/stream.ts
+++ b/packages/common/src/stream.ts
@@ -115,6 +115,7 @@ export interface LogEntry {
   cid: CID
   type: CommitType
   timestamp?: number
+  expirationTime?: number
 }
 
 /**
--- a/packages/core/src/conflict-resolution.ts
+++ b/packages/core/src/conflict-resolution.ts
@@ -256,7 +256,7 @@ function checkForCacaoExpiration(state: StreamState): void {
           state
         ).toString()} has a CACAO that expired at ${entry.expirationTime}`
       )
-      err.expiredCommitCID = entry.cid
+      err.lastValidCommitCid = state.log[i - 1]?.cid || null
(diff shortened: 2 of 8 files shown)


packages/common/src/stream.ts
```diff
@@ -115,6 +115,7 @@ export interface LogEntry {
   cid: CID
   type: CommitType
   timestamp?: number
+  expirationTime?: number
```
oed, line 118: Could you add some documentation what each of these params are. I don't think it's very obvious for an onlooker.

packages/core/src/conflict-resolution.ts
```diff
@@ -237,6 +237,34 @@ export function commitAtTime(stateHolder: StreamStateHolder, timestamp: number):
...
+      timestamp = entry.timestamp
+    }
+    if (entry.expirationTime && entry.expirationTime < timestamp) {
+      err = new Error(
+        `CACAO expired: Commit ${entry.cid.toString()} of Stream ${StreamUtils.streamIdFromState(
+          state
+        ).toString()} has a CACAO that expired at ${entry.expirationTime}`
+      )
+      err.lastValidCommitCid = state.log[i - 1]?.cid || null
+      // Keep iterating through the whole log so that we find the oldest commit that is expired.
```
oed, line 260: Should we not break if we get to an anchor commit?

oed (commented):
Looks reasonable I think after a quick look.

stbrody:
will put up new PRs for this work, closing this draft

PR #2407, opened 2022-09-01
feat(core): moving rate limiting message from warning to verbose

willex:
## Description

To declutter production log messages, moving the pubsub rate limiting warning message from `warning` to `verbose` since rate limiting doesn't have any negative impact on actual data retention or message socialization.

## How Has This Been Tested?

- [x] Standard unit tests

## PR checklist

## References:

### DIFF
--- a/packages/core/src/pubsub/pubsub-ratelimit.ts
+++ b/packages/core/src/pubsub/pubsub-ratelimit.ts
@@ -100,7 +100,7 @@ export class PubsubRateLimit
       if (timeSinceOldestQuery < 1000) {
         // If it's been less than a second since the oldest query, sleep until it's been more than a
         // second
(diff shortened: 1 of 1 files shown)


oed (approved):
I'm not familiar with the debugging framework but I suppose this is correct.

ukstv (approved):

stbrody:
This makes me nervous. Now if our users are getting rate limited they won't know and neither will we when inspecting their logs. What's the motivation for making this change? What if instead of pushing this message to verbose we rate limited the message itself, so it would only log max once per minute for example?

ukstv:
Kind of continuing Discord discussion here. "Now if our users are getting rate limited they won't know" - what would be the next step for an app operator, if she sees that warning?

PR #2408, opened 2022-09-01
Fix prev.equals error

v-stickykeys:
## Description

This PR forces the `prev` property on commit data to be a CID to prevent the following error:

- When trying to update a tile document or a mid in the [indexing smoke tests](https://github.com/3box/ceramic-smoke-tests/blob/feat/indexing/src/__tests__/indexing-bug-repro.test.ts) we have been hitting this error.

```js
await midFromIndexer.replace(updatedContent)
```

```sh
HTTP request to 'http://localhost:7007/api/v0/commits' failed with status 'Internal Server Error': {"error":"commitData.commit.prev.equals is not a function"}
```

- It turns out the prev was not a CID, and therefore doesn't have the `equals` function attached. However it can be cast into a CID.

### Repro

1. Run the ceramic daemon (you can repro on any ceramic network)
```
CERAMIC_ENABLE_EXPERIMENTAL_COMPOSE_DB=true node packages/cli/bin/ceramic.js daemon
```

2. Run the smoke tests
```sh
git clone https://github.com/3box/ceramic-smoke-tests.git
cd ceramic-smoke-tests
git checkout feat/indexing
npm install
NODE_OPTIONS=--experimental-vm-modules jest src/__tests__/indexing-bug-repr
```

You will see the error in the smoke tests.

### Fix

1. Build js-ceramic from this branch
```
cd js-ceramic
git checkout fix/prev-cid-conflict-resolution
npm install
npm run build
```
2. Follow repro step 1 above to run the daemon
3. Follow repro step 2 above to run the smoke tests

### DIFF
--- a/packages/core/src/conflict-resolution.ts
+++ b/packages/core/src/conflict-resolution.ts
@@ -1,4 +1,4 @@
-import type { CID } from 'multiformats/cid'
+import { CID } from 'multiformats/cid'
 import {
   AnchorProof,
   AnchorStatus,
@@ -207,7 +207,7 @@ export async function fetchLog(
   const nextCommitData = await Utils.getCommitData(dispatcher, cid, stateLog.streamId, timestamp)
   // Update the running timestamp if it was updated via an anchor commit fetch
   timestamp = nextCommitData.timestamp
-  const prevCid: CID = nextCommitData.commit.prev
+  const prevCid: CID = CID.asCID(nextCommitData.commit.prev)
   if (!prevCid) {
     // Someone sent a tip that is a fake log, i.e. a log that at some point does not refer to a previous or genesis
     // commit.


oed:
Why is the prev CID not a CID when it is retrieved from IPLD? What is the type of the property when you inspect it before casting it?

pawartur:
> Why is the prev CID not a CID when it is retrieved from IPLD?

@oed , I've been debugging this the whole day trying to answer this question and I don't have an answer yet.

> What is the type of the property when you inspect it before casting it?

It's type `object`

oed:
@pawartur what is contained within the object?

pawartur:
> @pawartur what is contained within the object?

This is an example:
```
{"code":133,"version":1,"hash":{"0":18,"1":32,"2":252,"3":197,"4":76,"5":52,"6":79,"7":8,"8":148,"9":249,"10":101,"11":13,"12":242,"13":148,"14":48,"15":135,"16":33,"17":54,"18":174,"19":181,"20":201,"21":13,"22":84,"23":205,"24":16,"25":47,"26":54,"27":126,"28":132,"29":189,"30":189,"31":254,"32":9,"33":104}}
```

pawartur:
@v-stickykeys , we've done some more debugging with @oed and it turns out that the use of outdated `did` library which (I think) was serialising objects in a different way when signing.

We bumped the version of `dids` package in the smoke tests repo from `2.4.3` to latest (`3.3.0`) and the tests started passing :).

So I guess if there are any fixes in js-ceramic related to this, I guess we can check, if the dependencies on `dids` are correct here. I'll have a look at that.

ukstv:
So, can we close the PR now?

oed:
Yes!

PR #2410, opened 2022-09-05
feat: Make SYNC_ALWAYS rewrite and revalidate local state

ukstv:


oed (commented):
Should we add a test to make sure that it works when there are invalid commits?

ukstv:
> Should we add a test to make sure that it works when there are invalid commits?

That would be tricky. Let me think. What if one of the commits in the log is wrongly signed or belongs to a weird chain as an anchor proof? How do you think?

oed:
Ideally the case that we have seen. A CACAO that has been timed out in one of the commits.
Alternatively we can just put some bogus state in the state store and make sure it's updated to the correct state.

stbrody:
@ukstv I have a WIP test that induces the problematic state in https://github.com/ceramicnetwork/js-ceramic/pull/2406/files in `capability.test.ts` that might serve as a decent starting point.

packages/core/src/state-management/state-manager.ts
```diff
@@ -78,16 +78,27 @@ export class StateManager {
...
     const tip$ = this.dispatcher.messageBus.queryNetwork(state$.id)
-    // We do not expect this promise to return anything, so set `defaultValue` to `undefined`
-    await lastValueFrom(
-      tip$.pipe(
-        takeUntil(timer(timeoutMillis)),
-        concatMap((tip) => this._handleTip(state$, tip))
-      ),
-      { defaultValue: undefined }
-    )
+    if (hint) {
```
stbrody, line 83: This makes `sync` more complicated and adds logic outside of its core function of syncing the tip from pubsub.  IMO I think this would be easier to follow if the body of `sync` was left alone and if the call site in `repository.load` explicitly called `handleTip` right after calling `sync` in the SYNC_ALWAYS case

packages/core/src/state-management/repository.ts
```diff
@@ -180,17 +181,31 @@ export class Repository {
...
+        case SyncOptions.NEVER_SYNC: {
+          const [streamState$, alreadySynced] = await this._loadGenesis(streamId)
+          return [await this.stateManager.verifyLoneGenesis(streamState$), alreadySynced]
+        }
+        case SyncOptions.SYNC_ALWAYS: {
+          const [fromNetwork$, fromMemoryOrStore] = await Promise.all([
+            this.fromNetwork(streamId),
+            this.fromMemoryOrStore(streamId),
+          ])
+          await this.stateManager.sync(fromNetwork$, opts.syncTimeoutSeconds * 1000, fromMemoryOrStore?.tip)
```
stbrody, line 203:
```suggestion
          await this.stateManager.sync(fromNetwork$, opts.syncTimeoutSeconds * 1000)
          if (fromMemoryOrStore?.tip) {
            await this.stateManager.handleTip(fromMemoryOrStore?.tip)
          }
```

stbrody: Suggestion to keep the implementation of `sync` simpler and more focused on a single task

packages/core/src/state-management/__tests__/repository.test.ts
```diff
@@ -18,166 +25,219 @@ const STRING_MAP_SCHEMA = {
...
+
+afterEach(async () => {
+  jest.resetAllMocks()
+  await ceramic.close()
+})
 
-  beforeEach(async () => {
-    ceramic = await createCeramic(ipfs)
-    repository = ceramic.repository
+describe('#load', () => {
```
stbrody, line 50: it's hard to read the diff on this file.  Is it all formatting changes or are there real changes to the test logic?

stbrody (commented):
overall approach looks good, added one optional suggestion for code cleanup/clarity.

packages/core/src/state-management/__tests__/repository.test.ts
ukstv: 1. Remove single useless `describe` statement. Thus, formatting changes.
2. Add `sync: SYNC_ALWAYS` tests.

Existing tests are the same.

packages/core/src/state-management/state-manager.ts
ukstv: I would like to counter that proposal by noting, that it would make the loading process obviously longer.

packages/core/src/state-management/state-manager.ts
```diff
@@ -75,14 +75,17 @@ export class StateManager {
...
+   * @param hint - Tip to try while we are waiting for the network to respond.
    */
-  async sync(state$: RunningState, timeoutMillis: number): Promise<void> {
+  async sync(state$: RunningState, timeoutMillis: number, hint?: CID): Promise<void> {
     const tip$ = this.dispatcher.messageBus.queryNetwork(state$.id)
+    const tipSource$ = hint ? merge(tip$, of(hint)) : tip$
     // We do not expect this promise to return anything, so set `defaultValue` to `undefined`
     await lastValueFrom(
-      tip$.pipe(
+      tipSource$.pipe(
```
stbrody, line 88: do we have to worry about any synchronization issues with syncing and applying two different logs to the same state object in parallel? 
 Could one modify the state object out from under the other?  Isn't that the reason why we have the ExecutionQueue serializing log application in other places?

stbrody (commented):
looks good mod question about concurrency

packages/core/src/state-management/state-manager.ts
ukstv: Should not be the case, as concatMap AFAIK waits till the execution of the function is over. Getting a demonstration code ready.

ukstv: https://gist.github.com/ukstv/93ff950d370fa501d9ac9f0251c95313

It works. No concurrent calls.

packages/core/src/state-management/state-manager.ts
stbrody: Sorry, how does this suggestion make the loading process longer?  I had assumed it was because the existing code in the PR applied the two commits in parallel, but now you're saying it happens serially, so I'm no longer clear why it's faster than this suggestion

stbrody: I'm confused, below you said `concatMap` processes things serially (which is important because if it didn't there'd be concurrent modifications to the state object), but here you're saying it can apply the hint commit in parallel to loading the tip from the network.  Aren't those things mutually exclusive?

stbrody: Oh nevermind, I think I get it now.  We kick off the query to the network first, then we go to the `concatMap` of applying the tips one at a time to the state object.

ukstv: Every individual commit is processed serially. You could do that while still waiting for the network to reply.

ukstv: IMG_6905

packages/core/src/state-management/repository.ts
```diff
@@ -180,17 +181,35 @@ export class Repository {
...
+            await this.stateManager.sync(streamState$, opts.syncTimeoutSeconds * 1000)
+            return [await this.stateManager.verifyLoneGenesis(streamState$), true]
+          }
+        }
+        case SyncOptions.NEVER_SYNC: {
+          const [streamState$, alreadySynced] = await this._loadGenesis(streamId)
+          return [await this.stateManager.verifyLoneGenesis(streamState$), alreadySynced]
+        }
+        case SyncOptions.SYNC_ALWAYS: {
+          const [fromNetwork$, fromMemoryOrStore] = await Promise.all([
```
stbrody, line 199:
```suggestion
          // When SYNC_ALWAYS is provided, we want to reapply and re-validate
          // the stream state.  We effectively throw out our locally stored state
          // as its possible that the commits that were used to construct that
          // state are no longer valid (for example if the CACAOs used to author them
          // have expired since they were first applied to the cached state object).
          // But if we were the only node on the network that knew about the most
          // recent tip, we don't want to totally forget about that, so we pass the tip in 
          // to `sync` so that it gets considered alongside whatever tip we learn
          // about from the network.
          const [fromNetwork$, fromMemoryOrStore] = await Promise.all([
```

packages/core/src/state-management/state-manager.ts
ukstv: Huh, GitHub should autorefresh the page on a comment.

packages/core/src/state-management/state-manager.ts
```diff
@@ -75,14 +75,17 @@ export class StateManager {
...
-   * @param state$
-   * @param timeoutMillis
+   *
+   * @param state$ - Current stream state.
+   * @param timeoutMillis - How much time do we wait for a response from the network.
+   * @param hint - Tip to try while we are waiting for the network to respond.
    */
-  async sync(state$: RunningState, timeoutMillis: number): Promise<void> {
+  async sync(state$: RunningState, timeoutMillis: number, hint?: CID): Promise<void> {
     const tip$ = this.dispatcher.messageBus.queryNetwork(state$.id)
```
stbrody, line 84:
```suggestion
    // Begin querying the network for the tip immediately.
    const tip$ = this.dispatcher.messageBus.queryNetwork(state$.id)
```

packages/core/src/state-management/state-manager.ts
```diff
@@ -75,14 +75,17 @@ export class StateManager {
...
+    const tipSource$ = hint ? merge(tip$, of(hint)) : tip$
```
stbrody, line 85:
```suggestion
    // If a 'hint' is provided we can work on applying it while the tip is
    // fetched from the network
    const tipSource$ = hint ? merge(tip$, of(hint)) : tip$
```

stbrody, line 85: Does the order of the entries within `tipSource$` matter?  Should we swap `tip$` and `of(hint)` to make sure the hint is processed first?  Or will `pipe` already handle moving on to the hint if the `tip$` promise isn't ready yet?

stbrody (approved):
lgtm, thanks for bearing with me as I got my head around the change

packages/core/src/state-management/state-manager.ts
ukstv: The order for `merge` does not matter. We `applyTip` on a CID that is _emitted_ first, and it is most probably `of(hint)` here.

PR #2421, opened 2022-09-07
feat(common): Update type definitions to support simple relations

PaulLeCam:
Opening as draft as the actual logic and tests need to be implemented, but here are the changes to the type definitions I've been working with in the tools and client so far.

The changes to the index API types to remove the `readonly` modifiers are not strictly needed, but that would be useful to use these types to construct queries on the client side, and avoid having to use `@ts-ignore` in https://github.com/ceramicstudio/js-composedb/blob/main/packages/client/src/query.ts#L15


PaulLeCam:
@stbrody @willex these should cover all the changes to the types we agreed on today.
I updated the tests but haven't added actual support for the extra indexing APIs or other logic that could be needed for the model definition, so do you think this can be merged as-is or should be taken over by the related implementation tasks please?

packages/core/src/indexing/local-index-api.ts
```diff
@@ -43,6 +43,10 @@ export class LocalIndexApi implements IndexApi {
     await this.databaseIndexApi.indexStream(args)
   }
 
+  async queryCount(query: BaseQuery): Promise<number> {
```
ukstv, line 46: Do you expect it to return number of _all_ relevant records?

packages/core/src/__tests__/ceramic-api.test.ts
```diff
@@ -41,7 +41,7 @@ describe('Ceramic API', () => {
 
   const MODEL_DEFINITION: ModelDefinition = {
     name: 'MyModel',
-    accountRelation: ModelAccountRelation.LIST,
+    accountRelation: { type: 'list' },
```
ukstv, line 44: Are you sure that does not affect existing models?
I guess, there are no existing models, but still.

packages/stream-model-handler/src/__tests__/model-handler.test.ts
```diff
@@ -110,7 +110,7 @@ const CONTENT_WITH_INVALID_VIEWS: ModelDefinition = {
 
 const CONTENT_WITH_INVALID_SCHEMA = {
   name: 'MyModel',
-  accountRelation: ModelAccountRelation.LIST,
+  accountRelation: { type: 'list' },
```
ukstv, line 113: Technicality here. While I understand you love string literals, they will make maintenance harder. Could you make `list` or `single` an enum back, please?

ukstv (changes requested):

packages/core/src/__tests__/ceramic-api.test.ts
PaulLeCam: I expect it will affect existing models, this should be considered a breaking change.

packages/stream-model-handler/src/__tests__/model-handler.test.ts
PaulLeCam: The goal is that the `ModelAccountRelation` itself is an object type, with its `type` as discriminant so it would be completely restricted by the TS compiler. If we need to use an enum, are you suggesting we have both `ModelAccountRelation` and `ModelAccountRelationType` please?

Also, do you have examples of how enums make maintenance easier please? In my experience it's the opposite, enums are a pain to reuse in "type only" modules. For example to use them in ComposeDB packages, I need to add a runtime dependency (can't be a dev dependency) on `@ceramicnetwork/stream-model` simply to import the enum, which makes the dependency tree heavier only for type constraints.

ukstv: > The goal is that the `ModelAccountRelation` itself is an object type, with its `type` as discriminant so it would be completely restricted by the TS compiler. If we need to use an enum, are you suggesting we have both `ModelAccountRelation` and `ModelAccountRelationType` please?

TS Compiler is smart enough to treat union of interfaces with distinct `type` field as a discriminated union, exactly the same as with the strings.

```
enum ModelAccountRelationType {
  LIST = 'list',
  SINGLE = 'single'
}
type ListRelation = {type: ModelAccountRelationType.LIST}
const ListRelation: ListRelation = {type: ModelAccountRelationType.LIST}
type SingleRelation = {type: ModelAccountRelationType.SINGLE}
const SingleRelation: SingleRelation = {type: ModelAccountRelationType.SINGLE}
type AccountRelation = ListRelation | SingleRelation

function foo(listRel: ListRelation) {
}
foo(ListRelation) // error
```

Also, `switch` works exhaustive too.

> Also, do you have examples of how enums make maintenance easier please? In my experience it's the opposite, enums are a pain to reuse in "type only" modules. For example to use them in ComposeDB packages, I need to add a runtime dependency (can't be a dev dependency) on `@ceramicnetwork/stream-model` simply to import the enum, which makes the dependency tree heavier only for type constraints.

I am curious, why is it a big trouble? Is it one of the biggest dependencies?

Examples of how enums make life easier:
1. Initially Ceramic had 3 pubsub messages, typed through enum. Message processing used exhaustive switch statement. Then we decided to add 4th. Now the compiler suggested where to add code to process the 4th type. Less error prone, more clear code.
2. Once we added one more SyncOption to Ceramic: `NEVER_SYNC`. It is pretty easy to find code probably affected by this addition. In case of string literals, it would be just plain search. For example, here, if we change something related to `list` type, we would have to sieve through all the `list` words to find relevant places.

PaulLeCam: > > The goal is that the `ModelAccountRelation` itself is an object type, with its `type` as discriminant so it would be completely restricted by the TS compiler. If we need to use an enum, are you suggesting we have both `ModelAccountRelation` and `ModelAccountRelationType` please?
> 
> TS Compiler is smart enough to treat union of interfaces with distinct `type` field as a discriminated union, exactly the same as with the strings.
> 
> ```
> enum ModelAccountRelationType {
>   LIST = 'list',
>   SINGLE = 'single'
> }
> type ListRelation = {type: ModelAccountRelationType.LIST}
> const ListRelation: ListRelation = {type: ModelAccountRelationType.LIST}
> type SingleRelation = {type: ModelAccountRelationType.SINGLE}
> const SingleRelation: SingleRelation = {type: ModelAccountRelationType.SINGLE}
> type AccountRelation = ListRelation | SingleRelation
> 
> function foo(listRel: ListRelation) {
> }
> foo(ListRelation) // error
> ```
> 
> Also, `switch` works exhaustive too.

Yes I agree, that's where I don't see any advantage of using enums instead of regular type objects, considering TS will provide the same safety either way?

> > Also, do you have examples of how enums make maintenance easier please? In my experience it's the opposite, enums are a pain to reuse in "type only" modules. For example to use them in ComposeDB packages, I need to add a runtime dependency (can't be a dev dependency) on `@ceramicnetwork/stream-model` simply to import the enum, which makes the dependency tree heavier only for type constraints.
> 
> I am curious, why is it a big trouble? Is it one of the biggest dependencies?

Not the biggest, but to me the question is more the other way around: why add runtime dependencies that are not strictly needed?

> Examples of how enums make life easier:
> 
>     1. Initially Ceramic had 3 pubsub messages, typed through enum. Message processing used exhaustive switch statement. Then we decided to add 4th. Now the compiler suggested where to add code to process the 4th type. Less error prone, more clear code.
> 
>     2. Once we added one more SyncOption to Ceramic: `NEVER_SYNC`. It is pretty easy to find code probably affected by this addition. In case of string literals, it would be just plain search. For example, here, if we change something related to `list` type, we would have to sieve through all the `list` words to find relevant places.

Thanks, that's good examples. Indeed it makes it easier for search in the code-base, but I've been through similar cases refactoring the ComposeDB logic to add the extra relations and the experience is pretty much the same, the TS compiler will error if there are missing cases due to the added types, I don't think enums would have made a difference here?

To me really the biggest concern is the need for extra runtime dependency on this package. It's essentially "leaking" the type abstraction to any downstream app/package, leading to more risks of duplicate dependencies, mismatches, etc.
We also haven't made any effort to address overall bundle size recently but I think it will be something to put more effort on over time, so the more we can limit the number of runtime dependencies we need the better.

ukstv: Oh, okay. Let it be this way with the literals.

packages/core/src/indexing/local-index-api.ts
PaulLeCam: Yes all the records matching the query parameters (account, model and filter(s)).

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
```diff
@@ -228,7 +228,7 @@ export class ModelInstanceDocumentHandler implements StreamHandler<ModelInstance
    * @param header - the header to validate
    */
   async _validateHeader(model: Model, header: ModelInstanceDocumentHeader): Promise<void> {
-    if (model.content.accountRelation == ModelAccountRelation.SINGLE) {
+    if (model.content.accountRelation.type === 'single') {
```
stbrody, line 231: I didn't totally follow the whole discussion above about enums and bundle size, but can we have `ModelAccountRelationType` string constants so we're not using raw strings here?  Makes it clearer to readers of the code that this isn't an arbitrary string field but has a pre-defined set of supported values that need to be considered.

packages/stream-model/src/model.ts
```diff
@@ -72,10 +72,31 @@ async function throwReadOnlyError(): Promise<void> {
...
-  SINGLE = 'single',
-}
+export type ModelAccountRelation = { type: 'list' } | { type: 'single' }
+
+/**
+ * Identifies types of properties that are supported as relations by the indexing service.
+ *
+ * Currently supported types of relation properties:
+ * - 'account': references a DID property
+ * - 'document': references a SteamID property with associated 'model' the related document must use
```
stbrody, line 82:
```suggestion
 * - 'document': references a StreamID property with associated 'model' the related document must use
```

packages/stream-model/src/model.ts
```diff
@@ -72,10 +72,31 @@ async function throwReadOnlyError(): Promise<void> {
...
+export type ModelRelationDefinition = { type: 'account' } | { type: 'document'; model: string }
+
+/**
+ * A mapping between model's property names and types of relation properties
+ *
+ * It indicates which properties of a model are relation properties and of what type
+ */
+export type ModelRelationsDefinition = Record<string, ModelRelationDefinition>
+
+export type ModelDocumentViewDefinition = { type: 'documentAccount' } | { type: 'documentVersion' }
```
stbrody, line 94: `ModelDocumentMetadataViewDefinition`?  Too verbose?

stbrody (commented):
looking good!

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
PaulLeCam: That's what the `ModelAccountRelation` type ensures, here the `type` can't be any value, the TS compiler will throw if it's not `single` or `list`.

packages/common/src/index-api.ts
```diff
@@ -27,16 +27,20 @@ export type Pagination = ForwardPagination | BackwardPagination
...
 
+export type IndexQuery = BaseQuery & Pagination
+
 /**
  * API to query an index.
  */
 export interface IndexApi {
-  queryIndex(query: BaseQuery & Pagination): Promise<Page<StreamState>>
+  queryCount(query: BaseQuery): Promise<number>
+  queryIndex(query: IndexQuery): Promise<Page<StreamState>>
```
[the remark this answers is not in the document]
PaulLeCam: Yes sounds good to me, I'll update it.

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
stbrody: sure, but it's about more than the compiler catching errors, it's about code legibility.  When I see a raw string constant I wonder "what else can this be?  Is this string user-provided?  Do I need to handle arbitrary input?".  When I see a namespaced constant like `ModelAccountRelationType.SINGLE` - I know immediately that we're dealing with a fixed set of supported values

PaulLeCam: Yeah that's the conversation we had with Sergey in another comment above, the problem with enums is that they are not "just types", they are actual objects, and so to use them in consuming packages, we need to import the actual model package rather than just the types.
It makes the runtime/production dependency tree heavier than it needs to be, for no additional type safety, to me the cost is really not worth the benefits, but if you feel strongly about it I think we should discuss other alternatives, such as a dedicated package that only contains such constants so we don't need to import the rest of the code?

stbrody: yeah, at some point in the future maybe having a separate "constants" package would make sense, probably not worth it right now though.  I do still prefer the enum approach for legibility, but I also don't know how big of an impact on bundle size it actually makes in practice.  In any case I don't feel so strongly about this - if you think the bundle size difference is meaningful then I'm fine to leave this as-is

PaulLeCam: We're not tracking bundle size yet so I don't know either, the problem I see is that by the time we get there, we'll probably already have shipped a stable version of these packages and we won't be able to make these kinds of breaking changes anymore without having to bump a major version.
If you think we can get to the point of having a separate constants package by the time we release a first stable version of the model package I think it's OK, but otherwise better use types than enums I think.

packages/stream-model/src/model.ts
PaulLeCam: @stbrody what do you think about future support for a "document history" view that represents a collection of all the versions of the document please?
I don't think it would really fit the "metadata" naming convention, do you? Should we use `ModelDocumentMetadataViewDefinition` and `ModelDocumentViewDefinition = ModelDocumentMetadataViewDefinition | { type: 'documentHistory' }` for example?

stbrody: yeah, that would seem sufficiently different to me that it merits its own category.

stbrody (approved):
lgtm

stbrody:
Although please update the commit message to proper semver style.  "feat(common): ..."

ukstv (approved):
🏇 

I see nothing suspicious. LGTM.

PR #2428, opened 2022-09-13
fix(stream-tile-handler): load historical commits with CACAOs

oed:
Timestamp information was not properly propagated to the event log.

### DIFF
--- a/packages/core/src/__tests__/ceramic.test.ts
+++ b/packages/core/src/__tests__/ceramic.test.ts
@@ -175,7 +175,7 @@ describe('Ceramic integration', () => {
       let stream2 = await TileDocument.createFromGenesis(ceramic2, logCommits[0].value, {
         anchor: false,
         publish: false,
-        sync: 2, // NEVER_SYNC
+        sync: SyncOptions.NEVER_SYNC
       })
       for (let i = 1; i < logCommits.length; i++) {
         stream2 = await ceramic2.applyCommit(stream2.id, logCommits[i].value, {
--- a/packages/core/src/__tests__/ceramic.test.ts
+++ b/packages/core/src/__tests__/ceramic.test.ts
@@ -140,7 +140,12 @@ describe('Ceramic integration', () => {
       await TestUtils.waitForState(
         stream3,
         2000,
-        (state) => StreamUtils.statesEqual(state, stream1.state),
+        (state) => {
+          // The timestamp deleted here will only be present if the log
+          // is synced from the network. Log[1] is a signed commit.
+          delete state.log[1].timestamp
+          return StreamUtils.statesEqual(state, stream1.state)
+	},
         () => {
           throw new Error(`streamtype3.state should equal streamtype1.state`)
         }
@@ -170,6 +175,7 @@ describe('Ceramic integration', () => {
       let stream2 = await TileDocument.createFromGenesis(ceramic2, logCommits[0].value, {
         anchor: false,
         publish: false,
+        sync: 2, // NEVER_SYNC
       })
       for (let i = 1; i < logCommits.length; i++) {
         stream2 = await ceramic2.applyCommit(stream2.id, logCommits[i].value, {
--- a/packages/stream-model-instance-handler/src/model-instance-document-handler.ts
+++ b/packages/stream-model-instance-handler/src/model-instance-document-handler.ts
@@ -119,13 +119,15 @@ export class ModelInstanceDocumentHandler implements StreamHandler<ModelInstance
     await this._validateContent(modelStream, payload.data, true)
     await this._validateHeader(modelStream, payload.header)
 
+    const entry = { cid: commitData.cid, type: CommitType.SIGNED } as CommitData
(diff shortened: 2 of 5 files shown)


oed:
This likely needs to be fixed in MID as well?

packages/core/src/__tests__/ceramic.test.ts
```diff
@@ -140,7 +140,12 @@ describe('Ceramic integration', () => {
       await TestUtils.waitForState(
         stream3,
         2000,
-        (state) => StreamUtils.statesEqual(state, stream1.state),
+        (state) => {
+          // The timestamp deleted here will only be present if the log
```
stbrody, line 144: I don't quite understand why this is necessary.  The commits *are* being synced over the network, via pubsub UPDATE messages?

stbrody: Oh, is the issue that if we apply a signed commit, and then later on apply an anchor commit, we don't back-propagate the timestamp into the CommitDatas for earlier commits in the log?

If so, isn't that a problem for this whole approach?  If the stream state was updated one commit at a time instead of synced via the network all at once, then loading at a CommitID will fail again since the state log won't have the timestamp information where it's needed?  I think loading a tip and syncing the full log at once vs loading and applying commits one at a time as they come in through pubsub UPDATE messages, should both result in the same stream state.  If they don't, that's a bug IMO.  So we should probably make applying an anchor commit actively propagate the timestamp back to earlier entries in the log (until the previous anchorCommit)

packages/core/src/__tests__/ceramic.test.ts
```diff
@@ -170,6 +175,7 @@ describe('Ceramic integration', () => {
       let stream2 = await TileDocument.createFromGenesis(ceramic2, logCommits[0].value, {
         anchor: false,
         publish: false,
+        sync: 2, // NEVER_SYNC
```
stbrody, line 178:
```suggestion
        sync: SyncOptions.NEVER_SYNC
```

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
```diff
@@ -119,13 +119,15 @@ export class ModelInstanceDocumentHandler implements StreamHandler<ModelInstance
     await this._validateContent(modelStream, payload.data, true)
     await this._validateHeader(modelStream, payload.header)
 
+    const entry = { cid: commitData.cid, type: CommitType.GENESIS } as CommitData
```
stbrody, line 122:
```suggestion
    const entry = { cid: commitData.cid, type: CommitType.GENESIS } as LogEntry
```

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
```diff
@@ -169,7 +171,9 @@ export class ModelInstanceDocumentHandler implements StreamHandler<ModelInstance
     nextState.signature = SignatureStatus.SIGNED
     nextState.anchorStatus = AnchorStatus.NOT_REQUESTED
     nextState.content = newContent
-    nextState.log.push({ cid: commitData.cid, type: CommitType.SIGNED })
+    const entry = { cid: commitData.cid, type: CommitType.SIGNED } as CommitData
```
stbrody, line 174:
```suggestion
    const entry = { cid: commitData.cid, type: CommitType.SIGNED } as LogEntry
```

packages/stream-tile-handler/src/tile-document-handler.ts
```diff
@@ -100,13 +100,15 @@ export class TileDocumentHandler implements StreamHandler<TileDocument> {
       throw new Error('Exactly one controller must be specified')
     }
 
+    const entry = { cid: commitData.cid, type: CommitType.GENESIS } as CommitData
```
stbrody, line 103:
```suggestion
    const entry = { cid: commitData.cid, type: CommitType.GENESIS } as LogEntry
```

packages/stream-tile-handler/src/tile-document-handler.ts
```diff
@@ -171,7 +173,9 @@ export class TileDocumentHandler implements StreamHandler<TileDocument> {
     nextState.signature = SignatureStatus.SIGNED
     nextState.anchorStatus = AnchorStatus.NOT_REQUESTED
 
-    nextState.log.push({ cid: commitData.cid, type: CommitType.SIGNED })
+    const entry = { cid: commitData.cid, type: CommitType.SIGNED } as CommitData
```
stbrody, line 176:
```suggestion
    const entry = { cid: commitData.cid, type: CommitType.SIGNED } as LogEntry
```

stbrody (commented):
I think this really needs a unit test of the original bug.  Also need to update model_handler as well

packages/core/src/__tests__/ceramic.test.ts
oed: Yep, this is a good point.

stbrody:
Closing in favor of https://github.com/ceramicnetwork/js-ceramic/pull/2523

PR #2431, opened 2022-09-14
Allow empty states for indexed streams

pawartur:
## Allow empty states for indexed streams

It may happen that a stream is indexed, but it's not possible to retrieve it from the repository (e.g. because it wasn't pinned, was unpinned or perhaps other reasons). In such case, we don't want to (for now at least) an index query to throw an error, but we want to return nulls instead.

## How Has This Been Tested?

- [X] Ran the tests

## PR checklist

### DIFF
--- a/packages/common/src/index-api.ts
+++ b/packages/common/src/index-api.ts
@@ -36,7 +36,7 @@ export interface BaseQuery {
  * API to query an index.
  */
 export interface IndexApi {
-  queryIndex(query: BaseQuery & Pagination): Promise<Page<StreamState>>
+  queryIndex(query: BaseQuery & Pagination): Promise<Page<StreamState | null>>
 }
 
 export type Edge<T> = {
--- a/packages/core/src/indexing/__tests__/local-index-api.test.ts
+++ b/packages/core/src/indexing/__tests__/local-index-api.test.ts
@@ -68,7 +68,7 @@ describe('with database backend', () => {
     }
     const pageFn = jest.fn(async () => backendPage)
     const streamStateFn = jest.fn(async (streamId: any) => {
-      throw Error(`Can't get the stream state`)
+      return undefined
     })
     const fauxBackend = { page: pageFn } as unknown as DatabaseIndexApi
     const fauxRepository = { streamState: streamStateFn } as unknown as Repository
--- a/packages/core/src/indexing/__tests__/local-index-api.test.ts
+++ b/packages/core/src/indexing/__tests__/local-index-api.test.ts
@@ -51,14 +51,14 @@ describe('with database backend', () => {
   })
 
   test('return empty page from the database if cannot retrieve stream state', async () => {
-    const query = { model: 'foo', first: randomInt(100) }
+    const query = { model: 'foo', first: 1 }
     const backendPage: Page<string> = {
-      edges: Array.from({ length: query.first }).map(() => {
-        return {
+      edges: [
+        {
           cursor: randomString(32),
           node: null,
         }
-      }),
+      ],
       pageInfo: {
         hasPreviousPage: false,
         hasNextPage: false,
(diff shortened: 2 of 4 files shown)


stbrody:
Strongly disagree with this direction.  If data is missing from the state store that's a major bug and we should absolutely throw an error in that case.  Just returning null could mask a serious data persistence bug

stbrody:
Instead I'd rather focus on making the error message more descriptive when this happens

pawartur:
Yeah, I actually agree with you, @stbrody .

pawartur:
> Instead I'd rather focus on making the error message more descriptive when this happens

Although, I don't agree with this, tbh. My understanding is that for the majority of users trying to query the index, the error message will not be useful in any way, because they won't be able to do anything to fix the state. So I guess that we should do something more than just making the error message more comprehensive.

pawartur:
Ok, we've agreed that this is indeed the way to go. Looking forward to getting a review once tests pass :).

packages/core/src/indexing/local-index-api.ts
```diff
@@ -49,15 +49,22 @@ export class LocalIndexApi implements IndexApi {
...
-  async queryIndex(query: BaseQuery & Pagination): Promise<Page<StreamState>> {
+  async queryIndex(query: BaseQuery & Pagination): Promise<Page<StreamState | null>> {
     if (this.databaseIndexApi) {
       const page = await this.databaseIndexApi.page(query)
       const edges = await Promise.all(
         // For database queries we bypass the stream cache and repository loading queue
         page.edges.map(async (edge) => {
+          let node: StreamState = null
+          try {
+            node = await this.repository.streamState(edge.node)
```
ukstv, line 60: It does not throw on empty state. It returns `undefined`. So, instead of catching an error, here it should be comparison to `undefined`.

Even if the code here is changed, the change will not completely eliminate the error we see on the API.

LocalIndexAPI#queryIndex result is fed to https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/cli/src/ceramic-daemon.ts#L484 which throws the observed error. So, we should fix that part in CeramicDaemon.

TBH, that should have been caught by the compiler. I expected it to be caught really. `e.node` there is nullable, it is why it throws.

ukstv (changes requested):

packages/core/src/indexing/__tests__/local-index-api.test.ts
```diff
@@ -48,6 +49,49 @@ describe('with database backend', () => {
...
+        endCursor: 'endCursor',
+      },
+    }
+    const pageFn = jest.fn(async () => backendPage)
+    const streamStateFn = jest.fn(async (streamId: any) => {
+      return undefined
+    })
+    const fauxBackend = { page: pageFn } as unknown as DatabaseIndexApi
+    const fauxRepository = { streamState: streamStateFn } as unknown as Repository
+    const fauxLogger = {
```
ukstv, line 75: we should check if the message is logged. one more expect statement to check if `warn` is called , and it is ok by me

ukstv (approved):

packages/common/src/index-api.ts
```diff
@@ -36,7 +36,7 @@ export interface BaseQuery {
  * API to query an index.
```
stbrody, line 36: Can we add a comment when this could return null, and what that would mean

packages/core/src/indexing/local-index-api.ts
```diff
@@ -49,15 +49,21 @@ export class LocalIndexApi implements IndexApi {
...
-  async queryIndex(query: BaseQuery & Pagination): Promise<Page<StreamState>> {
+  async queryIndex(query: BaseQuery & Pagination): Promise<Page<StreamState | null>> {
     if (this.databaseIndexApi) {
       const page = await this.databaseIndexApi.page(query)
       const edges = await Promise.all(
         // For database queries we bypass the stream cache and repository loading queue
         page.edges.map(async (edge) => {
+          let node: StreamState = await this.repository.streamState(edge.node)
+          if (node === undefined) {
+            this.logger.warn(`Cannot get stream state from repository. Unable to serve query ${JSON.stringify(query)}`)
```
stbrody, line 60: Can we make this error message more informative to the node operator?  How about:
`Did not find stream state for streamid ${edge.node.toString()} in our state store when serving an indexed query.  This may indicate a problem with data persistence of your state store, which can result in data loss.  Please check that your state store is properly configured with strong persistence guarantees.  This query may have incomplete results.  Affected query: ${JSON.stringify(query)}`

stbrody (commented):
overall looks good, but have 2 small suggestions for a follow-up

PR #2432, opened 2022-09-15
Don't try to deserialize null stream states. Also added a few more co…

pawartur:
## Don't try to deserialize null stream states. Also added a few more comments and a better logger warn

As we made it possible for index queries to return null nodes in https://github.com/ceramicnetwork/js-ceramic/pull/2431 , it may happen now that an http-client tries to deserialise a null node that it had got from the index api. This PR handles this case on low level.

## How Has This Been Tested?

- [X] Wrote a test to check that StreamUtils.deserializeState doesn't fail when given a null state

## PR checklist

### DIFF
--- a/packages/common/src/utils/stream-utils.ts
+++ b/packages/common/src/utils/stream-utils.ts
@@ -138,10 +138,12 @@ export class StreamUtils {
   }
 
   /**
-   * Deserializes stream cloned from over the network transfer
+   * Deserializes stream cloned from over the network transfer. Returns null if given null as a param.
    * @param state - Stream cloned
    */
-  static deserializeState(state: any): StreamState {
+  static deserializeState(state: any | null): StreamState | null {
+    if (state === null) return null
+
     const cloned = cloneDeep(state)
 
     if (cloned.doctype) {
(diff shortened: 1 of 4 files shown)


packages/common/src/utils/stream-utils.ts
```diff
@@ -138,10 +138,12 @@ export class StreamUtils {
   }
 
   /**
-   * Deserializes stream cloned from over the network transfer
+   * Deserializes stream cloned from over the network transfer. Returns null if given null as a param.
    * @param state - Stream cloned
    */
-  static deserializeState(state: any): StreamState {
+  static deserializeState(state: any | null): StreamState | null {
+    if (state === null) return null
```
ukstv, line 145:
```suggestion
    if (!state) return null
```

ukstv: There should also be a test for this.

packages/common/src/utils/stream-utils.ts
```diff
@@ -138,10 +138,12 @@ export class StreamUtils {
   }
 
   /**
-   * Deserializes stream cloned from over the network transfer
+   * Deserializes stream cloned from over the network transfer. Returns null if given null as a param.
    * @param state - Stream cloned
    */
-  static deserializeState(state: any): StreamState {
+  static deserializeState(state: any | null): StreamState | null {
```
ukstv, line 144:
```suggestion
  static deserializeState(state: any): StreamState | null {
```

as `any` includes `null`

ukstv (changes requested):

ukstv (approved):

PR #2433, opened 2022-09-15
chore(core): Clean up indexing initialization

stbrody:
No functionality change yet, just code restructuring to begin to decouple the config file from the process of indexing new models.  This should be helpful both for adding relations-specific arguments as well as for the admin api


willex (approved):
Awesome, thank you for the AdminAPI prep here! 🚀

PR #2435, opened 2022-09-16
feat(core): Add functionality for building tables with columns for relations

stbrody:


packages/core/src/indexing/postgres/postgres-index-api.ts
```diff
@@ -63,10 +63,11 @@ export class PostgresIndexApi implements DatabaseIndexApi {
     await verifyTables(this.dbConnection, models, tableStructure)
   }
 
-  async indexModels(models: Array<StreamID>): Promise<void> {
-    await initTables(this.dbConnection, models, this.logger)
-    await this.verifyTables(models)
-    this.modelsToIndex.push(...models)
+  async indexModels(models: Array<IndexModelArgs>): Promise<void> {
+    const modelStreamIDs = models.map((args) => args.model)
```
stbrody, line 67: ignore postgres for now

packages/core/src/indexing/postgres/__tests__/postgres-index-api.test.ts
```diff
@@ -70,66 +81,177 @@ afterAll(async () => {
...
+
+  describe('table validation tests', () => {
+    /**
+     * This test exists mostly to validate the rest of the validation tests.
+     * Those tests test that validation fails in certain cases by changing one thing
+     * off from what we expect to pass validation.  This test confirms that we are in fact
+     * creating tables in this unit test in the correct way that could pass validation if not
+     * for the intentional changes we make in the following test.
+     *
+     * NOTE: If this test ever fails and needs updating, update the other validation tests
```
stbrody, line 163: This is pretty brittle and I don't love it.  I had some time while working on these tests where the test was passing because validation was failing, but it was failing for the wrong reason and so wasn't really testing what I thought it was.  I happened to notice and figure it out, but it seems pretty brittle and likely that in the future as this evolves we may accidentally introduce a regression in our test coverage without noticing.  Unfortunately there's no great way to test for *why* schema validation failed so the test could check that it fails for the right reason, so I don't see a good way out of this at the moment.

stbrody: I guess instead of having the schema validation just check the JSON string of the found schema against the expected schema object, it could go field-by-field and check that the schema matches exactly what we expect, throwing a specific error message for each type of failure from each type of schema mismatch, then tests could check for the specific failure message it expects.

I don't really think that's worth it to do at the moment though.  It would be better to find (or build) a tool to do that type of validation in a generic and reusable way.

packages/core/src/indexing/postgres/init-tables.ts
```diff
@@ -21,27 +25,36 @@ export async function listMidTables(dataSource: Knex): Promise<Array<string>> {
...
-  modelsToIndex: Array<StreamID>,
+  modelsToIndex: Array<IndexModelArgs>,
   logger: DiagnosticsLogger
 ) {
-  const expectedTables = modelsToIndex.map(asTableName).map((tableName) => {
+  for (const modelIndexArgs of modelsToIndex) {
+    const tableName = asTableName(modelIndexArgs.model)
     if (tableName.length > 63) {
-      logger.err(`Invalid model added to config file: ${tableName}`)
-      process.exit(-1)
```
stbrody, line 35: I got rid of this `process.exit` call and changed it to just throw a normal Error.  I was afraid of this as a potential DOS security risk - if someone sends a request to the admin API to index a model with a long streamid now you can force the node to crash.  They'd need to have signed the request with an admin key, but still made me nervous. Just failing the current request instead of the whole process seems totally reasonable to me anyway

stbrody:
Okay, this should be ready for full review now @willex @ukstv

packages/core/src/indexing/postgres/init-tables.ts
willex: Yeah, I'd agree that throwing the error is reasonable now especially in the context of the admin API - the exit was also tied to the config file update necessitating a node restart so that we catch any potential model misconfiguration actively.

willex (approved):
LGTM

PR #2441, opened 2022-09-19
feat: Ceramic asks CAS to anchor indefinitely until some ok response

ukstv:


packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -113,16 +119,21 @@ export class EthereumAnchorService implements AnchorService {
...
+          method: 'POST',
+          body: {
+            streamId: cidStreamPair.streamId.toString(),
+            docId: cidStreamPair.streamId.toString(),
+            cid: cidStreamPair.cid.toString(),
+          },
+        })
+      )
     ).pipe(
+      retry({
```
stbrody, line 134: I just did some local testing with this PR and it looks like it swallows the error message when the request to the CAS fails.  I think it's important that it node operators still be able to tell that they're having issues connecting with the CAS, so I think this needs to log an error whenever it fails to submit a request to the CAS

ukstv: makes sense.

stbrody (approved):

willex (approved):

PR #2442, opened 2022-09-20
Added stubs for the admin api /models endpoints

pawartur:
## Implement the AdminAPIs for model indexing feature, connect and them with DatabaseApis

* added the new `admin/models` endpoint to `ceramic-daemon` with support of `GET`, `POST,` `PUT` and `DELETE` methods
* **no authentication for these call yet**
* added a new `AdminApi` interface methods for handling CRUD of models to index to
* added the new `AdminApi` to `CeramicApi`
* implemented the `AdminApi` for `http-client` as `RemoteAdminApi` (it calls the new endpoint)
* added a method to stop indexing models to `LocalIndexApi`
* implemented the `AdminApi` for `core` client as `LocalAdminApi` (it calls the `LocalIndexApi` to update the list of indexed models

## How Has This Been Tested?

- [X] Added tests for `ceramic-daemon`
- [X] Added tests for `http-client`

## PR checklist


packages/cli/src/ceramic-daemon.ts
```diff
@@ -489,6 +495,31 @@ export class CeramicDaemon {
     })
   }
 
+  async getModelsFromIndex(req: Request, res: Response): Promise<void> {
```
stbrody, line 498: `getIndexedModels`?  `getModelsFromIndex` sounds like you're querying the index itself rather than the configuration about what **will be** indexed

packages/core/src/ceramic.ts
```diff
@@ -22,7 +22,7 @@ import {
   SyncOptions,
   AnchorValidator,
   AnchorStatus,
-  StreamState,
+  StreamState, AdminApi
```
stbrody, line 25: don't forget to run prettier whenever you're modifying files in js-ceramic

packages/http-client/src/remote-admin-api.ts
```diff
@@ -0,0 +1,52 @@
...
+      body: { models: modelsIDs },
+    })
+  }
+
+  async getModelsFromIndex(): Promise<Array<Stream>> {
+    // TODO: Is this the simplest way to deserialize and return a stream? Should this operate on StreamStates rather than streams?
+    const response= await fetchJson(this.getUrl())
+    return response.models.map((modelStreamState: StreamState) => {
+      return new Model(
+        new Document(StreamUtils.deserializeState(modelStreamState), this._apiUrl, 1000000),
```
stbrody, line 34: We should use the actual sync interval, not an arbitrary large number

stbrody (commented):
looking good

packages/cli/src/ceramic-daemon.ts
```diff
@@ -489,6 +495,31 @@ export class CeramicDaemon {
...
+    const indexedModelStreams = await this.ceramic.admin.getModelsFromIndex()
+    res.json({
+      models: indexedModelStreams.map(modelStream => StreamUtils.serializeState(modelStream.state))
```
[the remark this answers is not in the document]
pawartur: +1, @stbrody . But also happy to add the `is_indexed` field anytime, if we feel it's useful, @v-stickykeys , @willex . Please share your thoughts :).

packages/common/src/ceramic-api.ts
```diff
@@ -54,6 +54,37 @@ export interface CeramicSigner extends CeramicCommon {
...
 }
 
+/**
+ * Describes Ceramic Admin API functionality
+ */
+export interface AdminApi {
+  /**
+   * List indexed model streams
+   */
+  getModelsFromIndex(): Promise<Array<Stream>>
```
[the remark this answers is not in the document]
pawartur: That's a great call and something I wasn't sure about, thanks @stbrody !

Currently, if I understand correctly, `Model` is defined in `packages/stream-model` which depends on `packages/common` (where `CeramicAPI` is defined. So if I want to make the signatures more explicit, like you suggested, I either need to move the definition of `Model` to `packages/common` or do I deeper refactoring?

WDYT? Is there a better way of doing this?

packages/cli/src/ceramic-daemon.ts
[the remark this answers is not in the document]
willex: `is_indexed` is meant to provide a logical delete option & filter for actively indexed models on the node via the `ceramic_models` table. So, looking for actively indexed models against the DB we should filter with `is_indexed IS TRUE`.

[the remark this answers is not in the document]
stbrody: right, I get why we have this field in the database.  My question is why we would expose it in the HTTP api to our end users, vs just filtering out the models where is_indexed is false in the db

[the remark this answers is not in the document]
stbrody: For context, see the spec for GET requests: https://www.notion.so/threebox/Admin-API-for-dynamic-indexing-039ea68c9d374c1987b3e09cd745296a#2a22fa99a01844778045595d1c26c881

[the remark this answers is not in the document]
stbrody: - response body contains models and their fields from db
    - `{models: [{model, is_indexed}, ...]}`

packages/common/src/ceramic-api.ts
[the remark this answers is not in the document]
stbrody: this would be a good question for @PaulLeCam or @ukstv who understand Typescript better than I do.  I believe you should be able to have the CeramicApi interface work with Stream but have the actual `Ceramic` and `CeramicClient` implementations of that interface specify that this actually a Model.

I'm also not totally clear on the difference between this interface specifying that the return type is `Stream` vs that it is `T extends Stream`

packages/cli/src/ceramic-daemon.ts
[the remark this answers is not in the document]
willex: ah IC, I had a discussion around that with @pawartur earlier.

Since we allow explicit deletes (and potentially future implicit deletes through deployments as discussed with the `PUT` method) via the API it is a way to confirm that a previously indexed model was indeed deactivated without the need to directly connect to the DB. 

If we only show active ones by default we lose an important part of the audit trail that e.g. a model was deactivated by accident by a different developer or admin and thats why it doesn't show up etc. 

There is also the option that we drop `is_indexed` and just implement a filter flag on the `GET` endpoint to query for deactivated models as well.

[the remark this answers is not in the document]
stbrody: I'm not really convinced.  To me it's a pretty simple API to just see which models are currently indexed and nothing else.  Making apps inspect and filter on the is_indexed field seems to complicate the API for little value.

That said I don't feel *that* strongly about it.  If yall are all in agreement to include the field then I won't stand in the way

packages/core/src/ceramic.ts
```diff
@@ -270,6 +275,10 @@ export class Ceramic implements CeramicApi {
     return new LocalPinApi(this.repository, this._logger)
   }
 
+  private _buildAdminApi(indexApi: LocalIndexApi, logger: DiagnosticsLogger): AdminApi {
```
ukstv, line 278: why method? why not just `this.admin = new LocalAdminApi(indexApi, logger)`?

packages/core/src/indexing/local-index-api.ts
```diff
@@ -100,11 +100,12 @@ export class LocalIndexApi implements IndexApi {
...
 
-  async indexModels(models: Array<StreamID> | null): Promise<void> {
-    if (!models) {
-      return
-    }
+  indexedModels(): Array<StreamID> {
+    return this.databaseIndexApi?.getActiveModelsToIndex()
+  }
 
+  async indexModels(models: Array<StreamID> | null): Promise<void> {
```
ukstv, line 107: `null` and/or `undefined`?

packages/core/src/local-admin-api.ts
```diff
@@ -0,0 +1,37 @@
...
+    return Promise.resolve(this.indexApi.indexedModels() ?? [])
+  }
+
+  async removeModelsFromIndex(modelsIDs: Array<StreamID>): Promise<void> {
+    this.logger.log(LogStyle.info, `Local Admin Api will remove model ids to index ${modelsIDs}`)
+    await this.indexApi.stopIndexingModels(modelsIDs)
+    this.logger.log(LogStyle.info, `Local Admin Api did remove model ids to index ${modelsIDs}`)
+  }
+
+  async replaceModelsInIndex(modelsIDs: Array<StreamID>): Promise<void> {
```
ukstv, line 30: Non-actionable. I have a bad feeling about it. This `replace` thing might be a source of confusion.

packages/common/src/ceramic-api.ts
```diff
@@ -54,6 +54,37 @@ export interface CeramicSigner extends CeramicCommon {
...
+   * @param modelsIDs - array of model stream IDs to add to index
+   */
+  addModelsToIndex(modelsIDs: Array<StreamID>): Promise<void>
+
+  /**
+   * Removes model streams from index
+   *
+   * @param modelsIDs - array of model stream IDs to remove from index
+   */
+  removeModelsFromIndex(modelsIDs: Array<StreamID>): Promise<void>
```
ukstv, line 78: Do we _stop indexing_ or _remove_ the data from index? o_O

ukstv: IMO, it should be `stop`-something like in `LocalIndexApi`.

packages/cli/src/ceramic-daemon.ts
```diff
@@ -492,6 +499,61 @@ export class CeramicDaemon {
     })
   }
 
+  async getIndexedModels(req: Request, res: Response): Promise<void> {
+    const indexedModelStreamIDs = await this.ceramic.admin.getIndexedModels()
+    res.json({
+      models: indexedModelStreamIDs.map(modelStreamID => modelStreamID.toString())
```
ukstv, line 505:
```suggestion
      models: indexedModelStreamIDs.map(String)
```

nitpick, I know.

ukstv (approved):
looks good, with few nitpicks

packages/cli/src/ceramic-daemon.ts
```diff
@@ -299,6 +302,10 @@ export class CeramicDaemon {
     documentsRouter.getAsync('/:docid', this.stateOld.bind(this)) // Deprecated
     recordsRouter.getAsync('/:streamid', this.commits.bind(this)) // Deprecated
     collectionRouter.getAsync('/', this.getCollection.bind(this))
+    adminModelRouter.getAsync('/', this.getIndexedModels.bind(this))
+    adminModelRouter.postAsync('/', this.addModelsToIndex.bind(this))
```
stbrody, line 306: `startIndexingModels`, `stopIndexingModels, and `replaceIndexedModels` to match the names of the AdminApi methods?

packages/core/src/local-admin-api.ts
```diff
@@ -0,0 +1,37 @@
...
+export class LocalAdminApi implements AdminApi {
+  constructor(
+    private readonly indexApi: LocalIndexApi,
+    private readonly logger: DiagnosticsLogger
+  ) {}
+
+  async startIndexingModels(modelsIDs: Array<StreamID>): Promise<void> {
+    this.logger.log(LogStyle.info, `Local Admin Api will add model ids to index ${modelsIDs}`)
+    await this.indexApi.indexModels(modelsIDs)
+    this.logger.log(LogStyle.info, `Local Admin Api did add model ids to index ${modelsIDs}`)
```
stbrody, line 17: logging both before and after the operation feels like overkill to me.  I'd suggest doing it before only

packages/core/src/local-admin-api.ts
```diff
@@ -0,0 +1,37 @@
...
+ * AdminApi for Ceramic core.
+ */
+export class LocalAdminApi implements AdminApi {
+  constructor(
+    private readonly indexApi: LocalIndexApi,
+    private readonly logger: DiagnosticsLogger
+  ) {}
+
+  async startIndexingModels(modelsIDs: Array<StreamID>): Promise<void> {
+    this.logger.log(LogStyle.info, `Local Admin Api will add model ids to index ${modelsIDs}`)
```
stbrody, line 15:
```suggestion
    this.logger.log(LogStyle.info, `Adding models to index: ${modelsIDs}`)
```

stbrody: I don't think "Local Admin Api" is going to mean anything to node operators

packages/core/src/local-admin-api.ts
```diff
@@ -0,0 +1,37 @@
...
+    await this.indexApi.indexModels(modelsIDs)
+    this.logger.log(LogStyle.info, `Local Admin Api did add model ids to index ${modelsIDs}`)
+  }
+
+  getIndexedModels(): Promise<Array<StreamID>> {
+    return Promise.resolve(this.indexApi.indexedModels() ?? [])
+  }
+
+  async stopIndexingModels(modelsIDs: Array<StreamID>): Promise<void> {
+    this.logger.log(LogStyle.info, `Local Admin Api will remove model ids to index ${modelsIDs}`)
```
stbrody, line 25:
```suggestion
    this.logger.log(LogStyle.info, `Removing models from index: ${modelsIDs}`)
```

packages/core/src/local-admin-api.ts
```diff
@@ -0,0 +1,37 @@
...
+    await this.indexApi.stopIndexingModels(modelsIDs)
+    this.logger.log(LogStyle.info, `Local Admin Api did remove model ids to index ${modelsIDs}`)
+  }
+
+  async replaceIndexedModels(modelsIDs: Array<StreamID>): Promise<void> {
+    this.logger.log(LogStyle.info, `Local Admin Api will replace model ids to index ${modelsIDs}`)
```
stbrody, line 31:
```suggestion
    this.logger.log(LogStyle.info, `Replacing models to index. Currently indexed models: ${this.indexApi.indexedModels()}.  Models to index going forward: ${modelsIDs}`)
```

packages/core/src/local-admin-api.ts
```diff
@@ -0,0 +1,37 @@
...
+    // TODO: Is this done in a single transaction? If not, it should be.
```
stbrody, line 32: I know we've discussed this a bunch, but I feel like we never had a proper resolution.  I still don't understand why we are building this "replace" endpoint at all if we have no plan to use it in our tools.  It's just increasing the complexity and surface area of our APIs, and giving us more things we need to test and maintain for little value as far as I can tell.  Can we please get rid of it?

stbrody (changes requested):
I'm still concerned about the "replace" endpoint.  If someone feels still feels like this is worth doing, can we have a call to discuss and try to reach some consensus here?

stbrody (commented):
basically looking good, just need to remove PUT/replace support

stbrody (approved):
lgtm assuming tests pass

stbrody:
Feel free to go ahead and merge this @pawartur!!

PR #2449, opened 2022-09-22
feat(core): implement `ceramic_models` indexing config table

willex:
## Description

Implemented `ceramic_models` configuration table for both SQLite & Postgres. Prepares future dynamic configuration tables to be added (e.g. auth) and separates config inits from dynamic MID table creation. 

## How Has This Been Tested?

- [x] Manual local run of indexing unit tests
- [x] CI integration tests

## PR checklist

## References:


packages/core/src/indexing/postgres/init-tables.ts
```diff
@@ -34,10 +45,24 @@ function relationsDefinitionsToColumnInfo(relations?: ModelRelationsDefinition):
...
+/**
+ * Create Compose DB config tables
+ */
+export async function initConfigTables(dataSource: Knex, logger: DiagnosticsLogger) {
+  const configTables = await listConfigTables()
+  for (const table of configTables) {
+    const exists = await dataSource.schema.hasTable(table.tableName)
+    if (!exists) {
+      logger.imp(`Creating Compose DB config table: ${table.tableName}`)
+      await createConfigTable(dataSource, table.tableName)
```
willex, line 57: Wanted to create the table with the `CONFIG_TABLE_MODEL_INDEX_STRUCTURE` object and looked into `knex` but unfortunately no support for that - would've been nice to use the `.columInfo()` dump to do so. There is a method called `knex.schema.createTableLike` but that only takes another table as input.

pawartur (approved):

packages/cli/src/daemon-config.ts
```diff
@@ -221,10 +221,13 @@ export class IndexingConfig {
   allowQueriesBeforeHistoricalSync = false
 
   /**
+   * @deprecated
```
stbrody, line 224: wasn't this already deprecated in another PR?  Does this PR need to update the base branch so we only see the new changes that are specific to this PR?

willex: good catch - just cross checked against `develop`. yeah i think that just came in because this branch was originally a fork of your WIP and I overlooked that deprecation during the merge from `develop`. I'll correct that in a minute. This should be the only file thats affected.

packages/core/src/indexing/postgres/migrations/1-create-model-table.ts
```diff
@@ -72,3 +72,30 @@ export async function createModelTable(
...
+  switch (tableName) {
+    case 'ceramic_models':
+      await dataSource.schema.createTable(tableName, function (table) {
+        // create model indexing configuration table
+        table.bigIncrements('index_id')
+        table.string('model', 1024).notNullable()
+        table.boolean('is_indexed').notNullable().defaultTo(true)
+        table.dateTime('created_at').notNullable().defaultTo(dataSource.fn.now())
+        table.dateTime('updated_at').notNullable().defaultTo(dataSource.fn.now())
+        table.string('updated_by', 1024).notNullable()
```
stbrody, line 86: ugh, this is such a poor man's audit log.  Postgres already has built-in audit log capabilities for people who want it.  Maybe we should just have a `created_by` but not try to track updates ourselves?  Otherwise it's weird to me the idea that a user who makes a trivial change might override the data about the user who made the last substantially meaningful change.

packages/core/src/indexing/postgres/migrations/1-create-model-table.ts
```diff
@@ -72,3 +72,30 @@ export async function createModelTable(
...
+
+        table.index(['model'], `idx_ceramic_models_model`, {
```
stbrody, line 88: why do we need this index if we have the compound index below?  I think this is just redundant storage for no value

stbrody: Or actually I think we may just want the second index to be just on `is_indexed`.  Otherwise we can't efficiently do a query for all indexed models.  I can't think of a time where you'd want to query on `model` AND `is_indexed` together anyway

stbrody (changes requested):
most important change is to the indexes on the config table.  Assume comments on postgres files apply to sqlite also

packages/core/src/indexing/postgres/migrations/1-create-model-table.ts
```diff
@@ -72,3 +72,30 @@ export async function createModelTable(
     )
   })
 }
+
+export async function createConfigTable(dataSource: Knex, tableName: string) {
+  switch (tableName) {
+    case 'ceramic_models':
+      await dataSource.schema.createTable(tableName, function (table) {
+        // create model indexing configuration table
+        table.bigIncrements('index_id')
```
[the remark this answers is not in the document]
willex: yeah, I thought about calling it `model_index_id` - I'm not a big fan of just `id`'s cause then you end up with a bunch of them all over the place. Given that we won't have that many tables though I'm OK just having it plain `id`.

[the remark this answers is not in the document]
stbrody: I mean, that's kind of why I asked if we even need this column at all - this ID has no semantic meaning.  It's literally just "entry number in this table".  So for that I guess `id` is fine, but I also still think we could consider just having `model` as the primary key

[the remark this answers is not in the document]
willex: yeah, convinced. I'm going to drop the column and make `model` PK + unique constraint.

packages/core/src/indexing/postgres/migrations/1-create-model-table.ts
willex: It's mainly meant to have some basic quick audit capability for an Admin to follow up e.g. if a model was deactivated by a developer by mistake for follow ups. 

Yes, there is the Postgres audit log capability (`pgAudit`) but that needs to be separately activated and is specific to Postgres again (right now that big of an issue really). As far as I remember that would also rely on different auth connections to differentiate users though. Since developers connect via the node connection string, I think the audit log would be less useful than to set explicit access via the DID as initially proposed. Overall, the last update on the status (`is_indexed`) is really the only interesting and relevant on in combination with the `updated_at` which is an indication of when models stopped being indexed -> this potentially could be useful for historical sync as well to check against the mid table history.

packages/core/src/indexing/postgres/init-tables.ts
```diff
@@ -25,6 +31,11 @@ export async function listMidTables(dataSource: Knex): Promise<Array<string>> {
   return midTables
 }
 
+export async function listConfigTables(): Promise<Array<{ tableName; validSchema }>> {
```
ukstv, line 34: Types, please.

packages/core/src/indexing/postgres/migrations/1-create-model-table.ts
ukstv: Ugh, while I see the point of that, it is better to implement an audit log separately, IMO.

packages/core/src/indexing/sqlite/__tests__/sqlite-index-api.test.ts
```diff
@@ -71,9 +73,13 @@ describe('init', () => {
       // Built-in table verification should pass
       await expect(indexApi.verifyTables(modelsToIndexArgs([modelToIndex]))).resolves.not.toThrow()
 
-      // Also manually check table structure
-      const columns = await dbConnection.table(asTableName(modelToIndex)).columnInfo()
+      // Also manually check MID table structure
+      let columns = await dbConnection.table(asTableName(modelToIndex)).columnInfo()
```
ukstv, line 77:
```suggestion
      const midColumns = await dbConnection.table(asTableName(modelToIndex)).columnInfo()
```

packages/core/src/indexing/sqlite/__tests__/sqlite-index-api.test.ts
```diff
@@ -71,9 +73,13 @@ describe('init', () => {
...
       expect(JSON.stringify(columns)).toEqual(JSON.stringify(COMMON_TABLE_STRUCTURE))
+
+      // Also manually check config table structure
+      columns = await dbConnection.table(asTableName(INDEXED_MODEL_CONFIG_TABLE_NAME)).columnInfo()
```
ukstv, line 81:
```suggestion
      const configTableColumns = await dbConnection.table(asTableName(INDEXED_MODEL_CONFIG_TABLE_NAME)).columnInfo()
```

ukstv (approved):

packages/core/src/indexing/postgres/migrations/1-create-model-table.ts
willex: @stbrody @ukstv Let's jump on a sync on this one. We can implement a separate audit log but would need to extend the scope of the current spec then and decide if its part of this or next release milestone.

ukstv: Yup. Do we even need an audit log for January release?

ukstv: @stbrody I am sold we better keep it.

willex: Going forward with the current implementation but filed a ticket to extend history: https://linear.app/3boxlabs/issue/CDB-1872/spec-log-history-for-ceramic-models-table

packages/core/src/indexing/postgres/migrations/1-create-model-table.ts
```diff
@@ -72,3 +74,25 @@ export async function createModelTable(
...
+    case INDEXED_MODEL_CONFIG_TABLE_NAME:
+      await dataSource.schema.createTable(tableName, function (table) {
+        // create model indexing configuration table
+        table.string('model', 1024).unique().notNullable().primary()
+        table.boolean('is_indexed').notNullable().defaultTo(true)
+        table.dateTime('created_at').notNullable().defaultTo(dataSource.fn.now())
+        table.dateTime('updated_at').notNullable().defaultTo(dataSource.fn.now())
+        table.string('updated_by', 1024).notNullable()
+
+        table.index(['is_indexed'], `idx_ceramic_is_indexed`, {
```
stbrody, line 89: I guess making the `model` column "primary" means it already has an index created by default, and that's why we don't need to define one here?

packages/core/src/indexing/postgres/postgres-index-api.ts
```diff
@@ -28,28 +29,82 @@ export class PostgresIndexApi implements DatabaseIndexApi {
...
       })
   }
 
+  private async indexModelsInDatabase(models: Array<IndexModelArgs>): Promise<void> {
+    if (models.length === 0) return
+    await initMidTables(this.dbConnection, models, this.logger)
+    await this.verifyTables(models)
+    //
+    // : CDB-1866 - populate the updated_by field properly when auth is implemented
+    await this.dbConnection(INDEXED_MODEL_CONFIG_TABLE_NAME)
```
stbrody, line 69: nitpick: can we have a helper function for this as well?

stbrody (approved):
lgtm

PR #2453, opened 2022-09-23
Start populating the `ceramic_models` table and add stopIndexingModels(...) to `DatabaseIndexApi`

pawartur:
## Description

* made the `model` column unique in `ceramic_models` table
* started adding entries to `ceramic_models` table in `indexModels(...)`
* added `stopIndexingModels(...)` to `DatabaseIndexApi` and implemented it in postres and sqlite db apis
* **did not(!!!!)** implement deletion of model index tables on `stopIndexingModels(...)`
* update entries in `ceramic_models` table in `stopIndexingModels(...)`
* load the list of models to index in `init()` in `DatabaseIndexApi`

## How Has This Been Tested?

- [X] Added tests for postgres
- [X] Added tests for sqlite

## PR checklist

### DIFF
--- a/packages/core/src/indexing/postgres/init-tables.ts
+++ b/packages/core/src/indexing/postgres/init-tables.ts
@@ -111,7 +111,7 @@ export async function verifyTables(dataSource: Knex, modelsToIndex: Array<IndexM
   for (const tableName of tables) {
     const modelIndexArgs = modelsToIndex.find((model) => tableName == asTableName(model.model))
     if (!modelIndexArgs) {
-      // FIXME: This means that there's is a table for a model that is no longer indexed. Should this table have been deleted?
+      // TODO: CDB-1869 - This means that there's is a table for a model that is no longer indexed. Should this table have been deleted?
       continue
     }
 
--- a/packages/core/src/indexing/postgres/init-tables.ts
+++ b/packages/core/src/indexing/postgres/init-tables.ts
@@ -8,7 +8,7 @@ import { asTableName } from '../as-table-name.util.js'
 import { Knex } from 'knex'
 import { Model, ModelRelationsDefinition } from '@ceramicnetwork/stream-model'
 import { DiagnosticsLogger } from '@ceramicnetwork/common'
-import { IndexModelArgs } from '../database-index-api.js'
+import { INDEXED_MODEL_CONFIG_TABLE_NAME, IndexModelArgs } from '../database-index-api.js'
 import {
   COMMON_TABLE_STRUCTURE,
   CONFIG_TABLE_MODEL_INDEX_STRUCTURE,
@@ -33,7 +33,7 @@ export async function listMidTables(dataSource: Knex): Promise<Array<string>> {
 
 export async function listConfigTables(): Promise<Array<any>> {
   // TODO (CDB-1852): extend with ceramic_auth
-  return [{ tableName: 'ceramic_models', validSchema: CONFIG_TABLE_MODEL_INDEX_STRUCTURE }]
+  return [{ tableName: INDEXED_MODEL_CONFIG_TABLE_NAME, validSchema: CONFIG_TABLE_MODEL_INDEX_STRUCTURE }]
 }
 
 function relationsDefinitionsToColumnInfo(relations?: ModelRelationsDefinition): Array<ColumnInfo> {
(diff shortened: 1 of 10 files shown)


pawartur:
> @willex , do you think you could help me understand why I'm getting the `Schema verification failed for config table: ${table.tableName}. Please make sure node has been setup correctly.` error in postgres tests?

Ah, I think it's because I need to update `CONFIG_TABLE_MODEL_INDEX_STRUCTURE` for postgres after I made the `model` column unique... Would love your help with this anyways, @willex , if that's possible, as I don't have postgres set up locally yet...

packages/core/src/indexing/postgres/init-tables.ts
```diff
@@ -110,6 +110,10 @@ export async function verifyTables(dataSource: Knex, modelsToIndex: Array<IndexM
 
   for (const tableName of tables) {
     const modelIndexArgs = modelsToIndex.find((model) => tableName == asTableName(model.model))
+    if (!modelIndexArgs) {
+      // FIXME: This means that there's is a table for a model that is no longer indexed. Should this table have been deleted?
```
stbrody, line 114: We originally said that expected behavior would be that the index data sticks around, but now i'm wondering if that's a good idea.  Or least, if we then decide to *re-index* the model, if we should delete and re-create the index table at that point?  Otherwise there could be very old outdated data in the index that we would start suddenly serving to the application.  Also would need to think about how this would work with historical indexing...

packages/core/src/indexing/postgres/migrations/1-create-model-table.ts
```diff
@@ -75,11 +77,11 @@ export async function createModelTable(
 
 export async function createConfigTable(dataSource: Knex, tableName: string) {
   switch (tableName) {
-    case 'ceramic_models':
+    case INDEXED_MODEL_CONFIG_TABLE_NAME:
       await dataSource.schema.createTable(tableName, function (table) {
         // create model indexing configuration table
         table.bigIncrements('index_id')
-        table.string('model', 1024).notNullable()
+        table.string('model', 1024).unique().notNullable()
```
stbrody, line 84: good call

packages/core/src/indexing/postgres/postgres-index-api.ts
```diff
@@ -63,14 +64,62 @@ export class PostgresIndexApi implements DatabaseIndexApi {
   }
 
   async indexModels(models: Array<IndexModelArgs>): Promise<void> {
+    if (models.length === 0) return
+
     await initMidTables(this.dbConnection, models, this.logger)
     await this.verifyTables(models)
+    // FIXME: populate the updated_by field properly when auth is implemented
+    await this.dbConnection(INDEXED_MODEL_CONFIG_TABLE_NAME)
```
stbrody, line 72: please make a helper function for this so that this function is easier to read

stbrody: also note my comment on https://github.com/ceramicnetwork/js-ceramic/pull/2449 about whether we even want this functionality at all

packages/core/src/indexing/postgres/postgres-index-api.ts
```diff
@@ -63,14 +64,62 @@ export class PostgresIndexApi implements DatabaseIndexApi {
...
+        is_indexed: false,
+        updated_by: "<FIXME: PUT ADMIN DID WHEN AUTH IS IMPLEMENTED>"
+      })
+    const modelsAsStrings = models.map(streamID => streamID.toString())
+    this.modelsToIndex = this.modelsToIndex.filter(modelStreamID => !modelsAsStrings.includes(modelStreamID.toString()))
+  }
+
   async init(): Promise<void> {
     await initConfigTables(this.dbConnection, this.logger)
+    this.modelsToIndex = (await this.dbConnection(INDEXED_MODEL_CONFIG_TABLE_NAME)
```
stbrody, line 115: helper function please!  I think all things that do SQL should be in separate functions so we don't have sql interspersing random other higher-level code.  Having a line like `this.modelsToIndex = await this.getIndexedModelsFromDatabase()` will be much clearer to read IMHO

packages/core/src/indexing/postgres/init-tables.ts
stbrody: I think we'll want to think about this, but we don't need to hold this PR up over it, can you file at ticket to revisit the behavior here after devcon?

stbrody (approved):
lgtm with a few small suggestions for code cleanup

packages/core/src/indexing/postgres/postgres-index-api.ts
pawartur: > whether we even want this functionality at all

Just to make sure I understand, @stbrody , which functionality were you referring to :) ?

stbrody: I meant the poor man's audit log of maintaining the `updated_by` column.

PR #2462, opened 2022-09-26
Added authorization headers to remote admin api calls

pawartur:
## Implemented Basic Authorization for the new AdminAPI

* path + models + one-time code are signed on the client side by authenticated DID
* path + models + one-time code are checked on the backend side
* it's also checked that the one-time code is not older than 1 min
* the signature is also verified on the backend side and it's checked that the signer is one of DIDs on the new admin dids list
* the admin dids list is passed in daemon or ceramic config

## How Has This Been Tested?

- [X] Added a bunch of tests for http-client and ceramic-daemon

## PR checklist


packages/cli/src/ceramic-daemon.ts
```diff
@@ -529,6 +529,7 @@ export class CeramicDaemon {
     if (error) {
       res.status(422).json({ error: error })
     } else {
+      req.headers
```
oed, line 532: What does this do?

pawartur: Leftover, sorry.

pawartur: Ah, it's no longer here ;P

packages/http-client/src/remote-admin-api.ts
```diff
@@ -7,40 +7,57 @@ import { StreamID } from '@ceramicnetwork/streamid'
 export class RemoteAdminApi implements AdminApi {
   // Stored as a member to make it easier to inject a mock in unit tests
   private readonly _fetchJson: typeof fetchJson = fetchJson
+
   readonly baseURL = './admin/models'
 
   constructor(
-    private readonly _apiUrl: URL
+    private readonly _apiUrl: URL,
+    private readonly _context: Context
```
stbrody, line 15: jumping in here with a nitpick since it came up in discussion with @ukstv this morning on another PR.  We'd like to move away from the `Context` object as a way to pass dependencies through our stack.  Can you change this to take the DID instance directly instead?

packages/http-client/src/remote-admin-api.ts
```diff
@@ -7,40 +7,57 @@ import { StreamID } from '@ceramicnetwork/streamid'
...
   ) {}
 
   private getUrl(): URL {
     return new URL(this.baseURL, this._apiUrl)
   }
 
+  private async buildAuthorizationHeader(modelsIDs?: Array<StreamID>): Promise<string> {
+    const body = modelsIDs ? modelsIDs.map(streamID => streamID.toString()) : undefined
+    const jws = await this._context.did.createJWS({
+      timestamp: Date.now(),
```
stbrody, line 25: what's this timestamp for?  Warning bells always go off for me when I see timestamps being taken from the system clock, especially the system clock on the client machine, as we cannot trust that timestamp for just about anything

packages/http-client/src/remote-admin-api.ts
```diff
@@ -7,40 +7,57 @@ import { StreamID } from '@ceramicnetwork/streamid'
...
+      requestPath: this._apiUrl.pathname,
```
stbrody, line 26: Depending on what DID Provider is being used, this might result in a signature request popup right?  If so, we should make sure the contents of that signature request are at least vaguely human readable

packages/http-client/src/remote-admin-api.ts
```diff
@@ -7,40 +7,57 @@ import { StreamID } from '@ceramicnetwork/streamid'
...
+    private readonly _context: Context
   ) {}
 
   private getUrl(): URL {
     return new URL(this.baseURL, this._apiUrl)
   }
 
+  private async buildAuthorizationHeader(modelsIDs?: Array<StreamID>): Promise<string> {
+    const body = modelsIDs ? modelsIDs.map(streamID => streamID.toString()) : undefined
+    const jws = await this._context.did.createJWS({
```
stbrody, line 24: Depending on what DID Provider is being used, this might result in a signature request popup right?  If so, we should make sure the contents of that signature request are at least vaguely human readable

packages/http-client/src/remote-admin-api.ts
```diff
@@ -7,40 +7,57 @@ import { StreamID } from '@ceramicnetwork/streamid'
...
+      timestamp: Date.now(),
+      requestPath: this._apiUrl.pathname,
+      requestBody: body
+    })
+    return `${jws.signatures[0].protected}.${jws.payload}.${jws.signatures[0].signature}`
+  }
+
   async addModelsToIndex(modelsIDs: Array<StreamID>): Promise<void> {
     await this._fetchJson(this.getUrl(), {
+      headers: { 'Authorization:': `Basic ${await this.buildAuthorizationHeader(modelsIDs)}` },
```
stbrody, line 34: what does the word "basic" in this string do?

oed:
> Depending on what DID Provider is being used, this might result in a
signature request popup right? If so, we should make sure the contents of
that signature request are at least vaguely human readable

Not really. Right now we only support did key. Further work would be needed
to support did pkh. Even then this text wouldn’t be part of any pop up.

*Joel Thorstensson *
Tech co-founder,

packages/http-client/src/remote-admin-api.ts
oed: To prevent replay attacks. If someone gets ahold of your signature after the fact they could resend it to the node to reissue the same command. 
What we really should do here is to have a challenge response protocol, but this would increase the scope.

oed: Actually might be worth it. We would need to do the following:

1. Make all admin endpoints not require any JWS
2. Make all endpoints return a random challenge when requested
3. Add a special endpoint to respond to the challenge

Note that this would add additional complexity to the implementation because it needs to be stateful.

packages/http-client/src/remote-admin-api.ts
pawartur: Removed `_context`

packages/http-client/src/remote-admin-api.ts
pawartur: It's just following the standard for auth headers: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization

I thought that `Basic` is the best choice here.

packages/cli/src/ceramic-daemon.ts
```diff
@@ -524,36 +517,111 @@ export class CeramicDaemon {
...
-    if (error) {
-      res.status(422).json({ error: error })
+  private async _processAdminModelsMutationRequest(
+    req: Request,
+    res: Response,
+    successCallback: (actingDid: string, modelIDs: Array<StreamID>) => Promise<void>
+  ): Promise<void> {
+    const modelsValidation = this._validateModelIDStrings(req.body.models)
+    if (modelsValidation.error) {
+      res.status(422).json({ error: modelsValidation.error })
```
stbrody, line 576: is there a library with named constants for these http codes so they are more semantically meaningful to a reader?

packages/cli/src/daemon-config.ts
```diff
@@ -347,6 +347,12 @@ export class DaemonMetricsConfig {
 @jsonObject
 @toJson
 export class DaemonConfig {
+  /**
+   * An array of DIDs with access to Admin API (represented as strings)
+   */
+  @jsonArrayMember(String, { name: 'admin-dids' })
```
stbrody, line 353: seems like this should probably go into a sub-section of the config file rather than the top level.  I'd suggest putting it under `node` (ie in the `DaemonCeramicNodeConfig`)

packages/common/src/ceramic-api.ts
```diff
@@ -60,29 +60,34 @@ export interface CeramicSigner extends CeramicCommon {
 export interface AdminApi {
   /**
    * List indexed model streams
+   *
+   * @param actingDid - did that performs the operation
    */
-  getIndexedModels(): Promise<Array<StreamID>>
+  getIndexedModels(actingDid: DID | string): Promise<Array<StreamID>>
```
stbrody, line 66: will this ever be a DID instance?  Isn't it always a string, at least for now?  So can we tighten up the interface and make it explicitly always the DID string?

packages/core/src/local-admin-api.ts
```diff
@@ -7,27 +7,39 @@ import { LocalIndexApi } from './indexing/local-index-api.js'
...
+    private readonly adminDids: Array<string> | undefined,
     private readonly indexApi: LocalIndexApi,
     private readonly logger: DiagnosticsLogger
   ) {}
 
-  async startIndexingModels(modelsIDs: Array<StreamID>): Promise<void> {
+  private verifyActingDid(actingDid: string) {
+    if (!this.adminDids || !this.adminDids.some( adminDid => actingDid.startsWith(adminDid) )) {
+      this.logger.log(LogStyle.warn, `Unauthorized access attempt to Admin Api from did: ${actingDid}`)
+      throw Error('Unauthorized access')
```
stbrody, line 18:
```suggestion
      throw Error(`Unauthorized access: DID '${actingDid}' does not have admin access permission to this node`)
```

stbrody: users might not even know where the composedb gets its DID from, including the DID being used in the error message may help node operators figure out what's going on when they have a misconfiguration

packages/http-client/src/remote-admin-api.ts
stbrody: yeah, this does kind of seem worth doing IMO, much better than relying on a timestamp.  Could probably wait until after devcon though?

stbrody (commented):
mostly looking good, but I've got some nitpicks for code cleanup

packages/common/src/ceramic-api.ts
pawartur: It's a did on the client side (see `RemoteAdminApi` :).

packages/cli/src/ceramic-daemon.ts
```diff
@@ -524,36 +517,111 @@ export class CeramicDaemon {
...
+  private _compareStringArrays(left: Array<string>, right: Array<string>): boolean {
+    return left.length === right.length && left.every((value, index) => value === right[index])
+  }
+
+  private async _validateDIDJWSAuthHeader(
+    basePath: string,
+    authHeader: string | undefined,
+    forModels?: Array<string>
+  ): Promise<{ kid?: string, error?: string }> {
+    const errorResult = { status: 'error', error: 'Missing or invalid authorization signature' }
```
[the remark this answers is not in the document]
pawartur: Sure we can. I didn't want to do it initially, because I think that in security-related contexts it's usually better not to give out too much information that could help attackers. Happy to change this, if you're fine with this, though.

packages/common/src/ceramic-api.ts
stbrody: Ah right, and I do see now that in the actual Local/RemoteAdminApi classes you use just the proper specific type that is expected in each context.  Not sure exactly how Typescript's compiler handles that kind of type specialization in interface implementations, but hopefully it does what we want 🙂

packages/cli/src/ceramic-daemon.ts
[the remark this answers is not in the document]
stbrody: Personally, I'd say that giving out information about the node configuration is a different than telling a user what about their specific request caused it to fail.  I don't think telling the user that they used the wrong did, or had an expired authorization gives out any extra information about the node itself (but I'm not a security expert so maybe there's something subtle here I'm missing?).  My personal instinct certainly always leans towards giving devs as much information as possible to understand when they are doing something wrong.

packages/http-client/src/remote-admin-api.ts
oed: @stbrody Wdym? Either we do it now or we wait with adding admin api before devcon. Otherwise there will be a breaking change here.

oed: The simplest way would probably be to do this:

1. Add a new endpoint `/admin/getCode`. This returns a random string / number, stores it in memory with a timestamp
2. For all other `/admin/` endpoints remove the timestamp argument and add a new `code` argument. 
3. When authorizing a request make sure `code` is valid and was issued within T time (e.g. 1 min), then remove the code from memory

cc @pawartur

pawartur:
@oed , @stbrody , what would the flow with the new challenge-based approach be like: would I have to request a different challenge before each request to the admin api?

oed:
> would I have to request a different challenge before each request to the admin api?

Yes exactly!

packages/http-client/src/remote-admin-api.ts
stbrody: FWIW I don't think it'd be the worst thing in the world to have a breaking change here, if we wanted to get this out the door for devcon.  Especially since it's only to the admin API, no live apps would break, only their deployment process would, and they'd just need to upgrade their composedb cli to the newest version and then it would work again.

packages/cli/src/ceramic-daemon.ts
```diff
@@ -523,26 +543,110 @@ export class CeramicDaemon {
...
+  private async _parseDidJWSAuthHeader(
+    authHeader: string | undefined
+  ): Promise<AdminAPIAuthHeaderContents> {
+    const jwsString = authHeader.split("Authorization: Basic ")[1]
+    const result = await this.ceramic.did.verifyJWS(jwsString)
+    return {
+      kid: result.kid,
+      code: result.payload.code,
+      requestPath: result.payload.requestPath,
+      forModels: result.payload.requestBody ? result.payload.requestBody.models : undefined,
```
oed, line 568:
```suggestion
      forModels: result.payload.requestBody?.models,
```
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

packages/common/src/ceramic-api.ts
```diff
@@ -58,24 +58,37 @@ export interface CeramicSigner extends CeramicCommon {
  * Describes Ceramic Admin API functionality
  */
 export interface AdminApi {
+  /**
+   * Generate a code needed to perform other AdminApi operations
+   */
+  generateCode(): Promise<string>
```
oed, line 64: This method doesn't need to be on the CeramicApi. Can be only available as an http endpoint. The ceramic-daemon can keep track of issued codes.

packages/core/src/local-admin-api.ts
```diff
@@ -1,26 +1,76 @@
...
+type Timestamp = number
+
+type AdminCodeCache = {
+  [key: AdminCode]: Timestamp
+}
 
 /**
  * AdminApi for Ceramic core.
  */
 export class LocalAdminApi implements AdminApi {
```
oed, line 19: Let's not leak the `code` abstraction into the `AdminApi` interface!

packages/core/src/local-admin-api.ts
```diff
@@ -1,26 +1,70 @@
...
+    if (!this.adminDids || !this.adminDids.some( adminDid => actingDid.startsWith(adminDid) )) {
+      this.logger.log(LogStyle.warn, `Unauthorized access attempt to Admin Api from did: ${actingDid}`)
+      throw Error(`Unauthorized access: DID '${actingDid}' does not have admin access permission to this node`)
+    }
+  }
+
+  async generateCode(): Promise<string> {
+    const newCode = crypto.randomUUID()
+    const now = (new Date).getTime()
+    this.codeCache[newCode] = now
```
oed, line 38: I would probably just move the `codeCacahe` into the `ceramic-daemon.ts` file, but it doesn't matter that much.

oed (approved):
lgtm

packages/cli/src/ceramic-daemon.ts
```diff
@@ -41,6 +45,15 @@ const DEFAULT_PORT = 7007
 const HEALTHCHECK_RETRIES = 3
 const CALLER_NAME = 'js-ceramic'
 
+const ADMIN_CODE_EXPIRATION_TIMEOUT = 1000 * 60 * 1 // 1 min
+
+type AdminCode = string
+type Timestamp = number
+
+type AdminCodeCache = {
```
stbrody, line 53: I'm worried about this growing unbounded and old entries never being cleaned up.  Can we use an `lru.LRUMap` (see state-cache.ts for an example) with a fixed size (of... 500?) just to make sure that there's some upper bound on how much memory this can consume?

packages/cli/src/ceramic-daemon.ts
```diff
@@ -511,51 +569,107 @@ export class CeramicDaemon {
...
+      res.status(StatusCodes.UNPROCESSABLE_ENTITY).json({ error: jwsValidation.error })
     } else {
-      await this.ceramic.admin.startIndexingModels(modelIDStrings.map( modelIDString => StreamID.fromString(modelIDString) ))
-      res.status(200).json({ result: 'success' })
+      try {
+        this.verifyAndDiscardAdminCode(jwsValidation.code)
+        await successCallback(jwsValidation.kid, jwsValidation.models)
+        res.status(StatusCodes.OK).json({ result: 'success' })
+      } catch (e) {
+        res.status(StatusCodes.UNAUTHORIZED).json({ error: e.message })
```
stbrody, line 630: unauthorized is only the right error code if `verifyAndDiscardAdminCode` fails.  If successCallback fails it should be something else.

stbrody (commented):
lgtm with two small cleanups.  Feel free to merge this and address the two comments in a follow-up PR so we can get this in.

PR #2463, opened 2022-09-26
feat: `count` endpoint

ukstv:


packages/core/src/indexing/sqlite/sqlite-index-api.ts
```diff
@@ -47,7 +47,7 @@ export class SqliteIndexApi implements DatabaseIndexApi {
       .insert({
         stream_id: args.streamID.toString(),
         controller_did: args.controller.toString(),
-        stream_content: args.streamContent.toString(),
+        stream_content: JSON.stringify(args.streamContent),
```
willex, line 50: Good catch!

willex (approved):
LGTM! 🏇

packages/http-client/src/remote-index-api.ts
```diff
@@ -16,11 +16,15 @@ export class RemoteIndexApi implements IndexApi {
...
   constructor(apiUrl: URL) {
-    this._collectionURL = new URL('./collection', apiUrl)
+    this._collectionURL = new URL('./collection/', apiUrl)
   }
 
   async count(query: BaseQuery): Promise<number> {
-    throw new Error('Not implemented')
+    const queryURL = new URL('./count', this._collectionURL)
+    for (const key in query) {
+      queryURL.searchParams.set(key, query[key])
```
stbrody, line 25: this won't actually serialize the filter right.  You'll need to use the `serializeObjectToSearchParams` function that I added in https://github.com/ceramicnetwork/js-ceramic/pull/2455

stbrody (commented):
mostly looks good, only real comment is around using an object in the HTTP response instead of just the raw count scalar

packages/http-client/src/remote-index-api.ts
ukstv: Quite probable. As we discussed, you merge #2455, then I merge this one or rebase, so that we share the same function to handle the filters.

packages/http-client/src/__tests__/remote-index-api.test.ts
```diff
@@ -29,7 +29,7 @@ test('model in query', async () => {
   ;(indexApi as any)._fetchJson = fauxFetch
   const result = await indexApi.query({ model: MODEL, first: 5 })
   expect(result).toEqual(EMPTY_RESPONSE)
-  expect(fauxFetch).toBeCalledWith(new URL(`https://example.com/collection?model=${MODEL}&first=5`))
+  expect(fauxFetch).toBeCalledWith(new URL(`https://example.com/collection/?model=${MODEL}&first=5`))
```
stbrody, line 32: this change seems surprising to me?  I'd expect the query params string to come directly on the `collection` endpoint, not after a slash to a new sub-endpoint?

packages/http-client/src/remote-index-api.ts
```diff
@@ -17,19 +17,20 @@ export class RemoteIndexApi implements IndexApi {
   private readonly _collectionURL: URL
 
   constructor(apiUrl: URL) {
-    this._collectionURL = new URL('./collection', apiUrl)
+    this._collectionURL = new URL('./collection/', apiUrl)
```
stbrody, line 20: why is this necessary?

packages/http-client/src/remote-index-api.ts
```diff
@@ -17,19 +17,20 @@ export class RemoteIndexApi implements IndexApi {
   private readonly _collectionURL: URL
```
stbrody, line 17: should we add a `_countURL` here so we don't need to construct it inline each time?

stbrody (commented):
only question is why the GET query params come after a slash now instead of directly after the `collection` part of the URL.

packages/http-client/src/remote-index-api.ts
ukstv: For a server that does not matter. It matters for composing URLs. `./collection` means a "resource" named `collection` in the current directory of `apiURL`. If we want to have `./collection/count`, based on `this._collectionURL` using `new URL('./count', collectionURL)`, we have to indicate that `collection` is a "folder" not a "resource". Thus, a slash.

packages/http-client/src/remote-index-api.ts
ukstv: yup, to resolve the confusion

stbrody (approved):
LGTM!

PR #2464, opened 2022-09-26
clean: update doc comment for request anchor method

v-stickykeys:
## Description

The comment was copy pasted from another function so just updated it here.


ukstv (approved):
Approved of course!

PR #2473, opened 2022-09-30
fix:cacao depend refs

zachferland:


stbrody (commented):
doing a simple grep, I also see imports from 'ceramic-cacao' in `common/src/utils/signature_utils.ts` and `core/src/utils.ts` - do those need to be updated also?

zachferland:
yup should be no more references to ceramic-cacao in libs package.json or libs, looks like its all set to merge

stbrody (approved):

PR #2480, opened 2022-10-03
chore(core, stream-tests): Testing and improved error message for count queries on unindexed streams

stbrody:
Includes additional test coverage of count queries and queries with relations

### DIFF
--- a/packages/stream-model-instance/src/model-instance-document.ts
+++ b/packages/stream-model-instance/src/model-instance-document.ts
@@ -281,19 +281,20 @@ export class ModelInstanceDocument<T = Record<string, any>> extends Stream {
       throw new Error(`Must specify a 'model' when creating a ModelInstanceDocument`)
     }
 
-    if (!metadata.controller) {
+    let controller = metadata.controller
+    if (!controller) {
       if (signer.did) {
         await _ensureAuthenticated(signer)
         // When did has a parent, it has a capability, and the did issuer (parent) of the capability
         // is the stream controller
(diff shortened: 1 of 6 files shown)


packages/stream-model-instance/src/model-instance-document.ts
```diff
@@ -281,19 +281,20 @@ export class ModelInstanceDocument<T = Record<string, any>> extends Stream {
       throw new Error(`Must specify a 'model' when creating a ModelInstanceDocument`)
     }
 
-    if (!metadata.controller) {
+    let controller = metadata.controller
```
stbrody, line 284: fix to avoid modifying the passed in `metadata` object

packages/stream-model-instance/src/model-instance-document.ts
```diff
@@ -281,19 +281,20 @@ export class ModelInstanceDocument<T = Record<string, any>> extends Stream {
...
+    if (!controller) {
       if (signer.did) {
         await _ensureAuthenticated(signer)
         // When did has a parent, it has a capability, and the did issuer (parent) of the capability
         // is the stream controller
-        metadata.controller = signer.did.hasParent ? signer.did.parent : signer.did.id
+        controller = signer.did.hasParent ? signer.did.parent : signer.did.id
```
willex, line 290: As per our discussion yesterday, good catch!

stbrody: image

stbrody (commented):
oops, looks like I forgot to send my comments before

willex (approved):
LGTM - 🏇

PR #2481, opened 2022-10-03
feat(core,common,http-client): Standardize AdminAPI implementations to not take DID argument.

stbrody:
puts the `admin: AdminAPI` back into the common `CeramicApi` interface (effectively reverting https://github.com/ceramicnetwork/js-ceramic/commit/d83c739ef6e5679da485363db8bc477ec1d39540) and unifies the `LocalAdminApi` and `RemoteAdminApi` to have the same signatures by removing the DID arguments from both and taking the DID from the `CeramicClient` handle in the http-client implementation

### DIFF
--- a/packages/cli/src/daemon-config.ts
+++ b/packages/cli/src/daemon-config.ts
@@ -153,6 +153,12 @@ export class DaemonHTTPApiConfig {
     },
   })
   corsAllowedOrigins?: RegExp[]
+
+  /**
+   * An array of DIDs with access to Admin API (represented as strings)
+   */
+  @jsonArrayMember(String, { name: 'admin-dids' })
+  adminDids?: Array<string>
 }
 
 /**
@@ -271,12 +277,6 @@ export class DaemonDidResolversConfig {
 @jsonObject
 @toJson
 export class DaemonCeramicNodeConfig {
-  /**
-   * An array of DIDs with access to Admin API (represented as strings)
-   */
-  @jsonArrayMember(String, { name: 'admin-dids' })
-  adminDids: Array<string>
-
   /**
    * Whether to run the Ceramic node in read-only gateway mode.
    */
--- a/packages/core/src/local-admin-api.ts
+++ b/packages/core/src/local-admin-api.ts
@@ -1,4 +1,4 @@
-import { AdminApi, DiagnosticsLogger, LogStyle } from '@ceramicnetwork/common'
+import { AdminApi, DiagnosticsLogger } from '@ceramicnetwork/common'
 import { StreamID } from '@ceramicnetwork/streamid'
 import { LocalIndexApi } from './indexing/local-index-api.js'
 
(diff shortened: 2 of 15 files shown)


packages/cli/src/daemon-config.ts
```diff
@@ -153,6 +153,12 @@ export class DaemonHTTPApiConfig {
     },
   })
   corsAllowedOrigins?: RegExp[]
+
+  /**
+   * An array of DIDs with access to Admin API (represented as strings)
+   */
+  @jsonArrayMember(String, { name: 'admin-dids' })
```
stbrody, line 160: note this is a breaking change to the config file format (but only since this functionality went in last week)

packages/core/src/local-admin-api.ts
```diff
@@ -7,38 +7,19 @@ import { LocalIndexApi } from './indexing/local-index-api.js'
  */
 export class LocalAdminApi implements AdminApi {
```
stbrody, line 8: this class is such a thin wrapper around the `LocalIndexApi` it's debatable whether it should even exist, but for the client right now we have a separation between the `client.index` methods and the `client.admin` methods, so need to mirror that separation server-side so long as they all share the same interfaces

packages/cli/src/daemon-config.ts
stbrody, line 160: note this is a breaking change to the config file format (but only since this functionality went in last week)

Now that the Ceramic core implementation actually does no enforcement of access control on the admin api it made more sense to me that this option be considered part of the http-api config instead of the node config section of the config file

packages/http-client/src/ceramic-http-client.ts
```diff
@@ -92,7 +93,9 @@ export class CeramicClient implements CeramicApi {
 
     this.pin = new RemotePinApi(this._apiUrl)
     this.index = new RemoteIndexApi(this._apiUrl)
-    this.admin = new RemoteAdminApi(this._apiUrl)
+    this.admin = new RemoteAdminApi(this._apiUrl, () => {
+      return this.did
```
oed, line 97: Is `this.did` really in scope on this anon function?

oed (approved):
lgtm

pawartur (approved):

packages/http-client/src/ceramic-http-client.ts
stbrody: ooph good call.  This isn't actually tested at the moment.  Was going to add tests in a follow-up, but now I feel like it might be worth doing before making the RC.  Will try to put something together right now

stbrody: Actually, with my basic testing, it seemed to work as-written.  But I added `.bind(this)` to the end just to be safe anyway

packages/http-client/src/__tests__/remote-admin-api.test.ts
```diff
@@ -10,66 +10,74 @@ import { randomBytes } from '@stablelib/random'
 const FAUX_ENDPOINT = new URL('https://example.com')
 const MODEL = new StreamID(1, TestUtils.randomCID())
 const SUCCESS_RESPONSE = {
-  result: 'success'
+  result: 'success',
```
stbrody, line 13: note changes to this file are new since last review

packages/stream-tests/src/__tests__/admin-api.test.ts
```diff
@@ -0,0 +1,141 @@
+import { jest } from '@jest/globals'
```
stbrody, line 1: this whole file is new since the last review

stbrody (commented):
Uploaded a new version with additional testing

PR #2482, opened 2022-10-03
chore(cli,core,http-client): Prettier code cleanup

stbrody:
No functionality change, just running prettier on some recently modified files


stbrody:
Going to assume this is fine since it's formatting only and merge without approval

PR #2488, opened 2022-10-11
fix: Wrap AJV instances in LRU cache

ukstv:
Apparently, AJV instance is just a huge source of memory leaks when applied to on-demand schema validation. Better put AJV instances in LRU cache to isolate garbage.

### DIFF
--- a/packages/stream-model-instance-handler/src/model-instance-document-handler.ts
+++ b/packages/stream-model-instance-handler/src/model-instance-document-handler.ts
@@ -19,6 +19,7 @@ import {
 import { StreamID } from '@ceramicnetwork/streamid'
 import { SchemaValidation } from './schema-utils.js'
 import { Model } from '@ceramicnetwork/stream-model'
+import lru from 'lru_map'
 
 // Hardcoding the model streamtype id to avoid introducing a dependency on the stream-model package
 const MODEL_STREAM_TYPE_ID = 2
@@ -218,7 +219,11 @@ export class ModelInstanceDocumentHandler implements StreamHandler<ModelInstance
         )
       }
     } else {
-      await this._schemaValidator.validateSchema(content, model.content.schema)
+      await this._schemaValidator.validateSchema(
+        content,
+        model.content.schema,
+        model.commitId.toString()
+      )
     }
   }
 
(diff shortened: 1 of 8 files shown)


ukstv:
The tests will be green after https://github.com/ceramicnetwork/js-ceramic/pull/2489 merge

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
```diff
@@ -218,7 +218,11 @@ export class ModelInstanceDocumentHandler implements StreamHandler<ModelInstance
         )
       }
     } else {
-      await this._schemaValidator.validateSchema(content, model.content.schema)
+      await this._schemaValidator.validateSchema(
+        content,
+        model.content.schema,
+        model.commitId.toString()
+      )
```
pawartur, line 225: Maybe just pass the content and the model to `validateSchema(...)` and let it handle cache keys inside?

pawartur (approved):

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
ukstv: Hmm, good suggestion!

ukstv: Oh, sorry, no. I can not generalise it for all the StreamUtils :/ There is no Model for Tile.

packages/stream-model-instance-handler/src/schema-utils.ts
```diff
@@ -1,31 +1,42 @@
...
-    addFormats(this._validator)
+    this.validators = new lru.LRUMap(AJV_CACHE_SIZE)
   }
 
-  public validateSchema(content: Record<string, any>, schema: SchemaObject) {
-    const isValid = this._validator.validate(schema, content)
-
-    // Remove schema from the Ajv instance's cache, otherwise the ajv cache grows unbounded
-    this._validator.removeSchema(schema)
+  public validateSchema(content: Record<string, any>, schema: SchemaObject, schemaId: string) {
```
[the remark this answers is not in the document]
ukstv: > it'd be nice if schemaId was a CommitID instead of a string.

It only introduces overhead for Tiles: https://github.com/ceramicnetwork/js-ceramic/blob/ee5d72100646dde44cf99d6a22a22543383249f8/packages/stream-tile-handler/src/tile-document-handler.ts#L186 So, good idea, but not this time.

Merging the 3 SchemaValidation classes into one: there is a ticket for that.

PR #2491, opened 2022-10-11
feat(core): Allow setting network to mainnet

stbrody:
We can already run on mainnet, you just need to set the network to "elp" instead of "mainnet". This is a confusing requirement. This PR let's mainnet mean mainnet, and removes the need to use "elp".  It also has logic to preserve backwards compatibly for nodes that previously were running with "elp" and then switch to "mainnet", to ensure that their state store data isn't lost.

builds on https://github.com/ceramicnetwork/js-ceramic/pull/2490

### DIFF
--- a/packages/core/src/store/level-state-store.ts
+++ b/packages/core/src/store/level-state-store.ts
@@ -1,6 +1,11 @@
 import levelTs from 'level-ts'
 import type Level from 'level-ts'
-import { StreamState, StreamStateHolder, StreamUtils } from '@ceramicnetwork/common'
+import {
+  DiagnosticsLogger,
+  StreamState,
+  StreamStateHolder,
+  StreamUtils,
+} from '@ceramicnetwork/common'
 import { StateStore } from './state-store.js'
 import { StreamID } from '@ceramicnetwork/streamid'
 import * as fs from 'fs'
@@ -23,8 +28,11 @@ const LevelC = (levelTs as any).default as unknown as typeof Level
  */
 export class LevelStateStore implements StateStore {
   #store: Level
+  #logger: DiagnosticsLogger
 
-  constructor(private storeRoot: string) {}
+  constructor(private storeRoot: string, logger: DiagnosticsLogger) {
+    this.#logger = logger
+  }
 
   /**
    * Gets internal db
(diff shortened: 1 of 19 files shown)


oed:
Do we have a follow up ticket to update our docs?

packages/core/src/store/level-state-store.ts
```diff
@@ -41,16 +49,36 @@ export class LevelStateStore implements StateStore {
...
-    if (fs) {
-      fs.mkdirSync(storePath, { recursive: true }) // create dir if it doesn't exist
+    if (networkName == Networks.MAINNET || networkName == Networks.ELP) {
+      // If using "mainnet", check for data under an 'elp' directory first. This is because older
+      // versions of Ceramic only supported an 'elp' network as an alias for mainnet and stored
+      // state store data under 'elp' instead of 'mainnet' by mistake, and we don't want to lose
+      // that data if it exists.
+      const store = this._makeStore(Networks.ELP)
+      const pinnedStreamIds = await _listAll(store, 1)
+      if (pinnedStreamIds.length > 0) {
```
willex, line 72: Do we care that this will most likely stick around for a while? Should we capture this with a follow up ticket to migrate the state-store directory eventually?

willex (approved):

ukstv:
Is there an ETA for when this code can be removed, because there is no ELP anymore?

packages/core/src/store/level-state-store.ts
stbrody: I honestly don't know when we'll be able to remove support for this TBH.  Maybe if we force the state store to always be in the DB and stop supporting the S3 state store entirely.  But otherwise we might always have nodes out there still with data in the `elp` directory.  Curious if you have any thoughts on this @oed

stbrody:
> Do we have a follow up ticket to update our docs?

Filed https://linear.app/3boxlabs/issue/CDB-1947/update-docs-around-running-with-network=mainnet.  I'm not even sure how much the `--network=elp` thing was documented to begin with, but I'll make sure to do a quick pass over our docs to double check

packages/core/src/store/level-state-store.ts
oed: That's probably fine. Don't feel strongly.

PR #2503, opened 2022-10-19
chore(streamid): Strict TS config

ukstv:
- Throw/no-throw functions: throw by default, `NoThrow` function wraps a throwing one in a try-catch.
- `StreamRef.fromBytes` and `StreamRef.fromString` along with respective `StreamID` and `CommitID` parsing functions reuse same parsing logic. Control flow does not rely on errors. @stbrody Your precious debugging should be all right.

It is now slightly faster: https://github.com/ceramicnetwork/js-ceramic/blob/feature/cdb-1945-streamid-bench/packages/streamid/src/old-new.bench.ts

Screenshot 2022-10-24 at 16 49 48

One caveat. For some reason we allowed parsing StreamID from `/ceramic/<stream-id>?commit=<commit-id>` string. This is a drastic inconsistency with other formats, where presence of CommitID information leads to an error. I changed parsing here, so that we throw an error if commit is present in a string formatted like that.

Git blame leads to an obscure commit https://github.com/ceramicnetwork/js-ceramic/commit/e2dce7e30ed15f95eae0a96444b5800f56c4b812 which effectively made previous history opaque.

### DIFF
--- a/.changeset/orange-files-listen.md
+++ b/.changeset/orange-files-listen.md
@@ -1,0 +1,5 @@
+---
+"@ceramicnetwork/streamid": patch
+---
+
+chore(streamid): Stricter code. Also, now NoThrow functions really do not throw an error.
--- a/.changeset/orange-files-listen.md
+++ b/.changeset/orange-files-listen.md
@@ -1,5 +1,0 @@
----
-"@ceramicnetwork/streamid": patch
----
-
-chore(streamid): Stricter code. Also, now NoThrow functions really do not throw an error.
--- a/packages/streamid/src/commit-id.ts
+++ b/packages/streamid/src/commit-id.ts
@@ -59,11 +59,11 @@ function fromBytesNoThrow(bytes: Uint8Array): CommitID | Error {
  *
  * @param input - CID or string.
  */
-function parseCID(input: any): CID | undefined {
+function parseCID(input: any): CID | null {
   try {
     return typeof input === 'string' ? CID.parse(input) : CID.asCID(input)
   } catch {
-    return undefined
+    return null
   }
 }
 
@@ -76,7 +76,7 @@ function parseCID(input: any): CID | undefined {
  * @param genesis - genesis CID for stream
  * @param commit - representation of commit, be it CID, 0, `'0'`, `null`
  */
-function parseCommit(genesis: CID, commit: CID | string | number = null): CID | null {
+function parseCommit(genesis: CID, commit: CID | string | number | null = null): CID | null {
   if (!commit) return null
 
   const commitCID = parseCID(commit)
@@ -122,8 +122,9 @@ function fromString(input: string): CommitID {
 function fromStringNoThrow(input: string): CommitID | Error {
   const protocolFree = input.replace('ceramic://', '').replace('/ceramic/', '')
   if (protocolFree.includes('commit')) {
-    const commit = protocolFree.split('?')[1].split('=')[1]
+    const commit = protocolFree.split('?')[1]?.split('=')[1]
     const base = protocolFree.split('?')[0]
+    if (!base || !commit) return new Error(`Malformed commit string: ${input}`)
     return make(StreamID.fromString(base), commit)
   } else {
     return fromBytesNoThrow(base36.decode(protocolFree))
@@ -177,7 +178,11 @@ export class CommitID implements StreamRef {
    * new StreamID(<type>, <CID>|<cidStr>)
    * new StreamID(<type>, <CID>|<cidStr>, <CommitCID>|<CommitCidStr>)
    */
-  constructor(type: string | number, cid: CID | string, commit: CID | string | number = null) {
+  constructor(
+    type: string | number,
+    cid: CID | string,
+    commit: CID | string | number | null = null
+  ) {
     if (!type && type !== 0) throw new Error('constructor: type required')
     if (!cid) throw new Error('constructor: cid required')
     this.#type = typeof type === 'string' ? StreamType.codeByName(type) : type
--- a/packages/streamid/src/commit-id.ts
+++ b/packages/streamid/src/commit-id.ts
@@ -8,6 +8,7 @@ import { STREAMID_CODEC } from './constants.js'
 import { readCid, readVarint } from './reading-bytes.js'
 import { StreamID } from './stream-id.js'
 import { StreamRef } from './stream-ref.js'
+import { tryCatch } from './try-catch.util.js'
 
 /**
  * Parse CommitID from bytes representation.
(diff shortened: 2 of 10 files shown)


.changeset/orange-files-listen.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ceramicnetwork/streamid": patch
+---
+
+chore(streamid): Stricter code. Also, now NoThrow functions really do not throw an error.
```
stbrody, line 5: I think @smrz2001 reverted the use of changesets - can we take this file out for now?

packages/streamid/src/commit-id.ts
```diff
@@ -17,32 +18,14 @@ import { StreamRef } from './stream-ref.js'
...
-}
-
-/**
- * Same as fromBytes, but returns an Error instance rather than throwing if there is a problem
- * with the input.
- * Note that some exceptions can still be thrown in certain cases, if they come from lower-level
- * libraries like multibase, for example.
- * @param bytes
- */
-function fromBytesNoThrow(bytes: Uint8Array): CommitID | Error {
```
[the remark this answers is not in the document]
stbrody: same comment applies to the other noThrow methods throughout this PR

stbrody (changes requested):
main comment is avoiding codepaths that use thrown exceptions for control flow, as they make certain debugging tools much less useful (and also have a performance overhead)

packages/streamid/src/commit-id.ts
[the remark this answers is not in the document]
ukstv: @stbrody Hmm, okay. Concern accepted. Will think on how to address it. At the same time, are you aware that former `fromBytesNoThrow` throws on garbage input actually? :)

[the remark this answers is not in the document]
stbrody: Yes, I know it the old `NoThrow` versions could still throw in some edge cases.  Not ideal, but it does avoid throwing in the common cases which preserves my ability to use the debugger break-on-throw feature.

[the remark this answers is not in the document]
ukstv: Yes, maybe errors should have not been used for control flow indeed.

.changeset/orange-files-listen.md
ukstv: removed.

packages/streamid/src/stream-ref-parsing.ts
```diff
@@ -0,0 +1,120 @@
...
+      type: type,
+      genesis: genesis,
+    }
+  } else if (gnesisRemainder.length === 1) {
+    // Zero commit
+    return {
+      kind: 'commit-id',
+      type: type,
+      genesis: genesis,
+      commit: null,
```
stbrody, line 43: this is technically a small behavior change as now you're not actually extracting the final byte, you're just assuming it's '0', which is the only value that is supported.  I'm not sure if it really matters, but if we wanted to preserve existing behavior I think it'd be as simple as allowing the `commit` field here to have type 'string' as well and return `genesisRemainder[0]` for the commit here.

packages/streamid/src/commit-id.ts
```diff
@@ -17,11 +18,11 @@ import { StreamRef } from './stream-ref.js'
...
 function fromBytes(bytes: Uint8Array): CommitID {
-  const result = fromBytesNoThrow(bytes)
-  if (result instanceof Error) {
-    throw result
+  const parsed = parsing.fromBytes(bytes, 'CommitID')
+  if (parsed.kind === 'commit-id') {
+    return new CommitID(parsed.type, parsed.genesis, parsed.commit)
   }
-  return result
+  throw new Error(`No commit information provided`)
```
stbrody, line 25:
```suggestion
  throw new Error(`Error while parsing CommitID from byte array: no commit information provided`)
```

stbrody: The error messages from these errors can show in logs without much context about what was happening when the error was thrown.  We won't always have a stack trace.  So it's important to always try to put as much context into error messages as possible

packages/streamid/src/commit-id.ts
```diff
@@ -105,11 +44,11 @@ function parseCommit(genesis: CID, commit: CID | string | number = null): CID |
...
 function fromString(input: string): CommitID {
-  const result = fromStringNoThrow(input)
-  if (result instanceof Error) {
-    throw result
+  const parsed = parsing.fromString(input, 'CommitID')
+  if (parsed.kind === 'commit-id') {
+    return new CommitID(parsed.type, parsed.genesis, parsed.commit)
   }
-  return result
+  throw new Error(`No commit information provided`)
```
stbrody, line 51:
```suggestion
  throw new Error(`Error while parsing CommitID from string (${input}): no commit information provided`)
```

packages/streamid/src/stream-id.ts
```diff
@@ -19,11 +20,11 @@ import { StreamType } from './stream-type.js'
...
 function fromBytes(bytes: Uint8Array): StreamID {
-  const result = fromBytesNoThrow(bytes)
-  if (result instanceof Error) {
-    throw result
+  const parsed = parsing.fromBytes(bytes, 'StreamID')
+  if (parsed.kind === 'stream-id') {
+    return new StreamID(parsed.type, parsed.genesis)
   }
-  return result
+  throw new Error(`Invalid StreamID: contains commit`)
```
stbrody, line 27:
```suggestion
  throw new Error(`Invalid StreamID byte array: contains commit`)
```

packages/streamid/src/stream-id.ts
```diff
@@ -57,11 +46,11 @@ function fromBytesNoThrow(bytes: Uint8Array): StreamID | Error {
...
 function fromString(input: string): StreamID {
-  const result = fromStringNoThrow(input)
-  if (result instanceof Error) {
-    throw result
+  const parsed = parsing.fromString(input, 'StreamID')
+  if (parsed.kind === 'stream-id') {
+    return new StreamID(parsed.type, parsed.genesis)
   }
-  return result
+  throw new Error(`Invalid StreamID: contains commit`)
```
stbrody, line 53:
```suggestion
  throw new Error(`Invalid StreamID string (${input}): contains commit`)
```

stbrody (approved):
this is such a nice cleanup!  Thanks for doing the work to really make this nice and clean.  Left a few suggestions around a bit more detailed error messages, but otherwise looks great.

packages/streamid/src/stream-ref-parsing.ts
ukstv: I do not understand the suggestion. This function parses bytes. Byte-encoded CommitID can only contain `[0]` as `genesisRemainder` or a full `CID`. What is the intent behind the suggestion?

packages/streamid/src/commit-id.ts
ukstv: ack

packages/streamid/src/stream-id.ts
ukstv: okay, now error messages are more descriptive

packages/streamid/src/stream-ref-parsing.ts
stbrody: Your most recent commit made the behavior change even smaller, since now you check that `genesisRemainder[0] === 0` above.  I think now the only difference is if the final byte was something other than '0', in the old version we would throw the error `'Cannot specify commit as a number except to request commit 0 (the genesis commit)'`, whereas now we'll pass the extra byte into `readCid` and presumably get some error message out of that that it isn't a valid CID.

In the old version the genesisRemainder final byte gets passed into the CommitID constructor as the `commit` arg, which in turn got passed into the old `parseCommit` function which would then throw the error when it saw that the byte wasn't `'0'`.

So now I think the behavior is functionally the same, just potentially with a less descriptive error message for a certain category of malformed CommitID where it uses a single byte for the commit but that byte doesn't contain '0'.  Not a big deal.  If you want to preserve the existing behavior I think you could just change the line this comment is on to say `commit: genesisRemainder[0]`, since that will then get passed into the CommitID constructor, and then down into `parsing.parseCommit`, which will return the same error message that used to be thrown.

Up to you though, I don't think it's a big deal either way.

stbrody (approved):
still lgtm

packages/streamid/src/stream-ref-parsing.ts
ukstv: sweet!

PR #2505, opened 2022-10-19
chore(stream-tile): Make TS compiler strict

ukstv:
Note: `context.api` is mandatory now.

One more change: `StreamState#log` is `NonEmptyArray` now, which on type level gives assurance that `log[0]` is present. No run-time guarantees though, but it seems fine for now.

### DIFF
--- a/.changeset/tidy-eagles-try.md
+++ b/.changeset/tidy-eagles-try.md
@@ -1,0 +1,5 @@
+---
+"@fake-scope/fake-pkg": patch
+---
+
+chore(stream-tile): Make TS compiler strict
--- a/.changeset/tidy-eagles-try.md
+++ b/.changeset/tidy-eagles-try.md
@@ -1,5 +1,0 @@
----
-"@fake-scope/fake-pkg": patch
----
-
-chore(stream-tile): Make TS compiler strict
--- a/packages/common/src/context.ts
+++ b/packages/common/src/context.ts
@@ -12,5 +12,5 @@ export interface Context {
   anchorService?: AnchorService
   loggerProvider?: LoggerProvider
 
-  api?: CeramicApi // the self reference to the Ceramic API
+  api: CeramicApi // the self reference to the Ceramic API
 }
--- a/packages/stream-tile-handler/src/schema-utils.ts
+++ b/packages/stream-tile-handler/src/schema-utils.ts
@@ -2,18 +2,14 @@ import { TileDocument } from '@ceramicnetwork/stream-tile'
 import { CeramicApi } from '@ceramicnetwork/common'
 import { CommitID } from '@ceramicnetwork/streamid'
 import Ajv from 'ajv'
+import type { Schema } from 'ajv'
 import addFormats from 'ajv-formats'
 import lru from 'lru_map'
-import isPlainObject from 'lodash.isplainobject'
 
 function buildAjv(): Ajv {
   const validator = new Ajv({ allErrors: true, strictTypes: false, strictTuples: false })
   addFormats(validator)
   return validator
-}
-
-function isRecord(input: unknown): input is Record<string, any> {
-  return isPlainObject(input)
 }
 
 const AJV_CACHE_SIZE = 500
@@ -35,7 +31,6 @@ export class SchemaValidation {
     schemaStreamId: string
   ): Promise<void> {
     const schema = await this._loadSchemaById(ceramic, schemaStreamId)
-    if (!isRecord(schema)) throw new Error(`Invalid schema in stream ${schemaStreamId}: The contents of the schema stream are not an object`)
     this._validate(content, schema, schemaStreamId)
   }
 
@@ -49,20 +44,18 @@ export class SchemaValidation {
     } catch {
       throw new Error('Commit missing when loading schema document')
     }
-    return ceramic.loadStream<TileDocument<T>>(commitId).then((doc) => doc.content)
+    const stream = await ceramic.loadStream<TileDocument<T>>(commitId)
+    return stream.content
   }
 
-  private _validate(
-    content: Record<string, any>,
-    schema: Record<string, any>,
-    schemaStreamId: string
-  ): void {
+  private _validate(content: Record<string, any>, schema: unknown, schemaStreamId: string): void {
     let validator = this.validators.get(schemaStreamId)
     if (!validator) {
       validator = buildAjv()
       this.validators.set(schemaStreamId, validator)
     }
-    const isValid = validator.validate(schema, content)
+    // Type casting, because we know that validator would throw an error if schema is wrong.
+    const isValid = validator.validate(schema as Schema, content)
 
     if (!isValid) {
       const errorMessages = validator.errorsText()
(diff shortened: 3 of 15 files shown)


.changeset/tidy-eagles-try.md
```diff
@@ -0,0 +1,5 @@
+---
```
stbrody, line 1: can we remove the changeset?

packages/common/src/context.ts
```diff
@@ -12,5 +12,5 @@ export interface Context {
   anchorService?: AnchorService
   loggerProvider?: LoggerProvider
 
-  api?: CeramicApi // the self reference to the Ceramic API
+  api: CeramicApi // the self reference to the Ceramic API
```
stbrody, line 15: 👍

packages/stream-tile-handler/src/schema-utils.ts
```diff
@@ -30,10 +35,14 @@ export class SchemaValidation {
     schemaStreamId: string
   ): Promise<void> {
     const schema = await this._loadSchemaById(ceramic, schemaStreamId)
+    if (!isRecord(schema)) throw new Error(`Invalid schema ${schemaStreamId}: Not object`)
```
stbrody, line 38: I don't really understand this - what situation is this trying to catch/prevent?

packages/stream-tile/src/tile-document.ts
```diff
@@ -370,13 +373,14 @@ export class TileDocument<T = Record<string, any>> extends Stream {
     }
 
     const patch = jsonpatch.compare(this.content, newContent)
-    const commit: RawCommit = {
+    const genesisLogEntry = this.state.log[0]
+    if (!genesisLogEntry) throw new Error(`No genesis log entry`)
```
stbrody, line 377: given this function isn't static, but is called on an actual TileDocument instance, and the fact that you can't construct an actual TileDocument instance if there's no genesis commit, then it seems like this check shouldn't be necessary?

.changeset/tidy-eagles-try.md
ukstv: Would like to summon @smrz2001 here. 

I have a suspicion, that it is here to stay with us. Maybe the right process is to fill it right before the merge.

I think you can mark the file as viewed to hide it.

Screenshot 2022-10-20 at 21 50 51

packages/stream-tile-handler/src/schema-utils.ts
ukstv: Whatever is referenced by `schemaStreamId` is free to contain some garbage, which is definitely not a schema, like a single string.

.changeset/tidy-eagles-try.md
smrz2001: Hmm, I feel like we can probably avoid adding changesets for the time being.

We currently have two orthogonal decisions to make - how to make releases not painful + move to a more efficient package manager, the former being the more urgent, blocking issue.

If/when we decide to move to `pnpm` for the latter, having changesets will be necessary, but it seems like we can fix the release process with `lerna` without needing changesets yet.

packages/stream-tile/src/tile-document.ts
ukstv: IMO, this is exactly pitfall in reasoning highlighted by TS now.

See for example this code in temp file `./packages/http-client/src/foo.ts`.
```
import { AnchorStatus, SignatureStatus, StreamState, TestUtils } from '@ceramicnetwork/common'
import { TileDocument } from '@ceramicnetwork/stream-tile'
import { CeramicClient } from './ceramic-http-client.js'
import { Document } from './document.js'

const a: StreamState = {
  log: [],
  type: 0,
  content: undefined,
  metadata: undefined,
  signature: SignatureStatus.GENESIS,
  anchorStatus: AnchorStatus.NOT_REQUESTED
}

const ceramic = new CeramicClient()
const d = new Document(a, new URL('http://ya.ru'), 1000000)
console.log('document created')
const b = new TileDocument(TestUtils.runningState(a), ceramic.context)
console.log('tile-document created')
```

It compiles fine. The only reason it fails _in runtime_ is because `TestUtils.runningState` calculates stream id when creating `RunningStateLike` object for optimisation purposes basically. If not for that, there is literally nothing to prevent `this.state` in TileDocument from having zero log entries.

.changeset/tidy-eagles-try.md
smrz2001: Having said that, do we want to move to `pnpm` next? It offers many advantages over `lerna`, but also comes with the (non-zero but also not large) extra effort of working with changesets.

I have a `pnpm` prototype PR ready, and could probably make that happen soon, but would defer to your judgement on this, @ukstv @stbrody.

packages/stream-tile-handler/src/schema-utils.ts
stbrody: ah, so you make a TileDocument, put some random data in it that isn't actually a json-schema, then use that TileDocument's streamid as the schema field for another tile.  This is the first place where we'd notice.  Makes sense

packages/stream-tile/src/tile-document.ts
stbrody: hmm... that's a pretty big abuse of the API though.  Note the use of `TestUtils.runningState`, that's something regular users shouldn't know about or have access to.

It seems like really the http-client should be validating that the state object it receives from the daemon is valid, but then code within TileDocument (or the other Stream handles) should be able to assume they are operating with a valid StreamState.  Validation should happen at the edges, not interspersed everywhere

packages/stream-tile-handler/src/schema-utils.ts
ukstv: Sarcasm? :) Types could be a pain, but in general, if we do not try to overstep limits of a particular type system, they highlight risks like that.

Regarding overstepping: TypeScript types are definitely not enough to fully implement dependent types (="this array contains strictly N elements based on this constructor parameter `n`"), or [algebraic effects](https://overreacted.io/algebraic-effects-for-the-rest-of-us/). Fun fact: vanilla JS is better suited for algebraic effects. Some time here and there we would reach the limits. Mostly though, TS types are expressive enough.

packages/stream-tile/src/tile-document.ts
ukstv: Oh, Spencer, yes, I can not agree more. I use here `TestUtils.runningState` for illustrative purposes. What is said above about it, is equally true for our `RunningState`. It is a nice coincidence that `RunningState` tries to construct `id` of a stream when constructing an object. We could have made it lazily computed though. Then, we would have no guarantees at construction time, that log is not empty.

> Validation should happen at the edges, not interspersed everywhere

Agree. Remember we had a discussion once, how (de)serialization is linked to our StateStore and shape of StreamState? I guess now it becomes clear (to me, to you, to others) how this two topics are linked.

packages/stream-tile-handler/src/schema-utils.ts
stbrody: Nope, I wasn't being sarcastic!  I was genuinely trying to understand in what situation someone could trigger that error.  :slightly_smiling_face:

PR #2506, opened 2022-10-19
chore(model-instance-document-handler): Stricter TS compiler

ukstv:


.changeset/tough-tomatoes-yell.md
```diff
@@ -0,0 +1,5 @@
+---
```
stbrody, line 1: can we remove this changeset?

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
```diff
@@ -104,9 +104,11 @@ export class ModelInstanceDocumentHandler implements StreamHandler<ModelInstance
 
     const isSigned = StreamUtils.isSignedCommitData(commitData)
     if (isSigned) {
+      const did = context.did
+      if (!did) throw new Error(`No did is set`)
```
stbrody, line 108: same comment from other PR: if we made `did` required in Context would that let us get rid of these checks?

ukstv: Most probably, yes.

ukstv: For now, there is no programmatic guarantee anywhere, that did is present at compile time. Type system now just highlights the fact.

stbrody: I'd rather fix it once so we know which things are supposed to always be present and can check for nulls just at the boundaries of user input and then trust the types within deeper code, rather than force engineers to be super defensive and have null checks everywhere throughout our whole codebase

PR #2510, opened 2022-10-19
add metrics for IPFS timeouts and other things - PLAT-849

gvelez17:
## Description

We want to see charts of IPFS timeouts, other errors, and also pin success/failure

## How Has This Been Tested?

Tests pass

## PR checklist

## References:

### DIFF
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -2,7 +2,7 @@ import * as fs from 'fs'
 import express, { Request, Response } from 'express'
 import { Ceramic, CeramicConfig } from '@ceramicnetwork/core'
 import { RotatingFileStream } from '@ceramicnetwork/logger'
-import { Metrics } from '@ceramicnetwork/metrics'
+import { Metrics, METRIC_NAMES } from '@ceramicnetwork/metrics'
 import { buildIpfsConnection } from './build-ipfs-connection.util.js'
 import { S3StateStore } from './s3-state-store.js'
 import {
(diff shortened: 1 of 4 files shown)


packages/core/src/dispatcher.ts
```diff
@@ -238,11 +241,12 @@ export class Dispatcher {
           console.warn(
             `Timeout error while loading CID ${asCid.toString()} from IPFS. ${retries} retries remain`
           )
+          Metrics.count(METRIC_NAMES.IPFS_TIMEOUT, 1)
           if (retries > 0) {
             continue
           }
         }
-
+
```
smrz2001, line 249: Can revert this.

packages/cli/src/ceramic-daemon.ts
```diff
@@ -816,6 +816,7 @@ export class CeramicDaemon {
     const streamId = StreamID.fromString(req.params.streamid || req.params.docid)
     const { force } = req.body
     await this.ceramic.pin.add(streamId, force)
+    Metrics.count(METRIC_NAMES.STREAM_PINNED, 1)
```
smrz2001, line 819: Some of these flows are synchronous API calls, will the metrics code add any latency?

smrz2001 (approved):
LGTM apart from a couple of minor comments.

packages/cli/src/ceramic-daemon.ts
gvelez17: it should not - it should be counting internally and the exporter able to be scraped.  it does not have to do a network call to count.

PR #2511, opened 2022-10-20
chore(stream-model-instance): Stricter TS

ukstv:


packages/stream-model-instance/src/model-instance-document.ts
```diff
@@ -100,13 +101,17 @@ export class ModelInstanceDocument<T = Record<string, any>> extends Stream {
...
-  get content(): T {
+  override get content(): T {
     return super.content
   }
 
   get metadata(): ModelInstanceDocumentMetadata {
     const metadata = this.state$.value.metadata
-    return { controller: metadata.controllers[0], model: metadata.model }
+    const controller = metadata.controllers[0]
+    if (!controller) throw new Error(`Controller is not present`)
```
stbrody, line 111: that shouldn't be possible, if controller wasn't present the handler should have failed when the stream was loaded

packages/stream-model-instance/src/model-instance-document.ts
```diff
@@ -100,13 +101,17 @@ export class ModelInstanceDocument<T = Record<string, any>> extends Stream {
...
+    const model = metadata.model
+    if (!model) throw new Error(`Model is not present`)
```
stbrody, line 113: same for model, both of these should be checked by the model-instance-document-handler as part of loading the stream in the first place.  If it gets far enough to have a ModelInstanceDocument instance, this should already be guaranteed to be set

packages/stream-model-instance/src/model-instance-document.ts
```diff
@@ -242,10 +247,12 @@ export class ModelInstanceDocument<T = Record<string, any>> extends Stream {
    */
   private _makeRawCommit(newContent: T | null): RawCommit {
     const patch = jsonpatch.compare(this.content, newContent || {})
+    const genesisLogEntry = this.state.log[0]
+    if (!genesisLogEntry) throw new Error(`Genesis log entry is absent`)
```
stbrody, line 251: this also should be guaranteed to be set

stbrody (commented):
a lot of these changes don't seem like they should be necessary based on the logic of how the code works.  Is the problem that we're not using the type system right so we're getting compiler errors about things that aren't actually issues?

ukstv:
Type system is definitely mis-used. The changes here just highlight pitfalls in reasoning behind the types.

stbrody:
> Type system is definitely mis-used. The changes here just highlight pitfalls in reasoning behind the types.

what would it take to fix the types instead of having to band-aid every time we use them?

PR #2512, opened 2022-10-20
chore(stream-model-handler): Stricter TS

ukstv:


packages/stream-model-handler/src/__tests__/model-handler.test.ts
```diff
@@ -48,7 +51,6 @@ const hash = (data: string): CID => {
 }
 
 const FAKE_CID_1 = CID.parse('bafybeig6xv5nwphfmvcnektpnojts33jqcuam7bmye2pb54adnrtccjlsu')
-const FAKE_CID_2 = CID.parse('bafybeig6xv5nwphfmvcnektpnojts44jqcuam7bmye2pb54adnrtccjlsu')
```
stbrody, line 51: Nitpick, but can we rename FAKE_CID_3 and FAKE_CID_4 to FAKE_CID_2 and FAKE_CID_3?  It's weird to have a gap in the numbers between 1 and 3 like that

packages/stream-model-handler/src/model-handler.ts
```diff
@@ -100,6 +100,8 @@ export class ModelHandler implements StreamHandler<Model> {
    * @private
    */
   async _applyGenesis(commitData: CommitData, context: Context): Promise<StreamState> {
+    const did = context.did
```
stbrody, line 103: can we just make `did` required within Context?

PR #2513, opened 2022-10-20
chore(stream-model): Stricter TS

ukstv:


packages/stream-model/src/model.ts
```diff
@@ -155,12 +156,14 @@ export class Model extends Stream {
...
 
-  get content(): ModelDefinition {
+  override get content(): ModelDefinition {
     return super.content
   }
 
   get metadata(): ModelMetadata {
-    return { controller: this.state$.value.metadata.controllers[0], model: Model.MODEL }
+    const controller = this.state$.value.metadata.controllers[0]
+    if (!controller) throw new Error(`Controller must be present`)
```
stbrody, line 165: same comment as from the other PR on model-instance-document, this should always be set

PR #2516, opened 2022-10-21
feat(ci): new release workflow

smrz2001:
## Description

This PR addresses the long standing work of automating our release/RC/hotfix workflows while avoiding package version conflicts between them. The new workflow continues to use `lerna`, and thus does not require any changes to the main `js-ceramic` code or packaging.

## How Has This Been Tested?

I have tested these flows indirectly (i.e. without actually publishing packages) but will test them formally and check off the tests accordingly:
- [ ] Publish hotfix
- [ ] Publish RC
- [ ] Publish release

## References:

[Here's](https://discord.com/channels/484729862368526356/1032056162218147931) the relevant Discord discussion. The summary of the final approach is:
- Releases graduate the minor version (e.g. `2.2.0-rc.0` => `2.2.0`)
- Release candidates bump the minor version, pre-release style (e.g. `2.1.9` => `2.2.0-rc.0`)
- Hotfixes bump the patch version (e.g. `2.1.9` => `2.1.10`)

Fixes [PLAT-644](https://linear.app/3boxlabs/issue/PLAT-644/implement-release-workflow) and [PLAT-856](https://linear.app/3boxlabs/issue/PLAT-856/dont-do-releases-until-build-is-verified).


package.json
```diff
@@ -5,8 +5,9 @@
     "test": "npx lerna exec npm t",
     "build": "npx lerna run build",
     "docs": "npx typedoc",
-    "publish:release-candidate": "npx lerna publish --allow-branch release-candidate --dist-tag next --preid rc --no-verify-access --conventional-prerelease",
-    "publish:latest": "npx lerna publish --allow-branch main --create-release github --no-verify-access --conventional-graduate",
+    "publish:release-candidate": "npx lerna publish --allow-branch release-candidate --dist-tag next --preid rc --no-verify-access --conventional-prerelease bump preminor",
+    "publish:release": "npx lerna publish --allow-branch main --dist-tag latest --no-verify-access --conventional-graduate bump minor",
+    "publish:hotfix": "npx lerna publish --allow-branch $(git rev-parse --abbrev-ref HEAD) --no-verify-access --conventional-graduate bump patch",
```
stbrody, line 10: should we just drop the `--allow-branch` flag entirely if we're always taking the current branch anyway?

package.json
```diff
@@ -5,8 +5,9 @@
     "test": "npx lerna exec npm t",
     "build": "npx lerna run build",
     "docs": "npx typedoc",
-    "publish:release-candidate": "npx lerna publish --allow-branch release-candidate --dist-tag next --preid rc --no-verify-access --conventional-prerelease",
-    "publish:latest": "npx lerna publish --allow-branch main --create-release github --no-verify-access --conventional-graduate",
+    "publish:release-candidate": "npx lerna publish --allow-branch release-candidate --dist-tag next --preid rc --no-verify-access --conventional-prerelease bump preminor",
+    "publish:release": "npx lerna publish --allow-branch main --dist-tag latest --no-verify-access --conventional-graduate bump minor",
```
stbrody, line 9: was it intentional to drop creating the release in github?

stbrody:
It'd be nice to also update https://github.com/ceramicnetwork/js-ceramic/blob/develop/DEVELOPMENT.md with instructions on how to do a hotfix, but feel free to file a ticket and do in a separate PR

.github/workflows/npm-publish.yml
```diff
@@ -4,191 +4,238 @@ on:
...
           git config user.email "${{ github.actor }}@users.noreply.github.com"
           git config user.name ${{ github.actor }}
           git fetch --all
-      - name: Merge down from `main` -> `rc` -> `develop`
-        if: ${{ steps.set-env.outputs.main == 'true' }}
+      -
+        name: Merge down from main -> rc -> develop
         run: |
           git checkout -B release-candidate refs/remotes/origin/release-candidate
           git merge main
```
stbrody, line 52: Do we need a `git push` step here?  We have it below for the post-release merges.

.github/workflows/npm-publish.yml
```diff
@@ -4,191 +4,238 @@ on:
...
+          install-only: true
+      -
+        name: Verify Docker image
+        run: |
+          dagger project init
+          dagger project update
+          dagger project update "github.com/3box/pipeline-tools/ci@v${{ env.PIPELINE_TOOLS_VER }}"
+          dagger do verify -p ${{ env.DAGGER_PLAN }}
+      -
+        name: Initialize and build code
```
stbrody, line 72: we don't need to build the code before the docker image?

.github/workflows/npm-publish.yml
```diff
@@ -4,191 +4,238 @@ on:
...
-          PUBLISH_ARGS: ${{ steps.set-env.outputs.publishArgs }}
+          GH_TOKEN: ${{ env.GH_TOKEN }}
         run: |
           npm config set //registry.npmjs.org/:_authToken $NODE_AUTH_TOKEN
-          npm run publish:latest $PUBLISH_ARGS
-          git commit -m 'chore(core): bump version' -- packages/core/src/version.ts || true
-          git push
-      - name: Update and push code to `release-candidate` and `develop`
-        if: ${{ steps.set-env.outputs.main == 'true' }}
+          npx lerna publish --allow-branch main --dist-tag latest --no-verify-access --conventional-graduate bump minor --yes
```
stbrody, line 83: is there any way to re-use the `publish:release` command from the package.json so we don't need to keep the command in sync in 2 different places?

stbrody (commented):
I'm so incredibly excited for this

.github/workflows/npm-publish.yml
smrz2001: Yeah, I was thinking about that. The `npm` command will run from whichever branch is checked-out, so it might not be one that has the very latest `package.json` changes to any of the `publish` commands.

However, this is likely just temporary till all the branches have the latest file, and these commands are unlikely to change often.

For next week, we might have to make another RC (even if not necessary otherwise) so that these new commands are merged into the RC branch, _then_ make the release, _then_ make the new RC.

.github/workflows/npm-publish.yml
smrz2001: No, we build the code while making the Docker image, so not necessary. And this way, if verification fails for some reason, we won't have spent time building the code twice.

.github/workflows/npm-publish.yml
smrz2001: No, not needed here. The `publish` command will automatically push.

package.json
smrz2001: I did intentionally remove this because it's a PITA to fix tags if something gets screwed up. We have to go in and delete all 5000 tags from GitHub if we need to redo something. This shouldn't be necessary going forward, hopefully, but it might till any bugs get ironed out.

This will also stop releases from being available on GitHub going forward though. Do we care that much about them? Don't think people download packages from GitHub. I can add this back and hope we don't have to fix tags lol.

package.json
smrz2001: Yes 👍🏼

package.json
smrz2001: Actually, you know what, scratch that. The git tags will get pushed anyway... I'll just add this back.

smrz2001:
> I'm so incredibly excited for this

LOLOL, me too - both excited and terrified finally being able to add back the `--yes`.

.github/workflows/npm-publish.yml
stbrody: Could we separate the changes to the package.json from the github workflow changes into two PRs, and merge/release the first before the second?

.github/workflows/npm-publish.yml
stbrody: annoying that the two steps can share the build, but oh well

.github/workflows/npm-publish.yml
stbrody: but the publish happens on main, here we're merging to the release-candidate and develop branches

package.json
```diff
@@ -5,9 +5,9 @@
     "test": "npx lerna exec npm t",
     "build": "npx lerna run build",
     "docs": "npx typedoc",
-    "publish:release-candidate": "npx lerna publish --allow-branch release-candidate --dist-tag next --preid rc --no-verify-access --conventional-prerelease bump preminor",
-    "publish:release": "npx lerna publish --allow-branch main --dist-tag latest --no-verify-access --conventional-graduate bump minor",
-    "publish:hotfix": "npx lerna publish --allow-branch $(git rev-parse --abbrev-ref HEAD) --no-verify-access --conventional-graduate bump patch",
+    "publish:release-candidate": "npx lerna publish --dist-tag next --preid rc --no-verify-access --conventional-prerelease bump preminor --yes",
```
stbrody, line 8: can we leave the `--yes` out of the command so it still asks for confirmation when running manually, but have the github action pass in `--yes` as an additional argument?

DEVELOPMENT.md
```diff
@@ -94,14 +94,26 @@ $ git checkout master
...
+The main step that creates the release is `npm run publish:release`. This creates the release on NPM, as well as making a git commit bumping the version, new git tags, and a release on github. Make sure to set the [GH_TOKEN](https://github.com/lerna/lerna/tree/master/commands/version#--create-release-type) environment variable and log into npm before you run this command.
+
+After the release, don't forget to make a post in the #releases channel of the Ceramic discord to notify the community about the new release!
+
+### Hotfix release
+```
+export GH_TOKEN=<your github token>       # You need a valid github api token set to create the release on github
+$ git checkout <hotfix branch>
+$ npm install && npm run build            # Make sure to build Ceramic and install current dependencies before releasing
+$ npm adduser                             # login to npm accunt
```
stbrody, line 114:
```suggestion
$ npm adduser                             # login to npm account
```

stbrody: here and above

DEVELOPMENT.md
```diff
@@ -94,14 +94,26 @@ $ git checkout master
...
+$ npm run publish:hotfix                  # create and publish the release
+```
+The main step that creates the release is `npm run publish:hotfix`. This creates the release on NPM, as well as making a git commit bumping the version and new git tags on github. Make sure to set the [GH_TOKEN](https://github.com/lerna/lerna/tree/master/commands/version#--create-release-type) environment variable and log into npm before you run this command.
```
stbrody, line 117: Can we add some description of what a hotfix is, and explain that whatever branch it's run against will be released as a real full release, so it should generally be done from a recent fork of `main`?

package.json
smrz2001: --yes ;)

.github/workflows/npm-publish.yml
smrz2001: We could but we usually promote changes to rc/main through the release process. To get just the package.json changes up, we'd have to do the same thing - make the RC to get it to rc branch, then release to get to main, then another RC.

Or maybe I'm missing what you're thinking about how this could work?

.github/workflows/npm-publish.yml
smrz2001: Oh wait, I thought you were looking at a different spot. I intentionally left push out from this spot to avoid extra commits since each commit comes with CircleCI tests + image verification + deploys (since this happens on deployable branches).

This increases the downtime from redeploys when we could just wait a few more steps and then push.

Though, now that I'm thinking about it, each commit, even pushed together with other later, will still do all these steps :( The one advantage of waiting till the end is that the CD manager is likely to collapse multiple back-to-back deploys.

Not a huge advantage with our current workflows, I guess, but at least now you know the idea behind doing it this way :)

.github/workflows/npm-publish.yml
smrz2001: Actually, I understand what you mean. Is it worth the trouble though? We'd have to make RCs and releases twice to test the changes anyway, even if we merge just the package changes in the first round.

stbrody (approved):
🥳

PR #2521, opened 2022-10-25
chore: Consolidate anchor commit application logic

stbrody:
Also introduce @ceramicnetwork/stream-handler-common package to house code shared between the various streamtype handlers.

PR #2523, opened 2022-10-25
fix(stream-handler-common): Fix loading of historical commits with CACAOs

stbrody:
Builds on https://github.com/ceramicnetwork/js-ceramic/pull/2521

### DIFF
--- a/packages/stream-handler-common/src/anchor-commit-utils.ts
+++ b/packages/stream-handler-common/src/anchor-commit-utils.ts
@@ -36,7 +36,6 @@ export async function applyAnchorCommit(
 
   state.anchorStatus = AnchorStatus.ANCHORED
   state.anchorProof = commitData.proof
-
   state.log.push({
     cid: commitData.cid,
     type: CommitType.ANCHOR,
--- a/packages/stream-handler-common/src/anchor-commit-utils.ts
+++ b/packages/stream-handler-common/src/anchor-commit-utils.ts
@@ -16,6 +16,16 @@ export async function applyAnchorCommit(
 
   state.anchorStatus = AnchorStatus.ANCHORED
   state.anchorProof = commitData.proof
+
+  // Apply anchor timestamp to all commits that it anchors
+  for (let i = state.log.length - 1; i >= 0; i--) {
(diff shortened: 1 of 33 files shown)


stbrody:
There is still one edge case present here.  For already existing streams pinned in the state store that were created with old/existing versions of Ceramic, the state object in the state store won't have the timestamp set for the log entries.  This means if a user loads a stream created by an old version of Ceramic, then tries to load at a CommitID, they can hit this bug still.  They should be able to fix the issue on a stream-by-stream basis by loading the stream with the `SYNC_ALWAYS`, which as of Sergey's recent change now causes the node to throw out the stream state and fully re-sync it from the network, where the state will be built according to the new logic that back-propagates the anchor timestamp to all entries in the log.

So I guess I'm wondering if it makes sense to change the logic when applying anchor commits to iterate over not just the most recent segment of the log up to the last anchor commit, but actually iterate over the whole log, propagating anchor timestamps back all the way to genesis?  This would allow a more seamless upgrade, but at the cost of always iterating over the whole stream log every time we apply an anchor commit.

stbrody:
I guess even if the log is comparatively long, we're just iterating over the entries in memory, not doing any i/o or anything, so it should still be pretty cheap and fast, I think I'll go ahead and make the change.

packages/stream-handler-common/src/anchor-commit-utils.ts
```diff
@@ -16,12 +16,29 @@ export async function applyAnchorCommit(
...
+  // stopping, so we don't have to iterate over the whole log.
+  let timestamp
+  for (let i = state.log.length - 1; i >= 0; i--) {
+    const logEntry = state.log[i]
+    if (logEntry.type == CommitType.ANCHOR) {
+      timestamp = logEntry.timestamp
+    } else {
+      logEntry.timestamp = timestamp
+    }
+  }
```
oed, line 40: Maybe put this into a separate function?

oed (approved):
lgtm

We need to be careful about this logic when we start allowing the log to have multiple branches.

ukstv (approved):

PR #2525, opened 2022-10-27
Fix/address_prometheus_errors PLAT-1093

gvelez17:
## Description

PrometheusExporter threw errors in integration tests; need to catch; however, in the latest version of the PrometheusExporter the error seems not possible, so upgrading.

## How Has This Been Tested?

- [x ] ran locally
- [ ] need to run branch on dev or trigger integration tests

## PR checklist

## References:

### DIFF
--- a/packages/metrics/package.json
+++ b/packages/metrics/package.json
@@ -41,7 +41,7 @@
   "dependencies": {
     "@opentelemetry/exporter-prometheus": "^0.33.0",
     "@opentelemetry/resources": "1.2.0",
-    "@opentelemetry/sdk-metrics-base": "^0.28.0",
+    "@opentelemetry/sdk-metrics": "^0.33.0",
     "@opentelemetry/semantic-conventions": "^1.3.0"
   },
   "devDependencies": {
--- a/packages/metrics/package.json
+++ b/packages/metrics/package.json
@@ -39,7 +39,7 @@
     "clean": "npx rimraf ./lib"
   },
   "dependencies": {
-    "@opentelemetry/exporter-prometheus": "^0.28.0",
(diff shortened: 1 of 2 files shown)


gvelez17:
So a couple things here:

Examining the current source code, the above error does not seem possible
https://github.com/open-telemetry/opentelemetry-js/blob/0d4c71fc50f8be5fe85dc38eeb622580422d92cd/experimental/packages/opentelemetry-exporter-prometheus/src/PrometheusExporter.ts#L92

so upgraded from 0.28.0 to 0.33.0

However, prometheusExporter is still marked experimental.  It may be better now that we have the Collector to actually  refactor https://github.com/ceramicnetwork/ceramic-anchor-service/blob/develop/src/service-metrics.ts as a separate package and use it for both CAS and js-ceramic, using the Collector in place of the PrometheusExporter

However, that may take an additional day or two so submitting this for now.

One other note!  This PR will cause the NPM package version to update, how do we handle that correctly? @stbrody  this must be something we deal with all the time but I am confused as to how to get the new package number to generate prior to making the build

gvelez17:
hey @ukstv do I need to do something else to avoid the problem you fixed in https://github.com/ceramicnetwork/js-ceramic/pull/2524 ?  I'm not clear on how the packages are supposed to update

ukstv:
@gvelez17 This blog post https://michaelsoolee.com/npm-pre-1-caret-rules/ explains the situation pretty succinctly. Caret for `0.x.y` to NPM means only `y` is auto-incremented. So, considering 0.x schema used for prometheus packages, you'll have to bump their versions manually until they hit 1.0.

ukstv (approved):

PR #2526, opened 2022-10-27
chore: Log connected ipfs node multiaddr at node startup

ukstv:


packages/cli/src/ceramic-daemon.ts
```diff
@@ -276,15 +276,18 @@ export class CeramicDaemon {
...
-      )
+    const diagnosticsLogger = modules.loggerProvider.getDiagnosticsLogger()
+    diagnosticsLogger.imp(
+      `Starting Ceramic Daemon at version ${packageJson.version} with config: \n${JSON.stringify(
+        opts,
+        null,
+        2
+      )}`
+    )
+    const ipfsId = await ipfs.id()
```
stbrody, line 287: any reason you put this here instead of in `core`?  Seems like it would fit well in `Ceramic._init()`, or even in IpfsTopology.

stbrody (approved):
lgtm.  Optional suggestion to move the logging lower down into core or ipfs-topology, but up to you, it works here too

packages/cli/src/ceramic-daemon.ts
ukstv: The message is useful if you run Ceramic daemon on cloud. It gives you info on how to expose an IPFS node without fiddling with IPFS API. The daemon code is responsible for building IPFS connection. Let it handle exposing relevant diagnostic information as well.

IpfsTopology is about connecting the IPFS node to other nodes, so not quite a fit.
Core - it is responsible for too many things at once.

In some sense, it is an intersection of a single responsibility principle and Edge vs Kernel dichotomy.

PR #2528, opened 2022-10-28
chore: Handle cookies in browser

ukstv:
Cookies are remembered by browser.
Cookies do not work if a request is originated from Node.js

It would require a wrapper around one of the "native" fetch implementations. Due to sheer complexity of handling that in a cross-platform way, I better slightly postpone handling of cookies in Node.js. Expectation anyway is absolute most of the requests come from a browser.

So, treat this PR as a starter before moving to better more future-proof fetch based on undici.


stbrody (approved):
lgtm

PR #2543, opened 2022-11-04
feat: Provisionary dedupe of pinning

ukstv:
Call `this.repository.pin` only if `force` is true. Otherwise it is a fancy no-op.


stbrody (approved):

oed:
This needs merging?

ukstv:
It does. Figuring out what's wrong with the tests.

PR #2549, opened 2022-11-10
chore(core): prefix columns for user-defined fields

PaulLeCam:
## Description

Prefixes all the columns created based on user-defined fields in model definitions, to avoid collisions with columns used internally.

## How Has This Been Tested?

- [ ] Updated tests checking the created tables for models

## PR checklist


PaulLeCam:
@willex any suggestion on how to manage migrations for existing tables or should we simply ignore and tell people to re-create their DB please?

willex (commented):
Overall looks good. We want to be careful with this though and communicate where necessary (e.g. grant cohort, adoption, product) as this will break existing deployments.

willex (approved):

PR #2557, opened 2022-11-14
feat/CAS without a node - persistent storage - CDB-2008

pawartur:
## Description

1. Refactored StateStore, so that it's gonna be able to reuse the same level/s3 storages for state store and for anchor request store (injected to Repository):
  * created `IKVStore`  interface
  * created `LevelDBStore` that implements `IKVStore` and is a thin wrapper on LevelDB
  * created `S3Store` that implements `IKVStore` and is a thin wrapper on S3 level S3Store
  * created `StateStore` class that opens an instance of `IKVStore` and uses it to store and retrieve stream states
  * removed `LevelStateStore` and `S3StateStore`
  * started injecting `IKVStore` stores into `Repository`, rather than `StateStores` into `PinStoreFactory`
2. Implemented the first version of `AnchorRequestStore`:
  * added support for "use cases" `LevelDBStore` and `S3Store`:
    * for LevelDB, stream states are stored in `statestore/<network-name>` (the default use case) and other use cases are stored in `statestore/<network-name>-<use-case-name>`
    * for S3, stream states are stored in `ceramic/<network-name>/state-store` (the default use case) and other use cases are stored in `ceramic/<network-name>/state-store-<use-case-name>`
  * created `IObjectStore`
  * created `ObjectStore<TKeyObject, TValue>` class
  * made `StateStore` extend `ObjectStore`
  * added `AnchorRequestStore` extending `ObjectStore`
  * started creating an instance of `AnchorRequestStore` in core ceramic client and injecting it to repository and state manager

## How Has This Been Tested?

- [X] Added tests for LevelDBStoreWrapper
- [X] As part of the refactoring from 1., refactored all the tests and made sure they pass
- [X] Manually tested against S3 state storage by running the daemon with `CERAMIC_ENABLE_EXPERIMENTAL_COMPOSE_DB="true" node packages/cli/bin/ceramic.js daemon --state-store-s3-bucket <bucket-name>`, creating a composite and a few documents
  
## PR checklist


packages/cli/src/s3-store-wrapper.ts
```diff
@@ -0,0 +1,101 @@
...
+  private throwIfNotInitialized(): void {
+    if (!this.#store) throw new Error('You must call async init(), before you start using the S3Store')
+  }
+
+  private getLocation(subChannel?: string) {
+    const baseLocation = this.#bucketName + '/ceramic/' + this.networkName + '/state-store'
+    return subChannel ? `${baseLocation}/${subChannel}` : baseLocation
+  }
+
+  private getStore(subChannel?: string) {
```
ukstv, line 41: IMO, optional `subChannel` location could be handled better, if you do not have a division between `this.#store` and `this.#subStores`. You could enumerate the stores by full path, instead of `subChannel` name. In this case `this.#store` becomes `this.#subStores(this.getLocation(undefined))`.

And, you could simplify code even further by defining a `SublevelsMap` class and using it instead of current `this.#subStores`:
```
class SublevelsMap {
  #map: Map<string, LevelUp.LevelUp>
  #baseLocation: string
  constructor(baseLocation: string) {
    this.#baseLocation = baseLocation
    this.#map.set(this.#baseLocation, new LevelUp(...))
  }
  
  get(sublevelName?: string) {
    return this.#map.get(this.fullLocation(sublevelName))
  }
  
  private fullLocation(sublevelName?: string) {
    if (sublevelName) {
      return `${this.#baseLocation}/${sublevelName}`
    } else {
      return this.#baseLocation
    }
  }
}
```

packages/cli/src/s3-store-wrapper.ts
```diff
@@ -0,0 +1,101 @@
...
+import S3LevelDOWN from 's3leveldown'
+import toArray from 'stream-to-array'
+import PQueue from 'p-queue'
+
+/**
+ * Maximum GET/HEAD requests per second to AWS S3
+ */
+const MAX_LOAD_RPS = 4000
+
+export class S3StoreWrapper implements StoreWrapperInterface {
```
ukstv, line 12: Mind taking a shot and calling it `IStoreWrapper`?

packages/cli/src/s3-store-wrapper.ts
```diff
@@ -0,0 +1,101 @@
...
+    this.throwIfNotInitialized()
+    const bufArray = await toArray(
+      this.getStore(params?.subChannel).createKeyStream({
+        limit: params?.limit
+      })
+    )
+    return bufArray.map((buf) => buf.toString())
+  }
+
+  async get(key: string, subChannel?: string): Promise<any> {
```
ukstv, line 77:
```suggestion
  async get<T = unknown>(key: string, subChannel?: string): Promise<T> {
```

packages/cli/src/s3-store-wrapper.ts
```diff
@@ -0,0 +1,101 @@
...
+    this.throwIfNotInitialized()
+    return this.#loadingLimit.add(async () => {
+      return await this.getStore(subChannel).get(key)
+    })
+  }
+
+  async put(key: string, value: any, subChannel?: string): Promise<void> {
+    this.throwIfNotInitialized()
+    return await this.getStore(subChannel).put(key, value)
```
ukstv, line 86:
```suggestion
    return this.getStore(subChannel).put(key, value)
```

packages/cli/src/s3-store-wrapper.ts
```diff
@@ -0,0 +1,101 @@
...
+      this.getStore(params?.subChannel).createKeyStream({
+        limit: params?.limit
+      })
+    )
+    return bufArray.map((buf) => buf.toString())
+  }
+
+  async get(key: string, subChannel?: string): Promise<any> {
+    this.throwIfNotInitialized()
+    return this.#loadingLimit.add(async () => {
```
ukstv, line 79:
```suggestion
    return this.#loadingLimit.add(() => {
        return this.getStore(subChannel).get(key)
     })
```

ukstv: or at least no `return await`, just `return` as it is already an async closure.

packages/cli/src/s3-store-wrapper.ts
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  async put(key: string, value: any, subChannel?: string): Promise<void> {
+    this.throwIfNotInitialized()
+    return await this.getStore(subChannel).put(key, value)
+  }
+
+  async del(key: string, subChannel?: string): Promise<void> {
+    this.throwIfNotInitialized()
+    return await this.getStore(subChannel).del(key)
```
ukstv, line 91:
```suggestion
    return this.getStore(subChannel).del(key)
```

packages/cli/src/daemon-config.ts
```diff
@@ -77,7 +77,7 @@ export class DaemonIpfsConfig {
 }
 
 /**
- * Whether the Ceramic state store should store its data on the local filesystem or in Amazon S3.
+ * Whether the Ceramic storage should store its data on the local filesystem or in Amazon S3.
```
ukstv, line 80: IMO, that should remain as it is, if the CLI mentions the same concept as `state store`. And, IMO, it is still a state store.

packages/core/src/__tests__/ceramic-pinning.test.ts
```diff
@@ -187,221 +187,221 @@ describe('Ceramic stream pinning', () => {
...
+  //     pin: false,
+  //   })
+  //   const pinSpy = jest.spyOn(ipfs1.pin, 'add')
+  //   const unpinSpy = jest.spyOn(ipfs1.pin, 'rm')
+  //   const saveStateSpy = jest.spyOn(ceramic.repository.pinStore.stateStore, 'save')
+  //   const removeStateSpy = jest.spyOn(ceramic.repository.pinStore.stateStore, 'remove')
+  //
+  //   // Pin stream
+  //   await ceramic.pin.add(stream.id)
+  //   expect(pinSpy).toBeCalledTimes(2)
```
ukstv, line 392: No longer relevant or what happened to these tests?

packages/core/src/store/abstract-store.ts
```diff
@@ -0,0 +1,63 @@
+import { StoreWrapperInterface } from './store-wrapper-interface.js'
+import { StoreInterface } from './store-interface.js'
+
+export abstract class AbstractStore<O, V> implements StoreInterface<O, V> {
```
ukstv, line 4:
```suggestion
export abstract class AbstractStore<TKey, TValue> implements StoreInterface<TKey, TValue> {
```

packages/core/src/store/state-store.ts
```diff
@@ -1,11 +1,80 @@
...
+
+  deserialize(serialized: any): StreamState {
+    return StreamUtils.deserializeState(serialized)
+  }
+
+  /**
+   * Pin stream
+   * @param stream - Stream instance
+   */
+  async saveFromStream(stream: Stream): Promise<void> {
```
ukstv, line 44: `Stream` implements `StreamStateHolder`. `saveFromStreamStateHolder` could be removed if here you replace `stream: Stream` with `stream: StreamStateHolder`.

packages/core/src/store/level-db-store-wrapper.ts
```diff
@@ -0,0 +1,96 @@
...
+// a properly typed constructor of `Level` (thus `LevelC`) exported from level-ts package.
+//
+// See also https://github.com/nodejs/node/blob/master/doc/api/esm.md#commonjs-namespaces,
+const LevelC = (levelTs as any).default as unknown as typeof Level
+
+export class LevelDbStoreWrapper implements StoreWrapperInterface {
+  readonly networkName: string
+  readonly #storeRoot: string
+  store: Level
+  #subStores: Record<string, Level> = {}
```
ukstv, line 25: See comment about `SublevelMap` in S3Wrapper.

packages/core/src/store/level-db-store-wrapper.ts
```diff
@@ -0,0 +1,96 @@
...
+// Types provided by level-ts though make it appear as ESM default.
+//
+// So, here we untangle this mess, even if ugly.
+// We import type information separately from code information, and then make sure we can access
+// a properly typed constructor of `Level` (thus `LevelC`) exported from level-ts package.
+//
+// See also https://github.com/nodejs/node/blob/master/doc/api/esm.md#commonjs-namespaces,
+const LevelC = (levelTs as any).default as unknown as typeof Level
+
+export class LevelDbStoreWrapper implements StoreWrapperInterface {
```
ukstv, line 21: Also, names, maybe this: ?
- `StoreWrapperInterface` -> `IKVStore`,
- `StoreInterface` -> gone, because it is only used in `AbstractStore`,
- `AbstractStore` -> `ObjectStore`.

I also have a sort of phantom pain about using abstract classes in TS/JS. I recall a weird case, when a method implemented in a child class is not called from the parent class constructor. IMO, it might be slightly less error prone to pass `getKey`, `serialize`, and `deserialize` as parameters to the `AbstractStore` constructor. Then in a child class you'll have something like this:
```
constructor() {
super({
 getKey: (object: CommitID) => object.toString(),
 serialize: (value: AnchorRequestData) => JSON.stringify(value),
 deserialise: (serialized: any) => JSON.parse(serialized)
})
}
```

ukstv: Actually, as that would add some garbage to memory, IMO, the following more verbose version is slightly more appropriate, using `getKey` as an example. Imagine doing that to `serialise` and `deserialise`:

```ts
function getKey(object: CommitID) {
  return object.toString()
}

class AnchorRequestStore extends ObjectStore<CommitID, AnchorRequestData>  {
   constructor() {
    super(getKey, serialize, deserialise)
  }
}
```

ukstv: Side benefit: `AbstractStore` / `ObjectStore` is no longer abstract, and becomes properly testable in isolation.

packages/core/src/store/level-db-store-wrapper.ts
ukstv: ... and about `return await`.

ukstv (changes requested):
Changes I would like to discuss and/or see implemented:
1. I am not sure about the wording: `store`/`storage` - it should be uniform IMO.
2. Names of classes and interfaces - we could expect some wider discussion about `ISomething` name format, though.
3. Please, do not use abstract classes in TS/JS. There are some dangers there.
4. Code for handling sublevels/subchannels.
5. The commented out test - could you please clarify, what it means?

packages/cli/src/s3-store-wrapper.ts
ukstv: Also, is this thing called a sublevel or a subchannel in LevelDB parlance?  I use sublevel here, based on our previous discussion, but I really do not know, if subchannel is a more appropriate name for this.

packages/cli/src/s3-store-wrapper.ts
ukstv: Or, rather, see [comment ](https://github.com/ceramicnetwork/js-ceramic/pull/2557#discussion_r1024083401)

packages/cli/src/s3-store-wrapper.ts
pawartur: @ukstv , do you think it makes sense to have two slightly different `Map` classes for S3 store and for LevelDB store?

packages/core/src/store/abstract-store.ts
pawartur: @ukstv , `O` here is not the type of the key (this is generally `string`), but rather the type of object from which a key is derived (`StreamID` for `StateStore`, `CommitID` for `AnchorRequestStore`), and then classes which extend this class need to implement methods for deriving a key from the object.

Perhaps this behaviour is misleading (as the object of type `O` is not stored anywhere, it's just used to derive the key)? Should I remove this generalisation altogether and just have a generic value type (assuming that the key is `string`)? Or should I rename `O` to `TObject`?

packages/core/src/store/state-store.ts
pawartur: I added two different methods for saving from stream and from stream state holder to minimise the risk of regression (as I replaced the `LevelStateStore` and `S3StateStore` with a single class and one of the old classes had `save(stream: Stream)` and the other had `save(stateHolder: StreamStateHolder)`).

Do you feel confident that if I unify this, nothing will break :) ?

packages/cli/src/s3-store-wrapper.ts
pawartur: > @ukstv , do you think it makes sense to have two slightly different Map classes for S3 store and for LevelDB store?

I see from your comment below that you think it does make sense. Cool, I'm gonna implement these two map classes then :)

packages/core/src/store/abstract-store.ts
ukstv: Would vote for `TObject` or even `TKeyObject` (weak opinion, feel free to disregard the suggestion). To me semantically, it is still an advanced key :) But, again, that's a weak opinion, feel free to disregard.

packages/core/src/store/state-store.ts
ukstv: `save(stateHolder: StreamStateHolder)` is a more abstract version of `save(stream: Stream)`, so my expectation is that it should not break anything.

packages/cli/src/s3-store.ts
```diff
@@ -0,0 +1,104 @@
...
+    const result = await this.find({
+      limit: 1,
+      ...params,
+    })
+    return result.length > 0
+  }
+
+  async find(params?: StoreSearchParams): Promise<Array<any>> {
+    const bufArray = await toArray(
+      this.#storeMap.get(params?.subChannel).createKeyStream({
```
stbrody, line 78: when getting the base store (where pinned streams are stored), won't this now also try to return the names of the sub-directories for the other stores?  Do we need to do anything to filter those out?

packages/core/src/store/__tests__/level-state-store.test.ts
```diff
@@ -43,21 +44,24 @@ const streamFromState = function (state: StreamState) {
   return new FakeType(TestUtils.runningState(state), {})
 }
 
-describe('LevelDB state store', () => {
+describe('LevelDB-backed StateStore state store', () => {
```
stbrody, line 47:
```suggestion
describe('LevelDB-backed StateStore', () => {
```

packages/core/src/store/__tests__/level-store.test.ts
```diff
@@ -0,0 +1,40 @@
...
+    // add a small delay after creating the leveldb instance before trying to use it.
+    await TestUtils.delay(100)
+  })
+
+  afterEach(async () => {
+    await levelStore.close()
+    await tmpFolder.cleanup()
+  })
+
+  test("The main store is separated from a sub-channel", async () => {
```
stbrody, line 23: I think this file should also test listing all keys in the stores and making sure that the list of keys is separate across the sub-channels as well.  That's where I'm most expecting a bug to show up (especially with the S3 store): `pin.ls()` just iterates over all the keys in the store

packages/core/src/store/anchor-request-store.ts
```diff
@@ -0,0 +1,25 @@
...
+
+function serialize(value: AnchorRequestData): any {
+  return JSON.stringify(value)
+}
+
+function deserialize(serialized: any): AnchorRequestData {
+  return JSON.parse(serialized)
+}
+
+export class AnchorRequestStore extends ObjectStore<CommitID, AnchorRequestData> {
```
stbrody, line 20: can we add a comment what this does and what it's used for?

stbrody (commented):
mostly looks good, my big concern remains that the sub-stores are going to interfere with `pin.ls()` in the S3 version.  We absolutely need to make sure we test this in the e2e tests

pawartur:
@stbrody , I've changed the PR so that LevelDB uses a different instance, located in a different subdirectory in `.ceramic/statestore/` and that S3 uses a different path in `ceramic/<network-name>/`. See the PR description for details.

stbrody (commented):
I can't seem to leave comments on an empty file, but can we remove the empty `composite.note.encoded.json` file that got added to git?

packages/cli/src/s3-store.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  constructor(bucketName: string, networkName: string) {
+    this.networkName = networkName
+    this.#storeRoot = bucketName + '/ceramic/' + this.networkName
+    this.#map = new Map<string, LevelUp.LevelUp>()
+  }
+
+  createStore(useCaseName = DEFAULT_S3_STORE_USE_CASE_NAME) {
+    // Different S3 stores live at different urls (named based use-cases with the default being <bucketName + '/ceramic/' + this.networkName + '/state-store'>
+    // and others being `<bucketName + '/ceramic/' + this.networkName + '/state-store-<useCaseName>` with useCaseNames passed as params by owners of the store map) in #storeRoot
```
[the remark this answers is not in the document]
pawartur: Thanks for the review, @stbrody !

So I tested this again, but took screenshots and printed logs this time.

I had one item in `state-store-anchor-requests`:
Screenshot 2022-11-22 at 08 52 57

And five items in `state-store`:
Screenshot 2022-11-22 at 08 53 04

I ran `pinStore.ls()` and got this:
```
ARTUR WDOWIARSKI LISTING PINS FROM S3:  [
  'kjzl6hvfrbw6c6pr3ygkc6f0rz4grrq5dr7nrm9uaop5lsq0wzff6zs8m9h2ts3',
  'kjzl6kcym7w8y6jb61412tmjaew6cso9wjgwgwxj9yv4rlbqdluim6vrjp5kak6',
  'kjzl6kcym7w8y8wgo5nd0cgrtrv3jzqk4h5z4jpmq8ywwvvhkedi7sg76vk7ub0',
  'kjzl6kcym7w8yagng2yrzan53kmylph9e0njbjw1586hxwg7w8t9i7y4aemlnyw',
  'kjzl6kcym7w8yavmu5athebu6lrk90shb42rn9k17nbbsh8qapaaoa1jdw6vnft'
]
```

Then I deleted from `state-store` the item for Stream Id `kjzl6kcym7w8yagng2yrzan53kmylph9e0njbjw1586hxwg7w8t9i7y4aemlnyw` (the same that I had entry for in `state-store-anchor-requests`):
Screenshot 2022-11-22 at 08 53 15

Then I ran `pinStore.ls()` again and got this:
```
ARTUR WDOWIARSKI LISTING PINS FROM S3:  [
  'kjzl6hvfrbw6c6pr3ygkc6f0rz4grrq5dr7nrm9uaop5lsq0wzff6zs8m9h2ts3',
  'kjzl6kcym7w8y6jb61412tmjaew6cso9wjgwgwxj9yv4rlbqdluim6vrjp5kak6',
  'kjzl6kcym7w8y8wgo5nd0cgrtrv3jzqk4h5z4jpmq8ywwvvhkedi7sg76vk7ub0',
  'kjzl6kcym7w8yavmu5athebu6lrk90shb42rn9k17nbbsh8qapaaoa1jdw6vnft'
]
```

So `kjzl6kcym7w8yagng2yrzan53kmylph9e0njbjw1586hxwg7w8t9i7y4aemlnyw` was not on the list returned by `pinStore.ls()` after I manually deleted it from the`state-store`, although there was still a value for this stream id in `state-store-anchor-requests`.

So it seems that this setup works as expected.

(I was testing this on branch `feat/storing-anchor-requests-CDB-2009`, which is basically the same as this one, but has the code that actually stored anchor requests in the anchor request store and tests for it)

packages/cli/src/s3-store.ts
```diff
@@ -0,0 +1,111 @@
...
+    const fullLocation = this.getFullLocation(useCaseName)
+    const storePath = `${this.#storeRoot}/${fullLocation}`
+    // @ts-ignore
+    this.#map.set(fullLocation, new LevelUp(new S3LevelDOWN(storePath)))
```
ukstv, line 33: I am curious why @ts-ignore here?

packages/cli/src/s3-store.ts
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  private getDefaultLocation(): string {
```
ukstv, line 36: Why not just an instance variable?

ukstv (approved):

packages/cli/src/s3-store.ts
pawartur: Addressed.

stbrody (approved):
sorry, I read too fast yesterday and misunderstood what this is doing.  It actually looks good.  LGTM mod the handful of small cleanups suggested previously

stbrody:
> mod the handful of small cleanups suggested previously
Which it also looks like you addressed yesterday - thanks!

PR #2559, opened 2022-11-15
feat(core): Cache recently processed pubsub messages

stbrody:


packages/core/src/state-management/state-manager.ts
```diff
@@ -217,20 +234,22 @@ export class StateManager {
    * @param model - Model Stream ID
    */
   async handlePubsubUpdate(streamId: StreamID, tip: CID, model?: StreamID): Promise<void> {
-    let state$ = await this.fromMemoryOrStore(streamId)
+    if (this.pubsubCache.get(tip.toString()) === streamId.toString()) {
+      // This tip was already processed for this streamid recently, no need to re-process it.
+      return
+    } else {
```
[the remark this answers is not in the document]
stbrody: yep, all fair suggestions.  Will do

packages/core/src/state-management/state-manager.ts
```diff
@@ -217,20 +218,14 @@ export class StateManager {
...
     }
 
-    if (!state$) {
-      state$ = await this.load(streamId)
-    }
-    this.executionQ.forStream(streamId).add(async () => {
-      await this._handleTip(state$, tip)
-    })
-    await this.indexStreamIfNeeded(state$)
+    await this.applyCommit(streamId, tip, {})
```
stbrody, line 228: tests are currently failing consistently (`ceramic-anchor.test.ts - test the same doc anchored twice (different Ceramic instances), first one wins)`).  Reverting the changes to this one function makes it pass again consistently.  

The only difference between what this function was doing before and what applyCommit does is that when the state is already in the cache or state store, we get it without going through the loading queue, whereas when we use applyCommit we always go through the loading queue.  That should not affect correctness however.  The fact that that small change is causing the test to break suggests to me that there is a bug in the queue logic somewhere.  I wonder if this is related to the queue filling up errors we've been seeing.

stbrody: lol I'm dumb.  stateManager.applyCommit takes a commit body, not a CID, but I was passing it the commit CID.  Has nothing to do with the queues

packages/stream-tests/src/__tests__/cross-node-indexing.test.ts
```diff
@@ -59,170 +59,173 @@ const envs: Array<CrossNodeIndexingTestEnv> = [
...
-  let model: Model
-  let midMetadata: ModelInstanceDocumentMetadataArgs
-  let dbConnection: Knex
-
-  async function dropKnexTables() {
-    await dbConnection.schema.dropTableIfExists(INDEXED_MODEL_CONFIG_TABLE_NAME)
-    await dbConnection.schema.dropTableIfExists(Model.MODEL.toString())
-    await dbConnection.schema.dropTableIfExists(model.id.toString())
-  }
+describe.each(envs)(
```
stbrody, line 62: shouldn't be any functional changes here, just formatting

ukstv (approved):

PR #2560, opened 2022-11-15
Cleanups to code reading interview snippet to aid readability

stbrody:


ukstv:
I am definitely out of the loop here. Why trivialise the task?

stbrody:
> I am definitely out of the loop here. Why trivialise the task?

In discussions about the hiring process for this role over the last two weeks, there's been a growing sentiment that this code segment requires a lot of prior context about how Ceramic works in order to understand.  This means that we wind up spending a lot of time in the interview just clarifying how things work and establishing context before we are able to even get into the meat of the task - understanding how the decision of which log to choose is made, and identifying and fixing the bug.  The goal of this PR is to make the code before the "highlighted block" (which starts on line 430 in the modified version, starting with the line `if (isLocalAnchored && isRemoteAnchored)`) easier to understand so that we can move more quickly into talking about the highlighted block, where the core logic of deciding which log to keep lives

ukstv:
Understood. Let's see what other people think about it. I am on the fence about adding lots of comments here.

willex (approved):
As just discussed - I'm good trying it for both L3 & L4 for the next set of candidates to get a sense of the improvement. It does make it more accessible to L3 if we need to adjust in a meeting and we can counter by raising the bar for an L4 with the expectation of seeing the "special sauce" e.g. if they arrive at the solution quicker we follow up with:

- now that you’ve read the code - whats your first thought?
- how could we make the tie breaker less arbitrary - how can you improve it?
etc.

We can always course correct as well and revert if need be if we see this dilutes the exercise too much.

PR #2561, opened 2022-11-15
feat(pkg): allow "both" release option

smrz2001:
We need both `publish_main` and `publish_rc` to run (and in that order) for the "both" release option ("publish release then publish a new RC").

With job-level checks, either one job will run or the other but not both, so we allow both jobs to always run but skip the steps from one or the other if "main" or "rc" is the selected release option.

"hotfix" is always run on its own, and doesn't need to be modified here.

cc @federicorinaldi @3benbox for visibility.


stbrody (commented):
I'm a little concerned having the condition checked in each and every step, and with how easy it would be to add a step and forget to add the condition check.  But I'm not familiar enough with github workflows to suggest an alternative approach.  I'll defer to Val on if she thinks this is the best way.  At the least this approach seems functionally correct as far as I can tell

smrz2001:
> I'm a little concerned having the condition checked in each and every step, and with how easy it would be to add a step and forget to add the condition check.  

Yeah, that's why I'd kept the flows separated with a single check for each job.

One cleaner way would be to have distinct jobs as before (no falling through from "main" to "rc"), and just have another one for "both" instead of reusing the jobs like in this PR.

So 4 completely stand-alone jobs, "rc", "main", "both", "hotfix", with one check per job.

.github/workflows/reusable-release.yml
```diff
@@ -0,0 +1,248 @@
...
+      gh_token:
+        required: true
+      npm_token:
+        required: true
+
+env:
+  # Dagger
+  DAGGER_PLAN: cue.mod/pkg/github.com/3box/pipeline-tools/ci/plans/ceramic.cue
+  DAGGER_LOG_FORMAT: "plain"
+  PIPELINE_TOOLS_VER: "0.2.0"
```
v-stickykeys, line 22: Just wondering, is there a way to always pull the latest release in pipeline-tools instead? or is that bad practice I guess?

v-stickykeys (approved):
LGTM

.github/workflows/reusable-release.yml
smrz2001: Yeah, we could, I just get worried we'd pull in some breaking change and all the pipelines will break across Ceramic/CAS/IPFS, etc.

You know what, it's probably ok. The change I'm making right now in pipeline-tools will let us test workflows in feature branches so we don't have to worry about breaking other repos by pushing untested changes to develop.

PR #2562, opened 2022-11-16
update the dependency in core from metrics to observability

gvelez17:
## Description

The plan is to remove the metrics package entirely from js-ceramic repo, and replace with a dependency on the independent package @ceramicnetwork/observabillity

However, since the metrics package is used both in core and cli, and core is also used in cli, we need to do this one step at a time as there are dependency conflicts.  

## How Has This Been Tested?

- [x ] running tests

## PR checklist

## References:


smrz2001 (approved):
🚀

PR #2566, opened 2022-11-19
feat/storing anchor requests CDB-2009

pawartur:
## Description

* updated `AnchorRequestStore` to properly save, load, delete and list anchor requests
* made `StateManager` save anchor requests once they're made and remove them from store after an `ANCHORED` or `FAILED` response
* In the process, refactored StateManager.anchor(...), so that it doesn't return an `rxjs` `Subscription` anymore, a we noticed with @ukstv that actually in some places we were awaiting these subscriptions (which effectively did nothing) instead of adding handlers to them, which caused some unexpected behaviours in tests. Now `StateManager.anchor(...)` returns `Promise<void>` and the promise is fulfilled once an anchor has been requested.
* added tests for `AnchorRequestStore`
* added tests to  `state-manager.test.ts` checking that anchor requests are stored when an anchor is requested and deleted when the status changes to `FAILED` or `ANCHORED`, but not when it changes to `PROCESSING`.

## How Has This Been Tested?

- [X] All tests pass
- [X] Manually tested with S3 storage mode

## PR checklist


packages/cli/src/s3-store.ts
```diff
@@ -86,7 +86,16 @@ export class S3Store implements IKVStore {
     return result.length > 0
   }
 
-  async findKeys(params?: StoreSearchParams): Promise<Array<any>> {
+  async find(params?: StoreSearchParams): Promise<Array<IKVStoreFindResult>> {
+    const dataArray = await toArray(
```
ukstv, line 90: Now as we map over the array, maybe we should do `toArray` and `map` in one sweep? `toArray` here is nothing more than

```
let results = []
for await (const entry of stream) {
results.push(entry)
}
return results
```

packages/core/src/state-management/state-manager.ts
```diff
@@ -323,7 +324,7 @@ export class StateManager {
   /**
    * Request anchor for the latest stream state
    */
-  anchor(state$: RunningState): Subscription {
+  async anchor(state$: RunningState): Promise<Subscription> {
```
ukstv, line 327: The request to store the anchor should be part of the rxjs pipeline, IMO, retaining sync method signature.

packages/core/src/store/level-db-store.ts
```diff
@@ -96,16 +96,26 @@ export class LevelDbStore implements IKVStore {
...
+
+  async findKeys(params?: StoreSearchParams): Promise<Array<string>> {
+    const searchParams: Record<string, any> = {
       keys: true,
       values: false,
       limit: params?.limit,
     }
 
     // The return type of .stream is Array<{ key: , value: }>, but if values: false is used in params, then it actually returns Array<string>
-    return (await this.#storeMap.get(params?.useCaseName)).stream(seachParams) as unknown as Promise<
```
ukstv, line 107: Variable for `const a = await whatever` instead of `(await whatever).doSomething`, please?

ukstv (changes requested):

packages/core/src/store/level-db-store.ts
pawartur: Addressed

packages/core/src/__tests__/state-manager.test.ts
```diff
@@ -17,12 +17,32 @@ import { TileDocument } from '@ceramicnetwork/stream-tile'
...
+import { concatMap, map, tap } from 'rxjs/operators'
 import { MAX_RESPONSE_INTERVAL } from '../pubsub/message-bus.js'
 import cloneDeep from 'lodash.clonedeep'
 import { StateLink } from '../state-management/state-link.js'
+import { InMemoryAnchorService } from '../anchor/memory/in-memory-anchor-service.js'
+
+const waitForAnchorStatus = async (stream$: RunningState, status: AnchorStatus): Promise<void> => {
+  await new Promise<void>((resolve) => {
+    let resolved = false
+    stream$.subscribe()
```
ukstv, line 31: why is that?

pawartur: Thanks! I missed that left-over. It's removed now.

packages/core/src/__tests__/state-manager.test.ts
```diff
@@ -17,12 +17,30 @@ import { TileDocument } from '@ceramicnetwork/stream-tile'
...
+import { CommitID, StreamID } from '@ceramicnetwork/streamid'
 import { from, timer } from 'rxjs'
-import { concatMap, map } from 'rxjs/operators'
+import { concatMap, map, tap } from 'rxjs/operators'
 import { MAX_RESPONSE_INTERVAL } from '../pubsub/message-bus.js'
 import cloneDeep from 'lodash.clonedeep'
 import { StateLink } from '../state-management/state-link.js'
+import { InMemoryAnchorService } from '../anchor/memory/in-memory-anchor-service.js'
+
+const waitForAnchorStatus = async (stream$: RunningState, status: AnchorStatus): Promise<void> => {
```
stbrody, line 28: You should be able to use `TestUtils.waitForState` for this.  May even be worth adding a `waitForAnchor` to TestUtils that utilizes `waitForState`

packages/core/src/__tests__/state-manager.test.ts
```diff
@@ -71,75 +89,151 @@ describe('anchor', () => {
...
     fakeHandleTip.mockRejectedValueOnce(new Error('Handle tip failed'))
     fakeHandleTip.mockRejectedValueOnce(new Error('Handle tip failed'))
     fakeHandleTip.mockRejectedValueOnce(new Error('Handle tip failed'))
-    fakeHandleTip.mockImplementationOnce(realHandleTip)
+    fakeHandleTip.mockRejectedValueOnce(new Error('Handle tip failed'))
+    fakeHandleTip.mockRejectedValueOnce(new Error('Handle tip failed'))
+    fakeHandleTip.mockRejectedValueOnce(new Error('Handle tip failed'))
+
+    // wait for 3 secs to let the retry mechanism work
+    await TestUtils.delay(3000)
```
stbrody, line 162: Can we use `TestUtils.waitForState` for this to wait until the anchor fails, rather than using sleeping which is brittle?

packages/core/src/__tests__/state-manager.test.ts
```diff
@@ -71,75 +89,151 @@ describe('anchor', () => {
...
     const fakeHandleTip = jest.fn()
+    fakeHandleTip.bind(stateManager)
     ;(stateManager as any)._handleTip = fakeHandleTip
+
+    // Mock a bunch of throws as consecutive calls to fakeHandleTip
     fakeHandleTip.mockRejectedValueOnce(new Error('Handle tip failed'))
     fakeHandleTip.mockRejectedValueOnce(new Error('Handle tip failed'))
     fakeHandleTip.mockRejectedValueOnce(new Error('Handle tip failed'))
-    fakeHandleTip.mockImplementationOnce(realHandleTip)
+    fakeHandleTip.mockRejectedValueOnce(new Error('Handle tip failed'))
```
stbrody, line 157: is there significance to there being 6 mock calls here?  Or is it just supposed to be "more than 3 failures since we expect 3 retries"?  If the latter, we could just use `fakeHandleTip.mockRejectedValue` which will return the given failed value no matter how many times it is called

packages/core/src/__tests__/state-manager.test.ts
```diff
@@ -71,75 +89,151 @@ describe('anchor', () => {
...
+
+  test('anchor request is stored when created and deleted when anchored', async () => {
+    const stream = await TileDocument.create(ceramic, INITIAL_CONTENT, null, { anchor: false })
+    const stream$ = await ceramic.repository.load(stream.id, {})
+
+    // @ts-ignore anchorRequestStore is private
+    const anchorRequestStore = ceramic.repository.stateManager.anchorRequestStore
+
+    expect(await anchorRequestStore.load(stream.id)).toBeNull()
+    await ceramic.repository.stateManager.anchor(stream$)
```
stbrody, line 176: shouldn't this line be after we check for AnchorStatus of PENDING and that the request store has the pending request (ie right before we call `waitForAnchorStatus`)?  Otherwise isn't there a race condition if the anchor happens too quickly?

packages/core/src/__tests__/state-manager.test.ts
```diff
@@ -71,75 +89,151 @@ describe('anchor', () => {
...
+
+    expect(await anchorRequestStore.load(stream.id)).not.toBeNull()
+
+    // fail anchor
+    const anchorService = ceramicWithManualAnchoring.context
+      .anchorService as any as InMemoryAnchorService
+    await anchorService.failPendingAnchors()
+    await stream.sync()
+    expect(stream.state.anchorStatus).toEqual(AnchorStatus.FAILED)
+    await TestUtils.delay(100) // Needs a bit of time delete the request from the store
```
stbrody, line 204: I hate how racy this is.  It might be time for us to add a `TestUtils.assertSoon` method that takes a function and calls it periodically until some timeout, and errors if the function doesn't return true within the timeout.

stbrody: Okay I went ahead and filed CDB-2090.  This can go in with the sleep approach, but let's add a TODO referencing that ticket and I'll clean it up with the helper once I finish it.

stbrody, line 204:
```suggestion
    await TestUtils.delay(5000) // Needs a bit of time to delete the request from the store. TODO(CDB-2090): use less brittle approach to waiting for this condition
```

packages/core/src/anchor/memory/in-memory-anchor-service.ts
```diff
@@ -104,6 +104,18 @@ export class InMemoryAnchorService implements AnchorService, AnchorValidator {
     this.#queue = [] // reset
   }
 
+  /**
+   * Sets all pending anchors for PROCESSINGS. Useful for testing.
```
stbrody, line 108:
```suggestion
   * Sets all pending anchors to PROCESSING status. Useful for testing.
```

packages/core/src/anchor/memory/in-memory-anchor-service.ts
```diff
@@ -194,6 +206,18 @@ export class InMemoryAnchorService implements AnchorService, AnchorValidator {
     })
   }
 
+  _startProcessingCandidate(candidate: Candidate, message?: string): void {
```
stbrody, line 209: this should probably be added to `_process` also to make the full anchor process behave more like the real CAS

packages/core/src/state-management/state-manager.ts
```diff
@@ -324,17 +325,31 @@ export class StateManager {
...
-    const anchorStatus$ = this.anchorService.requestAnchor(state$.id, state$.tip)
-    return this._processAnchorResponse(state$, anchorStatus$)
+
+    return new Promise((resolve) => {
+      this._saveAnchorRequestForState(state$).then(() => {
+        const anchorStatus$ = this.anchorService.requestAnchor(state$.id, state$.tip)
+        let resolved = false
+        const observable$ = anchorStatus$.pipe(
+          tap((value) => {
+            if (!resolved && value.status === AnchorStatus.PENDING) {
```
stbrody, line 344: what is this doing?  Why do we need to wait for PENDING here?

stbrody: Can this whole block just be
```
await this._saveAnchorRequestForState(state$)
const anchorStatus$ = this.anchorService.requestAnchor(state$.id, state$.tip)
return this._processAnchorResponse(state$, anchorStatus$)
```
?

stbrody (commented):
most comments are just on cleanups to the tests, but there's one real question about the changes to the state manager's anchor method that I'd like to understand

packages/core/src/state-management/state-manager.ts
pawartur: > Can this whole block just be
```
await this._saveAnchorRequestForState(state$)
const anchorStatus$ = this.anchorService.requestAnchor(state$.id, state$.tip)
return this._processAnchorResponse(state$, anchorStatus$)
```

We don't want to do this like that, because then the return type of this method would be `Promise<Subscription>` and we don't want to mix promises with `rxjs` types.

If we don't want to wait for PENDING HERE, I think we can just change this code to:
```
await this._saveAnchorRequestForState(state$)
const anchorStatus$ = this.anchorService.requestAnchor(state$.id, state$.tip)
this._processAnchorResponse(state$, anchorStatus$)
```

, but I'm not sure, if it doesn't break some assumptions in tests or in other logic elsewhere in `js-ceramic`, tbh, I'd have to check.

packages/core/src/__tests__/state-manager.test.ts
pawartur: The way it works, as far as I can see (in current `develop`) is that whether `_handleTip(..)` is retried inside of `_handleAnchorCommit(...)` does not depend on the `anchorStatus` of the stream. The retries are made iff any error is thrown inside of this [block](https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/core/src/state-management/state-manager.ts#L274) , so effectively, retries are made when anything throws inside of `_handleTip(...)` or `publishTip(...)`.

Then it comes to the `anchorStatus`, at the time `_handleAnchorCommit(...)` is called the anchor had been already requested and in the case of this test, the commit had been successfully anchored.

So the only think this test tests is that, if  `_handleTip(...)` throws an error, it will be retried up to three times after `anchor(...)` is called.

Does that make sense, @stbrody ?

packages/core/src/__tests__/state-manager.test.ts
pawartur: You mean that
```
 await ceramic.repository.stateManager.anchor(stream$)
 ```
 
 should be after
 ```
 expect(stream$.value.anchorStatus).toEqual(AnchorStatus.PENDING)
 ```
 ?
 
 The status of the anchor won't be `PENDING`, before we call `anchor(...)`.
 
 If I misunderstand you, can you propose code change, please?

File: packages/core/src/state-management/state-manager.ts (hunk 12 above)
Comment (pawartur) [reply packages/core/src/state-management/state-manager.ts:344]: Having said all of the above, here are the available anchor states right now:
```
export enum AnchorStatus {
  NOT_REQUESTED,
  PENDING,
  PROCESSING,
  ANCHORED,
  FAILED,
}
```

What anchor state should the stream have, in your opinion, @stbrody , after `anchor(...)` is called, if it doesn't wait for `PENDING`?

File: packages/core/src/__tests__/state-manager.test.ts (hunk 6 above)
Comment (pawartur) [reply packages/core/src/__tests__/state-manager.test.ts:162]: Clarification: currently in `develop` `_handleAnchorCommit(...)` actually reverts the stream's anchor status to `FAILED` (check [here](https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/core/src/state-management/state-manager.ts#L305), but only if the anchored `tip` equals the current `state$.tip`, which is not the case during this test, because `state$.tip` actually points to the currently processed anchor commit.

File: packages/core/src/__tests__/state-manager.test.ts (hunk 9 above)
Comment (stbrody) [reply packages/core/src/__tests__/state-manager.test.ts:204]: PR here: https://github.com/ceramicnetwork/js-ceramic/pull/2585.  You should be able to use the new `waitForConditionOrTimeout` function added there to remove these calls to `TestUtils.delay`

File: packages/core/src/__tests__/state-manager.test.ts (hunk 6 above)
Comment (stbrody) [reply packages/core/src/__tests__/state-manager.test.ts:162]: nope I'm still confused.  Nothing should even be calling `_handleTip` by this point in the test.  The anchor has already been successfully applied as of line 147.  I don't really understand why the assertion on line 165 isn't failing, given that there shouldn't be any anchor-related code running by the time the mocks are set up as far as I can tell?

File: packages/core/src/state-management/state-manager.ts (hunk 12 above)
Comment (stbrody) [reply packages/core/src/state-management/state-manager.ts:344]: the current function gives no guarantee as to the anchor status after `anchor()` returns, it only guarantees that the anchor process has been kicked off.  

I guess the promise-y way to do what the current code does with a Subscription would be to return a Promise that resolves when the anchor request either succeeds or definitively fails.

Did you give up on trying to make this work in a Subscription-friendly way?  I'd be curious for thoughts from @ukstv on what the right behavior here should be.

File: packages/core/src/__tests__/state-manager.test.ts (hunk 6 above)
Comment (pawartur) [reply packages/core/src/__tests__/state-manager.test.ts]: In my understanding, the assertion in line 165 is not failing, because the current behaviour of the system is such that `_handleAnchorCommit(...)` is called for a given tip **after** its corresponding state had been set to `ANCHORED`.

So if I'm not mistaken, I guess your confusion may be coming from the fact that it's not super clear what it currently means for a tip to be anchored.

Please, have a look at this test as it currently is on [develop](https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/core/src/__tests__/state-manager.test.ts#L105) - the behaviour is the same as in this PR. 

The goal of this PR is not to change the current behaviour, but rather just add a single async step before an anchor request is made.

Can we accept this as is (and maybe add a ticket to change the behaviour of anchor-state-tracking), or do you think I should continue digging into this in this PR, @stbrody :) ?

File: packages/core/src/state-management/state-manager.ts (hunk 12 above)
Comment (pawartur) [reply packages/core/src/state-management/state-manager.ts:344]: Can you explain what you mean by "Subscription-friendly way", @stbrody ?

The goal of this PR is to add a single async step, before an anchor is requested. Currently in `develop` we use a sync `anchor(...)` method which returns an instance of `Subscription`

So, the way I see it, we have two options here:
1. Keep the `anchor(...)` method as is, but add the new async call in before every call to `anchor(...`)
2. Make `anchor(...)` async to do the additional async work inside it, so that we make sure it's always done when `anchor(...)` is called

If we go with option 2 (which is what I decided to do in this PR), we have yet two more options:
a. Maintain the current contract of `anchor(...)` and continue returning an instance of `Subscription`, just make it async. This would lead to a return type of `Promise<Subscription>`, though, which we pointed out we didn't want in this [comment](https://github.com/ceramicnetwork/js-ceramic/pull/2566#discussion_r1029340359)
b. Slightly change the behaviour of the system by not returning an instance of `Subscription` anymore. Instead we change the return type of `anchor(...)` to `Promise<void>`, so it's not possible to add handlers to this subscription anymore (something we seem only to be using in tests right now, e.g. [here](https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/core/src/__tests__/state-manager.test.ts#L92) )

Do these options make sense to you, @stbrody ? Do you see any other options that we missed? What would you propose?

File: packages/core/src/__tests__/state-manager.test.ts (hunk 8 above)
Comment (pawartur) [reply packages/core/src/__tests__/state-manager.test.ts:170]: > ah, so we're relying on the call to anchor() to send the stream through both PENDING and ANCHORED statuses? 

No, the only thing I wrote was that `stream$.value.anchorStatus` will not be `AnchorStatus.PENDING`, before the call to `await ceramic.repository.stateManager.anchor(stream$)`

I think I really don't understand the point you originally made in this comment thread, @stbrody . Can you please create a commit suggestion here with the particular contents of this test that you have in mind :) ?

File: packages/core/src/state-management/state-manager.ts (hunk 12 above)
Comment (pawartur) [reply packages/core/src/state-management/state-manager.ts:344]: @ukstv , what do you think about it :) ?

Comment (ukstv) [reply packages/core/src/state-management/state-manager.ts:344]: @stbrody I tend to align with @pawartur here, and I agree it is a bit tricky.

Original `anchor` was really about kicking off the process. Now we have some part (=store the anchor to the store) that must succeed before we go any further. Thus, here we ensure that we store the anchor, and then kick off the polling process.

`anchor` method is called from two places: `Ceramic#requestAnchor` and `StateManager#applyWriteOps`. Both run as part of incoming HTTP request processing. Storing an anchor resembles pinning and storing the state, so here we continue a tradition of making storage awaitable.

As for this: "... to return a Promise that resolves when the anchor request either succeeds or definitively fails", the code here goes exactly as you wrote. If we have stored the anchor request in the store, we can be sure of its future fate.

**Review by ukstv — APPROVED** [review]
It seems, that we have a one future task here: measure how much time it takes to store an anchor (created in Linear, not assigned).

File: packages/core/src/__tests__/state-manager.test.ts (hunk 8 above)
Comment (stbrody) [reply packages/core/src/__tests__/state-manager.test.ts:165]: I think you could fix this by using a Ceramic node with `anchorOnRequest: false` set for the InMemoryAnchorService, like you do in some of the other tests in this file.  That way you're guaranteed that it won't move past PENDING into ANCHORED until you explicitly trigger the anchor against the InMemoryAnchorService

File: packages/core/src/__tests__/state-manager.test.ts [hunk 13]
```diff
@@ -71,75 +72,159 @@ describe('anchor', () => {
...
+
+    // Mock the result of the original implementation as the third call - this one should return
+    // and stop the retrying mechanism
     fakeHandleTip.mockImplementationOnce(realHandleTip)
 
-    await new Promise<void>((resolve) => {
-      ceramic.repository.stateManager.anchor(stream$).add(resolve)
-    })
+    await ceramic.repository.stateManager.anchor(stream$)
+    expect(stream$.value.anchorStatus).toEqual(AnchorStatus.PENDING)
```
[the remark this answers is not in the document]
pawartur: I used the approach with `anchorOnRequest: false` for this test as well, have a look, @stbrody :)

packages/core/src/__tests__/state-manager.test.ts
```diff
@@ -36,26 +36,22 @@ const STRING_MAP_SCHEMA = {
   },
 }
 
-let ipfs: IpfsApi
-let ceramic: Ceramic
-let controllers: string[]
```
stbrody, line 39: the changes to this file are now so big that I've lost the ability to meaningfully review them. In the future, if you do a large refactor of a file like this, please do a separate PR with no behavior change, just the refactor, so that then there can be a follow-up PR with just the behavior change so it's easy to review the meat of what's different

packages/core/src/anchor/memory/in-memory-anchor-service.ts
```diff
@@ -310,27 +309,44 @@ export class InMemoryAnchorService implements AnchorService, AnchorValidator {
     }
   }
 
+  async _storeRecord(record: Record<string, unknown>): Promise<CID> {
+    let timeout: any
+
+    const putPromise = this.#dispatcher._ipfs.dag.put(record).finally(() => {
```
stbrody, line 315: can use `dispatcher.storeCommit`, instead of reaching into its internal ipfs handle

packages/core/src/anchor/memory/in-memory-anchor-service.ts
```diff
@@ -310,27 +309,44 @@ export class InMemoryAnchorService implements AnchorService, AnchorValidator {
...
-    })
-    const commitFound: boolean =
-      null != stream.state.log.find((logEntry) => logEntry.cid.equals(expectedCID))
-    if (!commitFound) {
-      throw new Error(
-        `Anchor commit not found in stream log after being applied to Ceramic node. This most likely means the commit was rejected by Ceramic's conflict resolution. StreamID: ${streamId.toString()}, found tip: ${stream.tip.toString()}`
-      )
+    const anchorCid = await this._storeRecord(commit as any)
+
+    const updateMessage = {
```
stbrody, line 339: you can just use `dispatcher.publishTip` for pretty much the whole rest of this function

packages/stream-tests/src/__tests__/model.test.ts
```diff
@@ -30,7 +30,7 @@ describe('Model API http-client tests', () => {
     daemon = new CeramicDaemon(core, DaemonConfig.fromObject({ 'http-api': { port } }))
     await daemon.listen()
     ceramic = new CeramicClient(apiUrl)
-    ceramic.setDID(core.did)
+    await ceramic.setDID(core.did)
```
stbrody, line 33:
```suggestion
    ceramic.did = core.did
```

stbrody: `setDid` method is deprecated

stbrody (commented):
core changes all looking very good.  A few small suggestions for some code cleanup utilizing existing methods

pawartur:
Ok, @stbrody , I've:
* I included your suggestion to use `anchorOnRequest: false` and adjusted the tests accordingly, thanks for that 🙂
* I refactored the `InMemoryAnchorService` to push the tip directly to ipfs rather than use `ceramic.applyCommit(...)` and adjusted the tests accordingly

Can you have a look, please?

packages/stream-tests/src/__tests__/model.test.ts
```diff
@@ -93,31 +101,35 @@ describe('Model API http-client tests', () => {
...
     expect(model.state.log[0].type).toEqual(CommitType.GENESIS)
     expect(model.state.log[1].type).toEqual(CommitType.ANCHOR)
-    expect(model.content).toEqual(MODEL_DEFINITION)
+    expect(JSON.stringify(model.content)).toEqual(JSON.stringify(modelContent))
   })
 
   test('Models are created deterministically', async () => {
-    const model1 = await Model.create(ceramic, MODEL_DEFINITION)
-    const model2 = await Model.create(ceramic, MODEL_DEFINITION)
+    const modelContent = modelContentWithName('DeterministicModel')
```
stbrody, line 108: Optional suggestion, you could use `expect.getState().currentTestName` to automatically get the test name and use it in the model name, so that you don't need to manually provide a different name for each model.  Could probably even do it within modelContentWithName directly and make it take no arguments.  Then would be nice to add a comment explaining why it matters that each test use a model with a different name.

stbrody: Another option, which I actually just did in a recent PR then undid at the last moment (see https://github.com/ceramicnetwork/js-ceramic/pull/2586#discussion_r1034495433) would be to create a new Ceramic instance for each test so that each test is totally independent of each other test (basically just means moving most of the beforeAll/afterAll into beforeEach/afterEach, for everything except IPFS setup/teardown).

packages/core/src/anchor/memory/in-memory-anchor-service.ts
```diff
@@ -287,34 +310,51 @@ export class InMemoryAnchorService implements AnchorService, AnchorValidator {
...
    */
   async _publishAnchorCommit(streamId: StreamID, commit: AnchorCommit): Promise<CID> {
-    const block = await Block.encode({ value: commit, codec, hasher })
-    const expectedCID = block.cid
-    const stream = await this.#ceramic.applyCommit(streamId, commit, {
-      publish: true,
-      anchor: false,
+    const anchorCid = await this._storeRecord(commit as any)
+    let resolved = false
+    await new Promise<void>((resolve) => {
```
stbrody, line 340: why do you need this wrapping Promise?  Can you not just `await this.#dispatcher.publishTip()`?

pawartur: `publishTip()` returns an instance of `Subscription` and they are not awaitable as far as I understand :)

stbrody: ah okay, do we even need to wait for it at all then?  The pubsub message is processed asynchronously anyway, so I'm not really sure what value we're getting from waiting here?

PR #2571, opened 2022-11-21
feat/Removed the support for models setting in ceramic and ceramic daemon configs

pawartur:
## How Has This Been Tested?

- [X] Ran all the tests

## PR checklist

### DIFF
--- a/packages/core/src/initialization/make-index-api.ts
+++ b/packages/core/src/initialization/make-index-api.ts
@@ -40,10 +40,5 @@ export function makeIndexApi(
     return undefined
   }
   const indexApi = buildIndexing(indexingConfig, logger, network)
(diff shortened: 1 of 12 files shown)


packages/core/src/initialization/make-index-api.ts
```diff
@@ -40,10 +40,5 @@ export function makeIndexApi(
     return undefined
   }
   const indexApi = buildIndexing(indexingConfig, logger, network)
-  if (isNonProduction(indexApi, network, indexingConfig.models)) {
```
stbrody, line 43: why'd we lose this warning?  We should just adapt it to not depend on models being set in the config file, but the intention behind this warning is still valid.  Can we put this back in a follow-up?

PR #2573, opened 2022-11-22
Feat/resend stored anchor requests on node startup CDB-2010

pawartur:
## Description

* Implemented the new `AnchorResumingService` class  with a single async `resumeRunningStatesFromAnchorRequestStore(...)` method which:
  * fetches stored anchor requests in batches
  * goes through each stored anchor request from each batch and adds a job to a concurrency queue, which
    * loads the stream related to the stored anchor request (which make sure that there's a running state for it)
    * re-requests the anchor for the same tip that was stored in the stored anchor
* started calling the new `resumeRunningStatesFromAnchorRequestStore(...)` method in a non-blocking way in `Ceramic._init(...)` method
* made sure that the concurrency queue owned by the new method is cleared when ceramic core is closed
* added tests that check that:
  * the new method is called by `Ceramic._init(...)` and is non-blocking
  * the new method re-requests anchors for stored anchor requests
* Started using the anchor request store to determine, if repository should start polling for anchor status updates when loading a running stream from the stream state store

## How Has This Been Tested?

- [X] Ran the automatic tests
- [X] Tested manually against S3


pawartur:
TODOs:
1. Don't make the resume method blocking on daemon startup
2. Handle kill signal gracefully (see https://github.com/ceramicnetwork/js-ceramic/blob/57a25fa37ccdac3d378de847b7ce2f83200f390b/packages/cli/src/daemon/handle-sigint-signal.ts)

packages/core/src/state-management/repository.ts
```diff
@@ -265,6 +284,36 @@ export class Repository {
...
+          const stream$ = await this.load(listResult.key, { sync: SyncOptions.PREFER_CACHE })
+          if ([AnchorStatus.FAILED, AnchorStatus.ANCHORED].includes(stream$.value.anchorStatus)) {
+            await this.anchorRequestStore.remove(stream$.id)
+            return
+          }
+          // TODO: CDB-2010 Do we want to do the actual anchor request here? Or do we just want to start polling? If the latter, than .load(..) doesn't seem to start polling as of now.
+          await this.stateManager.anchor(stream$)
+          if (listResult.value.timestamp > currentTimestamp) {
+            throw Error(
+              `System clock possibly moving backwards!!! Check your system clock and if it's accurate, delete your state store`
```
ukstv, line 304: Nice!

packages/core/src/state-management/repository.ts
```diff
@@ -265,6 +284,36 @@ export class Repository {
...
+  async resumeRunningStatesFromAnchorRequestStore(): Promise<void> {
+    const currentTimestamp = Date.now()
+    let gt: StreamID | undefined = undefined
+    let batch = new Array<AnchorRequestStoreListResult>()
+    do {
+      batch = await this.anchorRequestStore.list(RESUME_BATCH_SIZE, gt)
+      batch.map((listResult) => {
+        this.resumeQ.add(async () => {
+          const stream$ = await this.load(listResult.key, { sync: SyncOptions.PREFER_CACHE })
+          if ([AnchorStatus.FAILED, AnchorStatus.ANCHORED].includes(stream$.value.anchorStatus)) {
```
ukstv, line 296: I am a bit conflicted about this. Could you tell please how it relates to recovery mechanism in `fromStreamStateStore`?

packages/cli/src/ceramic-daemon.ts
```diff
@@ -295,14 +295,15 @@ export class CeramicDaemon {
...
-      await ceramic.repository.injectStateStore(s3Store)
+      await ceramic.repository.injectKeyValueStore(s3Store)
     }
     const did = new DID({ resolver: makeResolvers(ceramic, ceramicConfig, opts) })
     ceramic.did = did
     await ceramic._init(true)
 
     const daemon = new CeramicDaemon(ceramic, opts)
     await daemon.listen()
+    ceramic.repository.resumeRunningStatesFromAnchorRequestStore()
```
ukstv, line 306: Since it is `async` operation, mind adding an error handler here, please?

Non-actionable: How about handling shutdown gracefully?

ukstv: UPD: Graceful shutdown is here. Have not seen that.

packages/core/src/state-management/repository.ts
```diff
@@ -265,6 +284,36 @@ export class Repository {
     return state$
   }
 
+  async resumeRunningStatesFromAnchorRequestStore(): Promise<void> {
```
ukstv, line 287: Should it really be a part of repository?

ukstv: IMO it would better be a separate service that uses the repository.

ukstv (changes requested):
Would like to request some additional info, not really the changes.

1. Would like to have a short discussion about `resumeRunningStatesFromAnchorRequestStore` being part of repository. Maybe it could be own service. Repository is already too big.
2. Recovery in `fromStreamStateStore` and in `resumeRunningStatesFromAnchorRequestStore` - don't you think there is a duplication of requests to CAS?

packages/cli/src/s3-store.ts
```diff
@@ -85,11 +85,11 @@ export class S3Store implements IKVStore {
...
-    const dataArray = await toArray(
-      store.createReadStream({
-        limit: params?.limit,
-      })
-    )
+    const options = {
+      limit: params?.limit,
+    }
+    if (params.gt) (options as any).gt = params.gt
+    const dataArray = await toArray(store.createReadStream(options))
```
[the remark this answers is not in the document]
pawartur: Ah, no, wait! We don't want to use `Partial` here and pass it to `store.createReadStream` because:
1. We don't want to pass the `useCaseName` to `store.createReadStream`
2. `StoreSearchParams` is a type defined specifically for the params passed to the search methods in ikv stores, so I think it's better to rely on the (non-)optionality of particular fields as defined in the type itself rather than make them all optional with `Partial`

Does that make sense, @ukstv ?

packages/core/src/state-management/repository.ts
pawartur: I'm not sure how it relates to the recovery mechanism from the `fromStreamStateStore`, @ukstv . 

To me, this particular `if` block is just a sanity check, as there should be no anchor requests in storage for streams with anchor status `FAILED` or `ANCHORED` in the first place. I'm happy to remove this check here or replace it with a logger warning.

packages/core/src/state-management/repository.ts
pawartur: Good call, will do!

pawartur:
> Would like to request some additional info, not really the changes.
> 
> 1. Would like to have a short discussion about `resumeRunningStatesFromAnchorRequestStore` being part of repository. Maybe it could be own service. Repository is already too big.
> 2. Recovery in `fromStreamStateStore` and in `resumeRunningStatesFromAnchorRequestStore` - don't you think there is a duplication of requests to CAS?

Agree about 1.

When it comes to 2, we've agreed today with @ukstv that there should be no call to `anchor()` in the resuming method, but unfortunately we can't just remove it now, because we had found a bug that prevents streams to be properly updated in the stream state store after their anchor status is set to pending.

We will fix this bug in this branch and then remove the call to `anchor(...)` in the resuming method.

packages/cli/src/ceramic-daemon.ts
```diff
@@ -295,14 +297,19 @@ export class CeramicDaemon {
...
-      await ceramic.repository.injectStateStore(s3Store)
+      await ceramic.repository.injectKeyValueStore(s3Store)
     }
     const did = new DID({ resolver: makeResolvers(ceramic, ceramicConfig, opts) })
     ceramic.did = did
     await ceramic._init(true)
 
     const daemon = new CeramicDaemon(ceramic, opts)
     await daemon.listen()
+    daemon.anchorResumingService
```
stbrody, line 308: I'm curious, is there a reason that the anchorResumingService is a part of the CeramicDaemon instead of the Ceramic core client?

packages/core/src/state-management/__tests__/anchor-resuming-service.test.ts
```diff
@@ -0,0 +1,163 @@
...
+
+    // Set requestAnchor back to the proper implementation and resume anchors
+    anchorService.requestAnchor = realRequestAnchor
+    const anchorResumingService = new AnchorResumingService(
+      ceramic.loggerProvider.getDiagnosticsLogger()
+    )
+    await anchorResumingService.resumeRunningStatesFromAnchorRequestStore(ceramic.repository)
+
+    // Give the anchor service some time to work, sync the stream and check that the anchor had been requested
+    await TestUtils.delay(3000) // TODO(CDB-2090): use less brittle approach to waiting for this condition
```
stbrody, line 75: I'd love to not have to keep updating that PR for CDB-2090 as new PRs go in - any chance I could get a review on that soon and get it merged in so you could make the change here?

packages/core/src/state-management/anchor-resuming-service.ts
```diff
@@ -0,0 +1,64 @@
...
+   * @private
+   */
+  #shouldBeClosed = false
+
+  constructor(private readonly logger: DiagnosticsLogger) {
+    this.resumeQ = new PQueue({ concurrency: RESUME_QUEUE_CONCURRENCY })
+  }
+
+  async resumeRunningStatesFromAnchorRequestStore(repository: Repository): Promise<void> {
+    this.#shouldBeClosed = false
```
stbrody, line 28: is this line necessary?  It makes me think that shouldBeClosed could sometimes go from false to true back to false again.

packages/core/src/state-management/anchor-resuming-service.ts
```diff
@@ -0,0 +1,64 @@
...
+          const stream$ = await repository.load(listResult.key, {
+            sync: SyncOptions.PREFER_CACHE,
+          })
+          if ([AnchorStatus.FAILED, AnchorStatus.ANCHORED].includes(stream$.value.anchorStatus)) {
+            await repository.anchorRequestStore.remove(stream$.id)
+            return
+          }
+          // TODO: CDB-2010 Do we want to do the actual anchor request here? Or do we just want to start polling? If the latter, than .load(..) doesn't seem to start polling as of now.
+          await repository.stateManager.anchor(stream$)
+          if (listResult.value.timestamp > currentTimestamp) {
```
stbrody, line 45: can you help me understand what this check is doing and why it's necessary?

packages/core/src/state-management/anchor-resuming-service.ts
```diff
@@ -0,0 +1,64 @@
...
+            throw Error(
+              `System clock possibly moving backwards!!! Check your system clock and if it's accurate, delete your state store`
```
stbrody, line 47: We need to be extremely careful here as if users delete their stream state store that could result in major data loss.  Is the intention actually to have them delete the anchor request store?  Even if we want them to just delete the anchor request store, we have to be very careful how we tell users to do this because if they get confused or do it wrong they could delete important data.

packages/core/src/state-management/anchor-resuming-service.ts
```diff
@@ -0,0 +1,64 @@
...
+            )
+          }
+          this.logger.log(
+            LogStyle.info,
```
stbrody, line 51: Suggest using the verbose log level, as this could get quite noisy at startup otherwise if you have a large number of pending anchor requests.

packages/cli/src/ceramic-daemon.ts
pawartur: I did it this way, because every time I've been a part of discussing how this should work, we were saying that the resume should happen when the daemon is restarted.

Happy to move it to Ceramic core, if you think it should happen always when the core is client is started. I assume this should happen in `Ceramic._init(...)`, right?

packages/core/src/state-management/anchor-resuming-service.ts
pawartur: If the only way to get this running again after calling `close()` is to create a new instance, then it's not necessary. I'll remove it.

packages/core/src/state-management/anchor-resuming-service.ts
pawartur: When we were discussing this project during the spec review, you said that it was important for you to check, if the node's time is moving backwards. This is what this check is supposed to verify.

You think I should remove it or change it, @stbrody ?

packages/core/src/state-management/anchor-resuming-service.ts
pawartur: What would be a better message here, @stbrody (assuming that we want to keep this check)?

packages/cli/src/ceramic-daemon.ts
oed: That's right @pawartur. Agree that this makes more sense in core, because if you are using core directly this feature is still desirable.

packages/core/src/anchor/memory/in-memory-anchor-service.ts
```diff
@@ -296,16 +294,10 @@ export class InMemoryAnchorService implements AnchorService, AnchorValidator {
     const feed$ = this.#feed.pipe(
       filter((asr) => asr.streamId.equals(streamId) && asr.cid.equals(tip))
     )
-
     if (anchorResponse) {
       return concat(of<AnchorServiceResponse>(anchorResponse), feed$)
     } else {
-      return of<AnchorServiceResponse>({
```
stbrody, line 303: wow, that previous behavior definitely seems wrong.  Kind of wild it never was an issue before now.  Good catch

packages/core/src/ceramic.ts
```diff
@@ -544,6 +547,13 @@ export class Ceramic implements CeramicApi {
       await this._anchorValidator.init(this._supportedChains ? this._supportedChains[0] : null)
 
       await this._startupChecks()
+
+      // We're not awaiting here for purpose, it's not supposed to be blocking
```
stbrody, line 551:
```suggestion
      // We're not awaiting here on purpose, it's not supposed to be blocking
```

packages/core/src/state-management/anchor-resuming-service.ts
```diff
@@ -25,32 +25,33 @@ export class AnchorResumingService {
...
-            await repository.anchorRequestStore.remove(stream$.id)
-            return
-          }
-          // TODO: CDB-2010 Do we want to do the actual anchor request here? Or do we just want to start polling? If the latter, than .load(..) doesn't seem to start polling as of now.
-          await repository.stateManager.anchor(stream$)
-          if (listResult.value.timestamp > currentTimestamp) {
-            throw Error(
-              `System clock possibly moving backwards!!! Check your system clock and if it's accurate, delete your state store`
+          try {
+            await repository.fromMemoryOrStore(listResult.key)
```
stbrody, line 40: nice catch changing from `repository.load()` to `repository.fromMemoryOrStore()`.  The former might have generated a lot of unnecessary pubsub traffic

packages/core/src/state-management/state-manager.ts
```diff
@@ -191,7 +191,8 @@ export class StateManager {
     const shouldIndex =
       state$.state.metadata.model && this._index.shouldIndexStream(state$.state.metadata.model)
     if (isPinned || shouldIndex) {
-      await this.pinStore.add(state$)
+      // force: true to save the updated stream in the stream state store, even if there hadn't been any new pinned commits
```
stbrody, line 194: this will cause a lot of extra work to be done to re-pin every IPFS record in the entire stream log.  Is this to work around the issue where we don't update the state in the state store if the only thing that changed is the anchorStatus?  If so I think we need a more targeted fix for that.  Does that need to be fixed in this PR or could that be a follow-up?  Is it something we even need to worry about fixing right now?

packages/core/src/state-management/anchor-resuming-service.ts
stbrody: ah I see.  I guess I had imagined that being tracked and handled separately.  Tying the system clock check in with the data in the state store worries me for the reasons I mentioned in my other comment.  Can we have a separate ticket for the check around the time moving backward and do some brainstorming on the best and safest approach to address that?  That way we don't hold up this PR on trying to figure that all out

packages/core/src/indexing/__tests__/build-indexing.test.ts
```diff
@@ -56,7 +56,7 @@ test('build for postgres connection string', async () => {
   )
   expect(indexingApi).toBeInstanceOf(PostgresIndexApi)
   await kill()
-})
+}, 10000) // TODO: CDB-XXXX creating the indexing db sometimes takes a lot of time
```
stbrody, line 59: probably should either just remove this TODO or update it with an actual ticket number before merging.

stbrody (approved):
LGTM

ukstv (approved):

PR #2574, opened 2022-11-22
feat(core): Throw when loading or updating a stream with expired CACAOs in the log

stbrody:
This has the nice bonus of removing the need for the "lone genesis" check since we now check for expired CACAOs every time we load or update a stream anyway, which lets us simplify the code a bit

### DIFF
--- a/packages/common/src/utils/stream-utils.ts
+++ b/packages/common/src/utils/stream-utils.ts
@@ -310,6 +310,26 @@ export class StreamUtils {
     if (commitData?.capability?.p?.exp) {
       logEntry.expirationTime = Math.floor(Date.parse(commitData.capability.p.exp) / 1000)
     }
+    if (commitData.timestamp) {
+      logEntry.timestamp = commitData.timestamp
+    }
     return logEntry
+  }
+
+  /**
+   * Takes a StreamState and validates that none of the commits in its log are based on expired CACAOs.
+   */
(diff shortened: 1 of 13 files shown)


packages/common/src/utils/stream-utils.ts
```diff
@@ -310,6 +310,26 @@ export class StreamUtils {
...
+    for (const logEntry of state.log) {
+      const timestamp = logEntry.timestamp ?? now
+      if (logEntry.expirationTime && logEntry.expirationTime < timestamp) {
+        throw new Error(
+          `CACAO expired: Commit ${logEntry.cid.toString()} of Stream ${StreamUtils.streamIdFromState(
+            state
+          ).toString()} has a CACAO that expired at ${logEntry.expirationTime}`
+        )
+      }
+    }
```
oed, line 333: Why not break after the first anchor is detected?

oed (approved):
lgtm, but would like a ✅ from @ukstv as well!

packages/common/src/utils/stream-utils.ts
stbrody: That will only work for now since the CAS is loading the streams and validating the commits before anchoring.  Once CAS w/o Ceramic node is complete, a commit that has already expired could still be anchored

oed: ok

ukstv (approved):

PR #2585, opened 2022-11-28
chore: Add test function to wait for condition and use it to remove sleeps

stbrody:


stbrody:
okay @pawartur, I merged in develop to include your recent changes to the state-manager test, and updated this PR to use the new function to clean up how the conditions are waited on.  Seems to have cut the average test execution time for `state-manager.test.ts` in about half (from 40 to 20 seconds) on my machine!

PTAL

pawartur (approved):

packages/common/src/utils/test-utils.ts
```diff
@@ -25,6 +25,42 @@ class FakeRunningState extends BehaviorSubject<StreamState> implements RunningSt
 }
 
 export class TestUtils {
+  /**
+   * Wait up to 'timeoutMs' for the given predicate to return true.  Polls the given predicate once
+   * every 100ms.  Returns true if the predicate eventually returned true, or false if it timed out
```
stbrody, line 30:
```suggestion
   * every 100ms (plus however long it takes for the predicate itself to execute).  
   * Returns true if the predicate eventually returned true, or false if it timed out
```

PR #2586, opened 2022-11-28
feat(stream-model): Validate relations in Model handler

stbrody:


packages/stream-model/src/model.ts
```diff
@@ -223,6 +224,35 @@ export class Model extends Stream {
...
+   */
+  static assertRelationsValid(content: ModelDefinition) {
+    if (!content.relations) {
+      return
+    }
+
+    for (const [fieldName, relationDefinition] of Object.entries(content.relations)) {
+      switch (relationDefinition.type) {
+        case 'account':
+          return
```
PaulLeCam, line 239:
```suggestion
          continue
```

packages/stream-tests/src/__tests__/model.test.ts
```diff
@@ -20,9 +33,16 @@ describe('Model API http-client tests', () => {
...
 
   beforeAll(async () => {
+    ipfs = await createIPFS()
+  })
+
+  afterAll(async () => {
+    await ipfs.stop()
+  })
+
+  beforeEach(async () => {
```
PaulLeCam, line 43: Does it really need to create a new Ceramic instance for each test, can't the test suite reuse the same instance?

PaulLeCam (changes requested):

packages/stream-model/src/model.ts
stbrody: oops good catch

packages/stream-tests/src/__tests__/model.test.ts
stbrody: uh, I guess it can.  I changed it b/c Model creation is deterministic, so some of the tests are creating the same model and not actually really creating a new model, but just re-using the model created in an earlier test.  But I guess that doesn't actually really break any of the assumptions or invalidate anything that the tests are trying to test, at least not with the current set of tests, so I can put it back for now

PaulLeCam (approved):
Thanks for the changes!

PR #2588, opened 2022-11-30
chore(core): consolidate postgres and sqlite logic

PaulLeCam:
Mostly moving logic around, main changes are:
- Replacing the `init-tables` modules by `tables-manager` class with DB-specific extensions
- Changed the `DatabaseIndexApi` interface to a class with shared logic implementation and DB-specific extensions

Still need to consolidate the common index API tests into a single module, but should be good to start reviewing already.


stbrody:
This would be great to get in.  @willex lmk if you don't think you're going to be able to review this and want someone else to take it over

packages/core/src/indexing/migrations/1-create-model-table.ts
```diff
@@ -21,11 +26,28 @@ export type ColumnInfo = {
   type: ColumnType
 }
 
-export async function createModelTable(
+function createExtraColumns(
+  table: Knex.CreateTableBuilder,
+  tableName: string,
+  extraColumns: Array<ColumnInfo>
+): void {
+  for (const column of extraColumns) {
```
willex, line 34:
```suggestion
  // create unique index name <64 chars that are still capable of being referenced to MID table
  const indexName = tableName.substring(tableName.length - 10)
  for (const column of extraColumns) {
```

packages/core/src/indexing/migrations/1-create-model-table.ts
```diff
@@ -21,11 +26,28 @@ export type ColumnInfo = {
...
+    switch (column.type) {
+      case ColumnType.STRING:
+        table.string(column.name, 1024).notNullable()
+        table.index([column.name], `idx_${tableName}_${column.name}`)
```
willex, line 38:
```suggestion
        table.index([column.name], `idx_${indexName}_${column.name}`)
```

packages/core/src/indexing/migrations/1-create-model-table.ts
```diff
@@ -39,16 +61,7 @@ export async function createModelTable(
     table.dateTime('created_at').notNullable().defaultTo(dataSource.fn.now())
     table.dateTime('updated_at').notNullable().defaultTo(dataSource.fn.now())
 
-    for (const column of extraColumns) {
-      switch (column.type) {
-        case ColumnType.STRING:
-          table.string(column.name, 1024).notNullable()
-          table.index([column.name], `idx_${tableName}_${column.name}`)
```
willex, line 46: same here as above - `indexName` vs. `tableName`

packages/core/src/indexing/migrations/1-create-model-table.ts
```diff
@@ -75,6 +88,35 @@ export async function createModelTable(
...
+    table.string('stream_content').notNullable()
+    table.string('tip').notNullable()
+    table.integer('last_anchored_at').nullable()
+    table.integer('first_anchored_at').nullable()
+    table.integer('created_at').notNullable()
+    table.integer('updated_at').notNullable()
+
+    createExtraColumns(table, tableName, extraColumns)
+
+    table.index(['last_anchored_at'], `idx_${tableName}_last_anchored_at`)
```
willex, line 108: `indexName` vs. `tableName`

packages/core/src/indexing/migrations/1-create-model-table.ts
```diff
@@ -100,6 +142,5 @@ export async function createConfigTable(dataSource: Knex, tableName: string, net
       break
     default:
       throw new Error(`Invalid config table creation requested: ${tableName}`)
-      process.exit(-1)
```
willex, line 103: Why not exit here; seems like this would be a good exit point during node startup?

willex (changes requested):
Looks good overall - added some comments for minor changes.

packages/core/src/indexing/migrations/1-create-model-table.ts
PaulLeCam: It's unreachable code, nothing gets executed after a `throw`

willex (approved):
Awesome, thanks!

PR #2589, opened 2022-12-01
Fix interference between tests

stbrody:
Seems that the reason that the tests interfered with each other had to do with the fact that the InMemoryAnchorService was getting multiple anchor requests for the same commit (the genesis commit for the Model).  When it picked which request to anchor, it would randomly choose one of the two identical requests to anchor, but mark the other as failed due to conflict resolution, which would send an event on the Observable polling the anchor service saying the request was failed.

I still don't totally understand then why the test would *hang* rather than erroring that the anchoring had failed, but I think at this point that's enough time spent digging on this.


pawartur (approved):
Thanks :) !!

PR #2595, opened 2022-12-05
code reading interview: rm comments from highlighted block

stbrody:
In response to Sergey's recent email proposing that we've made it *too* easy with all the recent comments.  I feel pretty fine about removing the comments from the highlighted block related to picking which log to keep.  I have no reservations about removing the comments there as proposed in this PR.  If there's a belief that that's still not enough and we should remove more comments from the earlier part of the function, we might want to have a call to discuss that in more detail, as that seems to me like a change that would come with more tradeoffs.


stbrody:
added a note to clarify that there's no way to save both branches, only one can survive

stephhuynh18 (approved):

PR #2599, opened 2022-12-08
Blockchain listener

PaulLeCam:
- Extracts the constants and shared utility functions (parsing root CIDs from transactions and events) from [`ethereum-anchor-validator.ts` in the `core` package](https://github.com/ceramicnetwork/js-ceramic/blob/51726a40a88cf885673da77b3b093f8897d59e67/packages/core/src/anchor/ethereum/ethereum-anchor-validator.ts) to a new `anchor-contract` package.
- Adds a new `anchor-listener` package using the new `anchor-contract` package exposing functions to load blocks and root CIDs from contract logs, and a listener for new blocks.

@stephhuynh18 can you have a look and let me know if the APIs for the [listener](https://github.com/ceramicnetwork/js-ceramic/compare/feat/CDB-2083-blockchain-listener?expand=1#diff-8edd94f3b92ba95835a94cf05acb3a4aeacaad9480c03c81346d669230316d14) and [loader](https://github.com/ceramicnetwork/js-ceramic/compare/feat/CDB-2083-blockchain-listener?expand=1#diff-a595567767b17ad1fa5edd92c52ec571d9aa39cc7f8577654aab63fc0a543753) work for you please?

Also, I'm wondering how to perform integration tests for [extracting the root CID from transactions and events](https://github.com/ceramicnetwork/js-ceramic/compare/feat/CDB-2083-blockchain-listener?expand=1#diff-b76991375e899b996101d60a0a88b305b51124e035a8a17ef287f772f7f7b7d4): any pointer for how the [transaction data used in this test](https://github.com/ceramicnetwork/js-ceramic/blob/51726a40a88cf885673da77b3b093f8897d59e67/packages/core/src/anchor/ethereum/__tests__/ethereum-anchor-validator.test.ts#L20) is generated please?


packages/anchor-listener/src/listener.ts
```diff
@@ -0,0 +1,97 @@
...
+  firstValueFrom,
+  fromEvent,
+  map,
+  mergeMap,
+  scan,
+} from 'rxjs'
+
+import { createRootCIDsLoader, mapLoadBlock } from './loader.js'
+
+export type ProcessedBlockEvent = {
```
stephhuynh18, line 16: To re create the ipfs anchor proof I will need the following information: 
{
      blockNumber: tx.blockNumber,
      blockTimestamp: tx.blockTimestamp,
      root: merkleRootCid,
      chainId: tx.chain,
      txHash: txHashCid,
    }

This is going to change because of this project found here: https://linear.app/3boxlabs/project/graceful-block-reorgs-f1707041a92d. I believe we will be only relying on transaction hash. 

I possible could we provide additional information?

packages/anchor-listener/src/listener.ts
```diff
@@ -0,0 +1,97 @@
...
+  type: 'processed-block'
+  block: Block
+  roots: Array<CID>
+}
+
+export type ReorganizationEvent = {
+  type: 'reorganization'
+  latestProcessedBlock: string
```
stephhuynh18, line 24: I'm not entirely sure what lastProcessedBlock and newBlock represent

packages/anchor-listener/src/listener.ts
```diff
@@ -0,0 +1,97 @@
...
+  confirmations: number
+  network: keyof typeof ANCHOR_CONTRACT_ADDRESSES
+  provider: Provider
+  retryConfig?: RetryConfig
+}
+
+type State = { status: 'process' | 'reorganized'; previousHash: string; block: Block }
+type Seed = { status: 'initial' }
+
+export function createBlockListener({
```
stephhuynh18, line 41: just wanted to mention taking a glance at the implementation. I have read (and seen although it has been a few years) that blocks may come out of order.

packages/anchor-listener/src/loader.ts
```diff
@@ -0,0 +1,48 @@
...
+): OperatorFunction<number, Block> {
+  return pipe(mergeMap((blockNumber) => createBlockLoader(provider, blockNumber, retryConfig)))
+}
+
+export function createRootCIDsLoader(
+  provider: Provider,
+  filter: Filter,
+  retryConfig: RetryConfig = { count: 3 }
+): Observable<Array<CID>> {
+  return from(provider.getLogs(filter)).pipe(
```
stephhuynh18, line 44: > Keep in mind that many backends will discard old events, and that requests which are too broad may get dropped as they require too many resources to execute the query.
https://docs.ethers.io/v5/api/providers/provider/#Provider-getLogs

The above is found on the ethers documentation. I wonder if the batching(? this may not be the correct word) should be handled here or by the user of this package

packages/anchor-listener/src/listener.ts
PaulLeCam: Yes looks like we can get the transaction hash from the log entry (https://docs.ethers.io/v5/api/providers/types/#providers-Log) so shouldn't be an issue for this. We can get the block number as well but not the block timestamp though it seems.

Also looking at the `Log` interface from ethers, there's a `removed` flag, should we take it into account please? For example, ignoring logs that are removed?

packages/anchor-listener/src/listener.ts
PaulLeCam: Here `newBlock` would provide the full `Block` object from ethers: https://docs.ethers.io/v5/api/providers/types/#providers-Block and `latestProcessedBlock` would be the block hash of the `processed-block` event previously emitted, but flagged as a reorganization because `newBlock.parentHash != latestProcessedBlock`, so the consumer should handle the reorganization as needed, if I understand the purpose?

We can provide different information if needed, for example should we extract the root CIDs from the new block in this case?

packages/anchor-listener/src/listener.ts
PaulLeCam: If so that should trigger the `reorganization` event, now? The logic to detect these is just comparing the newly loaded block's `parentHash` with the hash of the previously processed block.
Shall we change this behavior? If so, what logic should we apply to detect reorganizations please?

packages/anchor-listener/src/loader.ts
PaulLeCam: Yes I was wondering about that too.
I think that can be handled here based on the `fromBlock` and `toBlock` parameters of the `filter` argument and possible error response, but it depends if the consumer wants to handle these errors directly or not I guess?
A simple logic could be to only try to load half the number of blocks on error, until loading is successful?

packages/anchor-listener/src/listener.ts
oed: I think just making a call to `eth_getBlockByNumber` can be used to fetch the timestamp. 
This is a bit extra overhead, but will be removed when the [Graceful block reorgs](https://linear.app/3boxlabs/project/graceful-block-reorgs-f1707041a92d) is implemented. 

cc @stbrody

oed: Ah, thanks @stephhuynh18 you already pointed this out!

packages/anchor-listener/src/listener.ts
stephhuynh18: sorry for the long text, this is me thinking out loud haha

packages/anchor-listener/src/loader.ts
stephhuynh18: > A simple logic could be to only try to load half the number of blocks on error, until loading is successful?
I really like that! We can add functionality in the future if needed

packages/anchor-listener/src/listener.ts
stephhuynh18: In our case I only need the re-orged block number so this is great.

packages/anchor-listener/src/listener.ts
stephhuynh18: > Also looking at the `Log` interface from ethers, there's a `removed` flag, should we take it into account please? For example, ignoring logs that are removed?

I am not familiar with the flag. Based on this comment: 
```
log.removed is only works with filters API e.g. [eth_getFilterChanges](https://eth.wiki/json-rpc/API#eth_getfilterchanges)

It is always false for eth_getLogs
```
https://github.com/ethers-io/ethers.js/issues/385#issuecomment-1076735812 
I don't think we can use this

packages/anchor-listener/src/loader.ts
PaulLeCam: @stephhuynh18 I changed the type of the event emitted from the listener to https://github.com/ceramicnetwork/js-ceramic/pull/2599/files#diff-8edd94f3b92ba95835a94cf05acb3a4aeacaad9480c03c81346d669230316d14R32 does that make sense please?
The idea is that if the `reorganized` flag is `true`, the consumer can handle what it wants to do based on the `block` and `expectedParentHash` values, what do you think please?

packages/anchor-listener/src/listener.ts
PaulLeCam: Hopefully it's handled as expected now in the listener logic, does it look OK to you please?

packages/anchor-listener/src/listener.ts
PaulLeCam: Should be addressed now with providing `AnchorProof` objects in the event.

packages/anchor-listener/src/listener.ts
stephhuynh18: the logic looks okay to me!

stbrody: Genuine question: does it matter if we process blocks out of order?  We might learn about newer documents or updates before older ones, but does that matter?  We've never really defined what the consistency model across documents is for Ceramic.  I think even if we do the work to keep updates in order here, when we eventually switch away from blockchain anchors as the mechanism for syncing towards a peer-to-peer sync based on looking up providers for a Model in the DHT, then I think we'll struggle to maintain that kind of consistency or ordering guarantee.  It might be good to get our users used to a looser consistency model now.

CC @oed @aarongoldman

packages/anchor-listener/src/listener.ts
```diff
@@ -0,0 +1,131 @@
...
+    map((state) => state.continuous),
+    // Only push non-empty lists of blocks
+    filter((blocks) => blocks.length !== 0)
+  )
+}
+
+type ProcessingState = { status: 'process' | 'reorganized'; previousHash: string; block: Block }
+type ProcessingSeed = { status: 'initial' }
+
+export function mapProcessBlock(
```
stbrody, line 86: do all these functions in this file and in loader.ts all need to be exported?  Can we either wrap them in a class with a small public API vs private helper methods, or do something else to make it clearer which functions are really intended to be consumed by users of this package vs which are mostly internal?

stbrody (commented):
@PaulLeCam I'm mostly happy to defer to Steph and Joel on reviewing this, but pretty please can we get some more comments?  Lots of types and functions with no comments right now describing what they do or how to use them.

packages/anchor-listener/src/listener.ts
PaulLeCam: All exported functions should be safe to use by consumers as needed, there's no need to make anything "private".
We don't have to export them all but I guess I don't see the downside, in case they can be useful for the consumer logic?

packages/anchor-listener/src/listener.ts
oed: I have not followed the entire conversation here, but I don't see any reason we couldn't process multiple blocks in parallel.

PaulLeCam: The listener needs to keep track of the block order in order to detect possible reorganizations of the chain, that should be handled by the consumer.
If we processed blocks out of order, we'd need more logic to track previously seen blocks and possible mismatches in the chain ancestry.

packages/anchor-listener/src/listener.ts
stbrody: I've always believed it is good engineering best practice to leave APIs as small and narrow as possible to achieve the needed goals.  Putting functionality into the public API "just in case someone needs it" increases the surface area that needs to be maintained and makes it harder to refactor in the future.  I'd imagined one single very simple function call at the top level that you point at a contract address (possibly with some optional configuration like starting block number) and that then spits out merkle root CIDs.  Then everything about how it actually gets that information is an implementation detail.

We can always expose more functionality later if someone needs it, it's a lot harder to take functionality away once people start relying on it.

packages/anchor-listener/src/listener.ts
stbrody: > If we processed blocks out of order, we'd need more logic to track previously seen blocks and possible mismatches in the chain ancestry.

sorry, I don't really follow - can you elaborate on what this means?

Isn't whether or not we allow blocks to be processed in parallel a separate concern from detecting block reorgs?

PaulLeCam: The listener logic only keeps track of the hash of the previously seen block, to check if `block.parentHash = previousHash` when receiving new blocks to flag reorganizations.
The loading of blocks and anchors proofs can be done in parallel, but then the data gets pushed ordered by block number.

packages/anchor-listener/src/listener.ts
PaulLeCam: I agree for "high-level" APIs I prefer a smaller surface, but these packages are mainly low-level utilities that can be useful in different cases. If people use them, to me it's more a good sign that the library is useful to the community rather than a downside, no?
Anyway, I reduced the exports to the 3 highest level functions, as you wrote we can always export more if needed.

packages/anchor-listener/src/listener.ts
stbrody: okay, that makes sense.  I believe we do still hold off delivering newer blocks until we hear about the older blocks to fill in the gap, but that should be a small amount of time and relatively rare, so probably not a big deal

oed: In theory we could process everything except the most recent days worth of blocks in parallel since reorgs are extremely unlikely. However, adding this logic is probably more effort than it's worth right now.

packages/anchor-listener/src/listener.ts
```diff
@@ -0,0 +1,152 @@
...
+    scan<number, OrderingState, OrderingState>(
+      (state, blockNumber) => {
+        if (state.latestContinuous != null) {
+          if (blockNumber <= state.latestContinuous) {
+            // Already processed block number
+            return { ...state, continuous: [] }
+          }
+          if (blockNumber > state.latestContinuous + 1) {
+            // Future block number, keep track
+            return {
```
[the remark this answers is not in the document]
stbrody: FWIW, I think this is what I was trying to get at with my comment here: https://github.com/ceramicnetwork/js-ceramic/pull/2599#discussion_r1052660029

I feel like it would be perfectly safe to start processing blocks as soon as we hear about them, no need to wait.  But happy to defer to Steph and Paul to figure out if it makes sense to wait to make reorg handling easier.

packages/anchor-utils/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+```ts
+import {
+  ANCHOR_CONTRACT_ADDRESSES,
+  type SupportedNetwork,
+} from "@ceramicnetwork/anchor-utils";
+
+function getContractAddress(network: SupportedNetwork): string {
+  const address = ANCHOR_CONTRACT_ADDRESSES[network];
+  if (address == null) {
```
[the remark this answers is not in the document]
PaulLeCam: Yes exactly, it's just an additional runtime check in case the type check is bypassed

packages/anchor-listener/src/loader.ts
```diff
@@ -0,0 +1,267 @@
...
+  provider,
+  chainId,
+  fromBlock,
+  toBlock,
+  retryConfig,
+  blockLoadBuffer,
+}: BlocksProofsLoaderParams): Observable<BlockProofs> {
+  const retry = retryConfig ?? { count: 3 }
+  return range(fromBlock, toBlock - fromBlock + 1).pipe(
+    bufferCount(blockLoadBuffer ?? 5),
```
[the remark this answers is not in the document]
PaulLeCam: I expect we'd get `blockLoadBuffer` + 1 slices pushed, with the last slice possibly smaller than `blockLoadBuffer`, at least that's my understanding of how the `bufferCount()` operator works.

packages/anchor-listener/src/loader.ts
```diff
@@ -0,0 +1,267 @@
...
+  targetAncestorHash,
+  retryConfig,
+  maxConcurrency,
+}: AncestorBlocksProofsLoaderParams): Observable<BlockProofs> {
+  const retry = retryConfig ?? { count: 3 }
+  return createBlockLoader(provider, initialBlock, retry).pipe(
+    expand((block) => {
+      return block.parentHash === targetAncestorHash
+        ? EMPTY
+        : createBlockLoader(provider, block.parentHash, retry)
```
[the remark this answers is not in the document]
PaulLeCam: Yes it should, it simply calls ethers' `getBlock()` method that supports block hashes as far as I understand.

packages/anchor-listener/src/loader.ts
```diff
@@ -0,0 +1,267 @@
...
+  initialBlock,
+  targetAncestorHash,
+  retryConfig,
+  maxConcurrency,
+}: AncestorBlocksProofsLoaderParams): Observable<BlockProofs> {
+  const retry = retryConfig ?? { count: 3 }
+  return createBlockLoader(provider, initialBlock, retry).pipe(
+    expand((block) => {
+      return block.parentHash === targetAncestorHash
+        ? EMPTY
```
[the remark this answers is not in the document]
PaulLeCam: When returning `EMPTY` the `expand()` operator completes the observable, so basically ends the recursive block loading process.
The `mapLoadBlockProofs()` operator loads the anchor proofs for each input `Block` provided, and returns a `BlockProofs` object.

packages/anchor-listener/src/listener.ts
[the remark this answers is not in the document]
PaulLeCam: Yes I agree that doesn't really make sense anymore, I originally thought we wouldn't provide the blocks and anchors when reorganizations are detected so I wanted to avoid overfetching, but we need to provide them anyways so we can simplify the process, I'll update the logic.

packages/anchor-utils/src/__tests__/cid.test.ts
```diff
@@ -0,0 +1,36 @@
...
+  test('getCidFromAnchorEventLog()', () => {
+    const bytes = new Uint8Array([
+      0, 1, 2, 3, 4, 5, 6, 7, 0, 1, 2, 3, 4, 5, 6, 7, 0, 1, 2, 3, 4, 5, 6, 7, 0, 1, 2, 3, 4, 5, 6,
+      7,
+    ])
+    const log = { data: defaultAbiCoder.encode(['bytes32'], [bytes]) } as Log
+    const cid = getCidFromAnchorEventLog(log)
+    expect(cid.equals(createCID(bytes))).toBe(true)
+  })
+})
```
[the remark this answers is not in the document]
PaulLeCam: Yes I think more logic should go in this package, maybe it's something we can do as part of another ticket or at least PR though to avoid blocking this one if it's not critical?

stephhuynh18 (approved):

PR #2601, opened 2022-12-08
Use DID to authenticate to CAS by default

v-stickykeys:
## Description

By default, new Ceramic nodes will use DIDs to sign requests made to CAS. This will be disabled for existing Ceramic nodes but can be enabled in the daemon config.

(CAS will keep a registry of node DIDs, verify signatures against the registry, and track nonces per DID to prevent replay attacks. This work is implemented in the [CAS repo](https://github.com/ceramicnetwork/ceramic-anchor-service/pull/903).)

### Motivation

Remove the IP address requirement so that developers wanting to get onto mainnet can do so more easily, while still making it possible to uniquely identify requests made to CAS. The latter is essential for conflict resolution when we remove the Ceramic node in use by CAS.

### Changes

- Create an authenticated version of the Ethereum anchor service
- Read did seed from config file or env vars. Give deprecation warning if it is not present.
- Generate a random did seed for default daemon config and use it if no config file is present

## How Has This Been Tested?

Created unit tests and also tested live against a local CAS api gateway setup with serverless.
Tested e2e against dev environment (demoed in townhall)


v-stickykeys:
oop sorry for failed tests. will fix those

packages/cli/src/ceramic-cli-utils.ts
```diff
@@ -30,25 +31,42 @@ const DEFAULT_CLI_CONFIG_FILENAME = new URL('client.config.json', DEFAULT_CONFIG
...
+    'http-api': { 'cors-allowed-origins': [new RegExp('.*')], 'admin-dids': [] },
+    ipfs: { mode: IpfsMode.BUNDLED },
+    logger: { 'log-level': LogLevel.important, 'log-to-files': false },
+    metrics: {
+      'metrics-exporter-enabled': false,
+    },
+    network: { name: Networks.TESTNET_CLAY },
+    node: {
+      'did': did,
+      'did-seed': didSeed
```
oed, line 58:
```suggestion
      'did-private-key': didSeed
```

Using private key instead of seed is more precise here.

packages/cli/src/ceramic-cli-utils.ts
```diff
@@ -30,25 +31,42 @@ const DEFAULT_CLI_CONFIG_FILENAME = new URL('client.config.json', DEFAULT_CONFIG
...
+    },
+    'http-api': { 'cors-allowed-origins': [new RegExp('.*')], 'admin-dids': [] },
+    ipfs: { mode: IpfsMode.BUNDLED },
+    logger: { 'log-level': LogLevel.important, 'log-to-files': false },
+    metrics: {
+      'metrics-exporter-enabled': false,
+    },
+    network: { name: Networks.TESTNET_CLAY },
+    node: {
+      'did': did,
```
oed, line 57: Why is the did itself needed in the config? Makes it harder to manually configure a private key. Could we just output the DID when we start the node?

packages/cli/src/daemon-config.ts
```diff
@@ -260,6 +267,20 @@ export class DaemonDidResolversConfig {
...
+  @jsonMember(String)
+  did: string
+
+  /**
+   * Seed used to sign requests to CAS.
+   * This is randomly generated if a config file is not found.
+   * When specifying in a config file, use the name 'did-seed'.
+   */
+  @jsonMember(String, { name: 'did-seed' })
+  didSeed: URL
```
oed, line 282: Why is this a URL? I would have thought that it's a hex encoded string?

packages/cli/src/daemon/did-utils.ts
```diff
@@ -0,0 +1,49 @@
...
+    return new Ed25519Provider(seed)
+}
+
+export function generateSeed(): string {
+    return u8a.toString(randomBytes(32), 'base16')
+}
+
+export function generateSeedUrl(): URL {
+    const seed = generateSeed()
+    const url = `inplace:ed25519#${seed}`
```
oed, line 25: Ok, this explains my previous question. 
Suggestion:
Encode this as a multibase string prefixed by the `ed25519-priv` multicode: `0x1300`.

e.g.
`multibase(varint.encode(0x1300), seed)`

References:
https://github.com/multiformats/multicodec/blob/master/table.csv#L160
https://github.com/multiformats/multibase
https://www.npmjs.com/package/varint

packages/cli/src/ceramic-cli-utils.ts
v-stickykeys: yeah i like this idea

packages/cli/src/ceramic-cli-utils.ts
v-stickykeys: ? it is the seed that generates the private key iiuc
https://github.com/ceramicnetwork/js-did/blob/27a124597af9f61147be48ecca907978579d60ae/packages/key-did-provider-ed25519/src/index.ts#L93-L94

packages/cli/src/daemon/did-utils.ts
v-stickykeys: would keep this as a url. at least that way we can accept seeds from different locations in the future.

but yeah can multibase encode the `inplace` ones

packages/cli/src/ceramic-cli-utils.ts
oed: Good point. I thought that the seed and the private key was equivalent in this case, but it seems like their not (seed is 32 bytes, private key is 64 bytes). 

Two options:
- Maybe call it `private-seed`, or keep `seed` but don't use the multicode varint format as suggested below (that requires the actual private key to be used
- Use [`extractPublicKeyFromSecretKey`](https://www.stablelib.com/modules/_stablelib_ed25519.html#extractPublicKeyFromSecretKey) instead and use the suggested private key encoding as suggested below.

packages/cli/src/daemon/did-utils.ts
oed: Is there a requirement for this is a URL? 
Using the `0x1300` only make sense if we store the actual private key, not the seed.

packages/cli/src/ceramic-cli-utils.ts
v-stickykeys: do you think there is any advantage between one or the other in terms of storing the seed vs storing the private key?

oed: It's mainly easier to understand what a private key is than what a seed is. I don't feel strongly though.

packages/cli/src/daemon-config.ts
```diff
@@ -194,6 +194,13 @@ export class DaemonAnchorConfig {
   @jsonMember(String, { name: 'anchor-service-url' })
   anchorServiceUrl?: string
 
+  /**
+   * Controls the authentication method Ceramic uses to make requests to the Ceramic Anchor Service.
+   * When specifying in a config file, use the name 'auth-method'.
+   */
+  @jsonMember(String, { name: 'auth-method' })
```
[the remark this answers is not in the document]
v-stickykeys: right now it is just `did` or empty string for unauthenticated. a node operator might change this depending on what anchor service they are using.

packages/cli/src/daemon-config.ts
```diff
@@ -260,6 +267,20 @@ export class DaemonDidResolversConfig {
...
+   */
+  @jsonMember(String)
+  did: string
+
+  /**
+   * Seed used to sign requests to CAS.
+   * This is randomly generated if a config file is not found.
+   * When specifying in a config file, use the name 'did-seed'.
+   */
+  @jsonMember(String, { name: 'did-seed' })
```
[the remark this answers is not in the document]
v-stickykeys: it's a good question. s3 actually be default will read creds from a file. and then otherwise from env var. kubo keeps the secret used to generate peer ids in the config file as well--this design is modeled after how kubo does it today.

packages/cli/src/daemon-config.ts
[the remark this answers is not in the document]
stbrody: we only support one anchor service per network though - can we make the decision automatically and reduce how much configuration we expose to node operators?

packages/core/src/anchor/auth/did-anchor-service-auth.ts
```diff
@@ -0,0 +1,105 @@
...
+   */
+  set ceramic(ceramic: CeramicApi) {
+    this._ceramic = ceramic
+  }
+
+  get nonce(): number {
+    return this._nonce
+  }
+
+  init = async (): Promise<void> => {
```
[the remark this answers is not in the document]
v-stickykeys: it does matter for how `this.` is interpreted. using an arrow function was necessary for one of the functions here so i decided to change all of them to arrow functions

packages/core/src/anchor/auth/did-anchor-service-auth.ts
```diff
@@ -0,0 +1,105 @@
...
+    this._updateNonce()
+    const { authorization } = await this.signRequest(url, opts?.body, this._nonce)
+    let headers: any = { authorization }
+    opts?.headers && (headers = {...opts.headers, headers })
+    return await this._sendRequest(url, { ...opts, headers})
+  }
+
+  signRequest = async (url: URL | string, body?: any, nonce?: number): Promise<{jws: DagJWS, authorization: string}> => {
+    let payload: any = { url }
+    body && (payload = { ...payload, body})
```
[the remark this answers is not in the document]
v-stickykeys: If body evaluates to truthy, add it to payload. I think it actually could be `body && (payload.body = body)`

packages/core/src/anchor/auth/did-anchor-service-auth.ts
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  /**
+   * Increments nonce in memory.
+   */
+  private _updateNonce = async (): Promise<void> => {
+    this._nonce = this._nonce + 1
+  }
+
+  private _sendRequest = async (url: URL| string, opts?: FetchOpts): Promise<any> => {
```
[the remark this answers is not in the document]
v-stickykeys: good point. I will add tests for this. the idea I have for this is, from ceramic every request must increment the nonce. skipping nonces (e.g. send 1 then send 3) is fine, repeating a nonce (e.g. send 1 then send 1 again) is never ok.

CAS is responsible for storing the nonce whether the request fails or not. if storing the nonce fails then CAS is broken.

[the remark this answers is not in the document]
v-stickykeys: the nonce is used to prevent replay attack and identify unique requests. using an expiry timestamp would not allow replay attack to be prevented. unless each timestamp must be unique. a bit tough depending on precision but also might as well use a nonce in that case.

I think the risk of replay attack is small to be fair but taking similar precaution as with the admin api (except in this case I don't want to do a challenge response because that makes unnecessary extra requests)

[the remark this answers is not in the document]
oed: Another approach might be to use a random nonce? This would remove the need to make an additional request to retrieve a nonce and the work of keeping track of it on the js-ceramic side.

Also FYI, @ukstv's work on CAS w/o Ceramic node will add a timestamp to the request regardless.

packages/core/src/anchor/auth/did-anchor-service-auth.ts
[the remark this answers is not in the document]
stbrody: Ah okay.  Personally I'd vote for 
```
if (body) {
  payload.body = body
}
```
as it's more clear to me and I don't like any code that makes the reader have to think to understand it, but your choice

packages/core/src/anchor/auth/did-anchor-service-auth.ts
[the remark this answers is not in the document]
stbrody: Sorry, I don't really understand the issue.  Issues with `this` tracking make me worried something else might be going on.  Maybe we could sync on this sometime and you could show me what's happening?

packages/core/src/anchor/auth/did-anchor-service-auth.ts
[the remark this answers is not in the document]
stbrody: between challenge-response like the AdminAPI does, timestamp, or random nonce, I have no particular strong feelings.  I do think the incrementing nonce approach though does add extra fragility to the system, so I think one of the other 3 would be better if it can achieve the same goals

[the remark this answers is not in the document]
v-stickykeys: How does random nonce work @oed ?

That sounds like the same as challenge response or is it different?
Or is it just that you can pick any number as long as you haven't used it before? That means we have to track every nonce ever used.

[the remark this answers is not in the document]
v-stickykeys: I'll be updating this to the random nonce implementation and the nonces will be tracked by CAS

packages/cli/src/daemon-config.ts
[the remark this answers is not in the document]
v-stickykeys: I think this is a good way to easily disable authenticating every request. I'm not really positive when this would be useful, in some potential debugging or failure scenario, but it feels like a good idea to have an easy way to turn this off. 

Don't feel too strongly though either way.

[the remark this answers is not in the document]
v-stickykeys: well actually this could be a good way to feature flag also.

packages/cli/src/daemon/did-utils.ts
v-stickykeys: I think storing seed is actually nice because composedb cli has `composedb did:generate-private-key` which we can point people to. (though it also produces a seed, not a private key)

packages/core/src/anchor/auth/did-anchor-service-auth.ts
```diff
@@ -0,0 +1,74 @@
...
+    const { request } = await this.signRequest({url, opts})
+    return await this._sendRequest(request)
+  }
+
+  async signRequest(request: FetchRequestParams): Promise<{request: FetchRequestParams, jws: DagJWS}> {
+    const payload: any = { url: request.url, nonce: crypto.randomUUID() }
+    if (request.opts) {
+      if (request.opts.body) {
+        payload.body = request.opts.body
+      }
```
oed, line 51:
```suggestion
      if (request.opts.body) {
        const car = CarFactory.fromBytes(u8a.fromString(request.opts.body, 'binary'))
        payload.body = car.roots[0].toString()
      }
```
There is only a body if a CAR file is present (assuming this PR is on top of CAR file PR). This way we sign the CID of the root.

Using https://github.com/ceramicnetwork/cartonne

v-stickykeys: ty! I actually wonder if using a hash over the body in general is a better approach than including the full body as you mentioned

oed: It could be, but in the case of a CAR file that would be redundant since the root hash is already a hash over the entire body.

packages/core/src/anchor/auth/did-anchor-service-auth.ts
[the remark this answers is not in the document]
v-stickykeys: This is no longer relevant because we changed the nonce implementation to use a random nonce. Can change these to non arrow functions.

v-stickykeys:
Unclear why some tests are failing now but will resolve during today's pairing session.

packages/cli/src/ceramic-cli-utils.ts
```diff
@@ -120,6 +145,9 @@ export class CeramicCliUtils {
       config.metrics.metricsExporterEnabled = process.env.CERAMIC_METRICS_EXPORTER_ENABLED == 'true'
     if (process.env.COLLECTOR_HOSTNAME)
       config.metrics.collectorHost = process.env.COLLECTOR_HOSTNAME
+    if (process.env.CERAMIC_NODE_DID_SEED) {
+      config.node.privateSeed = process.env.CERAMIC_NODE_PRIVATE_SEED_URL
```
stbrody, line 149: is this env var supposed to be different than the one in the enclosing if-statement? SEED vs SEED_URL

stbrody (commented):
I'm kind of confused throughout this PR on the relationship between the private seed and the seed URL.  Is the URL a path to the local file system with a file that contains the actual private seed?  Or is it actually the private seed itself?  Can we make this clearer with comments and naming conventions throughout?

packages/cli/src/ceramic-cli-utils.ts
v-stickykeys: thanks this was a typo

v-stickykeys:
> I'm kind of confused throughout this PR on the relationship between the private seed and the seed URL. Is the URL a path to the local file system with a file that contains the actual private seed? Or is it actually the private seed itself? Can we make this clearer with comments and naming conventions throughout?

Yes I see how this can be confusing. I will update the name to be consistent. It should be `private seed url`.

By default, it is "inplace" meaning the seed itself is contained in the url.

packages/cli/src/daemon-config.ts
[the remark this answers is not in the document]
v-stickykeys: this is proving useful because we want auth disabled for Clay. keeping unless any other objections come up

packages/cli/src/daemon-config.ts
```diff
@@ -420,6 +451,12 @@ export class DaemonConfig {
   static async fromFile(filepath: URL): Promise<DaemonConfig> {
     const content = await readFile(filepath, { encoding: 'utf8' })
     const config = DaemonConfig.fromString(content)
+    // TODO: Replace with commented lines below when we fully deprecate IP address access to mainnet
+    if (!config.node) config.anchor.authMethod = AnchorServiceAuthMethods.NONE
```
v-stickykeys, line 455: @pawartur Here we allow the node to start even if the private seed url is not in the config

packages/core/src/ceramic.ts
```diff
@@ -437,22 +440,39 @@ export class Ceramic implements CeramicApi {
...
 
+      if (config.anchorServiceAuthMethod != AnchorServiceAuthMethods.NONE) {
+        try {
+          anchorServiceAuth = new AnchorServiceAuthMethodClasses[AnchorServiceAuthMethods.DID](anchorServiceUrl, logger)
+        } catch (error) {
+          throw new Error(`Auth method for anchor service failed to instantiate: ${config.anchorServiceAuthMethod}`)
+        }
+      } else {
+        logger.warn(
+          `DEPRECATION WARNING: IP address authentication will still work but will soon be deprecated. Add a private seed url to your daemon config.`
```
v-stickykeys, line 457: @pawartur Here we provide a deprecation warning if the auth method is not set.

willex (approved):
LGTM! 🚀

PR #2607, opened 2022-12-13
chore(core): Return blockTimestamp from anchor commit validation

stbrody:


packages/core/src/anchor/memory/in-memory-anchor-service.ts
```diff
@@ -39,7 +39,12 @@ interface InMemoryAnchorConfig {
   verifySignatures?: boolean
 }
 
-const SAMPLE_ETH_TX_HASH = 'bagjqcgzaday6dzalvmy5ady2m5a5legq5zrbsnlxfc2bfxej532ds7htpova'
+// Caches recent anchor txn hashes and the timestamp when they were anchored
+// This is intentionally global and not a member of InMemoryAnchorService. This is so that when
```
stbrody, line 43: I'm not thrilled about this, but it was the only way to get the tests passing.  It does kind of make the InMemoryAnchorService work more like the real anchor service, so it kind of makes sense.  Probably a better solution would be to make tests that use multiple Ceramic nodes share the same InMemoryAnchorService instance, but that would be a bigger change.

packages/core/src/__tests__/ceramic.test.ts
```diff
@@ -141,8 +141,8 @@ describe('Ceramic integration', () => {
         stream3,
         2000,
         (state) => StreamUtils.statesEqual(state, stream1.state),
-        () => {
-          throw new Error(`streamtype3.state should equal streamtype1.state`)
+        (state) => {
+          expect(state).toEqual(stream1.state)
```
oed, line 145: Can this actually be equal at some point? Why not use `statesEqual` as in the happy path if so?

packages/stream-model-handler/src/__tests__/model-handler.test.ts
```diff
@@ -474,8 +474,6 @@ describe('ModelHandler', () => {
 
     // apply anchor
     const anchorProof = {
-      blockNumber: 123456,
-      blockTimestamp: 1615799679,
       chainId: 'fakechain:123',
     }
```
oed, line 478: These anchor proofs are a bit silly now. ideally they should include a txHash, etc. not super important though.

oed (approved):

packages/core/src/__tests__/ceramic.test.ts
stbrody: because this way gives a nicer error message.  The way it was written just tells you that they don't equal, the new way will print out the diff of what's actually different between the two

packages/stream-model-handler/src/__tests__/model-handler.test.ts
stbrody: yeah personally I like using the bare minimum required in a unit test.  Makes it clear to readers what's actually being used.  I could add txHash to make it look like a real anchorProof, but then readers might think that's actually relevant to the test, which it isn't

PR #2609, opened 2022-12-14
Make sure that the database api is initialised correctly on startup

pawartur:
## Description

This is just a quick fix for the following problem:

`PostgresIndexApi` and `SqliteIndexApi` need their internal `modelsIndexedFields` in-memory caches to be able to correctly index documents with relations. However, currently the only way `modelsIndexedFields` caches are populated is when `LocalIndexApi` calls `.indexModels(...)` on its `databaseIndexApi`. It wasn't always doing it, though, which causes crashes when attempting to index a document with relations in Postgres (Sqlite wasn't throwing in this case, which made it more difficult to detect).

This PR doesn't have tests, as adding them for this band-aid fix would take significant time, which could be better spent working on a proper fix.

## How Has This Been Tested?

- [X] Tested manually that indexing a document with relations works with a Postgres indexing db

## PR checklist


stbrody:
Nice find and fix.  I took a crack at writing a test for this, PR here: https://github.com/ceramicnetwork/js-ceramic/pull/2620

ukstv (approved):

stbrody (approved):
LGTM

PR #2612, opened 2022-12-15
Use clearer syntax for code reading interview

stbrody:


packages/core/src/document.ts
```diff
@@ -432,8 +432,8 @@ class Document extends EventEmitter {
     const isRemoteAnchored = remoteState.anchorStatus === AnchorStatus.ANCHORED
 
     if (isLocalAnchored && isRemoteAnchored) {
-      const { anchorProof: localProof } = localState
```
stbrody, line 435: this syntax is very javascript-specific and the goal of this interview isn't to evaluate javascript knowledge

PR #2616, opened 2022-12-18
Update stream content in DB on stream updates

serafettin:
## Description

Indexing streams to DB works well for new streams but not for the updates.
It only updates `last_anchored_at` and `updated_at` fields for existing streams.
This commit simply adds `stream_content` field for both SQLite and PostgreSQL indexers.

### DIFF
--- a/packages/core/src/indexing/postgres/postgres-index-api.ts
+++ b/packages/core/src/indexing/postgres/postgres-index-api.ts
@@ -73,6 +73,7 @@ export class PostgresIndexApi implements DatabaseIndexApi {
       .onConflict('stream_id')
       .merge({
         stream_content: indexingArgs.streamContent,
+        tip: indexingArgs.tip.toString(),
         last_anchored_at: indexingArgs.lastAnchor,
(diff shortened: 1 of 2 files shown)


packages/core/src/indexing/postgres/postgres-index-api.ts
```diff
@@ -72,6 +72,7 @@ export class PostgresIndexApi implements DatabaseIndexApi {
       .insert(indexedData)
       .onConflict('stream_id')
       .merge({
+        stream_content: indexingArgs.streamContent,
```
stbrody, line 75: good catch!  This does indeed seem like a bug to me.  I think we probably also want to add `tip` into this merge to be updated as well, here and in the sqlite version.  Would you mind adding that change in as well?

serafettin: Of course. I've added the tip.

stbrody:
@serafettin sorry for the delay, getting ready to merge this in.  Can you please merge in the latest changes from the develop branch so it's ready to be merged?

stbrody:
Nevermind, since it applied cleanly I could just merge directly.

Thanks for your contribution!

PR #2621, opened 2022-12-20
feat(core): job queue interface

stephhuynh18:
## Description

Minimal job queue implementation and interfaces

Also needed to fix broken existing tests


packages/core/src/sync/sync-api.ts
```diff
@@ -0,0 +1,52 @@
...
+  pauseModelSync(models: string | string[]): Promise<void>
+}
+
+interface IpfsService {
+  retrieveFromIPFS(cid: CID | string, path?: string): Promise<any>
+  storeCommit(data: any, streamId?: StreamID): Promise<CID>
+  storeRecord(record: Record<string, unknown>): Promise<CID>
+}
+
+interface Protocol {
```
stbrody, line 22: Protocol is a pretty generic name.  Consider something more descriptive or at least a comment explaining what this represents

packages/core/src/sync/sync-api.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  async startModelSync(
+    startBlock: number,
+    endBlock: number,
+    models: string | string[]
+  ): Promise<void> {
+    // add to job queue
+  }
+
+  async pauseModelSync(models: string | string[]): Promise<void> {
```
stbrody, line 49: do we need the ability to pause individual models yet?  Should we have a single shutdown() function that stops all syncs?

stbrody (approved):
lgtm

packages/core/src/indexing/__tests__/build-indexing.test.ts
```diff
@@ -1,17 +1,30 @@
...
 const diagnosticsLogger = new LoggerProvider().getDiagnosticsLogger()
 // @ts-ignore default import
 const getDatabase = pgTest.default
 
+jest.setTimeout(150000)
+
 describe('sqlite', () => {
   let databaseFolder: tmp.DirectoryResult
 
+  beforeAll(async () => {
```
PaulLeCam, line 20: Why do we need to setup Postgres when running the SQLite tests please?
does this only affect the test setup, or does it become necessary to run Ceramic even with historical indexing disabled?

PaulLeCam (approved):
Interfaces LGTM, I'm just wondering about the logic to make historical sync Postgres-only, that would be handled at a different level I suppose?

packages/core/src/indexing/__tests__/build-indexing.test.ts
stephhuynh18: oooo good catch! You only need postgres if using postgres =)

packages/core/src/sync/sync-api.ts
stephhuynh18: Yea i don't think we need the ability yet so removed and added shutdown instead.

PR #2622, opened 2022-12-20
feat(core): Validate anchors and extract timestamp information before commit application

stbrody:
I went with this approach, where we validate the anchor commits right after fetching the log, but before applying them, and remember any validation errors so they can be thrown later during commit application.

This was the shortest path to making the changes needed for Graceful Block Reorgs while preserving existing functionality and not making major changes to how signatures are verified or how CACAO timeouts are handled.

That means this is a temporary fix.  We will hit against this problem again when we move log syncing to happen in a single pass where we go from oldest commit to newest commit, without the preliminary pass that first loads commits from newer to older.  We will need to make sure the scope for StreamSync includes time to think about how CACAO timeouts get validated and where the proper separation of concern boundaries lie between the CACAO/DIDs library and js-ceramic.  We might even want to consider a project here sooner so that we can ensure that the CACAO and DIDs libraries have the right APIs and boundaries around how timestamp information is handled asap before they get adopted outside of js-ceramic.

### DIFF
--- a/packages/core/src/conflict-resolution.ts
+++ b/packages/core/src/conflict-resolution.ts
@@ -417,8 +417,8 @@ export class ConflictResolution {
     }
 
     const stateLog = HistoryLog.fromState(this.dispatcher, initialState)
-    const log = await fetchLog(this.dispatcher, tip, stateLog).then((log) =>
-      verifyAnchorAndApplyTimestamps(this.logger, this.dispatcher, this.anchorValidator, log)
+    const logWithoutTimestamps = await fetchLog(this.dispatcher, tip, stateLog)
+    const log = await verifyAnchorAndApplyTimestamps(this.logger, this.dispatcher, this.anchorValidator, logWithoutTimestamps)
     )
 
     if (log.length) {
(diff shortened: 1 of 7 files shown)


packages/core/src/conflict-resolution.ts
```diff
@@ -181,6 +181,39 @@ export class HistoryLog {
...
+      try {
+        timestamp = await verifyAnchorCommit(dispatcher, anchorValidator, commitData)
+      } catch (err) {
+        // Save the error for now to be thrown when trying to actually apply the anchor commit.
+        logger.warn(`Error when validating anchor commit: ${err}`)
+        commitData.anchorValidationError = err
+      }
+    }
+    commitData.timestamp = timestamp
+  }
```
oed, line 212: Imo, a map would be easier to read. This is fine though.

packages/core/src/conflict-resolution.ts
```diff
@@ -416,7 +417,10 @@ export class ConflictResolution {
     }
 
     const stateLog = HistoryLog.fromState(this.dispatcher, initialState)
-    const log = await fetchLog(this.dispatcher, tip, stateLog)
+    const log = await fetchLog(this.dispatcher, tip, stateLog).then((log) =>
+      verifyAnchorAndApplyTimestamps(this.logger, this.dispatcher, this.anchorValidator, log)
+    )
```
oed, line 422:
```suggestion
    const logWithoutTimestamps = await fetchLog(this.dispatcher, tip, stateLog)
    const log = await verifyAnchorAndApplyTimestamps(this.logger, this.dispatcher, this.anchorValidator, logWithoutTimestamps)
    )
```

Kind of weird to mix `await` and `.then`.

oed (approved):

PR #2623, opened 2022-12-20
chore(core): Remove all references to AnchorProof.blockTimestamp (or blockNumber)

stbrody:
builds on top of https://github.com/ceramicnetwork/js-ceramic/pull/2622


packages/common/src/utils/stream-utils.ts
```diff
@@ -327,9 +327,27 @@ export class StreamUtils {
...
+  static anchorTimestampFromState(state: StreamState): number | null {
+    for (let i = state.log.length - 1; i >= 0; i--) {
+      const entry = state.log[i]
+      if (entry.timestamp) {
+        return entry.timestamp
+      }
+    }
+
+    return null
+  }
```
oed, line 352: Any reason we want to return `null` rather than `undefined` here? Either way, this is simpler:
```suggestion
  static anchorTimestampFromState(state: StreamState): number | undefined {
    return state.log.findLast(e => e.timestamp)?.timestamp
  }
```

oed (approved):
lgtm, but we can use built in function

packages/common/src/utils/stream-utils.ts
stbrody: looks like findLast isn't supported in node v16

stbrody: https://app.circleci.com/pipelines/github/ceramicnetwork/js-ceramic/9286/workflows/5b7f5f00-1ea3-4f6a-a470-2c056956b76a/jobs/12266

oed: ok, annoying.

oed: yeah, support was added in node 18.

PR #2633, opened 2023-01-04
feat: Add sep key to model stream types

oed:
This PR adds the `sep` key to the genesis header as specified in CIP-120. While not used yet it will allow the future upgrade to support CIP-120 much less painful.

Note: This is a breaking change to deterministic MID documents.


ukstv (approved):

oed:
Need to fix tests.

stbrody (approved):
lgtm

oed:
Note that this will be a breaking change for any `SINGLE` documents on Clay testnet.

PR #2635, opened 2023-01-05
Wait for all relevant anchor status updates in the anchor resume test

pawartur:
## Description

This test had been failing on CircleCi occasionally. I wasn't able to reproduce the failure locally, so I increased the timeout a bit and made the test wait for all anchor status updates before checking them. Let's see, if this make CircleCI stop failing.


stbrody (approved):
LGTM

PR #2639, opened 2023-01-05
feat(core): merkle tree leaf loader

stephhuynh18:
Retrieves the leaf cid at a specifc index and the merkle tree metadata. 
Tested using real merkle root cids as well as these tests


packages/core/src/sync/utils.ts
```diff
@@ -0,0 +1,81 @@
...
+import { IpfsService, TreeMetadata } from './interfaces.js'
+
+const METADATA_PATH = '2'
+
+enum PathDirection {
+  L,
+  R,
+}
+
+const getPath = (index: number, leavesCount: number): PathDirection[] => {
```
stbrody, line 11: Do you have any thoughts on how we're going to test that the path logic we have here perfectly matches the logic that the CAS uses to build its merkle tree?

stbrody (approved):
lgtm, main concern is around testing with various tree structures as generated by the CAS

packages/core/src/sync/utils.ts
stephhuynh18: yes very valid concern. I have manually tested this using real merkle root cids retrieved from the QA cas logs. I just felt for a unit test I shouldn't be relying on data that was stored via external sources. 

My plan is to have a larger test for the sync service where I use a CAS to anchor some MIDs to a test ethereum node that the syncer is hooked up to.

stbrody: Sounds good.  I also imagine we could probably import this function into the CAS codebase, which already depends on js-ceramic, and extend our unit tests that already build merkle trees of various structures with different numbers of nodes to also test this function against them.

PR #2640, opened 2023-01-06
Fixed the implementation of S3Store.exists(...)

pawartur:
## Description

There was a typo in the implementation of `S3Store.exists(...)` which caused crashes when checking, if a running state should be resumed for a `StreamID`.

## How Has This Been Tested?

- [X] Tested manually with S3 and clay-testnet using a local node (after reproducing the error)

## PR checklist

### DIFF
--- a/packages/cli/src/s3-store.ts
+++ b/packages/cli/src/s3-store.ts
@@ -94,7 +94,8 @@ export class S3Store implements IKVStore {
   async exists(key: string, useCaseName?: string): Promise<boolean> {
     const store = await this.#storeMap.get(useCaseName)
     try {
-      return typeof (await store.get(key).toString()) === 'string'
+      const value = await store.get(key)
+      return value !== undefined
     } catch (e) {
       if (/Key not found in database/.test(e.toString())) {
         return false


packages/cli/src/s3-store.ts
```diff
@@ -94,7 +94,8 @@ export class S3Store implements IKVStore {
   async exists(key: string, useCaseName?: string): Promise<boolean> {
     const store = await this.#storeMap.get(useCaseName)
     try {
-      return typeof (await store.get(key).toString()) === 'string'
+      const value = await store.get(key)
+      return value !== undefined
```
ukstv, line 98: Maybe:

```suggestion
      return Boolean(value)
```

ukstv (approved):

packages/cli/src/s3-store.ts
pawartur: Thanks, but not sure about this. `Boolean(value)` also returns `false` for `""`, `0`, etc., right? I can imagine that this general S3 storage class may want to store `0`s as values for something some time :) .

ukstv: True. That's why "maybe" :)

oed: @pawartur please refer to javascript documentation. The cases you mentioned returns false.

stbrody (approved):
LGTM

PR #2641, opened 2023-01-09
Feat/car files in anchor requests cdb 2099

pawartur:
## Made `EthereumAnchorService` send CAR files in anchor requests, rather than stream ids and CIDs

- changed the requests' content-type to `application/vnd.ipld.car`
- started generating and attaching CAR files with the following fields:
  - {
        timestamp: <date ISO string>,
        streamId: <stream id bytes>,
        tip: <anchored commit's CID>,
      },
    { isRoot: true }
  - genesis commit block
  - genesis cimmit's link block (if present)
  - tip commit block
  - tip commit's link block (if present)
  - tip commit's CACAO block (if present)

## How Has This Been Tested?

- [X] Updated existing tests
- [X] Added a test that checks that the right CAR file is build and show how to parse the expected data from it

## PR checklist


packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -105,6 +113,13 @@ export class EthereumAnchorService implements AnchorService {
     return [this._chainId]
   }
 
+  private _carFileFromRequestAnchorParams(params: RequestAnchorParams): CAR {
+    const carFactory = new CARFactory()
+    const car = carFactory.build()
+    car.put({...params})
```
oed, line 119: This doesn't look right to me. We should put each individual IPLD object into the car file. 

Should include:
* genesisCommit 
  * genesisCommit JWS (if signed)
  * genesisCommit CACAO (if created with session key)
* Actual commit:
  * Data object
  * JWS
  * CACAO (if present)
* metadata object, including the following info:
  * timestamp
  * streamtype number (can be used with genesis commit to create streamid

Then I'm not sure if maybe we want to sign the metadata object? Anyway, the metadata object should be set as `isRoot`.

pawartur: Thanks, @oed !

With @ukstv 's help, I fixed the structure of the CAR file.

Still not passing the actual commit, though, because @ukstv  and me agreed that:
* we wouldn't be using it for verification on CAS side in the foreseeable future
* when we do decide to check the signatures of the whole stream when creating the merkle tree, in general case would would most likely have to send more than just the anchored commit anyway. 

So it's gonna be best just to sent the genesis commit for now and then design the next step properly. Does this make sense, @oed ?

When it comes to signing the metadata, as a part of the CAS Auth work, @v-stickykeys has implemented signing of the whole request body for authenticated requests and we agreed that this will be good enough in the current stage, although we might want to optimise it for better performance in the future. WDYT, @oed ?

oed: Yes, integration with @v-stickykeys work makes sense. I'm not sure I follow the logic of not including the signed commit and the associated payload and CACAOs though? I thought this was spec'd out already?

oed: > why do you think we need to send the actual commit and a CACAO as well

They are needed to verify the signature of the commit on CAS. Without this signature verification a Ceramic node could spam CAS with a lot of fake commits for a stream, making historical indexing super slow because now syncing nodes would need to load a lot of commits.



It's worth noting that the genesis commit might be one ipld object (which you support in your code). It may also be a data object, a DagJWS object, and a CACAO object in the case of a signed genesis.

pawartur: > They are needed to verify the signature of the commit on CAS. 

Ok, let me sync with @ukstv and see, if this signature verification had been planned for this project and if not, how much additional time it would take.

On a side note, is it sufficient to send just the last commit in an anchor request, if CAS is supposed to verify it's signature? Won't CAS need the whole stream?

> It's worth noting that the genesis commit might be one ipld object (which you support in your code). It may also be a data object, a DagJWS object, and a CACAO object in the case of a signed genesis.

This is how I get the genesis payload before putting it in the CAR object: `const genesisPayload = await this.dispatcher._ipfs.block.get(state$.value.log[0].cid)`. Will this only give me unsigned genesis commits? Should I retrieve signed genesis commits in a different way?

oed: > if this signature verification had been planned for this project and if not, how much additional time it would take.

I'm fine with adding this later as long as we are aware that it would be a quite high priority project.

For future compatibility it would be nice if we sent along the entire commit being anchored, so that CAS could start verifying signatures at some point in the future!

> On a side note, is it sufficient to send just the last commit in an anchor request, if CAS is supposed to verify it's signature? Won't CAS need the whole stream?

If we make the assumption that the stream doesn't allow controller changes (e.g. MIDs and caip10-links) then we won't need the entire stream. If we do allow controller changes (like in TileDocuments) then yes the entire stream would need to be loaded. A tradeoff here could be to verify signatures for all streamtypes except Tiles.



 > Will this only give me unsigned genesis commits? Should I retrieve signed genesis commits in a different way?

This will either give you an unsigned genesis commit, or it will give you the DagJWS object. You can detect this if the loaded data has a `link` field.

pawartur: Ok, @oed , thanks for the responses :). I'm gonna add the anchored commit to the CAR file straight away and then we'll aim at adding the verification within this project or right after.

oed:
Going to follow up with a full review, but quick note based on PR overview is that we will also need "genesis commit CACAO block" if present.

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -125,24 +163,25 @@ export class EthereumAnchorService implements AnchorService {
...
     return defer(() =>
       from(
         this.sendRequest(this.requestsApiEndpoint, {
           method: 'POST',
-          body: {
-            streamId: params.streamID.toString(),
-            cid: params.tip.toString(),
-            timestamp: params.timestampISO,
-          },
+          headers: { 'Content-Type': 'multipart/form-data' },
```
oed, line 175:
```suggestion
          headers: { 'Content-Type': 'application/vnd.ipld.car' },
```

See: https://www.iana.org/assignments/media-types/media-types.xhtml

Also, according to the spec: https://www.iana.org/assignments/media-types/application/vnd.ipld.car
Looks like we should be sending it as binary, not string.

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -125,24 +163,25 @@ export class EthereumAnchorService implements AnchorService {
    * @param params - a RequestAnchorParams object
    * @private
    */
-  private _makeAnchorRequest(params: RequestAnchorParams): Observable<AnchorServiceResponse> {
+  private _makeAnchorRequest(
+    streamID: StreamID,
+    cid: CID,
```
oed, line 168:
```suggestion
    tip: CID,
```

For readability

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -111,6 +113,42 @@ export class EthereumAnchorService implements AnchorService {
...
+      const cacaoBlock = new CarBlock(params.tipCacaoCid, params.tipCacaoBlock)
+      car.blocks.put(cacaoBlock)
+    }
+
+    car.put(
+      {
+        timestamp: params.timestampISO,
+        streamId: params.streamId.toString(),
+        genesisCid: params.genesisCid.toString(),
+        genesisLinkCid: params.genesisLinkCid?.toString() ?? '',
```
oed, line 142: These CIDs don't need to be included in the root object. 
* `genesisCid` you can get from the `StreamID.fromBytes(streamId).cid`
* `genesisLinkCid` you can get from loading `car.get(genesisCid).link`, if present

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -111,6 +113,42 @@ export class EthereumAnchorService implements AnchorService {
...
+    }
+    if (params.tipCacaoCid) {
+      const cacaoBlock = new CarBlock(params.tipCacaoCid, params.tipCacaoBlock)
+      car.blocks.put(cacaoBlock)
+    }
+
+    car.put(
+      {
+        timestamp: params.timestampISO,
+        streamId: params.streamId.toString(),
```
oed, line 140:
```suggestion
        streamId: params.streamId.bytes,
```

More compact to just store the bytes directly. IPLD supports this.

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -111,6 +113,42 @@ export class EthereumAnchorService implements AnchorService {
...
+        genesisCid: params.genesisCid.toString(),
+        genesisLinkCid: params.genesisLinkCid?.toString() ?? '',
+        tipCid: params.tip.toString(),
```
oed, line 143:
```suggestion
        tip: params.tip,
```
IPLD supports storing CID types directly. 
It's implied that "tip" is a CID.

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -111,6 +113,42 @@ export class EthereumAnchorService implements AnchorService {
...
+        tipLinkCid: params.tipLinkCid?.toString() ?? '',
+        tipCacaoCid: params.tipCacaoCid?.toString() ?? '',
```
oed, line 145: These params are not required. 
* `tipLinkCid` you can get from loading `car.get(tip).link`
* `tipCacaoCid` you can get from loading `decodeHeader(car.get(tip)).cap`, if present

packages/core/src/state-management/state-manager.ts
```diff
@@ -323,12 +324,16 @@ export class StateManager {
...
+    const genesisCID = state$.value.log[0].cid
+    const genesisCommit = await this.dispatcher.retrieveCommit(genesisCID, state$.id)
+    await this._saveAnchorRequestForState(state$, genesisCommit)
+
+    const requestAnchorParams = await this._buildRequestAnchorParams(
+      state$,
+      genesisCID,
+      genesisCommit
+    )
+    const anchorStatus$ = this.anchorService.requestAnchor(requestAnchorParams)
```
oed, line 336: This code could use some clean up. I would suggest something like this:
```suggestion
    const requestCar = await this._buildAnchorRequestCarFromState(state$)
    
    const genesisCID = state$.value.log[0].cid
    const genesisCommit = requestCar.get(genesisCID)
    await this._saveAnchorRequestForState(state$, genesisCommit)

    const anchorStatus$ = this.anchorService.requestAnchor(requestCar)
```

This removes the need for the `RequestAnchorParams`. I don't believe gathering the params and creating the CAR file needs to be two different steps, in two different files. Hopefully this should make the code more readable!

oed (changes requested):
Suggesting some cleanup to remove some redundancy in the code. 

I realize that IPLD can be a bit tricky to work with if you are not familiar with it, so hopefully you find this helpful!

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -125,24 +162,25 @@ export class EthereumAnchorService implements AnchorService {
...
       from(
         this.sendRequest(this.requestsApiEndpoint, {
           method: 'POST',
-          body: {
-            streamId: params.streamID.toString(),
-            cid: params.tip.toString(),
-            timestamp: params.timestampISO,
-          },
+          headers: { 'Content-Type': 'application/vnd.ipld.car' },
+          body: carFile.bytes,
```
oed, line 175: Fwiw, I'm not sure if fetch works with Uint8Arrays. I've never actually used it to send binary data. 🤞

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -125,24 +162,25 @@ export class EthereumAnchorService implements AnchorService {
...
-  private _makeAnchorRequest(params: RequestAnchorParams): Observable<AnchorServiceResponse> {
+  private _makeAnchorRequest(
+    streamID: StreamID,
+    tip: CID,
+    carFile: CAR
+  ): Observable<AnchorServiceResponse> {
     return defer(() =>
       from(
         this.sendRequest(this.requestsApiEndpoint, {
           method: 'POST',
```
oed, line 173: How do we distinguish new vs old requests on CAS side?
Can this be done based on `Content-Type` perhaps?

pawartur: Yes, we'll be distinguishing by `Content-Type`.

packages/core/src/anchor/anchor-request-car-file-reader.ts
```diff
@@ -0,0 +1,25 @@
...
+  get timestamp(): Date {
+    return new Date(this.root.timestamp)
+  }
+
+  get streamId(): StreamID {
+    return StreamID.fromBytes(this.root.streamId)
+  }
+
+  get tip(): CID {
+    return toCID(this.root.tip)
```
oed, line 23: Isn't `root.tip` already of type CID?

packages/core/src/state-management/state-manager.ts
```diff
@@ -340,14 +342,69 @@ export class StateManager {
     return this._processAnchorResponse(state$, anchorStatus$)
   }
 
-  private async _saveAnchorRequestForState(state$: RunningState): Promise<void> {
+  private async _buildAnchorRequestCARFile(streamId: StreamID, tip: CID): Promise<CAR> {
+    const carFactory = new CARFactory()
+    carFactory.codecs.add(DAG_JOSE)
```
oed, line 347: Not super important but why do you create a new factory every time you call this function? You could just have a global factory object.

packages/core/src/state-management/state-manager.ts
```diff
@@ -340,14 +342,69 @@ export class StateManager {
...
+        tip: tip.bytes,
+      },
+      { isRoot: true }
+    )
+
+    // Genesis block
+    const genesisCid = streamId.cid
+    const genesisBlock = await this.dispatcher._ipfs.block.get(genesisCid)
+    const genesisCarBlock = new CarBlock(genesisCid, genesisBlock)
+    car.blocks.put(genesisCarBlock)
```
oed, line 365: You do this lots of times in this method. Why not create a helper function? 

```ts

```



```suggestion
    const cidToBlock = cid => new CarBlock(cid, await this.dispatcher._ipfs.block.get(cid))

    car.blocks.put(cidToBlock(streamId.cid))
```

packages/core/src/state-management/state-manager.ts
```diff
@@ -323,12 +325,12 @@ export class StateManager {
...
-      streamID: state$.id,
-      tip: state$.tip,
-      timestampISO: (new Date()).toISOString(),
-    })
+    const genesisCID = state$.value.log[0].cid
+    const genesisCommit = await this.dispatcher.retrieveCommit(genesisCID, state$.id)
+    await this._saveAnchorRequestForState(state$, genesisCommit)
+
+    const carFile = await this._buildAnchorRequestCARFile(state$.id, state$.tip)
+    const anchorStatus$ = this.anchorService.requestAnchor(carFile)
```
oed, line 333: Looks like you will be retrieving the genesis commit from IPFS twice. Better just get it from the car file the second time.
```suggestion
    const carFile = await this._buildAnchorRequestCARFile(state$.id, state$.tip)
    const genesisCID = state$.value.log[0].cid
    const genesisCommit = carFile.get(genesisCID)
    await this._saveAnchorRequestForState(state$, genesisCommit)

    const anchorStatus$ = this.anchorService.requestAnchor(carFile)
```

packages/core/src/anchor/anchor-request-car-file-reader.ts
pawartur: Nope, `root.tip` is a byte array right now. Can we leave it like this?

packages/core/src/state-management/state-manager.ts
```diff
@@ -340,14 +346,57 @@ export class StateManager {
...
+  private async _buildAnchorRequestCARFile(streamId: StreamID, tip: CID): Promise<CAR> {
+    const car = this.carFactory.build()
+
+    // Root block
+    const timestampISO = new Date().toISOString()
+    car.put(
+      {
+        timestamp: timestampISO,
+        streamId: streamId.bytes,
+        tip: tip.bytes,
```
oed, line 358:
```suggestion
        tip: tip,
```

Please just use the CID directly here. IPLD does support putting CIDs directly, I thought I made this clear in a previous comment? 

The neat thing here is that (at least in ipfs) you can do things like `ipfs.dag.get(rootCid, { path: '/tip/link/id' }` to traverse the dag (and in this example case retrieve the genesis commit).

pawartur: Yeah, you had already mentioned that we should have stored CID's directly, it slipped my mind, sorry.

packages/core/src/state-management/state-manager.ts
```diff
@@ -340,14 +346,57 @@ export class StateManager {
...
+
+    // Genesis block
+    const genesisCid = streamId.cid
+    car.blocks.put(await cidToBlock(genesisCid))
+
+    // Tip block
+    car.blocks.put(await cidToBlock(tip))
+
+    // Genesis Link Block
+    const genesisCommit = await this.dispatcher.retrieveCommit(genesisCid, streamId)
```
oed, line 373: Why not this? Worth trying to minimize the possible calls out to IPFS as they can require a http request to kubo.
```suggestion
    const genesisCommit = car.get(genesisCid)
```

packages/core/src/state-management/state-manager.ts
```diff
@@ -340,14 +346,57 @@ export class StateManager {
...
+    if (StreamUtils.isSignedCommit(genesisCommit)) {
+      car.blocks.put(await cidToBlock(genesisCommit.link))
+    }
+
+    // Tip Link Block
+    const tipCommit = await this.dispatcher.retrieveCommit(tip, streamId)
```
oed, line 379: Same thing here

oed (commented):
lgtm. I assume we don't want to merge this to develop before we have Clay running a version of CAS that can process these requests?

pawartur:
> lgtm. I assume we don't want to merge this to develop before we have Clay running a version of CAS that can process these requests?

Yes, that's correct. It's blocked by merging and deploying https://github.com/ceramicnetwork/ceramic-anchor-service/pull/897

oed:
@pawartur doesn't look like anything has changed since I last reviewed. I'm good with this when @ukstv approves

pawartur:
> @pawartur doesn't look like anything has changed since I last reviewed. I'm good with this when @ukstv approves

Right, @oed ! That's because the last time you wrote this :) :

> lgtm. I assume we don't want to merge this to develop before we have Clay running a version of CAS that can process these requests?

The CAS part has been merged already, that's why I re-requested your review. Sorry, should have clarify this yesterday :).

Will ask @ukstv for a review now, thanks!

ukstv (approved):

oed (approved):

packages/common/src/utils/http-utils.ts
```diff
@@ -24,9 +24,11 @@ export enum HttpMethods {
 
 export async function fetchJson(url: URL | string, opts: FetchOpts = {}): Promise<any> {
   if (opts.body) {
+    const headers = { 'Content-Type': 'application/json', ...opts.headers }
+
     Object.assign(opts, {
-      body: JSON.stringify(opts.body),
-      headers: { ...opts.headers, 'Content-Type': 'application/json' },
+      body: headers['Content-Type'] == 'application/json' ? JSON.stringify(opts.body) : opts.body,
```
v-stickykeys, line 30:
```suggestion
      body: headers['Content-Type'].includes('application/json') ? JSON.stringify(opts.body) : opts.body,
```
remember that content-type header can include encoding or other things at the end. like
```
Content-Type: text/html; charset=utf-8
Content-Type: application/json; charset=utf-8
Content-Type: multipart/form-data; boundary=something
```

PR #2646, opened 2023-01-10
feat(core): create rebuild anchor worker

stephhuynh18:
Also locally tested using a real merkle root from QA. I can successfully recreate the anchor commits, and apply them so that they get indexed.


packages/core/src/sync/workers/rebuild-anchor.ts
```diff
@@ -0,0 +1,95 @@
...
+  private async getModelForStream(streamId: StreamID): Promise<StreamID | null> {
+    const signedCommit = await this.ipfsService.retrieveCommit(
+      streamId.cid, // genesis commit CID
+      streamId
+    )
+
+    const genesisCommit = await this.ipfsService.retrieveCommit(signedCommit.link, streamId)
+
+    if (!genesisCommit.header.model) {
+      return
```
PaulLeCam, line 32: Seems this should return `null` here to match the method type definition?

packages/core/src/sync/workers/rebuild-anchor.ts
```diff
@@ -0,0 +1,95 @@
...
+    const metadata = await merkleTreeLeafLoader.getMetadata().catch(() => {
+      // TODO: add failure job for root cid
+    })
+    if (!metadata) {
+      return
+    }
+
+    const streams = metadata.streamIds
+
+    for (let i = 0; i < streams.length; i++) {
```
PaulLeCam, line 66: Does this need to be sequential or could this logic run in parallel please?

packages/core/package.json
```diff
@@ -41,6 +41,7 @@
     "clean": "npx rimraf ./lib"
   },
   "dependencies": {
+    "@ceramicnetwork/anchor-utils": "^1.1.0-rc.0",
```
PaulLeCam, line 44: Could it be set in `devDependencies` if the package is only used in tests please?

PaulLeCam (approved):

packages/core/src/sync/workers/rebuild-anchor.ts
stephhuynh18: very right i have changed it to run in parallel thanks!

PR #2647, opened 2023-01-10
fix(stream-tile, stream-model-instance): Enforce controller must be a string

stbrody:


packages/stream-tile-handler/src/tile-document-handler.ts
```diff
@@ -96,6 +96,11 @@ export class TileDocumentHandler implements StreamHandler<TileDocument> {
     if (!(payload.header.controllers && payload.header.controllers.length === 1)) {
       throw new Error('Exactly one controller must be specified')
     }
+    if (!StreamUtils.validDIDString(payload.header.controllers[0])) {
+      throw new Error(
+        `Attempting to create a TileDocument with an invalid DID string: ${payload.header.controllers[0]}`
```
oed, line 101: Should we add this to `stream-model-handler` as well?

oed (approved):
lgtm,
Still room for invalid DIDs but probably close enough.

If we feel like it here's a regex for DIDs:
https://github.com/decentralized-identity/did-resolver/blob/master/src/resolver.ts#L249

packages/stream-tile-handler/src/tile-document-handler.ts
stbrody: nope, Model genesis commits are always signed, so you get the check that the DID is valid for free from signature verification

PR #2651, opened 2023-01-11
chore(stream-tests): Test that you cannot index a non-model streamid

stbrody:


smrz2001:
Hey @stbrody, IIUC, this test isn't needed since we're already testing what it's doing, is that right?

stbrody:
@smrz2001 I'm sorry, I don't understand what you mean by "since we're already testing what it's doing", can you elaborate?

As far as I know we don't have any existing tests for this, which is why I put up this PR

smrz2001:
> @smrz2001 I'm sorry, I don't understand what you mean by "since we're already testing what it's doing", can you elaborate?
> 
> As far as I know we don't have any existing tests for this, which is why I put up this PR

Oh ok, I misunderstood one of your Discord messages and thought that you realized after writing this test that this was already covered by existing tests. My bad.

smrz2001 (approved):
🚢

PR #2655, opened 2023-01-17
feat(job): add ability to update a job's data

stephhuynh18:
Allows one to keep track of a job "state"


packages/core/src/state-management/job-queue.ts
```diff
@@ -1,10 +1,19 @@
-import { default as PgBoss } from 'pg-boss'
+import { default as PgBoss, type SendOptions } from 'pg-boss'
 import Pg from 'pg'
 import { fromEvent, firstValueFrom } from 'rxjs'
 
+export type Job = {
+  name: string
+  data: object
```
PaulLeCam, line 7: Any way we can refine the data needed by a job please?
If it differs between jobs, maybe we can make the `Job` type generic with a type parameter for the `data`? Or if the `data` is based on the job's `name`, can we define the mapping to benefit from the type validation?

packages/core/src/state-management/job-queue.ts
```diff
@@ -74,6 +107,22 @@ export class JobQueue implements IJobQueue {
     await this.dbConnection.end()
   }
 
+  /**
+   * Updates the job data for a particular job
+   * @param jobId id of the job
+   * @param data data to update
+   * @returns promise that results to a boolean that is true of the
+   */
+  async updateJob(jobId: string, data: object): Promise<void> {
```
PaulLeCam, line 116: What happens if we try to update a job that is already running please?

packages/core/src/state-management/job-queue.ts
```diff
@@ -74,6 +107,22 @@ export class JobQueue implements IJobQueue {
     await this.dbConnection.end()
   }
 
+  /**
+   * Updates the job data for a particular job
+   * @param jobId id of the job
+   * @param data data to update
+   * @returns promise that results to a boolean that is true of the
```
PaulLeCam, line 114: There seems to be a mismatch between this comment and return type

PaulLeCam (approved):

packages/core/src/state-management/job-queue.ts
stephhuynh18: From my tests it does not affect an already running job as the data is read at the beginning of the job

PaulLeCam (approved):
Thanks for the changes!

PR #2657, opened 2023-01-17
fix(cli): Export S3Store type

stbrody:
This is critical to fix the integration tests, as they cannot run the newest version of Ceramic currently, as they cannot build Ceramic with the S3 state store as the S3 store implementation moved and stopped being exported in https://github.com/ceramicnetwork/js-ceramic/commit/2754643e7503a9f5103cc5af14d1614a3ee07245#diff-2fb448d11702324cc59d657d5945d59cd325540eb87b71f2e66a5bc01ada834c


pawartur (commented):
Thanks :) !

PR #2658, opened 2023-01-18
feat(core): store indexing sync state

PaulLeCam:
Still WIP as more pieces need to be plugged-in and tested, but can you have a look please @stephhuynh18 to let me know if it matches what you have in mind and how to plug the workers logic? Thanks!


PaulLeCam:
@stephhuynh18 should be ready for review now if you can have a look please?
Only open question is setting the initial block number to do a full sync, do we already know what block number to use please?
Thanks!

stephhuynh18 (approved):
LGTM

Also we do not know the number yet. It will probably be a number based on chainId (that can be changed now or later) I'll add a ticket to figure this out.

PR #2668, opened 2023-01-26
feat(core) CDB-1655: Verify Indices

dbcfd:
Add checks when a node starts up to verify that indices are present. Added 2 additional tests to cover behavior.

Additionally fix code that was using too long of a name, switching from tableName to indexName.


dbcfd:
closing for https://github.com/ceramicnetwork/js-ceramic/pull/2672

PR #2670, opened 2023-01-26
Make the s3 endpoint and optional parameter and change its name to make it clearer

pawartur:
## How Has This Been Tested?

- [X] Tested manually that S3 state store works

## PR checklist


pawartur:
@stbrody , I'm not very happy with this change, because I needed to update the order of arguments, which may cause some existing code to break. But I didn't see how I could make it better without a bigger refactoring (which may still be a better idea...). Curious about your thoughts? Do you see a simple and safer way of fixing this?

stbrody (approved):
LGTM.  I don't think anyone is building manually against the CLI package so I think a breaking change to the API here is probably fine.  Thanks!

PR #2673, opened 2023-01-26
feat(core): Resumes active jobs on start of job queue

stephhuynh18:


packages/core/src/state-management/job-queue.ts
```diff
@@ -37,22 +38,45 @@ export class JobQueue<T extends Record<any, any>> implements IJobQueue<T> {
...
 
     this.queue = new PgBoss({ db: new PgWrapper(this.dbConnection) })
+    this.queue.on('error', (err) => {
+      this.logger.err(`Error received by queue: ${err}`)
+    })
+  }
+
+  async _getActiveJobIds(): Promise<string[]> {
+    const result = await this.dbConnection.query(
+      `SELECT id FROM pgboss.job WHERE state = 'active' and name IN (${['job1', 'job2']
```
PaulLeCam, line 54: Are `job1` and `job2` constants here please? Can't jobs have other names?

PaulLeCam (approved):
LGTM just wondering about the job names here

packages/core/src/state-management/job-queue.ts
stephhuynh18: yea good catch i missed these when i cleaning up (its supposed to be this.jobs)

PR #2674, opened 2023-01-30
chore(http-client): ensure client can be bundled by webpack

PaulLeCam:
I noticed the HTTP client can't be bundled with webpack because of some deep dependency, which is likely to affect lots of users.
This change adds a step to the `test` script to attempt to bundle the HTTP client with webpack, which should help surface and address the current issue, and avoid similar issues to happen later.


willex (approved):
LGTM - thanks for surfacing!

PR #2679, opened 2023-01-30
feat(core): handle blocks reorganizations when creating sync jobs

PaulLeCam:


packages/core/src/sync/sync-api.ts
```diff
@@ -137,18 +142,28 @@ export class SyncApi implements ISyncApi {
...
-   * @param block
+   * @param event: BlockProofsListenerEvent
    */
-  async _handleBlock(block: Block): Promise<void> {
+  async _handleBlockProofs(event: BlockProofsListenerEvent): Promise<void> {
+    const { block, reorganized } = event
+
+    let fromBlock = block.number
+    if (reorganized) {
+      const previousBlock = await this.provider.getBlock(event.expectedParentHash)
```
stephhuynh18, line 159: Ex. previousHash = block500Hash 
We receive block501. There was a reorg and block501.parent !== previousHash.

If we assume block501 is a block on the main chain, than it's parent will also be a block the main chain. This means that block500Hash would not exist anymore and `fromBlock` would end up being `block.number` (501 in this case). 

We also do not know how far back the reorg happened. In this example the fork could have happened at 490. The most correct way to handle this would be to keep track of all block hashes we have seen and walk back from block501 until we find one that matches one we seen. But we aren't keeping track of all block hashes and this is probably overkill. 

So the easy solution would be to just sync the last X blocks. X can be a constant we set, and can make configurable if needed.

PaulLeCam: Sounds good, thanks for the suggestion!
How many blocks should we go back to then please? Should we use the same value as the number of confirmations?

stephhuynh18: I think number of confirmations should be fine. We can always change it later

PaulLeCam: OK thanks! That's what I used in the latest change if you can have a look please?

stephhuynh18 (approved):

PR #2682, opened 2023-01-30
bug(core) CDB-833: Retry Getting Transaction

dbcfd:
Brings in exponential backoff package to provide exponential backoff behavior when we fail to get a transaction from the anchor. Adds 1 test and updates 1 test to verify this behavior.


packages/core/src/anchor/ethereum/ethereum-anchor-validator.ts
```diff
@@ -171,34 +172,42 @@ export class EthereumAnchorValidator implements AnchorValidator {
...
-        if (!this.ethereumRpcEndpoint) {
-          throw new Error(
-            `Failed to load transaction data for transaction ${txHash}. Try providing an ethereum rpc endpoint`
-          )
-        } else {
-          throw new Error(`Failed to load transaction data for transaction ${txHash}`)
+      const transaction: TransactionResponse = await backOff(async () => {
+        const tx = await this._getTransaction(provider, txHash)
+        if (!tx) {
+          if (!this.ethereumRpcEndpoint) {
```
dbcfd, line 178: @stbrody Due to lack of familiarity, can this endpoint change, or should this check be pulled out of the backoff code?

stbrody: nope, can't change in a running instance, requires a restart to update

packages/core/src/anchor/ethereum/ethereum-anchor-validator.ts
```diff
@@ -171,34 +172,42 @@ export class EthereumAnchorValidator implements AnchorValidator {
...
+            throw new Error(
```
stbrody, line 179: how does the `backOff` function handle if an error is thrown?  Ideally we want to log these errors when they are thrown even if it succeeds on retry

packages/core/src/anchor/ethereum/ethereum-anchor-validator.ts
```diff
@@ -171,34 +172,42 @@ export class EthereumAnchorValidator implements AnchorValidator {
...
-
-      if (!transaction) {
-        if (!this.ethereumRpcEndpoint) {
-          throw new Error(
-            `Failed to load transaction data for transaction ${txHash}. Try providing an ethereum rpc endpoint`
-          )
-        } else {
-          throw new Error(`Failed to load transaction data for transaction ${txHash}`)
+      const transaction: TransactionResponse = await backOff(async () => {
+        const tx = await this._getTransaction(provider, txHash)
```
stbrody, line 176: could we make helper functions for the bodies of the two backOff statements to improve readability?

stbrody (commented):
note tests are failing and need fixing

packages/core/src/anchor/ethereum/ethereum-anchor-validator.ts
```diff
@@ -171,34 +172,42 @@ export class EthereumAnchorValidator implements AnchorValidator {
...
-      const transaction: TransactionResponse = await this._getTransaction(provider, txHash)
-
-      if (!transaction) {
-        if (!this.ethereumRpcEndpoint) {
-          throw new Error(
-            `Failed to load transaction data for transaction ${txHash}. Try providing an ethereum rpc endpoint`
-          )
-        } else {
-          throw new Error(`Failed to load transaction data for transaction ${txHash}`)
+      const transaction: TransactionResponse = await backOff(async () => {
```
stbrody, line 175: we probably want to also set some limits on the backoff behavior.  it looks like there are arguments we can pass to limit the max number of retries and/or how long it retries for

packages/core/package.json
```diff
@@ -65,6 +65,7 @@
     "ajv-formats": "^2.1.1",
     "await-semaphore": "^0.1.3",
     "dids": "^3.4.0",
+    "exponential-backoff": "^3.1.0",
```
stbrody, line 68: Adding @ukstv for any thoughts on the quality of this package and it's a good dependency to add.

stbrody:
@dbcfd what's the status of this PR?  Should we close it?

packages/core/src/anchor/ethereum/ethereum-anchor-validator.ts
dbcfd: Backoff by default is 10 tries. Do we want to lower that?

PR #2683, opened 2023-01-31
feat(core): Limit how many pin requests can be sent to ipfs at once

stbrody:


packages/core/src/store/pin-store.ts
```diff
@@ -5,18 +5,29 @@ import { RunningState } from '../state-management/running-state.js'
 import { Model } from '@ceramicnetwork/stream-model'
 import { IKVStore } from './ikv-store.js'
 import { StreamStateStore } from './stream-state-store.js'
+import { Semaphore } from 'await-semaphore'
+
+const DEFAULT_CONCURRENT_PIN_LIMIT = 10
```
stbrody, line 10: not really sure what's a reasonable default here.  @v-stickykeys do you have any intuition here?

ukstv (approved):

PR #2692, opened 2023-02-07
feat(core) CDB-1655: Verify Indices

dbcfd:
Add checks when a node starts up to verify that indices are present.

Additionally fix code that was using too long of a name, switching from tableName to indexName.


stbrody:
@dbcfd please make sure you are running `prettier` to auto-format all your code before putting up for review.  Most editors/IDEs have a built-in integration to run prettier on save, which I'd recommend setting up

willex (approved):
LGTM!

PR #2693, opened 2023-02-07
feat(core): enable syncing on ceramic start

stephhuynh18:
Notes:
I attempted to make the merkle files into its own package but the serialization/deserialization used in the CAS makes this a little difficult. So i am using a simplified version very similar to the one found in the CAS in preparation to replace it.

I also have placed this in core/sync because I have mocked quite a few things. 
- provider because i don't want to have to launch a ganache and deploy a contract
- in memory anchor service which does not make merkle trees or use a provider to emit blocks

I am also temporarily using a passed in config var instead of retrieving the config var from the db as Alex is working on this
The providersCache is ripped directly out of the Ethereum Anchor Validator. I will consolidate in a future PR

If this is too much... oh dear. I tried. I can try again


packages/core/src/ceramic.ts
```diff
@@ -922,6 +952,7 @@ export class Ceramic implements CeramicApi {
    */
   async close(): Promise<void> {
     this._logger.imp('Closing Ceramic instance')
+    await this.syncApi.shutdown()
```
PaulLeCam, line 955: Does the sync API need to be shutdown first please?
If not, maybe the synchronous calls can be done first, then the async ones? And possibly in parallel if order doesn't matter?

packages/core/src/local-admin-api.ts
```diff
@@ -1,17 +1,17 @@
...
  */
 export class LocalAdminApi implements AdminApi {
-  constructor(
-    private readonly indexApi: LocalIndexApi
-  ) {}
+  constructor(private readonly indexApi: LocalIndexApi, private readonly syncApi: SyncApi) {}
 
   async startIndexingModels(modelsIDs: Array<StreamID>): Promise<void> {
     await this.indexApi.indexModels(modelsIDs)
+    await this.syncApi.startModelSync(modelsIDs)
```
PaulLeCam, line 14: Could these be called in parallel please?

packages/core/src/sync/sync-api.ts
```diff
@@ -186,21 +198,37 @@ export class SyncApi implements ISyncApi {
    * Also keeps in sync with new anchors.
    */
   async startModelSync(
-    startBlock: number,
-    endBlock: number,
-    models: string | string[]
+    models: StreamID | StreamID[],
```
PaulLeCam, line 201: Do these need to be stream IDs please?
Looks like they get converted to string, so couldn't the input be a string directly?

packages/core/src/sync/sync-api.ts
```diff
@@ -186,21 +198,37 @@ export class SyncApi implements ISyncApi {
...
+    startBlock = INITIAL_INDEXING_BLOCK,
+    endBlock?
   ): Promise<void> {
-    const modelIds = Array.isArray(models) ? models : [models]
+    if (!this.syncConfig.on) return
```
PaulLeCam, line 205: Maybe this should throw an error considering this method shouldn't be called in the first place if sync is off, no?

PaulLeCam (approved):
Would definitely need a larger refactor to avoid putting so much in core but the problem is not caused by this PR so I think it's OK to move ahead with these changes.

packages/core/src/sync/sync-api.ts
stephhuynh18: This function is being called from the AdminApi where it has no concept if sync should be on or not, similar to how local index api is used. 

Adding a model to the index currently does not throw even if  `process.env.CERAMIC_ENABLE_EXPERIMENTAL_COMPOSE_DB != 'true'`. I am maintaining that consistency for now.

I am not sure what an admin api user is going to expect or what the cli(?) user would expect.

PaulLeCam: OK thanks, indeed if the method callers don't do these checks better have this check here.

PR #2698, opened 2023-02-10
Prettier Check

dbcfd:
Adds a prettier check to CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,6 +26,7 @@ commands:
       - node/install-packages
       - run: npm run build
       - run: npm run lint
+      - run: npx prettier --check "packages/**/src/**/*.ts"
       - run:
           name: Test
           command: npm run test
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,7 +26,6 @@ commands:
       - node/install-packages
       - run: npm run build
       - run: npm run lint
-      - run: npx prettier --check "packages/**/src/**/*.ts"
       - run:
           name: Test
           command: npm run test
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -26,6 +26,7 @@ commands:
       - node/install-packages
       - run: npm run build
       - run: npm run lint
+      - run: npx prettier --check "packages/**/src/**/*.ts"
```
stbrody, line 29: I wonder if we should add this to the `lint` npm script?  @ukstv in case he has an opinion here.

ukstv: Yup. I am all for prettier. It should be part of `lint` npm script, so that a developer gets it automatically, without any extra commands.

ukstv: It could be `npx lerna run lint && npx prettier ...` inside root `package.json`.

ukstv (changes requested):
Good intent. Adds a bit friction to DX though in the current iteration: one more command to run (even if IDE _usually_ handles prettier for you). It's better to add prettier to existing `npm run lint` script.

Usually devs in JS ecosystem are well aware of `npm run lint`, and run it without any additional thoughts. If you hide prettier invocation in the lint script, it would be obvious for absolute most of the devs how to run it.

package.json
```diff
@@ -16,7 +16,7 @@
     "publish:release-candidate": "npx lerna publish --dist-tag next --preid rc --no-verify-access --conventional-prerelease bump preminor",
     "publish:release": "npx lerna publish --dist-tag latest --create-release github --no-verify-access --conventional-graduate bump minor",
     "publish:hotfix": "npx lerna publish --no-verify-access --conventional-graduate bump patch",
-    "lint": "npx lerna run lint",
+    "lint": "npx lerna run lint && npx prettier --check packages/**/src/**/*.ts",
```
dbcfd, line 19: @ukstv Added check here now and removed it from ci

ukstv (approved):

PR #2706, opened 2023-02-13
Update local development docs to specify how to setup with pyenv

dbcfd:


willex (approved):
Thanks for updating this!

PR #2713, opened 2023-02-14
feat(cli,http-client): Add node status admin API endpoint skeleton

stbrody:
No real functionality yet (beyond the admin api authorization), just returns an empty response body.

### DIFF
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -324,6 +324,7 @@ export class CeramicDaemon {
     const collectionRouter = ErrorHandlingRouter(this.diagnosticsLogger)
     const adminCodesRouter = ErrorHandlingRouter(this.diagnosticsLogger)
     const adminModelRouter = ErrorHandlingRouter(this.diagnosticsLogger)
+    const adminNodeStatusRouter = ErrorHandlingRouter(this.diagnosticsLogger)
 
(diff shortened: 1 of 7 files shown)


packages/cli/src/ceramic-daemon.ts
```diff
@@ -336,6 +337,7 @@ export class CeramicDaemon {
     baseRouter.use('/collection', collectionRouter)
     baseRouter.use('/admin/getCode', adminCodesRouter)
     baseRouter.use('/admin/models', adminModelRouter)
+    baseRouter.use('/admin/nodeStatus', adminNodeStatusRouter)
```
stbrody, line 340: I thought about putting this under the existing `node` router so the endpoint would be `node/status` instead of `admin/nodeStatus`.  But right now all the functionality that is gated by a JWS from an admin DID is under the `admin` router so I decided to keep with that pattern for now.  Don't feel strongly though and would be happy to move to `/node/status` if that seems better.

@oed @PaulLeCam

oed: How about `/admin/status`?

pawartur (approved):

packages/cli/src/ceramic-daemon.ts
stbrody: I'll do in the next PR to avoid merge conflicts

PR #2715, opened 2023-02-14
feat(core): initial nodeStatus output

stbrody:
Has the most basic information we want to start returning, just to have a starting point. There's lots more stuff we could add into this over time.  Notably this is still missing any information about composedb or the running historical syncs.

builds on https://github.com/ceramicnetwork/js-ceramic/pull/2713


packages/core/src/local-admin-api.ts
```diff
@@ -1,17 +1,22 @@
...
+
 /**
  * AdminApi for Ceramic core.
  */
 export class LocalAdminApi implements AdminApi {
-  constructor(private readonly indexApi: LocalIndexApi, private readonly syncApi: SyncApi) {}
+  constructor(
+    private readonly indexApi: LocalIndexApi,
+    private readonly syncApi: SyncApi,
+    private readonly nodeStatusFn: NodeStatusFn // TODO: circular dependency back into Ceramic
```
pawartur, line 15: Can we add a ticket to track this todo?

pawartur (approved):
LGTM with a couple of minor comments, thanks :) !

packages/common/src/node-status-interface.ts
```diff
@@ -0,0 +1,53 @@
...
+export interface AnchorNodeStatus {
+  /**
+   * The URL of the Ceramic Anchor Service used to request anchors.
+   */
+  anchorServiceUrl: string
+
+  /**
+   * The endpoint used to validate anchor transactions (generally this is an ethereumRpcEndpoint).
+   */
+  anchorValidationEndpoint: string | null
```
oed, line 52: Can we just call this `ethereumRpcEndpoint` for clarity?

packages/common/src/node-status-interface.ts
```diff
@@ -0,0 +1,53 @@
...
+export interface IpfsNodeStatus {
+  /**
+   * PeerId of the connected ipfs node
+   */
+  peerId: string
+
+  /**
+   * IPFS Swarm multiaddrs of the connected ipfs node
+   */
+  addresses: Array<string>
```
oed, line 40: Nice

oed (approved):

packages/common/src/node-status-interface.ts
```diff
@@ -0,0 +1,53 @@
...
+}
+
+export interface AnchorNodeStatus {
```
oed, line 43:
```suggestion
export interface AnchorNodeStatus {
  /**
   * The ethereum chainId used for anchors.
   */
  chainId: string
```

What about adding this as well?

packages/common/src/node-status-interface.ts
stbrody: Yeah we can.  We have all these interfaces and indirection layers around the anchor service and anchor validator that are designed to make it blockchain agnostic and able to support other chains in the future, so exposing the ethereumRpcEndpoint in the generic AnchorValidator class felt like a layer violation.  But if realistically we're never going to support another chain I can go ahead and do that.

oed: Having a Ceramic node validate anchors on multiple chains seems problematic since this would require a node operator to run a lot of different nodes. I think for simplicity and decentralization we should stick to using one chain per ceramic network.

By decentralization here I mean that if you need to run chain1-node, chain2-node, and ceramic-node, there is less people that can feasibly do this which reduces the decentralization of the network.

packages/common/src/anchor-service.ts
```diff
@@ -95,6 +95,11 @@ export interface AnchorService {
  * Describes behavior for validation anchor commit inclusion on chain
  */
 export interface AnchorValidator {
+  /**
+   * The rpc endpoint used to validate anchor transactions.
+   */
+  validationRpcEndpoint: string | null
```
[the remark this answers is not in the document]
stbrody: not necessarily, it might just be using the default provider from the ethers.js library (which is rate limited) instead of a dedicated ethereum rpc endpoint.

packages/common/src/node-status-interface.ts
```diff
@@ -0,0 +1,53 @@
+/**
+ * A JSON object containing various diagnostic and introspection information from a running
+ * Ceramic node.
+ */
+export interface NodeStatusResponse {
+  /**
+   * A random UUID that is generated each time a node starts up.  Can be used to detect when
+   * a node restarts.
+   */
+  runId: string
```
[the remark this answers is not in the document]
stbrody: That's a good idea, I'll add an `uptimeMs` field

packages/common/src/node-status-interface.ts
stbrody: okay will change

packages/common/src/anchor-service.ts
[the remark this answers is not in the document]
pawartur: Thanks! Could you add it this short explanation to the comment in line 99?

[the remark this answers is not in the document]
pawartur: Ah, you've already done it :). Sorry for not checking and thanks!

pawartur (approved):

PR #2724, opened 2023-02-17
feat(core) CDB-1630: Error on index query if sync hasn't completed

dbcfd:


dbcfd:
@stephhuynh18 @stbrody This should be ready for final review now.

packages/core/src/indexing/database-index-api.ts
```diff
@@ -201,17 +208,26 @@ export class DatabaseIndexApi<DateType = Date | number> {
...
+    if (foundModelToIndex == undefined) {
       const err = new Error(`Query failed: Model ${model} is not indexed on this node`)
       this.logger.debug(err)
       throw err
+    } else {
+      await this.assertModelIndexable(modelStreamId)
+    }
+  }
+
+  async assertModelIndexable(modelStreamId: StreamID | string): Promise<void> {
```
stbrody, line 225: I don't really understand the division in names between `assertModelQueryable` and `assertModelIndexable`.  This function only runs on models that are already being actively indexed and just checks whether queries should be allowed through or not, so seems weird to me to call it assertModelsIndexable when the caller (assertModelsQueryable) has already asserted that the model is actually actively indexed

packages/core/src/indexing/build-indexing.ts
```diff
@@ -2,6 +2,7 @@ import { type DatabaseIndexApi, PostgresIndexApi, SqliteIndexApi } from './datab
 import knex from 'knex'
 import { DiagnosticsLogger, Networks } from '@ceramicnetwork/common'
 import * as fs from 'fs'
+import { ISyncQueryApi } from '../sync/interfaces.js'
```
stephhuynh18, line 5: What is this used for?

packages/core/src/sync/interfaces.ts
```diff
@@ -10,7 +10,11 @@ export type SyncConfig = {
   chainId: SupportedNetwork
 }
 
-export interface ISyncApi {
+export interface ISyncQueryApi {
```
stbrody, line 13: can we add some comments for what these interfaces represent?

stbrody (approved):
LGTM mod some nitpicks on naming.

Would still like @stephhuynh18 to check that the way `syncComplete` is checking whether the historical sync has completed is correct

packages/core/src/ceramic.ts
```diff
@@ -302,6 +302,7 @@ export class Ceramic implements CeramicApi {
       this.repository.index,
       this._logger
     )
+    this.repository.index.setSyncQueryApi(this.syncApi)
```
stephhuynh18, line 305: There is a way to get rid of the circular dependency: 

`SyncApi` only uses the indexApi to retrieve the indexable models. We can pass the indexable models in `SyncApi.init()`

packages/core/src/sync/sync-api.ts
```diff
@@ -265,6 +265,10 @@ export class SyncApi implements ISyncApi {
     })
   }
 
+  async syncComplete(model: string): Promise<boolean> {
+    return !this.modelsToSync.has(model)
```
stephhuynh18, line 269: Unfortunately figuring out which models are currently syncing is a bit more complicated. There are two types of syncs: 1. continuous sync, 2. historic sync. 

We want to prevent queries if there are historic syncs going on for that model. The continuous sync is always running for all indexed models

packages/core/src/initialization/make-index-api.ts
```diff
@@ -2,6 +2,7 @@ import type { DiagnosticsLogger } from '@ceramicnetwork/common'
 import { Networks } from '@ceramicnetwork/common'
 import { buildIndexing, IndexingConfig } from '../indexing/build-indexing.js'
 import { DatabaseIndexApi, SqliteIndexApi } from '../indexing/database-index-api.js'
+import { ISyncQueryApi } from '../sync/interfaces.js'
```
stephhuynh18, line 5: Where is this used?

packages/core/src/ceramic.ts
stbrody: ooh good call.  We'd also have to update it any time the set of indexed models changes, but that should be doable and would indeed resolve the circular dependency.  Probably fine to leave for a follow-up though?

packages/core/src/sync/sync-api.ts
stbrody: Any ideas on how we could test this change effectively so a nuance like this would be caught?

stephhuynh18: We can check the active/pending jobs and see if the models is included, we can keep a running list of models that having pending history syncs and update when syncs are done. We can have a table with previous sync information for each model.

stephhuynh18: Easiest thing to do is to not allow ANY queries if history syncs are running. So basically if there are any active or pending history sync jobs

stbrody: I think disallowing all queries could cause issues.  Imagine an app has been running a while with some set of indexed models.  Now it wants to start using a new model, so it adds that model and starts the historical sync, but doesn't update the app logic to use the new model until the sync is complete.  But now the whole app will fail for the queries on all the existing models that were already working.

packages/core/src/indexing/build-indexing.ts
dbcfd: Unused, should probably make that an error in lint. Originally was plumbing through the dependency.

packages/core/src/ceramic.ts
dbcfd: That actually only removes one of the circular dependencies. The main one is from 

`this.repository.stateManager.handleUpdate.bind(this.repository.stateManager),`

Since you have to setDeps on repository, which creates a stateManager, which needs a local index. And the local index would need a sync api.

packages/core/src/indexing/database-index-api.ts
dbcfd: `assertModelQueryable` no longer calls `assertModelIndexable`. That along the lines of what you were thinking?

packages/core/src/sync/workers/sync.ts
```diff
@@ -66,7 +67,7 @@ export class SyncWorker implements Worker<SyncJobData> {
    */
   async handler(job: PgBoss.Job) {
     const jobData = job.data as SyncJobData
-    const { fromBlock, toBlock, models } = jobData
+    const { fromBlock, toBlock, models, syncCompleteCallback } = jobData
```
stephhuynh18, line 70: Interesting but unfortunately pg.boss serializes the data when its stored in postgres so syncCompleteCallback will not be a function but rather a stringified function. 

You could pass a callback to the worker itself

packages/core/src/indexing/database-index-api.ts
stbrody: except it's still weird to call this `assertModelIndexable`.  It only runs on models that are already being indexed

packages/core/src/sync/sync-api.ts
```diff
@@ -262,11 +266,16 @@ export class SyncApi implements ISyncApi {
       fromBlock: startBlock,
       toBlock: endBlock,
       models: modelIds,
+      syncCompleteCallback: this.modelSyncComplete,
     })
   }
 
+  modelSyncComplete(model: string) {
+    this.modelsToHistoricSync.delete(model)
```
stephhuynh18, line 274: a history sync can occur in the follow situations:
1. When a node is restart and needs to catch up
2. A node is syncing for the first time
3. A reorg happened and we need to re-sync X blocks
4. We add a new model

a sync is complete for a model when it does not have a sync running due to 2 or 4. I am not sure about 1 or 3.  A model can have multiple history syncs at once. Ex. A node restarts twice so it has two different catchup syncs. 

So basically I don't think modelsToHistoricSync should be a set, rather than a mapping with counts

packages/core/src/indexing/database-index-api.ts
stbrody: maybe just call it `assertNoOngoingSyncs` and put it back into being called by `assertModelsQueryable`?

stbrody: you could even take the current body of `assertModelsQueryable` and move it to a new helper called `assertModelsAreIndexed`.  Then the body of `assertModelsQueryable` would just be 2 lines long, calling the two helper functions for the two checks that determine if a model is queryable (that it is indexed and that there are no ongoing syncs)

stbrody (commented):
added comments on the helper function structure, but will defer to @stephhuynh18 on the correctness of the ongoing sync checks

packages/core/src/sync/sync-api.ts
```diff
@@ -77,7 +82,7 @@ export interface SyncStatus {
 
 export class SyncApi implements ISyncApi {
   public readonly modelsToSync = new Set<string>()
-
+  public readonly modelsToHistoricSync = new Map<string, number>()
```
stephhuynh18, line 85: This needs to be incremented and decremented for every history sync job that prevents queries.  Right now its incremented only during `startModelSync` and decremented for ALL `history sync` (catch up, reorg, first time, and new model. This will result in the counts being messed up.

packages/core/src/sync/sync-api.ts
stbrody: That's actually a really interesting question.  I think I agree with blocking cases 2 and 4 but letting queries go through in cases 1 and 3.  @oed @3avi @michaelsena for any other opinions.

packages/core/src/sync/sync-api.ts
```diff
@@ -242,14 +265,17 @@ export class SyncApi implements ISyncApi {
...
     if (reorganized) {
       await this._addSyncJob(HISTORY_SYNC_JOB, {
+        jobType: SyncJobType.Reorg,
         fromBlock: block.number - BLOCK_CONFIRMATIONS,
         toBlock: block.number,
         models: Array.from(this.modelsToSync),
       })
     } else {
       await this._addSyncJob(CONTINUOUS_SYNC_JOB, {
+        jobType: SyncJobType.Reorg,
```
stephhuynh18, line 278: Not important but can we change this to "new block". This is the case when a new block is heard

PR #2725, opened 2023-02-17
feat: Handle REPLACED status from CAS

ukstv:


packages/core/src/__tests__/state-manager.test.ts
```diff
@@ -603,6 +606,17 @@ describe('anchor', () => {
   describe('With anchorOnRequest == false', () => {
     let inMemoryAnchorService: InMemoryAnchorService
 
+    function expectAnchorStatus(stream: Stream, status: AnchorStatus) {
+      return TestUtils.waitForState(
+        stream,
+        1000,
+        (s) => s.anchorStatus === status,
+        () => {
+          throw new Error(`Expect PENDING`)
```
stbrody, line 615:
```suggestion
          throw new Error(`Expected anchor status ${status} but found ${s.anchorStatus}`)
```

stbrody: Will also have to update the previous line to make the error function take `s` as an argument.

packages/core/src/state-management/state-manager.ts
```diff
@@ -360,18 +354,16 @@ export class StateManager {
...
-            // The one exception is if the AnchorServiceResponse indicates that the old commit
-            // is now anchored, in which case we still want to try to process the anchor commit
-            // and let the stream's conflict resolution mechanism decide whether or not to update
-            // the stream's state.
-            return
-          }
+          // We don't want to change a stream's state due to changes to the anchor
+          // status of a commit that is no longer the tip of the stream, so we early return
+          // in most cases when receiving a response to an old anchor request.
+          // The one exception is if the AnchorServiceResponse indicates that the old commit
```
stbrody, line 360: this isn't the one exception anymore right, that's why you moved the if-statement this used to be contained in?  Can we update this comment to reflect the new reality?

stbrody (approved):
lgtm

packages/core/src/state-management/state-manager.ts
ukstv: In general, the logic of the comment makes sense, and does no harm.

It is an exception, because of the way CAS works _now_. If something changes down the road, the behaviour stated in the comment and in code would still make sense.

packages/core/src/__tests__/state-manager.test.ts
ukstv: ack.

PR #2727, opened 2023-02-17
Feature/cdb 1068 pin api access control

zachferland:
Move pin api to admin api, ~~draft until all todos are done~~

- [x] local api implemented,  moves ceramic.pin -> ceramic.admin.pin
- [x] daemon endpoints added/moved, behind admin api, jws middleware validation, basic tests 
- [x] implement remote admin pins api, tests
- [x] any higher level / integration tests 

~~do we want to keep existing pins paths, and add unsupported warnings?~~ 

~~does this need to be added CLI as well?~~ see other story for this now

### DIFF
--- a/packages/cli/src/__tests__/admin-api.test.ts
+++ b/packages/cli/src/__tests__/admin-api.test.ts
@@ -89,6 +89,16 @@ describe('admin api', () => {
     return `${jws.signatures[0].protected}.${jws.payload}.${jws.signatures[0].signature}`
   }
 
+  async function buildJWSPins(did: DID, code: string, requestPath:string): Promise<string> {
+    const body =  undefined
+    const jws = await did.createJWS({
+      code: code,
+      requestPath: '/api/v0/admin/pins',
+      requestBody: body,
+    })
+    return `${jws.signatures[0].protected}.${jws.payload}.${jws.signatures[0].signature}`
+  }
+
   it('admin models API CRUD test', async () => {
     const adminURLString = `http://localhost:${daemon.port}/api/v0/admin/models`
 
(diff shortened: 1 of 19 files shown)


packages/cli/src/ceramic-cli-utils.ts
```diff
@@ -384,7 +384,7 @@ export class CeramicCliUtils {
     const id = StreamID.fromString(streamId)
 
     await CeramicCliUtils._runWithCeramicClient(async (ceramic: CeramicApi) => {
-      const result = await ceramic.pin.add(id)
+      const result = await ceramic.admin.pin.add(id)
```
oed, line 387: Not sure if we really intended to put all pin apis behind admin api, but I think it makes sense to make this symmetrical. Streams are pinned by default on creation anyway so it shouldn't be a problem.

packages/cli/src/ceramic-daemon.ts
```diff
@@ -738,6 +741,32 @@ export class CeramicDaemon {
     }
   }
 
+  async validateAdminRequest(req: Request, res: Response, next:NextFunction): Promise<void> {
+    if (!req.headers.authorization) {
+      res.status(StatusCodes.UNAUTHORIZED).json({ error: 'Missing authorization header' })
+      return
+    }
+    const jwsString = req.headers.authorization.split('Basic ')[1]
```
gvelez17, line 749: is this a little brittle?  maybe use https://www.npmjs.com/package/basic-auth ?

zachferland: probably, less familiar with the codebase here, so mostly tried to align as closely as possible with existing admin api and minimize changes in any other places (some parts are somewhat redundant now tho)

zachferland:
Open questions: 
- do we want to keep existing http pins paths, and add unsupported warnings? (similar to other deprecated paths)
- is jws in header for all pins requests fine? saw after some do have optional args in body, but keeping in headers seems good 
- breaking core api, any other concerns? 

some things could be refactored differently now, but since less familiar with codebase now, so aimed to minimize changes

zachferland:
Added original pins http paths back with a unsupported method response

packages/cli/src/__tests__/admin-api.test.ts
```diff
@@ -130,6 +140,96 @@ describe('admin api', () => {
...
   })
 
+  it('admin pin API CRUD test', async () => {
+    const adminPinURLBaseString = `http://localhost:${daemon.port}/api/v0/admin/pins`
+
+    const fetchCode = async (): Promise<string> => {
+      return (await fetchJson(`http://localhost:${daemon.port}/api/v0/admin/getCode`)).code
+    }
+
+    // Get list of pins
```
pawartur, line 150: Just a nit, but I guess it would be cools to have separate tests for separate cases: list/single/delete/add

packages/cli/src/ceramic-daemon.ts
pawartur: @zachferland , can you add a ticket to unify the code behind `admin/models` with what you wrote in this PR? I think that we'll have some (almost) duplicated code when we merge this PR.

packages/http-client/src/remote-pin-api.ts
```diff
@@ -1,39 +1,80 @@
...
+    const jws = await actingDid.createJWS({
+      code: code,
+      requestPath: url.pathname,
+    })
+    return `${jws.signatures[0].protected}.${jws.payload}.${jws.signatures[0].signature}`
+  }
+
+  readonly getCodePath = './admin/getCode'
+
+  private getCodeUrl(): URL {
```
pawartur, line 27: Could you also add a TODO here to unify this with `RemoteAdminApi`? Have you considered moving this code there and perhaps use composition to encapsulate the `/models` and `/pins` functionality? I understand that this PR is under time pressure, that's why you choose this approach, right?

pawartur (approved):
Accepted with a couple of comments :)

oed:
@pawartur plz see https://github.com/ceramicnetwork/js-ceramic/pull/2735 which tries to fix issues in this PR

pawartur:
> @pawartur plz see #2735 which tries to fix issues in this PR

Thanks, @oed !

oed:
This should be closed now I think?

stbrody:
superceded by https://github.com/ceramicnetwork/js-ceramic/pull/2735

PR #2732, opened 2023-02-20
Feature prevent reindexing model cdb1869

gzalz:


gzalz:
There are some admin api tests failing which seemed to be present on develop commit a6e96b4b when I created the branch.

packages/core/src/indexing/database-index-api.ts
```diff
@@ -197,6 +207,15 @@ export class DatabaseIndexApi<DateType = Date | number> {
       return StreamID.fromString(result.model)
     })
   }
+  private async getPreviouslyIndexedModelsFromDatabase(): Promise<Array<StreamID>> {
+    return (
+      await this.dbConnection(INDEXED_MODEL_CONFIG_TABLE_NAME).select('model').where({
+        is_indexed: false,
```
dbcfd, line 213: Shouldn't this be true if we're getting the models that were previously indexed?

packages/core/src/indexing/database-index-api.ts
```diff
@@ -79,6 +80,14 @@ export class DatabaseIndexApi<DateType = Date | number> {
...
+    for (const modelArgs of this.previouslyIndexedModels) {
+      const modelPreviouslyIndexed = this.previouslyIndexedModels.some(function (streamId) {
+        return String(streamId) === String(modelArgs)
+      })
+      if (modelPreviouslyIndexed){
+        throw new Error(`Cannot re-index model ${modelArgs.model.toString()}, data may not be up-to-date`)
+      }
+    }
     await this.indexModelsInDatabase(models)
     for (const modelArgs of models) {
```
dbcfd, line 92: Previously updated models should be updated here?

packages/core/src/indexing/database-index-api.ts
```diff
@@ -55,6 +55,7 @@ type IndexedData<DateType> = {
 export class DatabaseIndexApi<DateType = Date | number> {
   private readonly insertionOrder: InsertionOrder
   private modelsToIndex: Array<StreamID> = []
+  private previouslyIndexedModels = []
```
dbcfd, line 58: Do we need to save this? Since indexing models also hits the database, should we just query it first?

dbcfd (changes requested):

packages/core/src/indexing/database-index-api.ts
gzalz: We could certainly query the database instead of maintaining another copy. I'll go ahead and implement that.

packages/core/src/indexing/database-index-api.ts
gzalz: I was under the impression that the models to prevent re-indexing for had their indexing explicitly stopped and had the _is_indexed_ field set to false based on the ticket description: "Prevent models that are already set in ceramic_models:is_indexed to false and the corresponding MID table exists the DB don't allow to re-enable this model."

willex: Previously would be `false` - currently is `true`.

packages/core/src/indexing/database-index-api.ts
gzalz: That seems consistent with a DX I would want. I'm wondering if this is out of scope of the ticket since that was previously the description and then crossed out in favor of disallowing re-indexing streams that have had their is_indexed field set to false. I'd like to understand the historical context behind striking through the previous description and will seek some clarity on that.

packages/core/src/indexing/database-index-api.ts
```diff
@@ -79,6 +80,14 @@ export class DatabaseIndexApi<DateType = Date | number> {
    * @param models
    */
   async indexModels(models: Array<IndexModelArgs>): Promise<void> {
+    for (const modelArgs of this.previouslyIndexedModels) {
+      const modelPreviouslyIndexed = this.previouslyIndexedModels.some(function (streamId) {
+        return String(streamId) === String(modelArgs)
+      })
+      if (modelPreviouslyIndexed){
```
willex, line 87: Can you open a new ticket in HDS to 'Handle model history sync after re-indexing` and add `TODO` as part of a follow up. Thx!

packages/core/src/indexing/database-index-api.ts
willex: Yes, thats a good DX behavior for now.

dbcfd (approved):

willex (approved):
LGTM!

PR #2734, opened 2023-02-21
fix: align hotfix workflow with release workflow and push/deploy image

smrz2001:
Will test the workflows before merging

### DIFF
--- a/.github/workflows/reusable-release.yml
+++ b/.github/workflows/reusable-release.yml
@@ -237,6 +237,16 @@ jobs:
           npm config set //registry.npmjs.org/:_authToken $NODE_AUTH_TOKEN
           npm run publish:hotfix -- --yes
       -
+        name: Get version
+        run: echo "VERSION=$(dagger do -l error --output-format json version -p ${{ env.DAGGER_PLAN }} | jq -r '.version')" >> $GITHUB_ENV
+      -
+        # Pretend to be the main branch since a hotfix is a patch for the main branch
+        name: Push Docker image
+        run: dagger do push -w "actions:push:\"${{ env.AWS_REGION }}\":prod:main:\"${{ env.SHA }}\":\"${{ env.SHA_TAG }}\":\"${{ env.VERSION }}\":_" -p ${{ env.DAGGER_PLAN }}
+      -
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -15,7 +15,7 @@
     "docs": "npx typedoc",
     "publish:release-candidate": "npx lerna publish --dist-tag next --preid rc --no-verify-access --conventional-prerelease bump preminor",
     "publish:release": "npx lerna publish --dist-tag latest --create-release github --no-verify-access --conventional-graduate bump minor",
-    "publish:hotfix": "npx lerna publish --no-verify-access --conventional-graduate bump patch",
+    "publish:hotfix": "npx lerna publish --dist-tag latest --no-verify-access --conventional-graduate bump patch",
```
smrz2001, line 18: Don't think we should publish to GitHub like the release workflow does since that can result in tag conflicts. The `dist-tag` is the only diff I see that could explain why it wasn't working before. Will test and check.

.github/workflows/reusable-release.yml
```diff
@@ -236,6 +236,16 @@ jobs:
...
           npm run publish:hotfix -- --yes
+      -
+        name: Get version
+        run: echo "VERSION=$(dagger do -l error --output-format json version -p ${{ env.DAGGER_PLAN }} | jq -r '.version')" >> $GITHUB_ENV
+      -
+        # Pretend to be the main branch since a hotfix is a patch for the main branch
+        name: Push Docker image
+        run: dagger do push -w "actions:push:\"${{ env.AWS_REGION }}\":prod:main:\"${{ env.SHA }}\":\"${{ env.SHA_TAG }}\":\"${{ env.VERSION }}\":_" -p ${{ env.DAGGER_PLAN }}
+      -
+        name: Create deployment job
```
smrz2001, line 247: Do we want to deploy a hotfix from this workflow?

package.json
smrz2001: Actually, the problem appears to have been not fetching more git history. The `--dist-tag latest` bit is useful regardless since we'd want this hotfix publish to replace the previous release.

PR #2735, opened 2023-02-22
feat: Pin access control

ukstv:
Rebased version of #2727

### DIFF
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -816,8 +816,7 @@ export class CeramicDaemon {
       })
       return
     }
-    const requestPath = req.url === '/' ? req.baseUrl : req.baseUrl + req.url
-    const jwsValidation = await this._validateAdminApiJWS(requestPath, jwsString, false)
+    const jwsValidation = await this._validateAdminApiJWS(req.baseUrl, jwsString, false)
     if (jwsValidation.error) {
       res.status(StatusCodes.UNAUTHORIZED).json({ error: jwsValidation.error })
     } else {
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -1,5 +1,5 @@
 import * as fs from 'fs'
-import express, { Request, Response } from 'express'
+import express, { Request, Response, NextFunction } from 'express'
 import { Ceramic, CeramicConfig } from '@ceramicnetwork/core'
 import { RotatingFileStream } from '@ceramicnetwork/logger'
 import { ServiceMetrics as Metrics } from '@ceramicnetwork/observability'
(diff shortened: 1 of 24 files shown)


zachferland:
thanks @ukstv! let me know if there is anything else i should follow up on here

ukstv:
@zachferland As far as I could see, the code is rebased all right, and I believe the original branch has gone through the thorough review.

It would be good to have you skim through the PR diff, so that you could spot if anything went wrong during the rebase.

zachferland:
yeah rebase looks good to me

zachferland (approved):

packages/cli/src/ceramic-daemon.ts
```diff
@@ -341,33 +341,35 @@ export class CeramicDaemon {
...
     baseRouter.use('/multiqueries', multiqueriesRouter)
     baseRouter.use('/node', nodeRouter)
-    baseRouter.use('/pins', pinsRouter)
     baseRouter.use('/records', recordsRouter)
     baseRouter.use('/streams', streamsRouter)
     baseRouter.use('/collection', collectionRouter)
     baseRouter.use('/admin/getCode', adminCodesRouter)
     baseRouter.use('/admin/models', adminModelRouter)
     baseRouter.use('/admin/status', adminNodeStatusRouter)
+    // Admin Pins Validate JWS Middleware
```
stbrody, line 363: what does this mean?  Also what does it do when you call `.use` on a Router twice with the same path?

Should the other Admin API endpoints use this same technique?  Why is this endpoint doing something different than the other ones?

packages/cli/src/ceramic-daemon.ts
```diff
@@ -799,6 +804,32 @@ export class CeramicDaemon {
     })
   }
 
+  async validateAdminRequest(req: Request, res: Response, next: NextFunction): Promise<void> {
```
stbrody, line 807: this approach of separating the authorization logic from the actual core request logic is really nice.  Can we do it for the other admin API GET requests also?  Would also let us delete a lot of now-duplicated code

packages/core/src/local-admin-api.ts
```diff
@@ -12,7 +12,8 @@ export class LocalAdminApi implements AdminApi {
   constructor(
     private readonly indexApi: LocalIndexApi,
     private readonly syncApi: SyncApi,
-    private readonly nodeStatusFn: NodeStatusFn // TODO(CDB-2293): circular dependency back into Ceramic
+    private readonly nodeStatusFn: NodeStatusFn, // TODO(CDB-2293): circular dependency back into Ceramic private readonly pinApi: PinApi
```
stbrody, line 15:
```suggestion
    private readonly nodeStatusFn: NodeStatusFn, // TODO(CDB-2293): circular dependency back into Ceramic
```

stbrody (approved):
since I know we want to get this in for the release today, fine to leave all comments for a follow-up

packages/cli/src/ceramic-daemon.ts
zachferland: yeah its a common expressjs pattern - https://expressjs.com/en/guide/using-middleware.html, and you can register multiple and as mentioned below it uses next()

likely would make sense to use for all admin api validation, so that you can pull that verification out of core functions 

there is a could handful of things duplicated a bit here, I mostly tried to minimize changes to the existing adminapi since im less familiar with the codebase here now and shorter timeline, but these things could be revisited

packages/cli/src/ceramic-daemon.ts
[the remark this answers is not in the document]
zachferland: yeah a general one should look something like that, the pins requests ended all being the same and using headers (since no body, except some optional args), this is one point of inconsistency though

zachferland: yeah! commented above

packages/http-client/src/remote-admin-api.ts
```diff
@@ -104,4 +99,10 @@ export class RemoteAdminApi implements AdminApi {
       },
     })
   }
+
+  get pin(): PinApi {
+    if (this._pinApi) return this._pinApi
+    this._pinApi = new RemotePinApi(this._apiUrl, this._getDidFn)
```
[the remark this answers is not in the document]
zachferland: yeah can/should be moved to constructor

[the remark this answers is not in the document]
ukstv: Okay, since I do not see any commits to address this from @zachferland, I go ahead and add another change.

ukstv:
@zachferland @stbrody If by any chance I am AFK in bed after the tests become green (unlikely), feel free to squash and merge this PR, please.

PR #2736, opened 2023-02-22
feat (core, cli): enable node to run without cdb indexing

willex:
## Description

Introduces the option to run a node without Compose DB Indexing via:

- CLI flag: `--disable-composedb`
- ENV Variable: `CERAMIC_DISABLE_COMPOSE_DB="true"`

main changes are in the following files:

- `make-index-api.ts`
- `ceramic-cli-utils.ts`
- `daemon-config.ts`

the rest are mostly `.test` files where the ENV variable changed.

Additionally, this also removes the ability to run indexing against `main-net` with SQLite. 

PR addresses:
`CDB-2262`, `CDB-2144` & `CDB-2234`

## How Has This Been Tested?

- [ ] local tests confirming SQLite file is not created when starting node first time with indexing disabled
- [ ] local tests confirming config file doesn't include new indexing param
- [ ] running tests locally

## PR checklist

## References:

Docs updates are needed to include the new CLI & env options.


3avi:
LGTM from product/dx side -- removes sqlite indexing on mainnet and provides two paths to run node without indexing

packages/cli/src/bin/ceramic.ts
```diff
@@ -47,6 +47,7 @@ program
     '--sync-override <string>',
     'Global forced mode for syncing all streams. One of: "prefer-cache", "sync-always", or "never-sync". Defaults to "prefer-cache". Deprecated.'
   )
+  .option('--composedb-enabled <boolean>', 'Run node with Compose DB indexing enabled.')
```
stbrody, line 50: we've had problems in the past with setting boolean flags to false.  I think the option parsing library we use will interpret the boolean as "true" if the flag is provided at all.  We might want to consider making this a `--disable-composedb` flag that doesn't take any arguments for simplicity

packages/cli/src/ceramic-cli-utils.ts
```diff
@@ -206,6 +209,15 @@ export class CeramicCliUtils {
       if (syncOverride) {
         config.node.syncOverride = syncOverride
       }
+
+      config.indexing.composedbEnabled = fromBooleanInput(enableComposedb, true)
+      if (process.env.CERAMIC_ENABLE_COMPOSE_DB) {
```
stbrody, line 214: since composedb should be enabled by default, should this env var be `CERAMIC_DISABLE_COMPOSE_DB`?

packages/core/src/initialization/make-index-api.ts
```diff
@@ -27,13 +11,26 @@ export function makeIndexApi(
...
-    return undefined
-  }
   if (!indexingConfig) {
     logger.warn(`Indexing is not configured. Please add the indexing settings to your config file`)
     return undefined
   }
+
+  if (indexingConfig.composedbEnabled === false) {
+    logger.warn(
+      'Indexing is actively disabled. Change the corresponding CLI flag or ENV variable to enable it.'
```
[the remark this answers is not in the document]
stbrody: also the config file can disable it also.  We could update this message to mention that also, but to me that seems to be getting overly verbose.  I don't think we should have to re-explain how to set config flags here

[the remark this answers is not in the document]
stbrody: note the main change I wanted to suggest here though was just to include the word "ComposeDB" in the message

packages/core/src/initialization/make-index-api.ts
```diff
@@ -27,13 +11,26 @@ export function makeIndexApi(
...
+    )
+    return undefined
+  }
+
+  // TODO: extend with dynamic config check if MAINNET or sync is enabled
```
stbrody, line 26: what does "dynamic config check" mean?  Also can we link a linear ticket for the TODO?

packages/core/src/initialization/make-index-api.ts
```diff
@@ -27,13 +11,26 @@ export function makeIndexApi(
...
+  if (network === Networks.MAINNET && indexingConfig.db.toLowerCase().startsWith('sqlite')) {
+    logger.err(
+      'SQLite is not supported for indexing in production use. Please setup a Postgres instance and update the config file.'
```
stbrody, line 29:
```suggestion
      'SQLite is not supported for the ComposeDB indexing database in production use. Please setup a Postgres instance and update the config file.'
```

stbrody: ComposeDB is the feature name people know (and included more than just indexing), so I'm trying to make sure we always talk about ComposeDB not just "indexing"

stbrody (commented):
looks good.  A few small comments but they're all fairly trivial.  I think the most important question is whether the boolean direction of the flags should be inverted so they become "disable" flags instead of needing to set "enable" flags to false.

stbrody (commented):
putting this back out of my court as I already sent comments

packages/cli/src/bin/ceramic.ts
willex: OK, if there have been issues in the past with it I'll change the flag and address the downstream comments accordingly.

stbrody: talked about this with @oed and he's in alignment with this approach.  He suggests we should also change the config file option to disable-composedb so all 3 options are symmetrical

packages/core/src/initialization/make-index-api.ts
willex: Yeah that makes sense, I also standardized around `Compose DB`.

packages/cli/src/ceramic-cli-utils.ts
```diff
@@ -2,6 +2,7 @@ import os from 'os'
 import pc from 'picocolors'
 import { randomBytes } from '@stablelib/random'
 import * as u8a from 'uint8arrays'
+import { fromBooleanInput } from '@ceramicnetwork/ipfs-daemon'
```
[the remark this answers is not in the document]
willex: I couldn't move it since it had to the part in the same folder. My guess because of it being a package - so unless there are strong feelings around it I'll just keep as is.

packages/cli/src/ceramic-cli-utils.ts
```diff
@@ -132,7 +133,7 @@ export class CeramicCliUtils {
     pubsubTopic: string,
     corsAllowedOrigins: string,
     syncOverride: string,
-    enableComposedb: string | undefined
+    disableComposedb: string | undefined
```
stbrody, line 136:
```suggestion
    disableComposedb: boolean
```

stbrody: can look at how the other boolean flags are handled here

packages/cli/src/ceramic-cli-utils.ts
```diff
@@ -209,11 +210,13 @@ export class CeramicCliUtils {
...
       }
+      if (disableComposedb) {
+        config.indexing.disableComposedb = true
+      }
 
-      config.indexing.composedbEnabled = fromBooleanInput(enableComposedb, true)
-      if (process.env.CERAMIC_ENABLE_COMPOSE_DB) {
-        config.indexing.composedbEnabled = fromBooleanInput(
-          process.env.CERAMIC_ENABLE_COMPOSE_DB,
+      if (process.env.CERAMIC_DISABLE_COMPOSE_DB) {
```
stbrody, line 217:
```suggestion
      if (process.env.CERAMIC_DISABLE_COMPOSE_DB === 'true') {
```

stbrody: so it wont trigger with random input

packages/cli/src/ceramic-cli-utils.ts
```diff
@@ -209,11 +210,13 @@ export class CeramicCliUtils {
...
+        config.indexing.disableComposedb = fromBooleanInput(
```
stbrody, line 218: if you make the change above you can just set to true here (which is how we handle other boolean flags, so would be consistent).  Or you can leave it like this but make the default value be the current value in the config object

stbrody (commented):
a few minor comments on how options are handled.  very close though

packages/cli/src/ceramic-cli-utils.ts
willex: ah, yeah that was a leftover from the previous input from CLI

packages/cli/src/ceramic-cli-utils.ts
willex: yeah, I'll do how the others are handled. that way I remove the dependency to the helper function as well.

stbrody (approved):
lgtm assuming tess pass

PR #2737, opened 2023-02-22
fix: fix postgresql database url not accepted

pradel:
## Description

When trying to run the ceramic node with a postgres database, I faced an issue with the env variable being injected by the provider. The format was `postgresql://janedoe:mypassword@localhost:5432/mydb?schema=sample` but the node is expecting a string with the protocol `postgres`. Most of the cloud providers inject an URL with `postgresql` as a prefix so I think both should be accepted to make it work out of the box with render, supabase etc.. 

## PR checklist

## References:

- Prisma documentation https://www.prisma.io/docs/reference/database-reference/connection-urls

### DIFF
--- a/packages/core/src/indexing/build-indexing.ts
+++ b/packages/core/src/indexing/build-indexing.ts
@@ -73,7 +73,8 @@ export function buildIndexing(
         network
       )
     }
-    case 'postgres': {
+    case 'postgres':
+    case 'postgresql': {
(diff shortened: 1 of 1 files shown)


stbrody (approved):
seems reasonable to me

willex (approved):
Great catch! Thanks for submitting 💪

PR #2738, opened 2023-02-22
fix: re-indexing error on startup

gzalz:
Moved re-indexing to the CLI call as it prevented node startup otherwise.


willex (approved):
Lets Goooooo! 💨

PR #2739, opened 2023-02-22
fix(core): logging and job queue

stephhuynh18:
Use the interface of Job queue when possible
Add comments to status
Address comments for logger

### DIFF
--- a/packages/core/src/sync/sync-api.ts
+++ b/packages/core/src/sync/sync-api.ts
@@ -25,7 +25,7 @@ import {
 import { RebuildAnchorWorker } from './workers/rebuild-anchor.js'
 import { SyncWorker, createHistorySyncJob, createContinuousSyncJob } from './workers/sync.js'
 
-const SYNC_STATUS_LOG_INTERVAL = 60000
+const SYNC_STATUS_LOG_INTERVAL = 1000
 export const BLOCK_CONFIRMATIONS = 20
 // TODO (CDB-2292): block number to be defined
 export const INITIAL_INDEXING_BLOCKS: Record<string, number> = {
@@ -66,6 +66,8 @@ export interface ActiveSyncStatus {
   startedAt: Date
 }
 export interface ContinuousSyncStatus {
+  // The first block recevied form the chain on node startup
+  startBlock: number
   // The latest block received from the chain
   latestBlock: number
   // The number of blocks we wait for before we process a block
@@ -102,6 +104,7 @@ export class SyncApi implements ISyncApi {
   private initialIndexingBlock: number
   private periodicStatusLogger: Subscription | undefined
   private currentBlock: number
+  private startBlock: number
 
   constructor(
     private readonly syncConfig: SyncConfig,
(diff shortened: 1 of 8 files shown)


packages/core/src/sync/sync-api.ts
```diff
@@ -51,22 +52,37 @@ export type SyncConfig = {
...
+  startBlock: number
+  // The block the sync is currently processing
   currentBlock: number
+  // The block the sync will end on
   endBlock: number
+  // Models that are being synced
   models: Array<string>
+  // Date when the sync was requested
   created: Date
+  // Date when the sync started
```
stbrody, line 65: could there ever be more than a few milliseconds difference between `created` and `startTime`?

packages/core/src/sync/sync-api.ts
```diff
@@ -51,22 +52,37 @@ export type SyncConfig = {
...
+  // The block the sync starts at
+  startBlock: number
+  // The block the sync is currently processing
   currentBlock: number
+  // The block the sync will end on
   endBlock: number
+  // Models that are being synced
   models: Array<string>
+  // Date when the sync was requested
   created: Date
```
stbrody, line 64: nitpick but the naming is inconsistent between `created` and `startTime`.  Maybe rename `created` to `createdTime` for symmetry?

stbrody: or `createdAt` and `startedAt`?

packages/core/src/sync/sync-api.ts
```diff
@@ -282,7 +298,8 @@ export class SyncApi implements ISyncApi {
       activeSyncs: historySyncJobs.map((job) => {
         const jobData = job.data as SyncJobData
         return {
-          currentBlock: jobData.fromBlock,
+          currentBlock: jobData.currentBlock || jobData.fromBlock,
+          startBlock: jobData.fromBlock,
```
stbrody, line 302: startBlock for the continuousSync below?

stbrody (approved):
nice cleanup, thanks!

packages/core/src/sync/sync-api.ts
stephhuynh18: We can only have one History Sync happening at a time (this may change), we currently can only have one history sync being worked on at once (that may change in the future). But if a model is added, a history sync job is queued up for later. We do not error if a user tries to index + sync a model while a sync is going on. I can change that

packages/core/src/sync/sync-api.ts
stephhuynh18: I'm not sure what the start block would represent in a continuous sync. I would guess the user would want to know how far from the latest block we are which is current block. What does start block mean to you?

packages/core/src/sync/sync-api.ts
stbrody: ah interesting.  Yeah I thought we were going to error if a user tried to add a model while a sync was ongoing.  Seems like more of a product question though.  @3avi @michaelsena

packages/core/src/sync/sync-api.ts
stbrody: whatever block the continuous sync started at.  Presumably this will be whatever the current ethereum block is when the node starts up

PR #2740, opened 2023-02-22
fix: apply multiaddr patch

smrz2001:


packages/cli/package.json
```diff
@@ -33,6 +33,7 @@
   },
   "scripts": {
     "test": "NODE_OPTIONS=--experimental-vm-modules npx jest --silent --coverage --runInBand --forceExit",
+    "postinstall": "shx cp -r ../../node_modules/multiaddr node_modules/multiaddr && patch-package",
```
smrz2001, line 36: There are 2 other packages that use the outdated IPFS HTTP client - `ipfs-daemon` and `pinning-ipfs-backend`. IMO, only the Ceramic CLI has the crash vulnerability, but I can also patch the others to be safe. WDYT?

stbrody: I guess if you test with the bundled ipfs and do some pinning/unpinning requests and it all works then it's probably fine?

stbrody (approved):
I can see the intent and it seems reasonable, but will have to defer to @ukstv on if the changes here actually do what they intend in a good way

ukstv (approved):
If it works, it is approved.

Seems legit/innocent enough to me.

PR #2741, opened 2023-02-22
feat(core): Add stop model sync to admin api

stephhuynh18:


packages/core/src/sync/sync-api.ts
```diff
@@ -365,6 +365,19 @@ export class SyncApi implements ISyncApi {
     })
   }
 
+  /**
+   * Stop models from being included in the continuous sync
+   * TODO (CDB-2303): Remove existing history sync jobs as well
```
stbrody, line 370: could we log a warning that the sync will keep running to completion, and mention that this is a temporary state, and link a TODO to delete the log message when this is done?

stbrody (approved):
lgtm.  Seems like this should go into the ComposeDB beta RC we're cutting today, yes?

CC @oed

willex (approved):
RC📦✅

stbrody:
FYI test is failing @stephhuynh18

PR #2747, opened 2023-02-24
fix: validate daemon config after applying env overrides

nathanielc:
## Description
Validation of the private seed url was inside the fromFile function meaning that the config file must always have a value even if it was overwritten by the env later. This changes the logic to allow it to be missing from the file but still require that it be set somehow.

## How Has This Been Tested?

Running locally we no longer get an error about it being missing from the file.

## PR checklist

## References:


oed (approved):
To go in next RC.

PR #2748, opened 2023-02-24
chore(cli): authenticate CLI commands interacting with the pin APIs

PaulLeCam:
Tested manually but I couldn't find where to update the test suite to check the CLI interactions.


stbrody (approved):
LGTM

Unfortunately we do not have any automated tests of the Ceramic CLI currently.

https://linear.app/3boxlabs/issue/CDB-270/create-clidaemon-tests

packages/cli/src/bin/ceramic.ts
```diff
@@ -207,24 +207,24 @@ const pin = program.command('pin')
 pin.description(`('Ceramic local pinning API ${pc.red(pc.bold('[Deprecated]'))}`)
 
 pin
-  .command('add <streamId>')
-  .description(`Pin stream`)
-  .action(async (streamId) => {
-    await CeramicCliUtils.pinAdd(streamId)
+  .command('add <privateKey> <streamId>')
```
oed, line 210: This doesn't seem right. We already have a DID in the cli!

ukstv:
@PaulLeCam We can later add a private key to a client config file, right?

packages/cli/src/ceramic-cli-utils.ts
```diff
@@ -414,38 +414,44 @@ export class CeramicCliUtils {
...
+   * @param privateKey - Admin DID private key
    * @param streamId - Stream ID
    */
-  static async pinAdd(streamId: string): Promise<void> {
+  static async pinAdd(privateKey: string, streamId: string): Promise<void> {
     const id = StreamID.fromString(streamId)
 
-    await CeramicCliUtils._runWithCeramicClient(async (ceramic: CeramicApi) => {
+    await CeramicCliUtils._runWithCeramicClient(async (ceramic: CeramicClient) => {
+      await CeramicCliUtils._authenticateClient(ceramic, privateKey)
```
oed, line 424: `CeramicCliUtils._runWithCeramicClient` already have a DID: https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/cli/src/ceramic-cli-utils.ts#L501

oed (changes requested):

packages/cli/src/ceramic-cli-utils.ts
oed: err, here: https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/cli/src/ceramic-cli-utils.ts#L499

oed: Imo already confusing enough that the CLI and the DID used for CAS Auth have different ways to provide seeds. No need to add yet another way to create a DID when running js-ceramic.

PaulLeCam:
@ukstv I think the config file only contains the DID string, not the private key, does it?
I guess we can always add the private key to the config as well if we want at some point.

ukstv:
> @ukstv I think the config file only contains the DID string, not the private key, does it?
> I guess we can always add the private key to the config as well if we want at some point.

https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/cli/src/ceramic-cli-utils.ts#L499 as @oed pointed out contains a seed for `did:key`.

packages/cli/src/ceramic-cli-utils.ts
PaulLeCam: Isn't that different? My understanding is that the CLI stores a seed locally, but that doesn't mean it's necessarily an admin DID?
If an operator wants to interact with pins on a remote server, they should be able to provide the admin DID private key from the client, no?

PaulLeCam:
Yes I've seen this seed used by the CLI to create streams for example, but it's not necessarily an admin DID, or the only admin DID, is it?

stbrody:
I guess it seems like maybe we should add an `adminSeed` to the client config file alongside the existing `seed`?  `seed` is used for creating and updating tile documents, `adminSeed` is used for admin operations like the pin api?

oed:
> I guess it seems like maybe we should add an adminSeed to the client config file alongside the existing seed? seed is used for creating and updating tile documents, adminSeed is used for admin operations like the pin api?

Why do we need two separate things here? Can we please use the same one to avoid confusion?

PaulLeCam:
I just checked the behavior, the CLI generates a `client.config.json` file with a `seed` when running commands such as `ceramic create`, but it does not add the matching DID to the `http-api.admin-dids` array in the `daemon.config.json` file.
These seems like 2 independent behaviors to me, I don't think we should make assumptions about a possible relation between the node admins config and the client seed, should we?

> I guess it seems like maybe we should add an `adminSeed` to the client config file alongside the existing `seed`? `seed` is used for creating and updating tile documents, `adminSeed` is used for admin operations like the pin api?

`adminSeed` would be server-dependent though, no? If you connect to your local node, staging or production environment, you could have different admin DIDs so I wonder what the DX expectations would be like?

oed:
It's exactly this kind of conversation I'd like to avoid by keeping things simple.

Right now (before this PR) we have:

1. daemon config: adminDID - the DID that can be used to take admin actions
2. daemon config: cas-authDID - the seed for the DID used to send signed requests to CAS
3. cli config: clientSideDID - the DID used to sign things from the CLI

1 and 2 should definitely be different.

Seems very logical to me that the clientSideDID could be used as the admin DID. Otherwise we would introduce yet another thing here.

packages/cli/src/ceramic-cli-utils.ts
oed: > If an operator wants to interact with pins on a remote server, they should be able to provide the admin DID private key from the client, no?

yes, they should configure this for their cli client with the right seed to do this.

PaulLeCam:
@willex and I discussed it further, we thought it best for now to make the DX as similar as possible to the ComposeDB CLI which uses the `--did-private-key` flag for commands using admin auth, so I updated the PR to use this flag instead of the argument.
Eventually, we'll need to revisit the whole DX to ensure we properly cover the different user expectations during development and node operation flows, but that will require proper time and focus.
Does that work for you please @ukstv @stbrody @oed?

PaulLeCam:
Behavior updated again following a chat with @oed, the commands will default to using the client seed if the private key flag is not set.

packages/cli/src/bin/ceramic.ts
```diff
@@ -204,27 +204,30 @@ schemas
   })
 
 const pin = program.command('pin')
-pin.description(`('Ceramic local pinning API ${pc.red(pc.bold('[Deprecated]'))}`)
+pin.description(`Ceramic local pinning API ${pc.red(pc.bold('[Deprecated]'))}`)
```
oed, line 207: Should this really say deprecated?

oed (approved):

packages/cli/src/bin/ceramic.ts
PaulLeCam: Probably not, I can remove this.
Shall I also remove the deprecation notice for other commands? IIRC we log a message about using the Glaze CLI, which we don't maintain anymore.

PaulLeCam: Updated in new commit

oed (approved):

PR #2749, opened 2023-02-24
fix: added disallow HDS with SQLite check

willex:
## Description

Historical data sync is not supported with SQLite. Added addt. check to inform dev to switch.

## How Has This Been Tested?

- [ ] Tested locally switching between modes with both SQLite & Postgres

## PR checklist

### DIFF
--- a/packages/core/src/initialization/make-index-api.ts
+++ b/packages/core/src/initialization/make-index-api.ts
@@ -26,6 +26,17 @@ export function makeIndexApi(
   }
 
   const indexApi = buildIndexing(indexingConfig, logger, network)
+  // TODO(CDB-2078): replace env var with config option from ceramic_config
+  if (
+    process.env.CERAMIC_ENABLE_EXPERIMENTAL_SYNC === 'true' &&
+    indexApi instanceof SqliteIndexApi
+  ) {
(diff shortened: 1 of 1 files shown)


dbcfd:
@willex Can we create an issue to track cleaning up the usage of environmental variables within the code?

dbcfd (approved):

packages/core/src/initialization/make-index-api.ts
```diff
@@ -26,6 +26,17 @@ export function makeIndexApi(
   }
 
   const indexApi = buildIndexing(indexingConfig, logger, network)
+  // TODO(CDB-2078): replace env var with config option from ceramic_config
+  if (
+    process.env.CERAMIC_ENABLE_EXPERIMENTAL_SYNC === 'true' &&
```
stbrody, line 31: I thought we were getting rid of this env var?  Why is this even still required?  Isn't HDS on by default now?

willex: not yet on by default until qualification has been run on main-net. Bit concerned around Infura API calls as well atm with current non-optimization. If necessary I can remove it as part of a final RC build if necessary.

PR #2750, opened 2023-02-24
fix: Disallow pinning and unpinning through CRUD APIs

stbrody:


zachferland:
seems reasonable to me, so still pins by default on create, then all other applys/updates you cant pass pin/unpin option

zachferland (approved):

packages/core/src/__tests__/ceramic-pinning.test.ts
```diff
@@ -271,17 +294,56 @@ describe('Ceramic stream pinning', () => {
...
     await expect(TestUtils.isPinned(ceramic, stream.id)).resolves.toBeTruthy()
-    await TileDocument.load(ceramic, stream.id, { sync: SyncOptions.NEVER_SYNC, pin: false })
+    // pin:false flag will be ignored
+    TileDocument.load(ceramic, stream.id, { sync: SyncOptions.NEVER_SYNC, pin: false })
+    await expect(TestUtils.isPinned(ceramic, stream.id)).resolves.toBeTruthy()
+
+    await ceramic.close()
+  })
+
+  it('Existing stream cannot be pinned and unpinned on deterministic load', async () => {
```
oed, line 318: Why not pinned on load? 
Scenario: 
1. Someone creates a deterministic stream for user A without pinning
2. User A now creates their stream and tries to pin it
3. Operation 2 fails, this seems problematic

oed (commented):
At airport and haven't reviewed all code yet, but I have a question outlined below. Seems a bit too strict at first look.

stbrody:
So my first version of this PR only forbid unpinning through the CRUD APIs but still allowed pinning.  But that doesn't match what we did with the regular explicit pin API.  We moved all 3 pin operations, including `pin` and `ls`, behind admin DID protection.  So I matched here what we did there.  @oed

packages/core/src/__tests__/ceramic-pinning.test.ts
oed: @stbrody this needs to be fixed imo

oed:
Don't feel strongly @stbrody

packages/core/src/__tests__/ceramic-pinning.test.ts
stbrody: This should work I believe, so long as there's still only a single commit on the stream, you can pin with the `deterministic` api.  Once an update has been added, then you can no longer pin it

oed: Ok, then I'm fine with this PR 👍

stbrody: Added a test case to demonstrate

oed (approved):

PR #2754, opened 2023-02-26
chore: Update error message when starting on mainet without postgres to mention disabling ComposeDB

stbrody:
Also update all places where we say "Compose DB" to say "ComposeDB" to match our branding and marketing material.

### DIFF
--- a/packages/core/src/initialization/make-index-api.ts
+++ b/packages/core/src/initialization/make-index-api.ts
@@ -13,14 +13,14 @@ export function makeIndexApi(
 ): DatabaseIndexApi | undefined {
   if (!indexingConfig) {
     logger.warn(
-      `Compose DB Indexing is not configured. Please add the indexing settings to your config file`
+      `ComposeDB Indexing is not configured. Please add the indexing settings to your config file`
     )
     return undefined
   }
 
   if (indexingConfig.disableComposedb === true) {
     logger.warn(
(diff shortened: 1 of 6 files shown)


packages/core/src/initialization/make-index-api.ts
```diff
@@ -32,15 +32,15 @@ export function makeIndexApi(
...
+      'SQLite is not supported for the ComposeDB indexing database with historical syncing enabled. Please setup a Postgres instance and update the config file.'
     )
     return undefined
   }
 
   // TODO(CDB-2078): extend with addt. properties from startup if MAINNET or sync is enabled
   if (network === Networks.MAINNET && indexApi instanceof SqliteIndexApi) {
     throw Error(
-      'SQLite is not supported for the Compose DB indexing database in production use. Please setup a Postgres instance and update the config file.'
+      'SQLite is not supported for the ComposeDB indexing database in production use. Please setup a Postgres instance and update the config file to use ComposeDB, or disable ComposeDB to start up without an indexing database.'
```
stbrody, line 43: I'm a little worried about existing Ceramic users like gitcoin trying to upgrade to the new version and suddenly their node doesn't start up because they don't have a postgres database.  I want to at least make clear that disabling composedb is an option for those that aren't using it yet.

oed (approved):
Good idea

PR #2757, opened 2023-02-27
feat(cli, core): add sync worker flag to config file with default values

willex:
## Description

Moves HDS configuration flag from experimental env variable (`CERAMIC_ENABLE_EXPERIMENTAL_SYNC`) to be set in the config file. This populates both `allow-queries-before-historical-sync` & `run-historical-sync-worker` by network specific default values.

This PR has default options `3` implemented as outlined in the product discussion: https://www.notion.so/threebox/HDS-Default-Behavior-f5314ebbaba54e1d86ad3af613056f2c

## How Has This Been Tested?

- [ ] Locally verified config file creation

## PR checklist

## References:

### DIFF
--- a/packages/cli/src/ceramic-cli-utils.ts
+++ b/packages/cli/src/ceramic-cli-utils.ts
@@ -27,7 +27,7 @@ import { handleHeapdumpSignal } from './daemon/handle-heapdump-signal.js'
 import { handleSigintSignal } from './daemon/handle-sigint-signal.js'
 import { generateSeedUrl } from './daemon/did-utils.js'
 import { TypedJSON } from 'typedjson'
-import { getNetworkDefaultConfig } from '../../core/lib/indexing/migrations/1-create-model-table.js'
+import { getDefaultCDBDatabaseConfig } from '../../core/lib/indexing/migrations/1-create-model-table.js'
 
 const HOMEDIR = new URL(`file://${os.homedir()}/`)
 const CWD = new URL(`file://${process.cwd()}/`)
@@ -47,7 +47,7 @@ const DEFAULT_INDEXING_DB_FILENAME = new URL('./indexing.sqlite', DEFAULT_CONFIG
  */
 const generateDefaultDaemonConfig = () => {
   const privateSeedUrl = generateSeedUrl()
-  const getIndexingConfig = getNetworkDefaultConfig(Networks.TESTNET_CLAY)
+  const defaultIndexingConfig = getDefaultCDBDatabaseConfig(Networks.TESTNET_CLAY)
 
   return DaemonConfig.fromObject({
     anchor: {
@@ -70,8 +70,9 @@ const generateDefaultDaemonConfig = () => {
     indexing: {
       db: `sqlite://${DEFAULT_INDEXING_DB_FILENAME.pathname}`,
       'disable-composedb': false,
-      'run-historical-sync-worker': getIndexingConfig.run_historical_sync_worker,
-      'allow-queries-before-historical-sync': getIndexingConfig.allow_queries_before_historical_sync,
+      'run-historical-sync-worker': defaultIndexingConfig.run_historical_sync_worker,
+      'allow-queries-before-historical-sync':
+        defaultIndexingConfig.allow_queries_before_historical_sync,
     },
   })
 }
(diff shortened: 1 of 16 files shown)


packages/cli/src/ceramic-cli-utils.ts
```diff
@@ -46,6 +47,7 @@ const DEFAULT_INDEXING_DB_FILENAME = new URL('./indexing.sqlite', DEFAULT_CONFIG
  */
 const generateDefaultDaemonConfig = () => {
   const privateSeedUrl = generateSeedUrl()
+  const getIndexingConfig = getNetworkDefaultConfig(Networks.TESTNET_CLAY)
```
stbrody, line 50: can we rename this to something that makes it clearer the difference between that function and this one?  Maybe call `getNetworkDefaultConfig` "getDefaultCDBDatabaseConfig"?  And change the variable to be called "defaultDBConfig"?

packages/cli/src/ceramic-cli-utils.ts
```diff
@@ -201,6 +205,10 @@ export class CeramicCliUtils {
       }
       if (network) {
         config.network.name = network
+        const getIndexingConfig = getNetworkDefaultConfig(network)
+        config.indexing.allowQueriesBeforeHistoricalSync =
+          getIndexingConfig.allow_queries_before_historical_sync
```
stbrody, line 210: this is making the setting in the database (the harder one for users to change) override the behavior from the config file (the thing users actually have access to).  I think the the order of precedence for config options should be:
Env vars -> CLI flags -> config file -> database config table.

Since in this case there is no env var for cli flag, once the config file option is removed then the database table options will take over as the only way to configure this, but as long as the config file option exists it seems like it should take precedence.

packages/core/src/indexing/migrations/1-create-model-table.ts
```diff
@@ -90,23 +90,27 @@ export function indices(tableName: string): TableIndices {
...
+        run_historical_sync_worker: false,
       }
       break
     }
     case 'testnet-clay':
       return {
         enable_historical_sync: true,
-        allow_queries_before_historical_sync: false,
-        run_historical_sync_worker: true,
+        allow_queries_before_historical_sync: true,
```
stbrody, line 112: lets set `allow_queries_before_historical_sync` to true by default on `dev-unstable` also.  It's the last remaining network where it isn't

stbrody (changes requested):

packages/cli/src/ceramic-cli-utils.ts
stbrody: wait, this isn't even getting it from the database, this is just overriding the value from the config file with the default values all the time

packages/cli/src/ceramic-cli-utils.ts
```diff
@@ -201,6 +205,10 @@ export class CeramicCliUtils {
       }
       if (network) {
         config.network.name = network
+        const getIndexingConfig = getNetworkDefaultConfig(network)
+        config.indexing.allowQueriesBeforeHistoricalSync =
```
stbrody, line 209:
```suggestion
        config.indexing.allowQueriesBeforeHistoricalSync = config.indexing.allowQueriesBeforeHistoricalSync ?? getIndexingConfig.allowQueriesBeforeHistoricalSync
```

stbrody: although I'd also like to rename `getIndexingConfig` to something like `defaultIndexingConfig`

packages/cli/src/daemon-config.ts
```diff
@@ -242,10 +243,21 @@ export class IndexingConfig {
...
   @jsonMember(Boolean, {
     name: 'disable-composedb',
   })
   disableComposedb = false
+
+  /**
+   * Enable Historical data sync worker for Compose DB indexing
+   */
+  @jsonMember(Boolean, {
+    name: 'run-historical-sync-worker',
```
stbrody, line 258: is the word "worker" adding any clarity here?  How about just `run-historical-sync` or `enable-historical-sync`?

stbrody (approved):
lgtm mod question about the flag name

stbrody (approved):

PR #2762, opened 2023-02-27
fix(cli): legacy pin api, empty 200 on add to support 3id-connect, ot…

zachferland:
Related context - https://discord.com/channels/484729862368526356/1075891586354925708/1079555831537881198

get /pins/streamid - throw error
get /pins/ - throw error 
post /pins/streamid - ~~empty 200~~ 200 with warning message, not pinned, req ignored/dropped
delete /pins/streamid - throw error

Not ideal as it doesnt communicate expected behavior, but worthwhile temporary tradeoff (ref discord). Likely less effect on pin.add as well, since streams are pinned by default on creation, people calling pin.add are likely calling it on stream already pinned, while rm/list still throw errors (and more likely to be used by node runners). With plan to change with a stable path after.

### DIFF
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -347,6 +347,7 @@ export class CeramicDaemon {
     const adminCodesRouter = ErrorHandlingRouter(this.diagnosticsLogger)
     const adminModelRouter = ErrorHandlingRouter(this.diagnosticsLogger)
     const adminPinsRouter = ErrorHandlingRouter(this.diagnosticsLogger)
+    const legacyPinsRouter = ErrorHandlingRouter(this.diagnosticsLogger)
     const adminNodeStatusRouter = ErrorHandlingRouter(this.diagnosticsLogger)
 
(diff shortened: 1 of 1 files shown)


packages/cli/src/ceramic-daemon.ts
```diff
@@ -968,6 +972,10 @@ export class CeramicDaemon {
       .json({ error: 'Method not supported: pin requests have moved to the admin API /admin/pins' })
   }
 
+  async _empty200(req: Request, res: Response): Promise<void> {
+    res.sendStatus(StatusCodes.OK)
```
stbrody, line 976: could we maybe still include a field in the response object with a warning message that this is deprecated and not doing anything?  People using tools are unlikely to see it, but at least if someone manually hits the pin endpoint they'll see something in the response object telling them what's up

zachferland: yeah that makes sense, adding now

zachferland:
Added message, in reference to comment

stbrody (approved):
lgtm

oed (approved):

PR #2763, opened 2023-02-27
chore: Slightly more relaxed check for absence (#2763)Co-authored-by: Spencer T Brody <spencer+github@3box.io>

ukstv:
Maybe it is not `undefined`.

### DIFF
--- a/packages/core/src/indexing/local-index-api.ts
+++ b/packages/core/src/indexing/local-index-api.ts
@@ -102,7 +102,9 @@ export class LocalIndexApi implements IndexApi {
           let node = await this.repository.streamState(edge.node)
           if (!node) {
             this.logger.warn(`
-            Did not find stream state for streamid ${edge.node} in our state store when serving an indexed query.
+            Did not find stream state for streamid ${
+              edge.node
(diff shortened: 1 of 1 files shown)


packages/core/src/indexing/local-index-api.ts
```diff
@@ -99,8 +99,8 @@ export class LocalIndexApi implements IndexApi {
       const edges = await Promise.all(
         // For database queries we bypass the stream cache and repository loading queue
         page.edges.map(async (edge) => {
-          let node: StreamState = await this.repository.streamState(edge.node)
-          if (node === undefined) {
+          let node = await this.repository.streamState(edge.node)
+          if (!node) {
             this.logger.warn(`
             Did not find stream state in our state store when serving an indexed query.
```
stbrody, line 105:
```suggestion
            Did not find stream state for streamid ${edge.node} in our state store when serving an indexed query.
```

stbrody: let's add this too while we're at it so we can tell what's the problematic streamid

stbrody (approved):
lgtm with suggestion for improving the log message

PR #2764, opened 2023-02-27
fix(core): warn instead of throw on pin opts for stream crud (#2764

zachferland:
Similar reasoning to https://github.com/ceramicnetwork/js-ceramic/pull/2762 

Currently causes error in composedb cli /etc, those could be easily fixed, but may also cause glaze/selfid and all related dependencies to also throw errors which used these options (would have to double check these, it looks like they may only be used on create, but this would still likely be the safest short term option) 

Changes the thrown error to console.warn(if we use that all) 

~~Just commented out lines in tests, but can remove or cleanup in another way.~~

### DIFF
--- a/packages/core/src/__tests__/ceramic-pinning.test.ts
+++ b/packages/core/src/__tests__/ceramic-pinning.test.ts
@@ -266,14 +266,14 @@ describe('Ceramic stream pinning', () => {
       pin: false,
     })
     await expect(TestUtils.isPinned(ceramic, stream.id)).resolves.toBeFalsy()
-    await expect(
-      stream.update({ foo: 'baz' }, null, { anchor: false, publish: false, pin: true })
-    ).rejects.toThrow(/Cannot pin or unpin streams through the CRUD APIs/)
+    // await expect(
+    //   stream.update({ foo: 'baz' }, null, { anchor: false, publish: false, pin: true })
+    // ).rejects.toThrow(/Cannot pin or unpin streams through the CRUD APIs/)
     await expect(TestUtils.isPinned(ceramic, stream.id)).resolves.toBeFalsy()
     await stream.sync()
-    await expect(
-      stream.update({ foo: 'foobarbaz' }, null, { anchor: false, publish: false, pin: false })
-    ).rejects.toThrow(/Cannot pin or unpin streams through the CRUD APIs/)
+    // await expect(
+    //   stream.update({ foo: 'foobarbaz' }, null, { anchor: false, publish: false, pin: false })
+    // ).rejects.toThrow(/Cannot pin or unpin streams through the CRUD APIs/)
     await expect(TestUtils.isPinned(ceramic, stream.id)).resolves.toBeFalsy()
     await ceramic.close()
   })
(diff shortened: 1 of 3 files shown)


packages/core/src/state-management/repository.ts
```diff
@@ -340,7 +340,8 @@ export class Repository {
     if (opts.pin !== undefined && opType !== OperationType.CREATE) {
       const pinStr = opts.pin ? 'pin' : 'unpin'
       const opStr = opType == OperationType.UPDATE ? 'update' : 'load'
-      throw new Error(
+      // An error should be thrown once fully deprecated
+      console.warn(
```
zachferland, line 344: return after warn

zachferland:
yeah makes sense, seems like this is low risk to warn & return in function that doesnt return anything, but may be missing something 

only worth doing if low risk, with plan to change back after, at very least would like a save few hours from time spent on testing glaze/selfid etc and would likely speed up composedb testing 

but in the mean time ill look a bit deeper

zachferland:
link - https://github.com/ceramicstudio/js-composedb/pull/101 

^ yeah i think thats all for composedb

zachferland:
Then in glaze it looks like its only 
https://github.com/ceramicstudio/js-glaze/blob/a1897bb354a3e511ea1bf4cd526c3105ff108478/packages/tile-loader/src/index.ts#L296
https://github.com/ceramicstudio/js-glaze/blob/a1897bb354a3e511ea1bf4cd526c3105ff108478/packages/tile-loader/src/index.ts#L312

will this not throw on create? right? its only passed on creating tiles

stbrody:
> seems like this is low risk to warn & return in function that doesnt return anything, but may be missing something

I'm coming around on this thinking about it more.  I definitely am worried about people upgrading to the new version and having their apps suddenly break.

packages/core/src/__tests__/ceramic-pinning.test.ts
```diff
@@ -266,14 +266,14 @@ describe('Ceramic stream pinning', () => {
       pin: false,
     })
     await expect(TestUtils.isPinned(ceramic, stream.id)).resolves.toBeFalsy()
-    await expect(
-      stream.update({ foo: 'baz' }, null, { anchor: false, publish: false, pin: true })
-    ).rejects.toThrow(/Cannot pin or unpin streams through the CRUD APIs/)
+    // await expect(
```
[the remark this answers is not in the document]
zachferland: yeah ill add a ticket and reference those lines

zachferland:
yeah think anything that helps minimize time spent debugging older libraries and supporting existing apps in the next week is worth it

packages/core/src/state-management/repository.ts
```diff
@@ -340,11 +340,13 @@ export class Repository {
     if (opts.pin !== undefined && opType !== OperationType.CREATE) {
       const pinStr = opts.pin ? 'pin' : 'unpin'
       const opStr = opType == OperationType.UPDATE ? 'update' : 'load'
-      throw new Error(
+      // An error should be thrown once fully deprecated
```
stbrody, line 343:
```suggestion
      // TODO(CDB-2314): An error should be thrown once fully deprecated
```

stbrody: can you also add TODOs around the commented out tests?  Otherwise I'm worried about new devs going into the tests, seeing those commented-out lines and not understanding why they are there or why they are commented-out

stbrody: on second thought, maybe we should just delete those lines from the tests for now instead of leaving them sitting there commented out

zachferland: ah yeah a linear ref todo, added that 

removed the commented out test lines

stbrody (approved):
lgtm

PR #2765, opened 2023-02-28
fix(core, anchor-listener): fix hds bugs

stephhuynh18:
1. not keeping track of ongoing syncs corretly
2. Must sync after the index is established
3. increase numbers


willex (approved):
Great catch!

PR #2766, opened 2023-02-28
correct example postgres connection string

gvelez17:
## Description

Tiny change to correct the format of the comment describing the postgres connection string.

Checked and verified manually.


packages/cli/src/daemon-config.ts
```diff
@@ -230,7 +230,7 @@ export class IndexingConfig {
    * Connection string to a database. Only sqlite and postgres are supported.
    * Examples:
    *  - `sqlite:///path/to/database.sqlite`,
-   *  - `postgres:///user:password@host:5432/database`
```
stbrody, line 233: isn't the existing version here more useful?  By clarifying where the username, host, and database go?  This change makes the example more concrete, but also less informative

PR #2768, opened 2023-02-28
feat: Add loadOpts to MultiQuery

gzalz:


packages/core/src/ceramic.ts
```diff
@@ -873,28 +873,28 @@ export class Ceramic implements CeramicApi {
       let stream
       try {
         stream = await promiseTimeout(
-          this.loadStream(streamId, { atTime: query.atTime }),
+          this.loadStream(streamId, { atTime: query.loadOpts?.atTime }),
```
stbrody, line 876: we should pass in the full loadOpts object here, not just the atTime.  That's the whole advantage of using loadOpts instead - now we can set more options than just atTime!

packages/common/src/ceramic-api.ts
```diff
@@ -200,5 +200,5 @@ export interface MultiQuery {
   /**
    * Load a previous version of the stream based on unix timestamp
    */
-  atTime?: number
```
stbrody, line 203: We can't remove the existing atTime parameter or we'll break all existing applications that are currently using it.  We need to make this change in a backwards-compatibility-preserving way.  So we should continue to support the existing atTime field, while adding support for LoadOpts.  We should, however, mark the old atTime field as deprecated by adding `@deprecated` to the doc comment here, with a note to use `opts` instead.

packages/common/src/ceramic-api.ts
```diff
@@ -200,5 +200,5 @@ export interface MultiQuery {
...
+  loadOpts?: LoadOpts
```
stbrody, line 203: loadStream just calls this `opts`, we should probably do the same here for consistency.

stbrody (changes requested):
this needs tests of the new functionality.  You can probably play around with the `sync` flag within LoadOpts to see that its getting passed through correctly.  This test might give some ideas on what could work to show that the sync flag is being respected: https://github.com/ceramicnetwork/js-ceramic/blob/eb179830da3b5396794e85a5df374c7fc1590829/packages/core/src/__tests__/ceramic.test.ts#L411-L421

although that test uses Ceramic core instances directly, and we probably want this test to use the http-client so we can ensure that the loadOpts are being serialized correctly.  So we'll need to write a test in the stream-tests package to do that.

I'd suggest start with the other suggestions I made in this PR, then maybe put some time on my calendar for us to pair program the test writing, as that seems likely to be the trickiest part to get right

packages/3id-did-resolver/src/index.ts
```diff
@@ -190,6 +190,7 @@ const resolve = async (
     query.push({
       streamId,
       atTime: verNfo.timestamp,
+      opts: { atTime: verNfo.timestamp },
```
ukstv, line 193: Please, create an issue to remove the deprecated `atTime` after sufficient number of nodes got updated.

packages/core/src/__tests__/ceramic-api.test.ts
```diff
@@ -1,9 +1,9 @@
 import { jest } from '@jest/globals'
 import tmp from 'tmp-promise'
-import type { Ceramic } from '../ceramic.js'
+import { Ceramic } from '../ceramic.js'
```
ukstv, line 3: Please, get the imports in a previous shape if you do not use the changes in the actual code below.

packages/stream-tests/src/__tests__/multiquery.test.ts
```diff
@@ -0,0 +1,285 @@
...
+import { createCeramic } from '../../../core/src/__tests__/create-ceramic.js'
+import { TileDocument } from '@ceramicnetwork/stream-tile'
+import { IpfsApi, SyncOptions, TestUtils } from '@ceramicnetwork/common'
+import { StreamID } from '@ceramicnetwork/streamid'
+import { createIPFS } from '@ceramicnetwork/ipfs-daemon'
+import getPort from 'get-port'
+import { CeramicDaemon, DaemonConfig } from '@ceramicnetwork/cli'
+import { CeramicClient } from '@ceramicnetwork/http-client'
+import { jest } from '@jest/globals'
+describe('multiquery API http-client tests', () => {
```
ukstv, line 12: LF before this line.

packages/stream-tests/src/__tests__/multiquery.test.ts
```diff
@@ -0,0 +1,285 @@
+import tmp from 'tmp-promise'
+import type { Ceramic } from '../../../core/src/ceramic.js'
```
ukstv, line 2:
```suggestion
import type { Ceramic } from '@ceramicnetwork/core'
```

And do similar thing on the following line. Cross-package imports are not allowed.

packages/stream-tests/src/__tests__/multiquery.test.ts
```diff
@@ -0,0 +1,285 @@
...
+  let streamA: TileDocument,
+    streamB: TileDocument,
+    streamC: TileDocument,
+    streamD: TileDocument,
+    streamE: TileDocument,
+    streamF: TileDocument
+  const notExistStreamId = StreamID.fromString(
+    'kjzl6cwe1jw1495fyn7770ujykvl1f8sskbzsevlux062ajragz9hp3akdqbmdg'
+  )
+  const streamFTimestamps = []
```
ukstv, line 27: You use these two variables in tests. No need to make them available to the whole file.

packages/stream-tests/src/__tests__/multiquery.test.ts
```diff
@@ -0,0 +1,285 @@
...
+        atTime: streamFTimestamps[2],
+      },
+      {
+        streamId: streamF.id,
+      },
+    ]
+    const streams = await ceramic.multiQuery(queries)
+
+    expect(Object.keys(streams).length).toEqual(4)
+    const states = Object.values(streams).map((stream) => stream.state)
```
ukstv, line 123: `expect(states.length).toEqual(4)` and no need to use `Object.keys` above.

packages/stream-tests/src/__tests__/multiquery.test.ts
```diff
@@ -0,0 +1,285 @@
...
+    // timestamp between the first and the second anchor commit
+    streamFTimestamps.push(Math.floor(Date.now() / 1000))
+    streamFStates.push(streamF.state)
+    await TestUtils.delay(1000)
+    await streamF.update({ ...streamF.content, update: 'newer stuff' })
+    await TestUtils.anchorUpdate(core, streamF)
+    await TestUtils.delay(1000)
+    // timestamp after the second anchor commit
+    streamFTimestamps.push(Math.floor(Date.now() / 1000))
+    streamFStates.push(streamF.state)
```
ukstv, line 215: Same code all over again? Consider wrapping that in a function.

ukstv (changes requested):
Mostly okay, mod cross-package imports, tracking when to remove a deprecated option, and reusing code.

packages/stream-tests/src/__tests__/multiquery.test.ts
```diff
@@ -0,0 +1,285 @@
...
+        'http-api': { port: port, 'admin-dids': [core.did.id.toString()] },
+        node: {},
+      })
+    )
+    await daemon.listen()
+    ceramic = new CeramicClient(apiUrl)
+    ceramic.did = core.did
+
+    streamF = await TileDocument.create(ceramic, { test: '321f' })
+    streamE = await TileDocument.create(ceramic, { f: streamF.id.toUrl() })
```
stbrody, line 52: only streamF is ever used in tests, no need to make these other streams.

stbrody: Actually no need to make streamF here globally either, you can just make a new stream at the beginning of each test.

stbrody (commented):
looking good, a few small cleanups suggested but nothing big

packages/core/src/ceramic.ts
```diff
@@ -870,31 +870,33 @@ export class Ceramic implements CeramicApi {
 
     const index = {}
     const walkNext = async (node: TrieNode, streamId: StreamID | CommitID) => {
+      const queryAtTime = query.opts?.atTime ? query.opts?.atTime : query.atTime
+      const opts = queryAtTime ? { atTime: queryAtTime, ...query.opts } : {... query.opts}
```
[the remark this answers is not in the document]
gzalz: If we use query.opts here rather than {... query.opts} directly we lose the SyncOpts that are not explicitly part of LoadOpts and the sync tests fail.

packages/core/src/__tests__/ceramic-api.test.ts
gzalz: This is an unused import.

packages/stream-tests/src/__tests__/multiquery.test.ts
gzalz: notExistStreamId is used in the beforeAll and I believe should stay put. I can make these changes for the streamF* vars.

packages/core/src/ceramic.ts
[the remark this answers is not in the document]
stbrody: Hmm sorry I don't understand.  `{...query.opts}` should create an object that is totally identical to `query.opts`. It's just doing a shallow copy of all the fields.  I don't see how it could change the results of the test at all.  Maybe I'm missing something?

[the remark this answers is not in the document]
gzalz: From what I understand: 

```
LoadOpts extends SyncOpts. Therefore the schema to LoadOpts is:
{
 atTime
 sync
 syncTimeoutSeconds
}
```

`{... query.opts}` should copy all those fields.
`{query.opts}` would drop all of the original opts Sync fields on the floor.

Is there an assumption here that is incorrect?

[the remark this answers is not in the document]
gzalz: Correction: 

```
{
 loadOpts
 sync
 syncTimeoutSeconds
}
```

packages/stream-tests/src/__tests__/multiquery.test.ts
gzalz: Oops! I was looking at the wrong test WRT to notExistStreamId

packages/core/src/ceramic.ts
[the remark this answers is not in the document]
stbrody: note the code suggestion is replacing `{... query.opts}` with `query.opts` not with `{query.opts}`.  Just using the opts object directly.  There should be no fields dropped as it's just using the object as-is.

stbrody:
Okay @ukstv I cleaned this up and addressed the outstanding comments.  Should be ready for review again

stbrody:
ping @ukstv

ukstv:
Pong @stbrody  :/ forgot about it. Gonna review today.

ukstv (approved):

PR #2769, opened 2023-02-28
fix(cli): legacy pin, must return json for 200 (#2769

zachferland:
Needs to be valid json response, since http clients uses fethJSON and json parse response - https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/common/src/utils/http-utils.ts#LL49C6-L49C6

otherwise json parse error will be returned 

will try to add a test now, should have the first time, pushing up now for visibility


stbrody (approved):
lgtm

oed (approved):

PR #2770, opened 2023-03-03
feat: Change MID content in the indexing DB on update

ukstv:


packages/core/src/indexing/database-index-api.ts
```diff
@@ -155,34 +156,24 @@ export class DatabaseIndexApi<DateType = Date | number> {
...
-    for (const field of this.modelsIndexedFields.get(indexingArgs.model.toString()) ?? []) {
+    const fields = this.modelsIndexedFields.get(indexingArgs.model.toString()) ?? []
+    for (const field of fields) {
       indexedData[addColumnPrefix(field)] = indexingArgs.streamContent[field]
     }
-
     await this.dbConnection(tableName).insert(indexedData).onConflict('stream_id').merge({
       last_anchored_at: indexedData.last_anchored_at,
       updated_at: indexedData.updated_at,
+      stream_content: indexedData.stream_content,
```
PaulLeCam, line 173: Don't we need to also add the extra fields to index, in case they are updated?
Maybe the logic could simply be `this.dbConnection(tableName).insert(indexedData).onConflict('stream_id').merge(indexedData)` to ensure no data is missing?

PaulLeCam (approved):
LGTM just wondering about updates for custom fields?

packages/core/src/indexing/database-index-api.ts
ukstv: Ohh, true.

PR #2779, opened 2023-03-14
chore: Make anchor resuming more robust

ukstv:


packages/core/src/pubsub/task-queue.ts
```diff
@@ -30,17 +30,23 @@ export interface TaskQueueLike {
  * PQueue with synchronous `add` and a common error-handler.
  */
 export class TaskQueue implements TaskQueueLike {
```
stbrody, line 32: nitpick: can we move this out of the `pubsub` directory?  Perhaps to a new `utils` directory?

stbrody (approved):
lgtm

PR #2786, opened 2023-03-21
feat(cli): inject latest git hash during builds

smrz2001:
Just one way to do it that seemed simple enough and similar to `genversion`. Let me know what you think.

This is what the file looks like:
```
export const commitHash = 'e458b1d2c64ae47ab7286ee531ef23309725ffff'
```
I haven't _used_ the hash anywhere yet. I just wanted to enable it first, then it can be used anywhere, like `version`.


packages/cli/package.json
```diff
@@ -33,7 +33,7 @@
   },
   "scripts": {
     "test": "NODE_OPTIONS=--experimental-vm-modules npx jest --silent --coverage --runInBand --forceExit",
-    "build": "genversion --es6 src/version.ts && npx tsc -p tsconfig.json",
+    "build": "genversion --es6 src/version.ts && node ../../tools/commitHash.js && npx tsc -p tsconfig.json",
```
stbrody, line 36: I'll defer to @ukstv on if its safe to reference something outside of the package like this.  Maybe it would be safer to move this tool into the `cli` package?

smrz2001: Yeah, I don't mind keeping it wherever. I put it in `tools` so it can be used from other packages as needed, but if we're only ever going to use it from the CLI, then I can move it in there.

smrz2001: Actually, it was more trouble to keep it outside lol, so I moved it into the `cli` package.

stbrody (approved):
lgtm

ukstv (approved):

PR #2787, opened 2023-03-22
feat: add composeDB section to node status endpoint

nathanielc:
## Description

This change adds composeDB specific information to the status admin API endpoint. Additionally a CLI command is added to export the status.

## How Has This Been Tested?

Unit tests added 

## PR checklist

## References:

Fixes: CDB-2289
Fixes: CDB-2290


stbrody:
FYI @nathanielc PR is failing lint check

stbrody (approved):
LGTM

nathanielc:
@stbrody Why did CI fail? The CI logs say that all the tests passed.

packages/core/src/ceramic.ts
```diff
@@ -701,12 +704,24 @@ export class Ceramic implements CeramicApi {
       chainId: this._anchorValidator.chainId,
     }
     const ipfsStatus = await this.dispatcher.ipfsNodeStatus()
+
+    let composeDB = undefined
+    if (!this._params.indexingConfig.disableComposedb) {
```
stbrody, line 709: I'd prefer us not save the `_params` - it's the job of the `IndexApi` to interpret the config so I think it would be better to ask questions of it than the config directly.

There isn't currently an `IndexApi.enabled` accessor, but you could easily add one that just checks if the underlying `databaseIndexApi` us defined or not.

stbrody (changes requested):
paul's commit I think is in the right direction to fix the test failure, but I put some suggestions to change the approach to avoid holding onto the params object for the life of the process and instead to ask questions about the current state of the world via the APIs that own the various sub-systems.

packages/core/src/indexing/local-index-api.ts
```diff
@@ -53,6 +54,7 @@ export class LocalIndexApi implements IndexApi {
     networkName: Networks
   ) {
     this.databaseIndexApi = makeIndexApi(indexingConfig, networkName, logger)
+    this.enabled = !indexingConfig?.disableComposedb
```
stbrody, line 57: this isn't quite right.  If `indexingConfig` is null this will set `enabled` to true.  The `makeIndexApi` function already has logic to interpret the indexingConfig.  I'd just set `enabled` based on whether `databaseIndexApi` is defined or not.

stbrody (approved):
lgtm mod one small fix

packages/core/src/indexing/local-index-api.ts
PaulLeCam: Thanks, should be fixed now

stbrody (approved):
still lgtm

3avi:
@PaulLeCam feel free to merge on monday!

PR #2793, opened 2023-03-27
rough out code to cache responses and backoff responding too much to …

gvelez17:
…queries

## Description

## How Has This Been Tested?

- [ ] Test B

## PR checklist

## References:

### DIFF
--- a/packages/core/src/dispatcher.ts
+++ b/packages/core/src/dispatcher.ts
@@ -403,13 +403,24 @@ export class Dispatcher {
       // TODO: Should we validate that the 'id' field is the correct hash of the rest of the message?
 
       const tip = streamState.log[streamState.log.length - 1].cid
-      // Build RESPONSE message and send it out on the pub/sub topic
-      // TODO: Handle 'paths' for multiquery support
-      const tipMap = new Map().set(streamId.toString(), tip)
-      this.publish({ typ: MsgType.RESPONSE, id, tips: tipMap })
+
+      const responseDelay = Math.floor(Math.random() * (this.responseDelay + 1))
+
+      // Push the message onto a time delay of 5 seconds
+      setTimeout(() => {
+        // Are we a good one to answer this message?
+        if (this._shouldAnswer(tip)) {
+          // Build RESPONSE message and send it out on the pub/sub topic
+          // TODO: Handle 'paths' for multiquery support
+          const tipMap = new Map().set(streamId.toString(), tip)
+          this.publish({ typ: MsgType.RESPONSE, id, tips: tipMap })
+          this._speedUpResponses()
+        } else {
+          this._slowDownResponses()
+        }
+      }, responseDelay)
     }
   }
-
   /**
    * Handles an incoming Response message from the pub/sub topic.
    * @param message
@@ -419,6 +430,9 @@ export class Dispatcher {
     const { id: queryId, tips } = message
     const outstandingQuery = this.messageBus.outstandingQueries.queryMap.get(queryId)
     const expectedStreamID = outstandingQuery?.streamID
+
+    this._noteResponse(message)
+
     if (expectedStreamID) {
       const newTip = tips.get(expectedStreamID.toString())
       if (!newTip) {
--- a/packages/core/src/dispatcher.ts
+++ b/packages/core/src/dispatcher.ts
@@ -39,6 +39,11 @@ const MAX_INTERVAL_WITHOUT_KEEPALIVE = 24 * 60 * 60 * 1000 // one day
 const IPFS_CACHE_SIZE = 1024 // maximum cache size of 256MB
 const IPFS_OFFLINE_GET_TIMEOUT = 200 // low timeout to work around lack of 'offline' flag support in js-ipfs
 const PUBSUB_CACHE_SIZE = 500
+const DEFAULT_RESPONSE_DELAY_MS = 1000 // 1 second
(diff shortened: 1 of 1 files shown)


packages/core/src/dispatcher.ts
```diff
@@ -366,14 +386,24 @@ export class Dispatcher {
   /**
    * Handle a new tip learned about from pubsub, either via an UPDATE or a RESPONSE message
    */
-  async _handleTip(tip: CID, streamId: StreamID, model?: StreamID) {
+  async _handleTip(tip: CID, streamId: StreamID, model?: StreamID, isResponse=false) {
```
stephhuynh18, line 389: this will only look at responses for our queries (see `_handleResponseMessage`)

packages/core/src/dispatcher.ts
```diff
@@ -39,6 +39,11 @@ const MAX_INTERVAL_WITHOUT_KEEPALIVE = 24 * 60 * 60 * 1000 // one day
 const IPFS_CACHE_SIZE = 1024 // maximum cache size of 256MB
 const IPFS_OFFLINE_GET_TIMEOUT = 200 // low timeout to work around lack of 'offline' flag support in js-ipfs
 const PUBSUB_CACHE_SIZE = 500
+const DEFAULT_RESPONSE_DELAY_MS = 1000 // 1 second
```
stephhuynh18, line 42: Best to ask sergey but the syncTimeoutSeconds (the amount of time we wait for responses to our query is defaulted to3 seconds). So these numbers need to reflect this

packages/core/src/dispatcher.ts
```diff
@@ -389,6 +419,51 @@ export class Dispatcher {
...
+     // if we know about every answer someone else gave, but we have a different answer
+     // then we know best
+     if (answers.every((answerTip) => streamState.log.includes(answerTip))) {
+         return true
+     }
+
+     // we didn't know about all the answers someone else gave so they probably know better
+     // note we have to prune attackers who are just saying junk, that should be separate thing
+     // the guy whose query it was will report them when he tries to anchor it and checks the sig
+     return false
```
stephhuynh18, line 464: What if we have a conflicting answer. Ex A->B and A->C

packages/core/src/dispatcher.ts
stephhuynh18: https://github.com/ceramicnetwork/js-ceramic/blob/90e77ee5b5bb14d3711c14ef6852bd5c9442f3ca/packages/core/src/state-management/repository.ts#L45 found here

packages/core/src/dispatcher.ts
gvelez17: ah!  how does it know

ukstv (changes requested):
First, Dispatcher is not the best place for this logic. It is already overwhelmed with various responsibilities. Maybe a better alternative would be yet another Observable akin to existing `PubsubRateLimit` or `PubsubKeepalive`.

Second, I am a bit worried about mixing random timeouts and strict logic like " ok we disagree, but whose answer is better?  now we have to look at the streamState". I mean, nothing is wrong by its own, it is just (un)fun stuff emerges on intersection of these two approaches in my experience.

It is a pretty complicated change, it would be good to have a memo in Notion explaining various control flow scenarios.

packages/core/src/dispatcher.ts
```diff
@@ -39,6 +39,11 @@ const MAX_INTERVAL_WITHOUT_KEEPALIVE = 24 * 60 * 60 * 1000 // one day
 const IPFS_CACHE_SIZE = 1024 // maximum cache size of 256MB
 const IPFS_OFFLINE_GET_TIMEOUT = 200 // low timeout to work around lack of 'offline' flag support in js-ipfs
 const PUBSUB_CACHE_SIZE = 500
+const DEFAULT_RESPONSE_DELAY_MS = 1000 // 1 second
+const MINIMUM_RESPONSE_DELAY_MS = 0
+const MAXIMUM_RESPONSE_DELAY_MS = 16040 // 16 seconds and change
```
gvelez17, line 44: we can reduce this - on average it will be this / # peers with the answer who are listening

packages/core/src/dispatcher.ts
```diff
@@ -403,13 +478,24 @@ export class Dispatcher {
...
-      const tipMap = new Map().set(streamId.toString(), tip)
-      this.publish({ typ: MsgType.RESPONSE, id, tips: tipMap })
+
+      const responseDelayMS = Math.floor(Math.random() * (this.responseDelayMS + 1))
+
+      // Push the message onto a random and variable length time delay
+      setTimeout(() => {
+        // Are we a good one to answer this message?
+        if (this._shouldAnswer(streamId, tip, streamState)) {
+          // Build RESPONSE message and send it out on the pub/sub topic
```
gvelez17, line 488: @ukstv @dbcfd @stbrody doesn't this address the issue of some nodes being more correct than others?  Right now we have the gateway node for instance responding to all queries, even though it probably doesn't know about them.  Even if conflict resolution is not perfect, right now we are going to fall over completely during the gitcoin round - i do think it would make sense to try to limit responses instead?

packages/core/src/dispatcher.ts
gvelez17: yeah we could add a case for this also

gvelez17:
Closing this since it is not the approach we will take.

PR #2796, opened 2023-03-30
fix(core): programmatically change PG sslmode to avoid self signed cert errors

willex:
## Description

This addresses a common issue that Postgres will throw a self signed cert error with cloud deployments/PaaS (e.g. Heroku, DigitalOcean) and terminate connections because of the reliance on self signed certificated within a VPC and ultimate fail a build. Trying to address the issue directly in the ORM library Knex with the SSL connection properties of: `{ rejectUnauthorized: false }` did not yield the desired effect and the only other recourse would've been to disable TLS verification completely (`process.env.NODE_TLS_REJECT_UNAUTHORIZED`), which is highly undesirable.

Updating the mode programmatically was chosen over the option to inform the user and do it manually or integrating it into the config file to avoid further friction during node deployments. Overall, this approach should not introduce any security risks as well as a MITM (man in the middle attack) between a node deployment and the Postgres instance within a VPC is fairly unlikely. 

**Before Fix:**
With a PaaS default connection URL example of: `postgres://ceramic:password@127.0.0.1:5432/ceramic?sslmode=required`

Screenshot 2023-03-29 at 22 13 15

**After Fix:**
Screenshot 2023-03-29 at 22 11 34

## How Has This Been Tested?

- [ ] Local replication of described issue
- [ ] tested all valid `?sslmode=XYZ` config options to verify functionality

## References:

This should most likely be listed or called out in the documentation referencing DigitalOcean & Heroku as source for the motivation.

### DIFF
--- a/packages/core/src/indexing/build-indexing.ts
+++ b/packages/core/src/indexing/build-indexing.ts
@@ -85,6 +85,18 @@ export function buildIndexing(
     case 'postgres':
     case 'postgresql': {
       logger.imp('Initializing PostgreSQL connection')
+
+      if (connectionString.searchParams.has('sslmode')) {
+        const uriSSLConfig = connectionString.searchParams.get('sslmode')
+
+        if (uriSSLConfig != 'disable' && uriSSLConfig != 'no-verify') {
+          connectionString.searchParams.set('sslmode', 'no-verify')
+          logger.warn(
+            `Changed "?sslmode=${uriSSLConfig}" to "no-verify" to avoid self signed cert errors during cloud deployments.`
+          )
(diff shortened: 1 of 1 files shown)


3benbox:
I'd like to see the option of providing a CA cert (which would validate the self-signed server cert) for operators that want to preserve database connection security.
It's reasonable to assume the database and credentials could be reused outside ceramic ;(
If we go that route, proving an option or client certs would be nice too
https://www.postgresql.org/docs/current/auth-cert.html

stbrody:
ping @willex what's the status of this PR?

PR #2808, opened 2023-04-13
chore: setMaxListeners for child AbortSignal

ukstv:
Should eliminate `MaxListenersExceededWarning`. Based on a hypothesis: ipfs-http-client reuses single passed AbortSignal too much adding listeners and never removing them.

### DIFF
--- a/packages/common/src/utils/abort-signal-utils.ts
+++ b/packages/common/src/utils/abort-signal-utils.ts
@@ -1,5 +1,6 @@
 import { timer, fromEvent, merge, Subscription } from 'rxjs'
 import { first } from 'rxjs/operators'
+import { setMaxListeners } from 'node:events'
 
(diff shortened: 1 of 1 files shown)


packages/common/src/utils/abort-signal-utils.ts
```diff
@@ -62,6 +63,7 @@ export async function abortable<T>(
   }
   original.addEventListener('abort', onAbort)
   if (original.aborted) controller.abort()
+  setMaxListeners(Infinity, controller.signal)
```
dbcfd, line 66: It seems like this might open us up to possibly memory leaks. Can we make the value passed in and configurable?

Alternate is actually to provide a metric somehow to make sure this isn't constantly growing, but not sure if there's an easy way to do that.

dbcfd (changes requested):

PR #2820, opened 2023-05-03
feat: Less calls to IPFS node

ukstv:
Draft till we have a proof of improvement.


packages/core/src/utils.ts
```diff
@@ -98,26 +98,6 @@ export class Utils {
...
 
-  /**
-   * Puts a block on IPFS
-   * @param cid the CID of the block to put
-   * @param block bytes array of block to put
-   * @param ipfsApi the IPFS Api instance to use
-   * @param signal AbortSignal
-   */
-  static async putIPFSBlock(
-    cid: CID | string,
```
dbcfd, line 109: Does this count as a stringly typed fix too 😆

dbcfd (approved):

packages/core/src/utils.ts
ukstv: LOL

packages/core/src/dispatcher.ts
```diff
@@ -138,40 +159,28 @@ export class Dispatcher {
...
+        const cid = carFile.put(jws, { codec: 'dag-jose', hasher: 'sha2-256', isRoot: true }) // Encode JWS itself
+        restrictCommitBlockSize(carFile, cid)
+        await all(this._ipfs.dag.import(carFile, { pinRoots: false }))
         return cid
       }
-      const cid = await this._shutdownSignal.abortable((signal) => {
-        return this._ipfs.dag.put(data, { signal: signal })
-      })
-      await this._restrictCommitSize(cid)
+      const cid = carFile.put(data)
```
oed, line 181: Shouldn't this be the root?
```suggestion
      const cid = carFile.put(data, { isRoot: true })
```

packages/core/src/dispatcher.ts
```diff
@@ -302,21 +311,16 @@ export class Dispatcher {
    * @param cid - Commit CID
    * @private
    */
-  async _restrictCommitSize(cid: CID | string): Promise<void> {
+  async _restrictCommitSize(cid: CID | string): Promise<number> {
     const asCid = typeof cid === 'string' ? CID.parse(cid) : cid
-    const stat = await this._shutdownSignal.abortable((signal) => {
+    const { size } = await this._shutdownSignal.abortable((signal) => {
       return this._ipfs.block.stat(asCid, {
```
oed, line 317: Shy still do this one remotely?

oed (commented):
This is a very nice cleanup 💥

packages/core/src/dispatcher.ts
ukstv: I believe we should. `_restrictCommitSize` is adjacent to [`dag.get`](https://github.com/ceramicnetwork/js-ceramic/pull/2820/files/3cd767b54de60bded408dc69cb59944006a8f17e#diff-e522280df0e1c4c8fe4f157393e23872535ee7bab6d33a2930c26b18f7411d19R209) call, which returns decoded IPLD structure. To check that w/o a call to IPFS we'd need to convert the structure back to block (useless) or decode the block on js-ceramic side (thus dealing with all of the codecs available on our own). A choice between scylla and charybdis IMO, which we better fly over.

To be honest, I am not yet sure about the extent of the performance improvement this PR brings. Measuring it now.

oed: If we want to minimize load on IPFS during high traffic this seems worth it? One call to `block.get` instead of two calls (`block.stat` and `dag.get`).
I assume we have the codecs in the js code anyway?

packages/core/src/__tests__/state-manager.test.ts
```diff
@@ -315,7 +315,9 @@ describe('anchor', () => {
         )
 
         // We fast-forward streamState2, because the commit is legit
-        expect(streamState2.state).toEqual(stream1.state)
+        expect(StreamUtils.serializeState(streamState2.state)).toEqual(
```
stbrody, line 318: How did this change to how we use IPFS result in a change that broke a comparison of Ceramic state objects?

packages/core/src/dispatcher.ts
```diff
@@ -138,40 +159,28 @@ export class Dispatcher {
...
-          return Utils.putIPFSBlock(linkCid, linkedBlock, this._ipfs, signal)
-        })
-        await this._restrictCommitSize(jws.link.toString())
-        await this._restrictCommitSize(cid)
+        carFile.blocks.put(new CarBlock(jws.link, linkedBlock)) // Encode payload
+        restrictCommitBlockSize(carFile, jws.link)
+        const cid = carFile.put(jws, { codec: 'dag-jose', hasher: 'sha2-256', isRoot: true }) // Encode JWS itself
+        restrictCommitBlockSize(carFile, cid)
+        await all(this._ipfs.dag.import(carFile, { pinRoots: false }))
         return cid
```
stbrody, line 178: do we need to bump the COMMITS_STORED Metric on this exit path as well?

packages/core/src/__tests__/state-manager.test.ts
ukstv: Improper coercion from Buffer to Uint8Arrays in a 2nd level dependency. Should be fixed.

ukstv:
Had to change FAKE CIDs in few places, as they used dag-pb codec.

ukstv:
Oh, there is one more positive side effect of this change. `ipfs.dag.get(cid, {path: '/long/path'}` would do traversal client-side, i.e multiple block.get calls. `ipfs.dag.resolve (if needed); ipfs.block.get` would do 1 or two calls max.

packages/core/src/__tests__/dispatcher-mock-ipfs.test.ts
```diff
@@ -12,10 +12,13 @@ import { StateManager } from '../state-management/state-manager.js'
...
 import { LevelDbStore } from '../store/level-db-store.js'
 import { StreamStateStore } from '../store/stream-state-store.js'
+import { CARFactory } from 'cartonne'
+import * as dagJoseCodec from 'dag-jose'
+import * as dagCborCodec from '@ipld/dag-cbor'
 
 const TOPIC = '/ceramic'
-const FAKE_CID = CID.parse('bafybeig6xv5nwphfmvcnektpnojts33jqcuam7bmye2pb54adnrtccjlsu')
-const FAKE_CID2 = CID.parse('bafybeig6xv5nwphfmvcnektpnojts44jqcuam7bmye2pb54adnrtccjlsu')
+const FAKE_CID = CID.parse('bafyreihbje3f5oj6vlszlqmnrtrsmuukrjurpsv6uus4siwgikldp6wnty')
```
ukstv, line 20: dag-pb, we do not use that.

PR #2821, opened 2023-05-04
Import codecs from CAS into dedicated package

PaulLeCam:
## Description

Imported codecs and related tests from CAS into new `@ceramicnetwork/codecs` package.


packages/codecs/src/did.ts
```diff
@@ -0,0 +1,30 @@
+import { refinement, string } from 'codeco'
+
+import type { Opaque } from './types.js'
```
ukstv, line 3: Since some time `ts-essentials` package provides working `Opaque` type. We'd need that in other places, so IMO, it makes sense now to use `Opaque` from `ts-essentials` as dev dependency here.

ukstv (approved):
Nice. The only change I would like to see is using `Opaque` from `ts-essentials`. My gut feeling is there are few other goodies there like `NonEmptyArray` that we could use more extensively in Ceramic codebase.

packages/codecs/src/did.ts
ukstv: To be honest, the only reason for using custom `Opaque` was that the one from `ts-essentials` [did not work](https://github.com/ts-essentials/ts-essentials/pull/348)

PaulLeCam: OK great that you got the upstream package fixed! I'll update the PR.
Shouldn't `ts-essentials` be a regular dependency rather than a dev one to ensure consuming packages also have it installed?

ukstv: `Opaque`, `NonEmptyArray` and most of the other goodies there are only type declarations. I conclude, it makes `ts-essentials` a dev dependency.

ukstv (approved):

PR #2822, opened 2023-05-09
Use codecs for pubsub

PaulLeCam:
@ukstv @stbrody any idea how to debug the tests please?
All I get is `ERROR: TypeError: terminated` when enabling the logs, which I expect is caused by the codec validation, but as there is no further information about the error I have no idea where it originates from?

### DIFF
--- a/packages/core/src/pubsub/pubsub-message.ts
+++ b/packages/core/src/pubsub/pubsub-message.ts
@@ -58,7 +58,7 @@ export const UpdateMessageCodec = co.intersection(
       stream: streamIdAsString,
       tip: cidAsString,
     }),
-    co.partial({ model: streamIdAsString }),
+    co.partial({ model: co.optional(streamIdAsString) }),
   ],
   'UpdateMessage'
 )
--- a/packages/core/src/pubsub/pubsub-message.ts
+++ b/packages/core/src/pubsub/pubsub-message.ts
@@ -51,13 +51,15 @@ export enum MsgType {
   KEEPALIVE,
 }
 
-export const UpdateMessageCodec = co.strict(
-  {
-    typ: co.literal(MsgType.UPDATE),
-    stream: streamIdAsString,
-    tip: cidAsString,
-    model: co.optional(streamIdAsString),
-  },
+export const UpdateMessageCodec = co.intersection(
+  [
+    co.strict({
+      typ: co.literal(MsgType.UPDATE),
+      stream: streamIdAsString,
+      tip: cidAsString,
+    }),
+    co.partial({ model: streamIdAsString }),
+  ],
   'UpdateMessage'
 )
 export type UpdateMessage = co.TypeOf<typeof UpdateMessageCodec>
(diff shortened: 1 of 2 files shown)


ukstv:
@PaulLeCam One idea is to call and console.log `validate` instead of doing `decode` to see what is considered wrong.

PaulLeCam:
@ukstv thanks but I'm mostly wondering where in the test suite the error is thrown?
All I get from Jest is an expectation failure with the stream contents, not the actual codec error so I don't know where the invalid message is created.

packages/core/src/pubsub/pubsub-message.ts
```diff
@@ -17,32 +51,60 @@ export enum MsgType {
   KEEPALIVE,
 }
 
+export const UpdateMessageCodec = co.intersection(
```
ukstv, line 54: `export const UpdateMessageCodec = co.sparse({
      typ: co.literal(MsgType.UPDATE),
      stream: streamIdAsString,
      tip: cidAsString,
      model: co.optional(streamIdAsString),
  },
  'UpdateMessage'
)`

packages/core/src/pubsub/pubsub-message.ts
```diff
@@ -17,32 +51,60 @@ export enum MsgType {
...
-}
+export const QueryMessageCodec = co.strict(
+  {
+    typ: co.literal(MsgType.QUERY),
+    id: co.string,
+    stream: streamIdAsString,
+  },
+  'QueryMessage'
+)
+export type QueryMessage = co.TypeOf<typeof QueryMessageCodec>
```
ukstv, line 81: @PaulLeCam Is it a style preference of yours to use different names for `QueryMessageCodec` and `QueryMessage` ? I personally find it easier if they both share the same name, say, `QueryMessage`, and I find it curious why do you think the separate names are better.

PaulLeCam: Main reason was to make the difference explicit, but I'll update 👍

ukstv (approved):
That's really nice.

There is one improvement I have in my mind out of the scope here, to make it codecs all the way down. Add two more codecs:
1. `jsonAsString` as a composable alternative to `JSON.stringify` + `JSON.parse`,
2. `stringToBytes` as a composable alternative to `TextDecoder` + `TextEncoder`. Then `serialize` and `deserialize` would be redundant.

PaulLeCam:
Thanks!
Yes good idea for the JSON and text encoders, I started porting the codecs to the DIDs package (cf https://github.com/ceramicnetwork/js-did/pull/153), I think the JSON codecs would be useful there.
For the text encoders, do you see them as part of the pubsub codecs or useful more generally in the `@ceramicnetwork/codecs` package please?

ukstv:
> For the text encoders, do you see them as part of the pubsub codecs or useful more generally in the @ceramicnetwork/codecs package please?

I have asked this question myself. I think, it makes sense to have them in pubsub codecs for now, as I find no other place where that could be useful from the top of my head. Whenever we find it useful in other places, one could move it to `common`.

PR #2826, opened 2023-05-18
Feat/build composedb container

3benbox:
Update the Docker file to include targets for ceramic and composedb for deployments.


Dockerfile.daemon
```diff
@@ -1,6 +1,6 @@
-FROM node:16
+FROM node:16 as ceramic
 
-RUN apt-get update && apt-get install -y netcat
+RUN apt-get update && apt-get install -y netcat gettext-base
```
smrz2001, line 3: What do we use gettext for?

smrz2001 (approved):
Just one qn otherwise LGTM.

Dockerfile.daemon
3benbox: This installs the `envsubt` command which is used in the container to template a config file.

PR #2828, opened 2023-05-23
feat(core): store CAR files that come back from CAS

stbrody:
Test coverage is pretty limited, but I think the easiest way to test this will be the CAS codebase in the ceramic integration test file


packages/codecs/src/ipld.ts
```diff
@@ -47,3 +47,27 @@ export const cidAsString = new Type<CID, string, string>(
   },
   (cid) => cid.toString()
 )
+
+/**
+ * codeco codec for CAR file encoded as a Uint8Array.
+ */
+// TODO(2820): Make this codec work
```
stbrody, line 54: @ukstv I may need your help to get this working right

stbrody: I filed https://linear.app/3boxlabs/issue/CDB-2520/use-a-codeco-codec-to-deserialize-the-car-file-from-a-uint8array for the a serialization/deserialization coded for CAR files.  Is that something you might be willing to take on?

packages/codecs/src/anchor.ts
```diff
@@ -66,6 +67,7 @@ export const AnchorServiceAnchored = strict(
     cid: cidAsString,
     message: string,
     anchorCommit: cidAsString,
+    carBytes: optional(uint8array), // TODO(CDB-2519): make this required once all CAS envs are returning car files
```
stbrody, line 70: We should decide what the right name for this field is as this is the contract between ceramic and CAS.  Maybe it should just be `car` instead of `carBytes`?  That or something like `witnessCar`?

packages/core/src/anchor/memory/in-memory-anchor-service.ts
```diff
@@ -366,6 +370,33 @@ export class InMemoryAnchorService implements AnchorService, AnchorValidator {
...
+   * the CAR file: the AnchorCommit, the AnchorProof for the batch, and the commit that was anchored.
+   */
+  async _buildWitnessCAR(proofCid: CID, anchorCommitCid: CID, updateCid: CID): Promise<CAR> {
+    const car = this.carFactory.build()
+
+    const cidToBlock = async (cid) => new CarBlock(cid, await this.#dispatcher.getIpfsBlock(cid))
+
+    car.blocks.put(await cidToBlock(proofCid))
+    car.blocks.put(await cidToBlock(anchorCommitCid))
+    car.blocks.put(await cidToBlock(updateCid))
```
stbrody, line 386: it's not strictly required for the CAR file to include the update that was anchored, as this node should already have that data since it was the one who requested that update be anchored in the first place.

I'd like to mirror whatever the CAS will do here - the leaf nodes of the anchor merkle tree are the updates that were anchored in the batch, so we'd probably have to be careful when constructing the CAR files in the CAS to exclude the updates if we want to save the storage and bandwidth costs from not including them.

packages/codecs/src/anchor.ts
oed: witnessCar makes sense to me!

ukstv: +1 to just `car`.

packages/codecs/src/ipld.ts
ukstv: @stbrody https://github.com/ceramicnetwork/codeco/tree/main#idea I have changed README to make it clearer how to create a codec. It would be sad if only Paul and I can do that.

ukstv (commented):
Makes sense overall.

packages/core/src/dispatcher.ts
```diff
@@ -156,6 +156,27 @@ export class Dispatcher {
...
+   * Stores all the blocks in the given CAR file into the local IPFS node.  Uses the blocks API
+   * instead of the dag.import API because the dag.import API requires the CAR file to be a fully
+   * self-contained graph but our CAR file may have references to IPLD objects that aren't included
+   * in the CAR file itself.
+   * @param car
+   */
+  async storeCarFile(car: CAR): Promise<void> {
+    return await this._shutdownSignal.abortable(async (signal) => {
+      for (const block of car.blocks) {
+        await this._ipfs.block.put(block.payload, { signal })
```
stbrody, line 175: note to self: pin the CIDs!

stbrody: oh wait, I think the normal stream pinning machinery should take care of pinning these later on.  So I don't think I actually need to worry about pinning here.

packages/codecs/src/anchor.ts
stbrody: changed to `witnessCar`

packages/codecs/src/ipld.ts
stbrody: I think I understand most of it, the only part I'm not sure of is how to implement the `is` function to tell if a given input is an instance of the CAR class.

packages/core/src/anchor/memory/in-memory-anchor-service.ts
```diff
@@ -68,6 +69,7 @@ export class InMemoryAnchorService implements AnchorService, AnchorValidator {
   readonly #anchorOnRequest: boolean
   readonly #verifySignatures: boolean
   readonly #feed: Subject<AnchorServiceResponse> = new Subject()
+  readonly carFactory = new CARFactory()
```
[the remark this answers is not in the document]
stbrody: done

packages/core/src/anchor/memory/in-memory-anchor-service.ts
stbrody: Removed the update from the CAR file

packages/codecs/src/ipld.ts
stbrody: okay, looked at a few other examples that Paul had done - is it as simple as just using `instanceOf`?  Anyway put up a version that does that

packages/codecs/src/ipld.ts
```diff
@@ -47,3 +49,24 @@ export const cidAsString = new Type<CID, string, string>(
...
+const carFactory = new CARFactory()
+carFactory.codecs.add(DAG_JOSE)
+
+/**
+ * codeco codec for CAR file encoded as a Uint8Array.
+ */
+export const carAsUint8Array = new Type<CAR, Uint8Array, Uint8Array>(
+  'CAR-as-uint8array',
+  (input: unknown): input is CAR => {
+    return input != null && input instanceof CAR
```
stbrody, line 62: is this right?

stbrody:
Okay, I think this is ready for review for real now @ukstv

packages/codecs/src/ipld.ts
ukstv: Yup.

ukstv (approved):

packages/core/src/dispatcher.ts
oed, line 175: Why don't we just use ipfs.dag.import so that we only need to make one http call?

ukstv: IPFS `ipfs.dag.import` assumes that a CAR file contains a _full subgraph_. This is not the case for us: we do not store commits previous to anchor commits.

ukstv: "Elsewhere" means when creating a commit, I guess. When we store a commit, (a) we have the full commit data, (b) we have previous commits available on the same node. So, it is a different thing.

ukstv: > Just because we have all the previous commits on the node doesn't mean we have the entire DAG because anchors link to multiple streams technically

Heck, you are right.

> What happens if you try to import a car file without a full DAG?

IPFS would hang till it downloads the DAG.

oed: Or maybe we need to patch IPFS?

oed: Looks like it's an option?
https://github.com/ipfs/js-ipfs/blob/master/docs/core-api/DAG.md#options-3

oed: Looks like you are already setting this to false

stbrody: Although, @ukstv do you know if I set that `pinRoots` flag to false, would I be able to use `dag.import` here safely?  It'd be nicer than having to send each block one at a time

oed: Ok, great! 🙏 🙏 

So should this PR change it's implementation to use `dag.import`?

ukstv: It should.
@stbrody

oed (approved):
`dag.import` 👌

PR #2829, opened 2023-05-24
CDB-2492: Add support for CDB indices

dbcfd:
Adds support for indices passed from CDB.

### DIFF
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -231,12 +231,12 @@ type AdminApiJWSValidationResult = {
 }
 
 type AdminApiModelDataMutationMethod = {
-  type: 'modelData',
+  type: 'modelData'
   method: (modelData: Array<ModelData>) => Promise<void>
 }
 
 type AdminApiModelMutationMethod = {
-  type: 'modelIDs',
+  type: 'modelIDs'
   method: (modelsIDs: Array<StreamID>) => Promise<void>
 }
 
@@ -796,7 +796,9 @@ export class CeramicDaemon {
       await successCallback.method(jwsValidation.models ?? [])
     } else {
       if (!jwsValidation.modelData || jwsValidation.modelData.length == 0) {
-        res.status(StatusCodes.BAD_REQUEST).json({error: 'Expected modelData to be present and contain elements'})
+        res
+          .status(StatusCodes.BAD_REQUEST)
+          .json({ error: 'Expected modelData to be present and contain elements' })
         return
       }
       await successCallback.method(jwsValidation.modelData)
@@ -855,9 +857,7 @@ export class CeramicDaemon {
     const body = {
       models: indexedModels.map((idx) => idx.streamID.toString()),
     }
-    res
-      .header("Deprecation", "Wed, 21 Jun 2023 23:59:59 GMT")
-      .json(body)
+    res.header('Deprecation', 'Wed, 21 Jun 2023 23:59:59 GMT').json(body)
   }
 
   async getIndexedModelData(req: Request, res: Response): Promise<void> {
@@ -873,7 +873,7 @@ export class CeramicDaemon {
           streamID: idx.streamID.toString(),
           indices: idx.indices,
         }
-      })
+      }),
     }
     res.json(body)
   }
@@ -907,23 +907,19 @@ export class CeramicDaemon {
   async startIndexingModels(req: Request, res: Response): Promise<void> {
     await this._processAdminModelsMutationRequest(
       req,
-      res.header("Deprecation", "Wed, 21 Jun 2023 23:59:59 GMT"),
+      res.header('Deprecation', 'Wed, 21 Jun 2023 23:59:59 GMT'),
       {
         type: 'modelIDs',
-        method: this.ceramic.admin.startIndexingModels.bind(this.ceramic.admin)
+        method: this.ceramic.admin.startIndexingModels.bind(this.ceramic.admin),
       }
     )
   }
 
   async startIndexingModelData(req: Request, res: Response): Promise<void> {
-    await this._processAdminModelsMutationRequest(
-      req,
-      res,
-      {
-        type: 'modelData',
-        method: this.ceramic.admin.startIndexingModelData.bind(this.ceramic.admin)
-      }
-    )
+    await this._processAdminModelsMutationRequest(req, res, {
+      type: 'modelData',
+      method: this.ceramic.admin.startIndexingModelData.bind(this.ceramic.admin),
+    })
   }
 
   async stopIndexingModels(req: Request, res: Response): Promise<void> {
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -828,7 +828,7 @@ export class CeramicDaemon {
 
     const indexedModels = await this.ceramic.admin.getIndexedModels()
     res.json({
-      models: indexedModels.map((modelStreamID) => modelStreamID.toString()),
+      models: indexedModels.map((idx) => idx.streamID.toString()),
     })
   }
 
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -730,9 +730,7 @@ export class CeramicDaemon {
         error: `At least one model must be specified, either in 'models' as a 'StreamID' or 'modelData' as a 'ModelData' consisting of 'StreamID' and other fields`,
       }
     } else {
-      const models = parsedJWS.models?.map((modelIDString) =>
-        StreamID.fromString(modelIDString)
-      )
+      const models = parsedJWS.models?.map((modelIDString) => StreamID.fromString(modelIDString))
       return {
         kid: parsedJWS.kid,
         code: parsedJWS.code,
@@ -834,8 +832,8 @@ export class CeramicDaemon {
       return
     }
 
-    const withModelData = req.query.withModelData
-    if (withModelData && withModelData.toString().toLowerCase() == 'true') {
+    const outputFormat = req.query.outputFormat?.toString() ?? 'onlystreamids'
+    if (outputFormat && outputFormat.toString().toLowerCase() == 'document') {
       const indexedModels = await this.ceramic.admin.getIndexedModelData()
       res.json({
         models: indexedModels.map((idx) => idx.streamID.toString()),
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -697,7 +697,8 @@ export class CeramicDaemon {
       kid: result.kid,
       code: result.payload.code,
       requestPath: result.payload.requestPath,
-      models: result.payload.requestBody ? result.payload.requestBody.models : undefined,
+      models: result.payload.requestBody?.models || [],
+      indices: result.payload.requestBody?.indices || [],
     }
   }
 
@@ -720,18 +721,22 @@ export class CeramicDaemon {
       }
     } else if (!parsedJWS.code) {
       return { error: 'Admin code is missing from the the jws block' }
-    } else if (shouldContainModels && (!parsedJWS.models || parsedJWS.models.length === 0)) {
+    } else if (
+      shouldContainModels &&
+      parsedJWS.models.length == 0 &&
+      parsedJWS.indices.length == 0
+    ) {
       return {
-        error: `The 'models' parameter is required and it has to be an array containing at least one model stream id`,
+        error: `At least one model must be specified, either in 'models' as a 'StreamID' or 'indices' as a 'ModelFieldIndex'`,
       }
     } else {
       return {
         kid: parsedJWS.kid,
         code: parsedJWS.code,
-        models: parsedJWS.models?.map((modelIDString) => StreamID.fromString(modelIDString)),
-        indices: parsedJWS.indices?.map((idx) => {
+        models: parsedJWS.models.map((modelIDString) => StreamID.fromString(modelIDString)),
+        indices: parsedJWS.indices.map((idx) => {
           return {
-            streamID: StreamID.fromString(idx.streamId),
+            streamID: StreamID.fromString(idx.streamID),
             indices: idx.indices,
           }
         }),
@@ -826,10 +831,24 @@ export class CeramicDaemon {
       return
     }
 
-    const indexedModels = await this.ceramic.admin.getIndexedModels()
-    res.json({
-      models: indexedModels.map((idx) => idx.streamID.toString()),
-    })
+    const withFieldIndices = req.query.withFieldIndices
+    if (withFieldIndices && withFieldIndices.toString().toLowerCase() == 'true') {
+      const indexedModels = await this.ceramic.admin.getIndexedModelsWithFieldIndices()
+      res.json({
+        models: indexedModels.map((idx) => idx.streamID.toString()),
+        indices: indexedModels.map((idx) => {
+          return {
+            streamID: idx.streamID.toString(),
+            indices: idx.indices,
+          }
+        }),
+      })
+    } else {
+      const indexedModels = await this.ceramic.admin.getIndexedModels()
+      res.json({
+        models: indexedModels.map((idx) => idx.toString()),
+      })
+    }
   }
 
   async validateAdminRequest(req: Request, res: Response, next: NextFunction): Promise<void> {
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -230,10 +230,15 @@ type AdminApiJWSValidationResult = {
   error?: string
 }
 
-type AdminApiModelMutationMethod = (
-  modelsIDs: Array<StreamID>,
-  indices?: Array<ModelData>
-) => Promise<void>
+type AdminApiModelDataMutationMethod = {
+  type: 'modelData',
+  method: (modelData: Array<ModelData>) => Promise<void>
+}
+
+type AdminApiModelMutationMethod = {
+  type: 'modelIDs',
+  method: (modelsIDs: Array<StreamID>) => Promise<void>
+}
 
 /**
  * Ceramic daemon implementation
@@ -364,6 +369,7 @@ export class CeramicDaemon {
     const collectionRouter = ErrorHandlingRouter(this.diagnosticsLogger)
     const adminCodesRouter = ErrorHandlingRouter(this.diagnosticsLogger)
     const adminModelRouter = ErrorHandlingRouter(this.diagnosticsLogger)
+    const adminModelDataRouter = ErrorHandlingRouter(this.diagnosticsLogger)
     const adminPinsRouter = ErrorHandlingRouter(this.diagnosticsLogger)
     const legacyPinsRouter = ErrorHandlingRouter(this.diagnosticsLogger)
     const adminNodeStatusRouter = ErrorHandlingRouter(this.diagnosticsLogger)
@@ -378,6 +384,7 @@ export class CeramicDaemon {
     baseRouter.use('/collection', collectionRouter)
     baseRouter.use('/admin/getCode', adminCodesRouter)
     baseRouter.use('/admin/models', adminModelRouter)
+    baseRouter.use('/admin/modelData', adminModelDataRouter)
     baseRouter.use('/admin/status', adminNodeStatusRouter)
     // Admin Pins Validate JWS Middleware
     baseRouter.use('/admin/pins', this.validateAdminRequest.bind(this))
@@ -408,6 +415,9 @@ export class CeramicDaemon {
     adminModelRouter.getAsync('/', this.getIndexedModels.bind(this))
     adminModelRouter.postAsync('/', this.startIndexingModels.bind(this))
     adminModelRouter.deleteAsync('/', this.stopIndexingModels.bind(this))
+    adminModelDataRouter.getAsync('/', this.getIndexedModelData.bind(this))
+    adminModelDataRouter.postAsync('/', this.startIndexingModelData.bind(this))
+    adminModelDataRouter.deleteAsync('/', this.stopIndexingModels.bind(this))
     adminPinsRouter.getAsync('/:streamid', this.listPinned.bind(this))
     adminPinsRouter.getAsync('/', this.listPinned.bind(this))
 
@@ -763,7 +773,7 @@ export class CeramicDaemon {
   private async _processAdminModelsMutationRequest(
     req: Request,
     res: Response,
-    successCallback: AdminApiModelMutationMethod
+    successCallback: AdminApiModelMutationMethod | AdminApiModelDataMutationMethod
   ): Promise<void> {
     // Parse request
     const jwsValidation = await this._validateAdminApiJWS(req.baseUrl, req.body.jws, true)
@@ -778,10 +788,19 @@ export class CeramicDaemon {
       this._verifyActingDid(jwsValidation.kid)
     } catch (e) {
       res.status(StatusCodes.UNAUTHORIZED).json({ error: e.message })
+      return
     }
 
     // Process request
-    await successCallback(jwsValidation.models ?? [], jwsValidation.modelData)
+    if (successCallback.type == 'modelIDs') {
+      await successCallback.method(jwsValidation.models ?? [])
+    } else {
+      if (!jwsValidation.modelData || jwsValidation.modelData.length == 0) {
+        res.status(StatusCodes.BAD_REQUEST).json({error: 'Expected modelData to be present and contain elements'})
+        return
+      }
+      await successCallback.method(jwsValidation.modelData)
+    }
     res.status(StatusCodes.OK).json({ result: 'success' })
   }
 
@@ -833,12 +852,23 @@ export class CeramicDaemon {
     }
 
     const indexedModels = await this.ceramic.admin.getIndexedModelData()
-    const body: Record<string, any> = {
+    const body = {
       models: indexedModels.map((idx) => idx.streamID.toString()),
     }
-    const outputFormat = req.query.outputFormat?.toString() ?? 'document'
-    if (outputFormat.toString().toLowerCase() != 'onlystreamids') {
-      body.modelData = indexedModels.map((idx) => {
+    res
+      .header("Deprecation", "Wed, 21 Jun 2023 23:59:59 GMT")
+      .json(body)
+  }
+
+  async getIndexedModelData(req: Request, res: Response): Promise<void> {
+    const authorized = await this._checkAdminAPIGETRequestAuthorization(req, res)
+    if (!authorized) {
+      return
+    }
+
+    const indexedModels = await this.ceramic.admin.getIndexedModelData()
+    const body = {
+      modelData: indexedModels.map((idx) => {
         return {
           streamID: idx.streamID.toString(),
           indices: idx.indices,
@@ -877,8 +907,22 @@ export class CeramicDaemon {
   async startIndexingModels(req: Request, res: Response): Promise<void> {
     await this._processAdminModelsMutationRequest(
       req,
+      res.header("Deprecation", "Wed, 21 Jun 2023 23:59:59 GMT"),
+      {
+        type: 'modelIDs',
+        method: this.ceramic.admin.startIndexingModels.bind(this.ceramic.admin)
+      }
+    )
+  }
+
+  async startIndexingModelData(req: Request, res: Response): Promise<void> {
+    await this._processAdminModelsMutationRequest(
+      req,
       res,
-      this.ceramic.admin.startIndexingModels.bind(this.ceramic.admin)
+      {
+        type: 'modelData',
+        method: this.ceramic.admin.startIndexingModelData.bind(this.ceramic.admin)
+      }
     )
   }
 
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -832,24 +832,20 @@ export class CeramicDaemon {
       return
     }
 
-    const outputFormat = req.query.outputFormat?.toString() ?? 'onlystreamids'
-    if (outputFormat && outputFormat.toString().toLowerCase() == 'document') {
-      const indexedModels = await this.ceramic.admin.getIndexedModelData()
-      res.json({
-        models: indexedModels.map((idx) => idx.streamID.toString()),
-        modelData: indexedModels.map((idx) => {
-          return {
-            streamID: idx.streamID.toString(),
-            indices: idx.indices,
-          }
-        }),
-      })
-    } else {
-      const indexedModels = await this.ceramic.admin.getIndexedModels()
-      res.json({
-        models: indexedModels.map((idx) => idx.toString()),
+    const indexedModels = await this.ceramic.admin.getIndexedModelData()
+    const body: Record<string, any> = {
+      models: indexedModels.map((idx) => idx.streamID.toString()),
+    }
+    const outputFormat = req.query.outputFormat?.toString() ?? 'document'
+    if (outputFormat.toString().toLowerCase() != 'onlystreamids') {
+      body.modelData = indexedModels.map((idx) => {
+        return {
+          streamID: idx.streamID.toString(),
+          indices: idx.indices,
+        }
       })
     }
+    res.json(body)
   }
 
   async validateAdminRequest(req: Request, res: Response, next: NextFunction): Promise<void> {
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -218,7 +218,7 @@ type AdminAPIJWSContents = {
   kid: string
   code: string
   requestPath: string
-  models: Array<string>
+  models?: Array<string>
   modelData?: Array<AdminAPIJWSModelFields>
 }
 
@@ -730,7 +730,7 @@ export class CeramicDaemon {
         error: `At least one model must be specified, either in 'models' as a 'StreamID' or 'modelData' as a 'ModelData' consisting of 'StreamID' and other fields`,
       }
     } else {
-      const models = (parsedJWS.models ?? []).map((modelIDString) =>
+      const models = parsedJWS.models?.map((modelIDString) =>
         StreamID.fromString(modelIDString)
       )
       return {
@@ -783,7 +783,7 @@ export class CeramicDaemon {
     }
 
     // Process request
-    await successCallback(jwsValidation.models, jwsValidation.modelData)
+    await successCallback(jwsValidation.models ?? [], jwsValidation.modelData)
     res.status(StatusCodes.OK).json({ result: 'success' })
   }
 
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -9,7 +9,7 @@ import {
   FieldsIndex,
   LoggerProvider,
   LogStyle,
-  ModelFieldsIndex,
+  ModelData,
   MultiQuery,
   StreamUtils,
   SyncOptions,
@@ -219,20 +219,20 @@ type AdminAPIJWSContents = {
   code: string
   requestPath: string
   models: Array<string>
-  indices?: Array<AdminAPIJWSModelFields>
+  modelData?: Array<AdminAPIJWSModelFields>
 }
 
 type AdminApiJWSValidationResult = {
   kid?: string
   code?: string
   models?: Array<StreamID>
-  indices?: Array<ModelFieldsIndex>
+  modelData?: Array<ModelData>
   error?: string
 }
 
 type AdminApiModelMutationMethod = (
   modelsIDs: Array<StreamID>,
-  indices?: Array<ModelFieldsIndex>
+  indices?: Array<ModelData>
 ) => Promise<void>
 
 /**
@@ -698,7 +698,7 @@ export class CeramicDaemon {
       code: result.payload.code,
       requestPath: result.payload.requestPath,
       models: result.payload.requestBody?.models || [],
-      indices: result.payload.requestBody?.indices || [],
+      modelData: result.payload.requestBody?.indices || [],
     }
   }
 
@@ -722,19 +722,21 @@ export class CeramicDaemon {
     } else if (!parsedJWS.code) {
       return { error: 'Admin code is missing from the the jws block' }
     } else if (
-      shouldContainModels &&
-      parsedJWS.models.length == 0 &&
-      parsedJWS.indices.length == 0
+      shouldContainModels && (
+        (!parsedJWS.models || parsedJWS.models.length == 0) &&
+        (!parsedJWS.modelData || parsedJWS.modelData.length == 0)
+      )
     ) {
       return {
-        error: `At least one model must be specified, either in 'models' as a 'StreamID' or 'indices' as a 'ModelFieldIndex'`,
+        error: `At least one model must be specified, either in 'models' as a 'StreamID' or 'modelData' as a 'ModelData' consisting of 'StreamID' and other fields`,
       }
     } else {
+      const models = (parsedJWS.models ?? []).map((modelIDString) => StreamID.fromString(modelIDString))
       return {
         kid: parsedJWS.kid,
         code: parsedJWS.code,
-        models: parsedJWS.models.map((modelIDString) => StreamID.fromString(modelIDString)),
-        indices: parsedJWS.indices.map((idx) => {
+        models: models,
+        modelData: parsedJWS.modelData?.map((idx) => {
           return {
             streamID: StreamID.fromString(idx.streamID),
             indices: idx.indices,
@@ -780,7 +782,7 @@ export class CeramicDaemon {
     }
 
     // Process request
-    await successCallback(jwsValidation.models, jwsValidation.indices)
+    await successCallback(jwsValidation.models, jwsValidation.modelData)
     res.status(StatusCodes.OK).json({ result: 'success' })
   }
 
@@ -831,15 +833,15 @@ export class CeramicDaemon {
       return
     }
 
-    const withFieldIndices = req.query.withFieldIndices
-    if (withFieldIndices && withFieldIndices.toString().toLowerCase() == 'true') {
-      const indexedModels = await this.ceramic.admin.getIndexedModelsWithFieldIndices()
+    const withModelData = req.query.withModelData
+    if (withModelData && withModelData.toString().toLowerCase() == 'true') {
+      const indexedModels = await this.ceramic.admin.getIndexedModelData()
       res.json({
         models: indexedModels.map((idx) => idx.streamID.toString()),
-        indices: indexedModels.map((idx) => {
+        modelData: indexedModels.map((idx) => {
           return {
             streamID: idx.streamID.toString(),
-            indices: idx.indices,
+            indices: idx.indices
           }
         }),
       })
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -722,16 +722,17 @@ export class CeramicDaemon {
     } else if (!parsedJWS.code) {
       return { error: 'Admin code is missing from the the jws block' }
     } else if (
-      shouldContainModels && (
-        (!parsedJWS.models || parsedJWS.models.length == 0) &&
-        (!parsedJWS.modelData || parsedJWS.modelData.length == 0)
-      )
+      shouldContainModels &&
+      (!parsedJWS.models || parsedJWS.models.length == 0) &&
+      (!parsedJWS.modelData || parsedJWS.modelData.length == 0)
     ) {
       return {
         error: `At least one model must be specified, either in 'models' as a 'StreamID' or 'modelData' as a 'ModelData' consisting of 'StreamID' and other fields`,
       }
     } else {
-      const models = (parsedJWS.models ?? []).map((modelIDString) => StreamID.fromString(modelIDString))
+      const models = (parsedJWS.models ?? []).map((modelIDString) =>
+        StreamID.fromString(modelIDString)
+      )
       return {
         kid: parsedJWS.kid,
         code: parsedJWS.code,
@@ -841,7 +842,7 @@ export class CeramicDaemon {
         modelData: indexedModels.map((idx) => {
           return {
             streamID: idx.streamID.toString(),
-            indices: idx.indices
+            indices: idx.indices,
           }
         }),
       })
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -6,8 +6,10 @@ import { ServiceMetrics as Metrics } from '@ceramicnetwork/observability'
 import { IpfsConnectionFactory } from './ipfs-connection-factory.js'
 import {
   DiagnosticsLogger,
+  FieldsIndex,
   LoggerProvider,
   LogStyle,
+  ModelFieldsIndex,
   MultiQuery,
   StreamUtils,
   SyncOptions,
@@ -202,6 +204,14 @@ function validatePort(inPort) {
 }
 
 /**
+ * Field index contents in JWS
+ */
+type AdminAPIJWSModelFields = {
+  streamID: string
+  indices: Array<FieldsIndex>
+}
+
+/**
  * Contents an authorization header signed by an admin DID
  */
 type AdminAPIJWSContents = {
@@ -209,16 +219,21 @@ type AdminAPIJWSContents = {
   code: string
   requestPath: string
   models: Array<string>
+  indices?: Array<AdminAPIJWSModelFields>
 }
 
 type AdminApiJWSValidationResult = {
   kid?: string
   code?: string
   models?: Array<StreamID>
+  indices?: Array<ModelFieldsIndex>
   error?: string
 }
 
-type AdminApiModelMutationMethod = (modelIDs: Array<StreamID>) => Promise<void>
+type AdminApiModelMutationMethod = (
+  modelsIDs: Array<StreamID>,
+  indices?: Array<ModelFieldsIndex>
+) => Promise<void>
 
 /**
  * Ceramic daemon implementation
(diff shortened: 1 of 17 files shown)


packages/core/src/indexing/database-index-api.ts
```diff
@@ -59,7 +81,7 @@ export abstract class DatabaseIndexApi<DateType = Date | number> {
   private modelsToIndex: Array<StreamID> = []
   // Maps Model streamIDs to the list of fields in the content of MIDs in that model that should be
   // indexed
-  private readonly modelsIndexedFields = new Map<string, Array<string>>()
+  private readonly modelsIndexedFields = new Map<StreamID, Array<string>>()
```
PaulLeCam, line 84: Does the mapping work as expected for different `StreamID` instances?

packages/core/src/indexing/database-index-api.ts
```diff
@@ -81,6 +103,25 @@ export abstract class DatabaseIndexApi<DateType = Date | number> {
     this.syncApi = api
   }
 
+  /**
+   * Add a composite index to the database
+   * @param model
+   * @param indices
+   */
+  async addCompositeIndex(model: StreamID, indices: Array<ModelIndex>): Promise<void> {
```
PaulLeCam, line 111:
```suggestion
  async addModelIndex(model: StreamID, indices: Array<ModelIndex>): Promise<void> {
```

Ceramic has no notion of composites, seems confusing to use the term here

packages/core/src/indexing/database-index-api.ts
```diff
@@ -81,6 +103,25 @@ export abstract class DatabaseIndexApi<DateType = Date | number> {
...
+    await this.dbConnection(INDEXED_MODEL_CONFIG_TABLE_NAME)
+      .update({ indices: JSON.stringify(indices) })
+      .where('model', model.toString())
+  }
+
+  async getCompositeIndex(model: StreamID): Promise<Array<ModelIndex> | undefined> {
```
PaulLeCam, line 117:
```suggestion
  async getModelIndex(model: StreamID): Promise<Array<ModelIndex> | undefined> {
```

packages/core/src/local-admin-api.ts
```diff
@@ -21,8 +22,12 @@ export class LocalAdminApi implements AdminApi {
   }
 
   async startIndexingModels(modelsIDs: Array<StreamID>): Promise<void> {
-    await this.indexApi.indexModels(modelsIDs)
-    await this.syncApi.startModelSync(modelsIDs.map((id) => id.toString()))
+    await this.startCompositeIndexing(new Map(modelsIDs.map((id) => [id, null])))
+  }
+
+  async startCompositeIndexing(models: Map<StreamID, Array<ModelIndex>>): Promise<void> {
```
PaulLeCam, line 28:
```suggestion
  async startModelIndexing(models: Map<StreamID, Array<ModelIndex>>): Promise<void> {
```

packages/core/src/indexing/database-index-api.ts
```diff
@@ -29,6 +31,25 @@ export interface IndexStreamArgs {
...
+ */
+export interface Field {
+  path: Array<string>
+}
+
+/**
+ * Index for a model
+ */
+export type ModelIndex = {
+  name: string
```
PaulLeCam, line 45: Is the `name` provided by the client please? If so, how should they be defined? What is the behavior if duplicate names are used?

packages/core/src/indexing/database-index-api.ts
```diff
@@ -16,8 +16,10 @@ import { TablesManager, PostgresTablesManager, SqliteTablesManager } from './tab
 import { addColumnPrefix } from './column-name.util.js'
 import { ISyncQueryApi } from '../sync/interfaces.js'
 import cloneDeep from 'lodash.clonedeep'
+import { indexNameFromTableName } from './migrations/1-create-model-table.js'
 
 export const INDEXED_MODEL_CONFIG_TABLE_NAME = 'ceramic_models'
+export const COMPOSITE_INDEX_TABLE_NAME = 'ceramic_composite_index'
```
PaulLeCam, line 22:
```suggestion
export const MODEL_INDEX_TABLE_NAME = 'ceramic_model_index'
```

packages/core/src/indexing/__tests__/database-index-api.test.ts
```diff
@@ -1033,24 +1119,37 @@ describe('sqlite', () => {
...
+              },
+            ],
+          },
+        ],
+      }
       const indexApi = new SqliteIndexApi(dbConnection, true, logger, Networks.INMEMORY)
       indexApi.setSyncQueryApi(new CompleteQueryApi())
       await indexApi.init()
-      await indexApi.indexModels(modelsToIndexArgs(modelsToIndex))
+      await indexApi.indexModels([modelToIndexArgs])
```
PaulLeCam, line 1142: Could you add tests to check the behavior when this API is called multiple times with the same arguments please?

packages/core/src/indexing/migrations/1-create-model-table.ts
```diff
@@ -214,6 +222,42 @@ export async function createSqliteModelTable(
...
+  indices: Array<ModelIndex>
+): Promise<void> {
+  for (const index of indices) {
+    const json_indices = []
+    for (const field of index.fields) {
+      const path = field.path.join('.')
+      json_indices.push(`json_extract(stream_content, '$.${path}')`)
+    }
+    const indexName = modelIndexName(index, tableName)
+    await dataSource.schema.raw(
```
PaulLeCam, line 255: What is the expected behavior when running this query please?
What happens if the DBMS has millions of rows to index, does this call hang? Or on the contrary, does it resolve before the DB is fully indexed?

PaulLeCam (changes requested):
LGTM overall, mostly wondering about the expected interactions + some comments about keeping the current naming for these APIs.

packages/cli/src/ceramic-daemon.ts
```diff
@@ -349,25 +351,30 @@ export class CeramicDaemon {
...
+    v0Router.use('/admin/getCode', adminCodesRouter)
+    v0Router.use('/admin/models', adminModelRouter)
+    v0Router.use('/admin/status', adminNodeStatusRouter)
     // Admin Pins Validate JWS Middleware
-    baseRouter.use('/admin/pins', this.validateAdminRequest.bind(this))
-    baseRouter.use('/admin/pins', adminPinsRouter)
+    v0Router.use('/admin/pins', this.validateAdminRequest.bind(this))
+    v0Router.use('/admin/pins', adminPinsRouter)
+
+    apiRouter.use('/v1', v1Router)
```
stbrody, line 372: I'd like to pull in @oed and @ukstv to confirm the use of v1 http APIs in this way.  We haven't yet ever tried to increase our HTTP api version, and I'm not sure if best practice would be to update all of our APIs at once vs having some apis on v0 while others are on v1

packages/cli/src/ceramic-daemon.ts
```diff
@@ -390,6 +397,7 @@ export class CeramicDaemon {
     adminModelRouter.getAsync('/', this.getIndexedModels.bind(this))
     adminModelRouter.postAsync('/', this.startIndexingModels.bind(this))
     adminModelRouter.deleteAsync('/', this.stopIndexingModels.bind(this))
+    adminModelRouter.postAsync('/', this.startModelIndexing.bind(this))
```
stbrody, line 400: I think this is supposed to be on the v1 router?

packages/common/src/ceramic-api.ts
```diff
@@ -77,6 +92,12 @@ export interface AdminApi {
    */
   startIndexingModels(modelsIDs: Array<StreamID>): Promise<void>
 
+  /**
+   * Index models, with custom indices
+   *
+   */
+  startModelIndexing(models: Map<StreamID, Array<ModelIndex> | undefined>): Promise<void>
```
stbrody, line 99: Well I really hate having both `startIndexingModels` and `startModelIndexing` with slightly different behavior.

I guess one option is to rename the new one something like `startIndexingModelsWithCustomIndexes`, but that's a mouthful, especially since the old version should eventually go away.

What I'd really like to do is rename the old function to `startIndexingModelsV0_deprecated` and call the new one `startIndexingModels`, but then that's a breaking change to any apps that build with the ceramic http client directly, so we'd probably need to do a major version bump.  Of course at that point if we're doing a major version bump we could just remove the old form of the API entirely and replace it wholesale with the new format.

@oed for thoughts on doing a major version bump.  CC @michaelsena

packages/common/src/ceramic-api.ts
```diff
@@ -77,6 +92,12 @@ export interface AdminApi {
    */
```
stbrody, line 92: Can we mark this old version as deprecated?

packages/core/src/indexing/database-index-api.ts
stbrody: Yeah, I'm pretty sure if you pass two different StreamID instances that represent the same actual StreamID, they'll go in as different fields in the Map, since I believe by default Objects use pointer memory location equality when they are keys in a map unless you specify a custom comparison function.  @ukstv would know for sure or you could do a simple test to confirm

packages/core/src/indexing/database-index-api.ts
```diff
@@ -29,6 +31,10 @@ export interface IndexStreamArgs {
   readonly firstAnchor: Date | null
 }
 
+export function modelIndexName(midx: ModelIndex, table: string): string {
```
stbrody, line 34: please add a comment on what this does and what the arguments represent

packages/core/src/indexing/database-index-api.ts
```diff
@@ -81,6 +88,25 @@ export abstract class DatabaseIndexApi<DateType = Date | number> {
     this.syncApi = api
   }
 
+  /**
+   * Add a composite index to the database
+   * @param model
+   * @param indices
+   */
+  async addModelIndex(model: StreamID, indices: Array<ModelIndex>): Promise<void> {
```
stbrody, line 96: I think the terminology throughout this PR gets really confusing regarding the difference between indexing a Model and building an index on a field or set of fields within the Model.  Can we change the names throughout to make this clearer?  For example, I think this function could be called `addFieldIndices` and the `ModelIndex` type could be called `FieldIndex`, or something like that

packages/core/src/indexing/database-index-api.ts
```diff
@@ -81,6 +88,25 @@ export abstract class DatabaseIndexApi<DateType = Date | number> {
     this.syncApi = api
   }
 
+  /**
+   * Add a composite index to the database
```
stbrody, line 92: Agreed with Paul on avoiding the term composite within js-ceramic as js-ceramic knows nothing of composites.

stbrody, line 92:
```suggestion
   * Add the given indices on fields within the Model to the database
```

packages/core/src/indexing/database-index-api.ts
```diff
@@ -97,7 +123,10 @@ export abstract class DatabaseIndexApi<DateType = Date | number> {
         this.modelsToIndex.push(modelArgs.model)
       }
       if (modelArgs.relations) {
-        this.modelsIndexedFields.set(modelArgs.model.toString(), Object.keys(modelArgs.relations))
+        this.modelsIndexedFields.set(modelArgs.model, Object.keys(modelArgs.relations))
```
stbrody, line 126: this `modelsIndexedFields` member variable is confusingly named now.  It sounds like it could contain every field with an index on it, but it no longer does.  It only contains the fields which get duplicated out of the content JSON into their own column and have indices on those columns.  Right now that is only for fields that are relations.  We should make sure the distinction between any field with an index on the JSON content vs a relation field that is extracted out to its own column is clear.
We could call this `modelsExtractedFields` or even just `modelsRelationFields`, or something like that.  Probably should also add some comments explaining how this works

packages/core/src/indexing/local-index-api.ts
```diff
@@ -28,16 +29,30 @@ async function _getIndexModelArgs(
...
+    model: modelStreamId,
+  }
+
   if (modelStreamId.type == Model.STREAM_TYPE_ID) {
     const modelState = await repository.load(modelStreamId, {})
     const content = modelState.state.next?.content ?? modelState.state.content
     Model.assertVersionValid(content, 'major')
     if (content.relations) {
-      return { model: modelStreamId, relations: content.relations }
+      opts = {
```
stbrody, line 41: there's no need to make a new opts object and copy the existing fields over.  This assignment to opts could instead just be `opts.relations = content.relations`

packages/core/src/indexing/local-index-api.ts
```diff
@@ -28,16 +29,30 @@ async function _getIndexModelArgs(
...
+        ...opts,
+        relations: content.relations,
+      }
+    }
+    const indices = this.databaseIndexApi.getCompositeIndex(modelStreamId)
```
stbrody, line 46: I don't understand the control flow here.  We're in the path that is supposed to be indexing the Model for the first time, but we're checking the database for state about the Model as if it's already indexed?

I'd have expected the list of indices to build would have to be part of the arguments to this function?

packages/core/src/indexing/local-index-api.ts
```diff
@@ -149,9 +164,17 @@ export class LocalIndexApi implements IndexApi {
...
 
     const indexModelsArgs = []
-    for (const modelStreamId of models) {
+    for (const entry of models.entries()) {
+      const modelStreamId = entry[0]
+      const modelIndices = entry[1]
       this.logger.imp(`Starting indexing for Model ${modelStreamId.toString()}`)
-      const indexModelArgs = await _getIndexModelArgs(modelStreamId, this.repository)
+      let indexModelArgs = await _getIndexModelArgs(modelStreamId, this.repository)
+      if (modelIndices) {
```
stbrody, line 172: can we push this down into `_getIndexModelArgs` so all the logic for building the args is contained?

packages/core/src/indexing/local-index-api.ts
stbrody: Okay, I see below you add the index args from the user after (and left a comment about pushing that down into this function).  I still don't understand why we need to check the database here for existing indices though?

packages/core/src/indexing/local-index-api.ts
```diff
@@ -186,7 +209,13 @@ export class LocalIndexApi implements IndexApi {
    */
   async populateDatabaseApiInternalState(): Promise<void> {
     const modelsToIndex = this.databaseIndexApi?.getActiveModelsToIndex()
```
stbrody, line 211: can we just update `getActiveModelsToIndex()` to give us all the information we need (or add a new function that gives us the info we need in a single call if `getActiveModelsToIndex` is used in other places that don't need the extra info).  That way we don't have to call into the databaseIndexApi multiple times back to back?

packages/common/src/ceramic-api.ts
```diff
@@ -7,6 +7,21 @@ import type { GenesisCommit } from './index.js'
...
+ */
+export interface Field {
+  path: Array<string>
+}
+
+/**
+ * Index for a model
+ */
+export type ModelIndex = {
+  name: string
```
stbrody, line 21: I feel like the `name` field should be optional, and we should let the underlying database apply whatever default it wants if the user doesn't explicitly set a name

stbrody (changes requested):
besides the comments I left inline, I also think we need to think carefully about what the behavior and flow should be to add new indices to an existing already-indexed model.  Might we worth writing up a very quick mini-document outlining the behavior and approach here

packages/common/src/ceramic-api.ts
oed: Is there no way we could just use the old api by adding an additional parameter? Looks to me like it chould be fairly straight forward to do so by having some extra logic in the [_validateAdminApiJWS](https://github.com/ceramicnetwork/js-ceramic/blob/37c2297b6f32e73543d7f0c06413a2b67c27cd5f/packages/cli/src/ceramic-daemon.ts#L694) method?


Then we just change the function signature of `startIndexingModels` to:

`startIndexingModels(modelsIDs: Array<StreamID>, indexes?: Map<StreamID, Array<ModelIndex>): Promise<void>`

packages/cli/src/ceramic-daemon.ts
oed: I'm not completely sure what best practice is either, but it feels a bit strange to use both `v0` and `v1` at the same time.

oed (changes requested):

packages/cli/src/ceramic-daemon.ts
dbcfd: There's a few ways to do it, one being slowly move over commands that are needed, updating clients as needed to the new methods, calling v0 until v1 api's are available. Another is to bring over all v0 api's except the ones we want to replace.

oed: See my comment below. I don't think we need to introduce a breaking change here at all?
Shouldn't we just be able to make the `v0` of the api parse the new and the old formatted JWS?

packages/common/src/ceramic-api.ts
dbcfd: Changing the function signature in this way is a little weird, since we either have to make sure the modelIDs match and discard indices that don't, or just allow indexing from either fields.

oed: Ok, I think we could do with only having the `startModelIndexing(models: Map<StreamID, Array<ModelIndex> | undefined>): Promise<void>` signature though? And deal with backwards compatibility in the [_validateAdminApiJWS](https://github.com/ceramicnetwork/js-ceramic/blob/37c2297b6f32e73543d7f0c06413a2b67c27cd5f/packages/cli/src/ceramic-daemon.ts#L694) method?

dbcfd: I think that's breaking for anyone using startModelIndexing. For now, just going to assume that you can index models from either field, joining those arrays together.

packages/http-client/src/remote-admin-api.ts
```diff
@@ -67,17 +93,26 @@ export class RemoteAdminApi implements AdminApi {
...
     await this._fetchJson(this.getModelsUrl(), {
       method: 'post',
       body: {
-        jws: await this.buildJWS(this._getDidFn(), code, this.getModelsUrl().pathname, modelsIDs),
+        jws: await this.buildModelsJWS(
+          this._getDidFn(),
+          code,
+          this.getModelsUrl().pathname,
+          modelIDs,
+          indices
```
oed, line 109: Looks like you need to pass the body as an object. Also, convert the models array to strings.

```suggestion
          {
            models: modelsIDs.map((streamID) => streamID.toString()),
            indices
          }
```

packages/http-client/src/remote-admin-api.ts
```diff
@@ -88,8 +123,11 @@ export class RemoteAdminApi implements AdminApi {
         )}`,
       },
     })
-    return response.models.map((modelStreamIDString: string) => {
-      return StreamID.fromString(modelStreamIDString)
+    return response.map((resp) => {
+      return {
+        models: resp.models.map((model) => model.streamID.toString()),
+        indices: resp.models,
+      }
```
oed, line 130: This looks like a breaking change to the http api?

oed (commented):
Ok, this is better. Looks like:

1. http api for `startIndexingModels` should not be backwards compatible
2. http api for `getIndexedModels` looks like it will have a new response format, so it will be a breaking change. Can we fix this somehow?

packages/http-client/src/remote-admin-api.ts
dbcfd: I think it's non-breaking. I think the original response was `{'models':[]}` and it's now `{'models':[],'indices':[]}`

packages/core/src/indexing/migrations/1-create-model-table.ts
dbcfd: As implemented currently, we only apply the index when a MID table is created, since models are immutable. The indices cannot be changed once the MID table is created. We will have to revisit this when models aren't immutable.

packages/http-client/src/remote-admin-api.ts
dbcfd: The test has this as the expected response ```const GET_RESPONSE = {
  models: [MODEL.toString()],
}```

dbcfd: The breaking change is actually elsewhere, where models should be an array of string StreamIDs, and indices should be the model indices

oed: Ok. Maybe we should do some cross testing for this PR?
* older client works with newer http api
* newer client works with older http api

packages/core/src/indexing/migrations/1-create-model-table.ts
dbcfd: We don't currently have anything that updates tables in js-ceramic. My inclination on this is to not add that functionality until needed, due to three reasons:
* We have immutable models, so developers would likely be creating new models or wrapping old ones, rather than creating an update on an existing model
* The indices do not prevent querying, only improve performance.
* ComposeDB can use the indices for querying/filtering even if they're not applied in js-ceramic

PaulLeCam: That seems like a pretty big departure from what we described in the RFC, notably the ability to add query filters on existing models and the fact the ComposeDB client would generate the query filters types based on the indices set in the composites rather than any field.
To me that would make the whole feature much harder to explain and less useful to developers, cc @oed @stbrody any thoughts here please?

stbrody: > The indices do not prevent querying, only improve performance.

When we discussed this on a call the other day I thought we said that indices are actually required or else the GraphQL runtime will reject the query? 

> ComposeDB can use the indices for querying/filtering even if they're not applied in js-ceramic

I don't follow what this means?

> That seems like a pretty big departure from what we described in the RFC, notably the ability to add query filters on existing models

Yeah it was also my understanding that we would support adding new indices to existing Models, which is why I requested in my other comment a document describing how that would work.

dbcfd: > When we discussed this on a call the other day I thought we said that indices are actually required or else the GraphQL runtime will reject the query?

That is correct. That will be implemented in js-composedb though.

>> ComposeDB can use the indices for querying/filtering even if they're not applied in js-ceramic
> I don't follow what this means?

ComposeDB can use the indices to create the necessary input types for querying/ordering. This PR isn't even necessary to do filtering/ordering, other than not adding the indices in js-ceramic would have a performance impact. The main point of this PR is to make sure that what is implemented in composedb works for communicating indices to js-ceramic and keep them outside of the model.

PaulLeCam: I think the main problems here are the gap between expectations and implementation behavior, and overall usability of the feature.

To summarize the issue, here is the current behavior in ComposeDB v0.4:

- Deploying composites is idempotent, if the models are already indexed on a node, no error is thrown
- Composites are mutable, models can be added to them and views can be added to existing models (at the composite level)

Now from my understanding of the changes here, we’d have:

- Idempotent composites deployment: still possible or throwing an error if fields indices are set?
- No possibility to add indices to already deployed models
- No possibility to add views to already deployed models if they contain indices (due to the previous constraint)

Ideally the behavior we'd want is:

- Idempotent composites deployment: whether models already exist or not, and field indices need to be added or not, the composite deployment logic should work and not throw errors of existing indices
- Ability to add indices to existing composites (with already deployed models)

So then when the ComposeDB client generates the supported query filters types, there should be no "gap" in functionalities/expectations between what the client queries and the Ceramic node can efficiently resolve, as long as the composite has been deployed to the node.

dbcfd: @PaulLeCam Please add this information to the project page so we can capture requirements for additional PR's to consider the feature completed.

packages/http-client/src/remote-admin-api.ts
dbcfd: I think this is handled by the admin-api in stream-tests. This test makes http calls against ceramic, using remote admin api.

stbrody: the stream-tests tests all use the same version of the client and server.  We don't have any mixed-version compatibility tests currently

packages/cli/src/ceramic-daemon.ts
```diff
@@ -705,15 +721,25 @@ export class CeramicDaemon {
...
-        error: `The 'models' parameter is required and it has to be an array containing at least one model stream id`,
+        error: `At least one model must be specified, either in 'models' as a 'StreamID' or 'indices' as a 'ModelFieldIndex'`,
       }
     } else {
       return {
         kid: parsedJWS.kid,
         code: parsedJWS.code,
-        models: parsedJWS.models?.map((modelIDString) => StreamID.fromString(modelIDString)),
+        models: parsedJWS.models.map((modelIDString) => StreamID.fromString(modelIDString)),
+        indices: parsedJWS.indices.map((idx) => {
```
stbrody, line 737: I'm okay with duplicating the model streamids in the input and output of these commands in the short term to ease backwards compatibility.  But I'd like us to think about what the path would be to eventually condensing down to a format without that duplication.  If we eventually dropped the `models` field entirely, just having a field called `indices` would be odd.  Maybe we could call this `indexedModels`, `modelsInfo`, `modelsData` or some other similarly general-purpose name, and mark everywhere that `models` is deprecated in favor of the new field?

I also considered proposing the name `modelsWithIndices` but then if we ever add more information to the output format we'd want to rename it again.  Having the output format be an object now means it should be a bit more future proof as we can add new fields to the object in the future without breaking backwards compatibility.

packages/common/src/ceramic-api.ts
```diff
@@ -71,12 +94,18 @@ export interface AdminApi {
    */
   getIndexedModels(): Promise<Array<StreamID>>
 
+  /**
+   * List indexed model streams with fields
+   */
+  getIndexedModelsWithFieldIndices(): Promise<Array<ModelFieldsIndex>>
```
stbrody, line 100: what if in the future there is extra information we want to return about models?  Then we'd need yet another method here.  What if we call this `getIndexedModelsInfo` or something like that, and it returns a `ModelInfo` which for now just has the streamid and the indices array, but could be extended with more data in the future?

packages/cli/src/ceramic-daemon.ts
```diff
@@ -805,10 +831,24 @@ export class CeramicDaemon {
...
+          return {
+            streamID: idx.streamID.toString(),
+            indices: idx.indices,
+          }
+        }),
+      })
+    } else {
+      const indexedModels = await this.ceramic.admin.getIndexedModels()
+      res.json({
+        models: indexedModels.map((idx) => idx.toString()),
```
stbrody, line 849: if we take my suggestion from above to make the `indices` top-level field name something more general and extensible, then we probably still want to return it here with the object-based output format, just with the `indices` array field omitted.

packages/core/src/indexing/database-index-api.ts
```diff
@@ -29,6 +32,19 @@ export interface IndexStreamArgs {
   readonly firstAnchor: Date | null
 }
 
+/**
+ * Create a valid name for a fields index, if the index has a name specified
```
stbrody, line 36: should this say "if the index *does not* have a name specified?"

packages/core/src/indexing/database-index-api.ts
```diff
@@ -81,6 +98,33 @@ export abstract class DatabaseIndexApi<DateType = Date | number> {
...
+   */
+  async getFieldsIndex(model: StreamID): Promise<Array<FieldsIndex> | undefined> {
+    const res = await this.dbConnection(INDEXED_MODEL_CONFIG_TABLE_NAME)
+      .first('indices')
+      .where('model', model.toString())
+
+    if (res) {
+      return JSON.parse(res.indices)
+    } else {
+      undefined
```
stbrody, line 124: should this just be an empty array?

packages/core/src/indexing/local-index-api.ts
stbrody: ping

ditto below with indices

packages/core/src/indexing/local-index-api.ts
```diff
@@ -17,27 +19,43 @@ import { Model } from '@ceramicnetwork/stream-model'
...
     const content = modelState.state.next?.content ?? modelState.state.content
     Model.assertVersionValid(content, 'major')
     if (content.relations) {
-      return { model: modelStreamId, relations: content.relations }
+      opts = {
+        ...opts,
+        relations: content.relations,
+      }
+    }
+    const indices = (await databaseIndexApi?.getFieldsIndex(modelStreamId)) ?? req.indices
```
stbrody, line 49: what's happening here?  If the database already has indices for this model we take those instead of the ones provided with the command?  How could that happen?  Even if it did, would ignoring the indices provided by the command be the right move?  Seems like it would be surprising behavior if you specified some indices and those were not used at all

packages/core/src/indexing/local-index-api.ts
```diff
@@ -136,34 +154,33 @@ export class LocalIndexApi implements IndexApi {
...
-          return streamId.equals(modelStreamId)
-        })
-        // TODO(CDB-2297): Handle a model's historical sync after re-indexing
-        if (modelNoLongerIndexed) {
-          throw new Error(
-            `Cannot re-index model ${modelStreamId.toString()}, data may not be up-to-date`
-          )
+    const indexModelsArgs = await Promise.all(
+      models.map(async (idx) => {
+        const modelStreamId = idx.streamID
```
stbrody, line 166: this is getting kind of hard to read - can we make a `_indexModel` helper function to clean this up a bit?

packages/core/src/local-admin-api.ts
```diff
@@ -20,13 +20,30 @@ export class LocalAdminApi implements AdminApi {
     return this.nodeStatusFn()
   }
 
-  async startIndexingModels(modelsIDs: Array<StreamID>): Promise<void> {
-    await this.indexApi.indexModels(modelsIDs)
-    await this.syncApi.startModelSync(modelsIDs.map((id) => id.toString()))
+  async startIndexingModels(
+    modelsIDs: Array<StreamID>,
+    indices?: Array<ModelFieldsIndex>
```
stbrody, line 25: why do we take both arguments here at all?  What do we do if they conflict?  Seems like we should just take a single arg that's an array of ModelFieldsIndex and it should be up to the calling code in the http-api to de-dupe and merge the arguments received on the endpoint.

packages/core/src/local-admin-api.ts
```diff
@@ -20,13 +20,30 @@ export class LocalAdminApi implements AdminApi {
...
+  ): Promise<void> {
+    let models = modelsIDs.map((id) => {
+      return {
+        streamID: id,
+        indices: undefined,
+      } as ModelFieldsIndex
+    })
+    if (indices) {
+      models = models.concat(indices)
```
stbrody, line 34: this doesn't seem right?  This will just add the array of objects with both streamID and indices to the end of the array after the objects that just have streamID and an undefined indices field.  Presumably we actually want to merge and dedupe these somehow?

packages/http-client/src/remote-admin-api.ts
```diff
@@ -64,9 +70,23 @@ export class RemoteAdminApi implements AdminApi {
...
+  async startIndexingModels(
+    modelsIDs: Array<StreamID>,
+    indices?: Array<ModelFieldsIndex>
+  ): Promise<void> {
     const code = await this.generateCode()
-    const body = modelIDsAsRequestBody(modelsIDs)
+    let body = modelIDsAsRequestBody(modelsIDs)
+    if (indices) {
+      body = {
+        ...body,
```
stbrody, line 81: again, no need to copy the body object, you can just say `body.indices = ...`

packages/http-client/src/remote-admin-api.ts
```diff
@@ -86,9 +106,23 @@ export class RemoteAdminApi implements AdminApi {
         )}`,
       },
     })
-    return response.models.map((modelStreamIDString: string) => {
-      return StreamID.fromString(modelStreamIDString)
+    return response.models.map((id) => StreamID.fromString(id))
```
stbrody, line 109: if we rename `indices` to `indexedModels` (or similar) and deprecate `models` in the output, we should update this to get the streamids from the non-deprecated field to future-proof it.

packages/http-client/src/remote-admin-api.ts
```diff
@@ -86,9 +106,23 @@ export class RemoteAdminApi implements AdminApi {
...
+  }
+
+  async getIndexedModelsWithFieldIndices(): Promise<Array<ModelFieldsIndex>> {
```
stbrody, line 112: do we need a new method?  Could we just add an optional `args` object to the existing `getIndexedModels` function?

stbrody (changes requested):
looking good, almost all of my comments are around future-proofing the APIs further

packages/core/src/local-admin-api.ts
dbcfd: So this was to maintain backwards compatibility with the RemoteAdminApi, which ComposeDB uses. Both LocalAdminApi and RemoteAdminApi extend from AdminApi, where this method is defined.

packages/core/src/local-admin-api.ts
dbcfd: They do get deduped eventually, and likely only one parameter would be passed here.

packages/http-client/src/remote-admin-api.ts
dbcfd: This would change the return type from `StreamID` to `ModelFieldsIndex`. If we're good with making this a breaking change in the API, we could simplify a bit of the code in this PR.

packages/core/src/indexing/local-index-api.ts
dbcfd: Can't do that with this specific type, since all fields are readonly. Or I can mark the fields not readonly.

stbrody:
okay yeah, I see your points about breaking the client APIs.  I definitely think we should at least do the work to future proof the http APIs, and then think about whether or not we want to break the client.  Personally I'd be in favor of a major version bump with a breaking change to the client to make it cleaner more future proof, but I understand there may be pushback to that.  @oed @PaulLeCam @michaelsena for opinions (also Michael can you tag James?  I can't find his github username)

PaulLeCam:
> If we don't want to do a breaking change right now, I wonder if there is a name we can think of for the new methods that is generic enough that we can mark the existing functions as deprecated and plan for the new ones to be the only API at some point in the future? I think the two methods that we want to replace with new versions are `startIndexingModels()` and `getIndexedModels()`

I like the idea of having new methods/API endpoints and deprecating the existing ones without making breaking changes now, we could have a dedicated project for a v3 of js-ceramic with multiple breaking changes at once.

packages/core/src/indexing/database-index-api.ts
dbcfd: Yep, originally was doing this to fallthrough, but request indices should override.

oed:
Let's definitely not break the http api. Don't feel strongly on the javascript api versioning.

packages/cli/src/ceramic-daemon.ts
```diff
@@ -831,12 +834,12 @@ export class CeramicDaemon {
       return
     }
 
-    const withFieldIndices = req.query.withFieldIndices
-    if (withFieldIndices && withFieldIndices.toString().toLowerCase() == 'true') {
-      const indexedModels = await this.ceramic.admin.getIndexedModelsWithFieldIndices()
+    const withModelData = req.query.withModelData
```
stbrody, line 837: hmm... if we're talking about users explicitly opting-in to certain types of data to be returned or not, then I'd think being specific about what data they want would be better.  I.e. `withIndices` instead of `withModelData`.  The problem is that this arg is doing more than just determining whether or not index data is returned, its actually controlling the **format** of the output.

What about making this field `outputFormat` with two possible values (and defaulting to old behavior if the param is omitted)?  Maybe something like `document` and `streamidsOnly`.  Or `legacy` and `withMetadata`.  Or maybe call it `outputVersion` with values `0` or `1`?

packages/cli/src/ceramic-daemon.ts
```diff
@@ -831,12 +834,12 @@ export class CeramicDaemon {
...
+    if (withModelData && withModelData.toString().toLowerCase() == 'true') {
+      const indexedModels = await this.ceramic.admin.getIndexedModelData()
       res.json({
         models: indexedModels.map((idx) => idx.streamID.toString()),
```
stbrody, line 841: since the client has to explicitly request the new format with an argument, we know if we get a request with this argument then we must be speaking to a new client.  So then changing the output format should be safe.  Meaning I think we don't need to have both `models` and `modelData` and duplicate information in the response.  I think we can just change the structure of `models` to be an array of documents instead of streamids, and new clients will know to expect that.

EDIT: I guess there would be an issue if you had a new client talking to an old server that didn't know about the new argument.  At some point we should probably standardize on whether we expect clients or servers to be upgraded first and only support cross-version compatibility in a single upgrade direction to simplify the surface area of what we need to maintain backwards compatibility for.  CC @oed @bigs

I guess duplicating the information for now is fine then to avoid having to make that decision right now

packages/common/src/ceramic-api.ts
```diff
@@ -95,17 +97,18 @@ export interface AdminApi {
...
   /**
    * Adds model streams to index
    *
-   * @param modelsIDs - array of model stream IDs to add to index
+   * @param modelsIDs - array of model stream IDs to add to index. This parameter is deprecated
+   * and indices should be specified instead
    * @param indices - array of model streams with field indices to index
    */
-  startIndexingModels(modelsIDs: Array<StreamID>, indices?: Array<ModelFieldsIndex>): Promise<void>
+  startIndexingModels(modelsIDs: Array<StreamID>, indices?: Array<ModelData>): Promise<void>
```
stbrody, line 111: should we rename the arg now to `modelData` instead of `indices`?

stbrody, line 111: also, I liked how above we split `getIndexedModels` and `getIndexedModelData` into two functions.  Can we do the same here?  `startIndexingModels` and `startIndexingModelData`?  Then the whole `startIndexingModels` method can just be marked as deprecated, and we don't have to worry about the case where the two args to this function specify a different set of models between them

packages/core/src/indexing/database-index-api.ts
```diff
@@ -118,10 +119,8 @@ export abstract class DatabaseIndexApi<DateType = Date | number> {
       .first('indices')
       .where('model', model.toString())
 
-    if (res) {
+    if (res && res.indices) {
       return JSON.parse(res.indices)
-    } else {
-      undefined
     }
```
stbrody, line 124: instead of falling through and implicitly returning `undefined`, can we change this to explicitly return `null` if there are no indices?  Or maybe just an empty array would also be fine?

packages/core/src/indexing/local-index-api.ts
```diff
@@ -46,7 +46,7 @@ async function _getIndexModelArgs(
         relations: content.relations,
       }
     }
-    const indices = (await databaseIndexApi?.getFieldsIndex(modelStreamId)) ?? req.indices
+    const indices = req.indices ?? (await databaseIndexApi?.getFieldsIndex(modelStreamId))
```
stbrody, line 49: Why do we need to check the database for existing indices here?  I think I asked about this in an earlier review round also and never got a response.

packages/core/src/indexing/local-index-api.ts
```diff
@@ -198,6 +209,17 @@ export class LocalIndexApi implements IndexApi {
     // back to the DatabaseIndexApi so that it can populate its internal state.
     // TODO(CDB-2132):  Fix this fragile and circular DatabaseApi initialization
     const modelsToIndex = this.databaseIndexApi.getIndexedModels()
+    for (const model of modelsToIndex) {
```
stbrody, line 212: can we make a helper function for this type checking?  Even better would be to use a codec, but maybe that's better saved for a follow-up to not hold up this PR any more

packages/core/src/local-admin-api.ts
```diff
@@ -20,30 +20,26 @@ export class LocalAdminApi implements AdminApi {
...
-      return {
-        streamID: id,
-        indices: undefined,
-      } as ModelFieldsIndex
-    })
-    if (indices) {
-      models = models.concat(indices)
+  async startIndexingModels(modelsIDs: Array<StreamID>, indices?: Array<ModelData>): Promise<void> {
+    const models = indices ?? []
+    for (const modelId of modelsIDs) {
```
stbrody, line 25: this still feels really ugly to me, but if we split up this method into two versions like I suggested elsewhere then the need for this goes away

packages/core/src/indexing/local-index-api.ts
stbrody: ping

packages/core/src/indexing/local-index-api.ts
```diff
@@ -181,6 +209,17 @@ export class LocalIndexApi implements IndexApi {
...
+      if (model.indices) {
+        for (const idx of model.indices) {
+          if (!idx.fields) {
+            throw new Error('Fields was undefined')
+          } else if (idx.fields instanceof String) {
```
PaulLeCam, line 217: `instanceof String` will return `false` for string literals, is this intentional? Otherwise the check should be `typeof idx.fields === 'string'`

PaulLeCam (changes requested):

packages/core/src/indexing/local-index-api.ts
dbcfd: This is for nodes that come back up with id's pinned for indexing. We can then validate that tables exist in the proper state or need to be migrated.

packages/core/src/indexing/database-index-api.ts
dbcfd: Empty array didn't work, because there are cases where we want to know if there are actually indices specified. Can change to explicit return.

packages/core/src/indexing/database-index-api.ts
```diff
@@ -113,6 +162,7 @@ export abstract class DatabaseIndexApi<DateType = Date | number> {
         models.map((indexModelArgs) => {
           return {
             model: indexModelArgs.model.toString(),
+            is_indexed: true,
```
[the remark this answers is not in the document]
dbcfd: This makes explicit behavior that was defaulted in the schema. Done for readability, maintenance, and in case for some reason the default changes.

packages/core/src/indexing/local-index-api.ts
dbcfd: Will remove, was debugging something

packages/cli/src/ceramic-daemon.ts
dbcfd: Given that we have a versioned endpoint, was planning on putting it under a v0. We can do another endpoint (/modelData?) and mark this one as deprecated, but that's a little weird with a versioned endpoint, rather than doing a /v1/models

packages/core/src/indexing/local-index-api.ts
stbrody: oh yeah, interesting. I don't usually see the `readonly` keyword used within interface types, usually only within classes.  Curious if @PaulLeCam has thoughts on best practices here

packages/core/src/indexing/local-index-api.ts
```diff
@@ -136,34 +154,44 @@ export class LocalIndexApi implements IndexApi {
...
+    modelData: ModelData
+  ): Promise<IndexModelArgs> {
+    const modelStreamId = modelData.streamID
+    this.logger.imp(`Starting indexing for Model ${modelStreamId.toString()}`)
+    const indexModelArgs = await _getIndexModelArgs(
+      modelData,
+      this.repository,
+      this.databaseIndexApi
+    )
+    if (modelsNoLongerIndexed) {
```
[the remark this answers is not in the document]
dbcfd: This was actually a holdover from previous code. I think tests are the only place breaking this type, but this PR just makes the name clearer, it doesn't add the check.

packages/core/src/indexing/local-index-api.ts
PaulLeCam: Same, I've only used `readonly` for classes getters, not for plain objects, but I think it's good to have immutable objects rather than having to clone them every time there are possible mutations.

packages/core/src/indexing/database-index-api.ts
```diff
@@ -29,6 +32,20 @@ export interface IndexStreamArgs {
...
 }
 
+/**
+ * Create a valid name for a fields index, if the index has a name specified, or generate a valid
+ * name if one is not specified
+ * @param idx Index to create a name for
+ * @param table Table to add index to
+ */
+export function fieldsIndexName(idx: FieldsIndex, table: string): string {
+  if (idx.name) {
```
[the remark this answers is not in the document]
PaulLeCam: cc @stbrody what do you think please? Can we remove the ability to set custom index names?

PaulLeCam (changes requested):
Thanks for the changes, mostly LGTM other than the type mismatch here, and I still think we should remove the ability to set custom index names to avoid conflicts.

packages/core/src/indexing/database-index-api.ts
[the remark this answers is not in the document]
stbrody: I don't feel strongly either way.  Happy to remove it.  I think if users ask for the ability to set custom index names we can always add it in later

packages/core/src/indexing/local-index-api.ts
PaulLeCam: No I really mean immutable. Objects that are passed as arguments to functions shouldn't be mutated IMO so it's good if they are clearly defined as such, it's much safer to have pure functions that create new objects as needed rather than mutate objects they receive as arguments, possibly affecting the logic of other functions.

The difference here is that we're building the object, so we're in the case where indeed the type should be writeable. Ideally we'd used a writeable type/interface definition, and use `Readonly<T>` in the function arguments so we can have the readonly type checks.

An alternative would be to use the `Writable` type from `type-fest` when building the object: https://github.com/sindresorhus/type-fest/blob/main/source/writable.d.ts

stbrody: ah yeah, agreed about arguments passed to functions, I was thinking about this case where we're just building up an object iteratively.  I think for this case simply removing `readonly` from the fields within the `IndexModelArgs` interface definition would suffice.  Then it should be up to the places that use specific instances of `IndexModelArgs` to decide whether the instance should be mutable or not.

PaulLeCam: Yeah sounds good to me to remove the `readonly` flags unless @ukstv has concerns about it?
I think we can incrementally add `Readonly<T>` to functions arguments as needed.

packages/http-client/src/remote-admin-api.ts
stbrody: ping.  We should prefer reading from the `modelData` field in the response and only fall back to reading from `models` if `modelData` is missing.  That way in the future we can eventually remove `models` from the response entirely without breaking updated clients

stbrody (commented):
looks like there were still a few old comments that hadn't actually gotten addressed.  We should make sure to go through all the open comment threads and make sure nothing slipped through the cracks before closing this out

packages/core/src/indexing/local-index-api.ts
stbrody: Sorry, I don't understand this answer.  What do you mean "with id's pinned for indexing"?

packages/cli/src/ceramic-daemon.ts
```diff
@@ -805,10 +832,20 @@ export class CeramicDaemon {
...
 
-    const indexedModelStreamIDs = await this.ceramic.admin.getIndexedModels()
-    res.json({
-      models: indexedModelStreamIDs.map((modelStreamID) => modelStreamID.toString()),
-    })
+    const indexedModels = await this.ceramic.admin.getIndexedModelData()
+    const body: Record<string, any> = {
+      models: indexedModels.map((idx) => idx.streamID.toString()),
+    }
+    const outputFormat = req.query.outputFormat?.toString() ?? 'document'
```
stbrody, line 839: I had been thinking that the default behavior should be the old behavior - i.e. `onlyStreamIds` - but that if you explicitly ask for the `document` format you *just* get the document structure back, without duplicating the `models` field as an array.

That said, I think Paul's suggestion to just use a different API endpoint is probably actually cleaner

packages/cli/src/ceramic-daemon.ts
stbrody: I think a `/modelData` endpoint would be fine and is probably simplest

stbrody:
Also we should figure out why the image verification step in CI is failing

dbcfd:
> Also we should figure out why the image verification step in CI is failing

I haven't merged in the latest develop yet with the CI changes. Was going to do that once everything looks good to limit what needs to be reviewed.

packages/core/src/indexing/database-index-api.ts
```diff
@@ -54,8 +50,8 @@ export function fieldsIndexName(idx: FieldsIndex, table: string): string {
  */
 export interface IndexModelArgs {
   readonly model: StreamID
```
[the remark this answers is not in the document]
dbcfd: Preferring to only change the fields that need to be mutated, so any users of model expecting it to be readonly do not have that assumption broken.

packages/cli/src/ceramic-daemon.ts
```diff
@@ -682,7 +697,8 @@ export class CeramicDaemon {
       kid: result.kid,
       code: result.payload.code,
       requestPath: result.payload.requestPath,
-      models: result.payload.requestBody ? result.payload.requestBody.models : undefined,
+      models: result.payload.requestBody?.models || [],
+      modelData: result.payload.requestBody?.indices || [],
```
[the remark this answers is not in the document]
dbcfd: It was, thanks for catching that.

packages/core/src/indexing/local-index-api.ts
```diff
@@ -17,27 +19,34 @@ import { Model } from '@ceramicnetwork/stream-model'
...
+
   if (modelStreamId.type == Model.STREAM_TYPE_ID) {
     const modelState = await repository.load(modelStreamId, {})
     const content = modelState.state.next?.content ?? modelState.state.content
     Model.assertVersionValid(content, 'major')
     if (content.relations) {
-      return { model: modelStreamId, relations: content.relations }
+      opts.relations = content.relations
     }
+    opts.indices = req.indices ?? (await databaseIndexApi?.getFieldsIndex(modelStreamId))
```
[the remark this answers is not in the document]
dbcfd: Code paths exist where we create index model args, without a request with composite level information. I think creating a new model instance is one of these code paths. This ensures that composite level indices are added to the request.

dbcfd:
Closing in favor of https://github.com/ceramicnetwork/js-ceramic/pull/2845

PR #2830, opened 2023-05-25
feat(core): finalize block threshold for switch to smart contract anchoring

stbrody:
Methodology for selecting block numbers:
- go to etherscan for the appropriate chain (eth mainnet, goerli, and gnosis).
- search for our anchor smart contract address: [0x231055A0852D67C7107Ad0d0DFeab60278fE6AdC](https://github.com/ceramicnetwork/js-ceramic/blob/94b51b6e11279a9e9daf007bcc3c9f00a283e6f1/packages/core/src/anchor/ethereum/ethereum-anchor-validator.ts#L63)
- Take the block number of the very first `anchorDagCbor` transaction (which is always the third overall transaction on the address - after contract creation and ownership transfer)

### DIFF
--- a/packages/core/src/anchor/ethereum/ethereum-anchor-validator.ts
+++ b/packages/core/src/anchor/ethereum/ethereum-anchor-validator.ts
@@ -45,20 +45,18 @@ const BASE_CHAIN_ID = 'eip155'
 const MAX_PROVIDERS_COUNT = 100
 const TRANSACTION_CACHE_SIZE = 50
 const BLOCK_CACHE_SIZE = 50
-const V0_PROOF_TYPE = 'raw'
 const V1_PROOF_TYPE = 'f(bytes32)' // See: https://namespaces.chainagnostic.org/eip155/caip168
 
 const ABI = ['function anchorDagCbor(bytes32)']
 
 const iface = new Interface(ABI)
 
-//TODO (NET-1659): Finalize block numbers and smart contract addresses once CAS is creating smart contract anchors
 const BLOCK_THRESHHOLDS = {
-  'eip155:1': 1000000000, //mainnet
+  'eip155:1': 16688195, //mainnet
   'eip155:3': 1000000000, //ropsten
-  'eip155:5': 1000000000, //goerli
-  'eip155:100': 1000000000, //gnosis
-  'eip155:1337': 1000000, //ganache
+  'eip155:5': 8498671, //goerli
+  'eip155:100': 26509835, //gnosis
+  'eip155:1337': 1, //ganache
 }
 const ANCHOR_CONTRACT_ADDRESS = '0x231055A0852D67C7107Ad0d0DFeab60278fE6AdC'
 
--- a/packages/core/src/anchor/ethereum/__tests__/ethereum-anchor-validator.test.ts
+++ b/packages/core/src/anchor/ethereum/__tests__/ethereum-anchor-validator.test.ts
@@ -19,7 +19,7 @@ const TEST_TRANSACTION = {
   blockNumber: 3,
   blockTimestamp: 1586784008,
   blockHash: '0x752fcd3593c140db9f206b421c47d875e0f92e425983f8c72ab04c0e969b072a',
-  to: '0xD3f84Cf6Be3DD0EB16dC89c972f7a27B441A39f2',
+  to: '0x231055A0852D67C7107Ad0d0DFeab60278fE6AdC',
 }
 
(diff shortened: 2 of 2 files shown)


packages/core/src/anchor/ethereum/ethereum-anchor-validator.ts
```diff
@@ -45,20 +45,18 @@ const BASE_CHAIN_ID = 'eip155'
...
 const BLOCK_THRESHHOLDS = {
-  'eip155:1': 1000000000, //mainnet
+  'eip155:1': 16688195, //mainnet
   'eip155:3': 1000000000, //ropsten
-  'eip155:5': 1000000000, //goerli
-  'eip155:100': 1000000000, //gnosis
-  'eip155:1337': 1000000, //ganache
+  'eip155:5': 8498671, //goerli
+  'eip155:100': 26509835, //gnosis
+  'eip155:1337': 1, //ganache
```
stbrody, line 59: This will mean that it is no longer possible to perform the old-style anchors with Ganache. Anyone testing with Ganache will need to deploy the anchor contract to it and use smart-contract style anchoring.  This makes tests more realistic but could add friction to people setting up test deployments using the `local` network with ganache.  FYI @oed let me know if you think I should hold off on this change for now.

@ukstv @stephhuynh18 please let me know if this will cause problems for the unit tests in the CAS codebase.

oed: @nathanielc would this add friction for you in your Keramik test setups?

packages/core/src/anchor/ethereum/ethereum-anchor-validator.ts
```diff
@@ -45,20 +45,18 @@ const BASE_CHAIN_ID = 'eip155'
...
 const BLOCK_CACHE_SIZE = 50
-const V0_PROOF_TYPE = 'raw'
 const V1_PROOF_TYPE = 'f(bytes32)' // See: https://namespaces.chainagnostic.org/eip155/caip168
 
 const ABI = ['function anchorDagCbor(bytes32)']
 
 const iface = new Interface(ABI)
 
-//TODO (NET-1659): Finalize block numbers and smart contract addresses once CAS is creating smart contract anchors
 const BLOCK_THRESHHOLDS = {
```
oed, line 54: Where is the threshold for the dev-unstable network? I believe it's on sepolia?

stbrody: dev-unstable is on goerli

packages/core/src/anchor/ethereum/ethereum-anchor-validator.ts
nathanielc: It will add friction,but if we can document the new process then we can add it as a step in keramik and then keep automation will solve it. 

Is adding the smart-contact to Ganache just a script I could run?

nathanielc:
I have a question about the larger design. Why does Ceramic need to be aware of the smart contract at all? Can it not rely on proofs that exist in the anchor commits to validate it was anchored without having to interact with the smart contract?

packages/core/src/anchor/ethereum/ethereum-anchor-validator.ts
oed: oh that's right!

oed:
@nathanielc So I think the reason it got introduced is because of [late publishing](https://developers.ceramic.network/protocol/streams/consensus/#late-publishing). The idea being that 3Box Labs running CAS would prevent anyone from anchoring events that wouldn't get published to the network.

My perspective on this is that any approach that is not a full blockchain or a semi-trusted set of nodes will never get around this problem. And I think I agree that at the core protocol level Ceramic should just accept any anchor. In order to mitigate the late publishing problem, streams should be able to diverge and converge without loosing specific events.

Note that right now enforcing the use of the CAS smart contract is what allows us to use historical data sync, which indexes the txs to the contract to synchronize past data. With the introduction of Recon, this is not longer needed.

packages/core/src/anchor/ethereum/ethereum-anchor-validator.ts
ukstv: I guess, it is all right from the CAS perspective. Means, we would remove some tests, and some code, and AFAIK that's it.

stbrody: If no one's worried about this change can I get some approvals on the PR?

PR #2834, opened 2023-06-03
CDB-2492: Indexing cleanup

dbcfd:
Cleanup parts of indexing in preparation for adding field indices.

### DIFF
--- a/packages/core/src/indexing/database-index-api.ts
+++ b/packages/core/src/indexing/database-index-api.ts
@@ -56,10 +56,10 @@ type IndexedData<DateType> = {
  */
 export abstract class DatabaseIndexApi<DateType = Date | number> {
   private readonly insertionOrder: InsertionOrder
-  private modelsToIndex: Array<StreamID> = []
+  private indexedModels: Array<StreamID> = []
   // Maps Model streamIDs to the list of fields in the content of MIDs in that model that should be
   // indexed
-  private readonly modelsIndexedFields = new Map<string, Array<string>>()
+  private readonly modelRelations = new Map<string, Array<string>>()
   tablesManager: TablesManager
   syncApi: ISyncQueryApi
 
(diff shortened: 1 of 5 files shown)


packages/core/src/indexing/database-index-api.ts
```diff
@@ -56,10 +56,10 @@ type IndexedData<DateType> = {
  */
 export abstract class DatabaseIndexApi<DateType = Date | number> {
   private readonly insertionOrder: InsertionOrder
-  private modelsToIndex: Array<StreamID> = []
+  private indexedModels: Array<StreamID> = []
   // Maps Model streamIDs to the list of fields in the content of MIDs in that model that should be
```
stbrody, line 60: comment needs updating now if we're limiting the meaning of this field

packages/core/src/indexing/local-index-api.ts
```diff
@@ -174,19 +173,12 @@ export class LocalIndexApi implements IndexApi {
...
-
-  /*
-  Gets currently indexed models from the database api and calls index's api indexModels()
-  which creates index model args and passes them back to the database api, so that it can populate
-  its internal state
-   */
-  async populateDatabaseApiInternalState(): Promise<void> {
-    const modelsToIndex = this.databaseIndexApi?.getActiveModelsToIndex()
-    await this.indexModels(modelsToIndex)
+    if (this.databaseIndexApi) {
```
stbrody, line 176: nitpick: would prefer early return with
```
if(!this.databaseIndexApi) {
  return
}
```

Early returns are nice for avoiding extra indentation layers and improving readability IMO

packages/core/src/indexing/local-index-api.ts
```diff
@@ -174,19 +173,12 @@ export class LocalIndexApi implements IndexApi {
...
+      await this.databaseIndexApi.init()
+      // FIXME: CDB-2132 - Fragile DatabaseApi initialisation
```
stbrody, line 178:
```suggestion
      // Load the set of indexed models from the database and pass them
      // back to the DatabaseIndexApi so that it can populate its internal state.
      // TODO(CDB-2132):  Fix this fragile and circular DatabaseApi initialization
```

stbrody (approved):
lgtm

PR #2835, opened 2023-06-03
CDB-2492: Prep remote admin api for field indices

dbcfd:


packages/http-client/src/remote-admin-api.ts
```diff
@@ -69,10 +66,13 @@ export class RemoteAdminApi implements AdminApi {
 
   async startIndexingModels(modelsIDs: Array<StreamID>): Promise<void> {
     const code = await this.generateCode()
+    const body = modelsIDs
```
dbcfd, line 69: This logic is repeated in this PR, but will be changed for the field indices PR.

stbrody: maybe we could have a `getRequestBody` function that for now takes the model list and gets updated in the follow-up PR with whatever extra arguments it needs?

Otherwise it feels weird to do a small PR like where the only thing it does is increase code duplication

stbrody (changes requested):

stbrody (approved):
lgtm

PR #2836, opened 2023-06-12
feat: add eventIDs

zachferland:
EventIDs for recon, ref corresponding rust pr - https://github.com/3box/rust-ceramic/pull/53/files#diff-2924eb293981196ffa6cd00553cc52fe96f0718b46b967ea3f2e436504ff0401

Placed in streamids packages, but could make sense to move elsewhere, needed to communicate later with rust ceramic/recon

### DIFF
--- a/packages/streamid/src/event-id.ts
+++ b/packages/streamid/src/event-id.ts
@@ -10,7 +10,7 @@ import { Memoize } from 'mapmoize'
 const TAG = Symbol.for('@ceramicnetwork/streamid/EventID')
 
 // Stream Codecs and Event Codec https://github.com/ceramicnetwork/CIPs/blob/main/tables/streamtypes.csv
-const EVENT_ID_CODEC = 0x71
+const EVENT_ID_CODEC = 0x05
 const STREAMID_CODEC_LEN = 2
 const EVENT_ID_CODEC_LEN = 1
 
--- a/packages/streamid/src/event-id.ts
+++ b/packages/streamid/src/event-id.ts
@@ -1,6 +1,6 @@
 import { CID } from 'multiformats/cid'
 import { base36 } from 'multiformats/bases/base36'
-import { hash as sha256 } from '@stablelib/sha256';
+import { hash as sha256 } from '@stablelib/sha256'
 import varint from 'varint'
 import * as cbor from 'cborg'
 import * as u8a from 'uint8arrays'
@@ -9,18 +9,20 @@ import { Memoize } from 'mapmoize'
 
 const TAG = Symbol.for('@ceramicnetwork/streamid/EventID')
 
-const DAG_CBOR_CODEC = 0x30
+// Stream Codecs and Event Codec https://github.com/ceramicnetwork/CIPs/blob/main/tables/streamtypes.csv
+const EVENT_ID_CODEC = 0x71
 const STREAMID_CODEC_LEN = 2
-const DAG_CBOR_CODEC_LEN = 1
+const EVENT_ID_CODEC_LEN = 1
 
+// Network Values https://cips.ceramic.network/tables/networkIds.csv
 const NETWORK: Record<string, number> = {
-  "mainnet":	0x00,
-  "testnet-clay":	0x01,
-  "dev-unstable":	0x02,
-  "inmemory":	0xFF
+  mainnet: 0x00,
+  'testnet-clay': 0x01,
+  'dev-unstable': 0x02,
+  inmemory: 0xff,
 }
 
-function networkByName(net: string):number {
+function networkByName(net: string): number {
   const netId = NETWORK[net]
   if (!netId && netId !== 0) throw new Error('Not a valid network name')
   return netId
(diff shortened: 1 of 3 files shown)


packages/streamid/src/event-id.ts
```diff
@@ -0,0 +1,202 @@
...
+import { hash as sha256 } from '@stablelib/sha256';
+import varint from 'varint'
+import * as cbor from 'cborg'
+import * as u8a from 'uint8arrays'
+import { STREAMID_CODEC } from './constants.js'
+import { Memoize } from 'mapmoize'
+
+const TAG = Symbol.for('@ceramicnetwork/streamid/EventID')
+
+const DAG_CBOR_CODEC = 0x30
```
oed, line 12:
```suggestion
const EVENT_IT_STREAM_TYPE = 0x30
```
This is no longer the dag cbor code.

packages/streamid/src/event-id.ts
```diff
@@ -0,0 +1,202 @@
...
+    this._controller = controller
+    this._init = init
+    this._eventHeight = eventHeight
+    this._event = event
+  }
+
+  /**
+   * Create a new EventID.
+   */
+  static create(networkID: number | string, separator: string, controller: string, init: CID | string, eventHeight: number, event: CID | string ): EventID {
```
oed, line 61: Can we allow `separator` to be a bytestring as well? In the case of a model, we will get it as a Uint8Array when we read it from the initevent.

cc @AaronGoldman thoughts about hashing either a string or a bytearray here?

AaronGoldman: The event is dag-cbor map with a field in it. I think we should take whatever is there. If it is a cbor Text string hash the utf-8 bytes, if it is a  Byte string hash the Uint8Array.

packages/streamid/src/event-id.ts
```diff
@@ -0,0 +1,202 @@
...
+    return new EventID(networkIDInt, separatorBytes, controllerBytes, initBytes, eventHeight, eventCid)
+  }
+
+  /**
+   * Bytes representation of EventID.
+   */
+  @Memoize()
+  get bytes(): Uint8Array {
+    const streamCodec = varint.encode(STREAMID_CODEC)
+    const dagcborCodec = varint.encode(DAG_CBOR_CODEC)
```
AaronGoldman, line 77: probably rename to `eventIdCodec` or  cip124EventIdCodec

packages/streamid/src/event-id.ts
AaronGoldman: nit: I usually link to the table in a comment next to magic number so later readers know where the magic is coming from.
https://github.com/ceramicnetwork/CIPs/blob/main/tables/streamtypes.csv

packages/streamid/src/event-id.ts
```diff
@@ -0,0 +1,202 @@
...
+      const networkID = varint.decode(bytes, netOffset)
+      const sepOffset = varint.decode.bytes + netOffset
+      const seperator = bytes.slice(sepOffset, sepOffset + 8)
+      const conOffset = sepOffset + 8 
+      const controller = bytes.slice(conOffset, conOffset + 8)
+      const initOffest = conOffset + 8 
+      const init = bytes.slice(initOffest, initOffest + 4)
+      const ehOffset = initOffest + 4 
+      const remaining = bytes.slice(ehOffset)
+      const eventHeight = cbor.decode(bytes.slice(ehOffset, ehOffset + remaining.length - 36))
```
AaronGoldman, line 99: I guess cbor decode does not tell you how many bytes it consumed but this seems a little fragile. I am also not sure if we need this function. The only question for an EventID is "What is the CID?" so we can look it up in IPFS. After that I don't know if we need full parse.

AaronGoldman (commented):
This looks good once we update the cip-124 eventID to the number from the table.
https://github.com/ceramicnetwork/CIPs/blob/main/tables/streamtypes.csv

Sorry for changing magic numbers as you were coding.

packages/streamid/src/event-id.ts
AaronGoldman: oh my PR is not merged I pointed you at a lying link. the number is going to be `0x05`
https://github.com/ceramicnetwork/CIPs/pull/139/files#diff-14311009753dce3ff50ded6e3fec3401bbd7e85f24071a0dba88bb390046bd13

packages/streamid/src/event-id.ts
zachferland: yeah agree, not great for decoding (var length), but as mentioned seemed like going from bytes to parts is not that important, other than grabbing cid off, which assume is always going to be 36 bytes in ceramic 

did we use cbor instead of varint to save a byte?

packages/streamid/src/event-id.ts
zachferland: update value

oed (approved):
All good on my end

packages/streamid/src/event-id.ts
AaronGoldman: We used cbor instead of varint here so that the sorting bytes is the same as sorting heights. Varint don't sort in the order that the actual ints do,

AaronGoldman (approved):

packages/streamid/src/event-id.ts
zachferland: ah ok, makes sense

PR #2838, opened 2023-06-15
feat: Add CAS Response as a codec

ukstv:
And drop internal `AnchorServiceResponse`.


packages/codecs/src/anchor.ts
```diff
@@ -1,120 +1,69 @@
...
+export const CompleteCASResponse = type(
   {
-    status: literal(AnchorStatus.FAILED),
-    streamId: streamIdAsString,
-    cid: cidAsString,
-    message: string,
+    ...NotCompleteCASResponse.props,
+    status: literal(RequestStatusName.COMPLETED),
+    anchorCommit: CommitPresentation,
+    witnessCar: uint8ArrayAsBase64.pipe(carAsUint8Array),
```
stbrody, line 51: should witnessCar be optional, at least temporarily until all the CAS environments are returning it?  ie CDB-2519

packages/core/src/__tests__/state-manager.test.ts
```diff
@@ -635,13 +635,13 @@ describe('anchor', () => {
...
       // Emulate CAS responses to the 1st commit
-      const fauxCASResponse$ = new Subject<AnchorServiceResponse>()
+      const fauxCASResponse$ = new Subject<CASResponse>()
       requestAnchorSpy.mockReturnValueOnce(fauxCASResponse$)
       // Subscription for the 1st commit
       const stillProcessingFirst = await ceramic.repository.stateManager.anchor(stream$)
       // The emulated CAS accepts the request
       fauxCASResponse$.next({
-        status: AnchorStatus.PENDING,
+        status: RequestStatusName.PENDING,
```
stbrody, line 644: Can we call this `AnchorRequestStatusName`?  Otherwise its not really clear reading this what the context for "Request" is.

packages/core/src/state-management/state-manager.ts
```diff
@@ -432,19 +434,19 @@ export class StateManager {
...
+            case RequestStatusName.PROCESSING: {
               if (!asr.cid.equals(state$.tip)) return
               state$.next({ ...state$.value, anchorStatus: AnchorStatus.PROCESSING })
               await this._updateStateIfPinned(state$)
               return
             }
-            case AnchorStatus.ANCHORED: {
-              await this._handleAnchorCommit(state$, asr.cid, asr.anchorCommit, asr.witnessCar)
+            case RequestStatusName.COMPLETED: {
+              await this._handleAnchorCommit(state$, asr.cid, asr.anchorCommit.cid, asr.witnessCar)
```
stbrody, line 444: what happened here?  Did the response format from the CAS change?  Or have we really never been properly processing the anchor commit CID from the anchor service response ever since the beginning of time?

stbrody (commented):
looking really good, a few small nitpicks around naming and commenting, and one more important question about the anchor service response format for completed anchors and how we get the anchor commit cid out from them and process it

packages/codecs/src/anchor.ts
```diff
@@ -1,120 +1,69 @@
...
-    prev: cidAsString,
-    proof: cidAsString,
-    path: string,
-  },
-  'AnchorCommit'
-)
-export type AnchorCommit = TypeOf<typeof AnchorCommit>
-
-export const AnchorServicePending = strict(
+export const CommitPresentation = sparse(
```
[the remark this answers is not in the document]
ukstv: `Part of CAS response that sends AnchorCommit content. Effectively a historical artefact.` and renamed. Sounds good to you?

packages/core/src/__tests__/state-manager.test.ts
ukstv: Yup.

packages/codecs/src/anchor.ts
ukstv: It is now.

packages/core/src/state-management/state-manager.ts
ukstv: It's not the CAS response has changed. We translated CAS Response into an internal representation. And we used just `anchorCommit.cid` part.

ukstv: https://github.com/ceramicnetwork/js-ceramic/pull/2838/files#diff-7920408ff457031025a51c6c64241f297dc91febb1a810e724475548c154c781L245

packages/codecs/src/anchor.ts
```diff
@@ -1,120 +1,69 @@
...
   {
-    status: literal(AnchorStatus.PENDING),
-    streamId: streamIdAsString,
-    cid: cidAsString,
-    message: string,
+    content: optional(
+      sparse({ path: optional(string), prev: string, proof: optional(string) }, 'content')
```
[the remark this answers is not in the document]
ukstv: This whole `content` thing is a misnomer. There is no code that uses it now, and it is like that since quite a long time ago. Only `cid` is read. So, I just deleted `content`.

packages/codecs/src/anchor.ts
```diff
@@ -12,43 +13,46 @@ export enum RequestStatusName {
...
+  literal(AnchorRequestStatusName.PROCESSING),
+  literal(AnchorRequestStatusName.FAILED),
+  literal(AnchorRequestStatusName.READY),
+  literal(AnchorRequestStatusName.REPLACED),
 ])
 export type NotCompleteStatusName = TypeOf<typeof NotCompleteStatusName>
 
 export const NotCompleteCASResponse = sparse(
   {
+    id: string,
```
stbrody, line 38: should this be `cidAsString`?

packages/core/src/state-management/state-manager.ts
stbrody: oh I see.  Weird.  We had the full anchor commit body but were throwing out most of it early in the pipeline.  Now we actually pass the full object through.  thanks for clarifying.

packages/codecs/src/anchor.ts
ukstv: It is request id, which is uuid string.

stbrody: oh whoops, good call.  I was thinking it was the `id` field of the AnchorCommit, which is the genesis CID.

PR #2839, opened 2023-06-16
feat(common): Update query type to support new filters and order parameters

PaulLeCam:
Needed by ComposeDB to create the new inputs and add them to APIs calls.

### DIFF
--- a/packages/common/src/index-api.ts
+++ b/packages/common/src/index-api.ts
@@ -27,31 +27,53 @@ export type Pagination = ForwardPagination | BackwardPagination
 /**
  * Boolean field value filter
  */
-export type BooleanValueFilter = { isNull: boolean } | { equalTo: boolean }
-
-/**
- * Enum field value filter
- */
-export type EnumValueFilter =
+export type BooleanValueFilter =
   | { isNull: boolean }
-  | { equalTo: string }
-  | { notEqualTo: string }
-  | { in: Array<string> }
-  | { notIn: Array<string> }
+  | { equalTo: boolean }
+  | { notEqualTo: boolean }
 
 /**
- * String or number field value filter
+ * Common type for multiple value filters
  */
-export type ScalarValueFilter<T extends string | number = string | number> =
+export type CommonValueFilter<T> =
   | { isNull: boolean }
   | { equalTo: T }
   | { notEqualTo: T }
   | { in: Array<T> }
   | { notIn: Array<T> }
-  | { lessThan: T }
-  | { lessThanOrEqualTo: T }
+
+/**
+ * Enum field value filter
+ */
+export type EnumValueFilter = CommonValueFilter<string>
+
+/**
+ * Supported greater than value conditions
+ */
+export type GreaterThanValueFilter<T extends string | number> =
   | { greaterThan: T }
   | { greaterThanOrEqualTo: T }
+
+/**
+ * Supported less than value conditions
+ */
+export type LessThanValueFilter<T extends string | number> =
+  | { lessThan: T }
+  | { lessThanOrEqualTo: T }
+
+/**
+ * Range value filter using a required greater than or less than value filter
+ * and optional other boundary
+ */
+export type RangeValueFilter<T extends string | number> =
+  | (GreaterThanValueFilter<T> & Partial<LessThanValueFilter<T>>)
+  | (LessThanValueFilter<T> & Partial<GreaterThanValueFilter<T>>)
+
+/**
+ * String or number field value filter
+ */
+export type ScalarValueFilter<T extends string | number = string | number> = CommonValueFilter<T> &
+  RangeValueFilter<T>
 
 /**
  * Any supported field value filter on an object
@@ -67,7 +89,7 @@ export type ObjectFilter = Record<string, AnyValueFilter>
  * Advanced query filters on a document fields
  */
 export type QueryFilters =
-  | { doc: ObjectFilter }
+  | { where: ObjectFilter }
   | { and: Array<QueryFilters> }
   | { or: Array<QueryFilters> }
   | { not: QueryFilters }
--- a/packages/common/src/index-api.ts
+++ b/packages/common/src/index-api.ts
@@ -25,12 +25,74 @@ export type BackwardPagination = {
 export type Pagination = ForwardPagination | BackwardPagination
 
 /**
+ * Boolean field value filter
+ */
+export type BooleanValueFilter = { isNull: boolean } | { equalTo: boolean }
+
+/**
+ * Enum field value filter
+ */
+export type EnumValueFilter =
+  | { isNull: boolean }
+  | { equalTo: string }
+  | { notEqualTo: string }
+  | { in: Array<string> }
+  | { notIn: Array<string> }
+
+/**
+ * String or number field value filter
+ */
+export type ScalarValueFilter<T extends string | number = string | number> =
+  | { isNull: boolean }
+  | { equalTo: T }
+  | { notEqualTo: T }
+  | { in: Array<T> }
+  | { notIn: Array<T> }
+  | { lessThan: T }
+  | { lessThanOrEqualTo: T }
+  | { greaterThan: T }
+  | { greaterThanOrEqualTo: T }
+
+/**
+ * Any supported field value filter on an object
+ */
+export type AnyValueFilter = BooleanValueFilter | EnumValueFilter | ScalarValueFilter
+
+/**
+ * Mapping of object keys to value filter
+ */
+export type ObjectFilter = Record<string, AnyValueFilter>
+
+/**
+ * Advanced query filters on a document fields
+ */
+export type QueryFilters =
+  | { doc: ObjectFilter }
+  | { and: Array<QueryFilters> }
+  | { or: Array<QueryFilters> }
+  | { not: QueryFilters }
+
+/**
+ * Field sort order, 'ASC' for ascending, 'DESC' for descending
+ */
+export type SortOrder = 'ASC' | 'DESC'
+
+/**
+ * Mapping of object keys to value sort order
+ */
+export type OrderBy = Record<string, SortOrder>
+
+/**
  * Base query to the index. Disregards pagination.
  */
 export type BaseQuery = {
   model: StreamID | string
   account?: string
+  // Deprecated filter for relations support in ComposeDB <= 0.4
   filter?: Record<string, string>
+  // New query filters and order by for ComposeDB >= 0.5
+  filters?: QueryFilters
+  orderBy?: OrderBy
 }
 
 export type PaginationQuery = BaseQuery & Pagination
--- a/packages/common/src/index-api.ts
+++ b/packages/common/src/index-api.ts
@@ -80,7 +80,7 @@ export type SortOrder = 'ASC' | 'DESC'
 /**
  * Mapping of object keys to value sort order
  */
-export type OrderBy = Record<string, SortOrder>
+export type Sorting = Record<string, SortOrder>
 
 /**
  * Base query to the index. Disregards pagination.
@@ -88,11 +88,12 @@ export type OrderBy = Record<string, SortOrder>
 export type BaseQuery = {
   model: StreamID | string
   account?: string
-  // Deprecated filter for relations support in ComposeDB <= 0.4
+  /**
+   * @deprecated relation filters used by ComposeDB <= 0.4
+   */
   filter?: Record<string, string>
-  // New query filters and order by for ComposeDB >= 0.5
-  filters?: QueryFilters
-  orderBy?: OrderBy
+  queryFilters?: QueryFilters
+  sorting?: Sorting
 }
 
 export type PaginationQuery = BaseQuery & Pagination


packages/common/src/index-api.ts
```diff
@@ -24,13 +24,75 @@ export type BackwardPagination = {
...
+ */
+export type BooleanValueFilter = { isNull: boolean } | { equalTo: boolean }
+
+/**
+ * Enum field value filter
+ */
+export type EnumValueFilter =
+  | { isNull: boolean }
+  | { equalTo: string }
+  | { notEqualTo: string }
```
dbcfd, line 38: Since we have a not filter, do we need notEqualTo or notIn?

packages/common/src/index-api.ts
```diff
@@ -24,13 +24,75 @@ export type BackwardPagination = {
...
+ * String or number field value filter
+ */
+export type ScalarValueFilter<T extends string | number = string | number> =
+  | { isNull: boolean }
+  | { equalTo: T }
+  | { notEqualTo: T }
+  | { in: Array<T> }
+  | { notIn: Array<T> }
+  | { lessThan: T }
+  | { lessThanOrEqualTo: T }
```
dbcfd, line 52: Similar to not, we have an or filter, so this seems redundant.

packages/common/src/index-api.ts
```diff
@@ -24,13 +24,75 @@ export type BackwardPagination = {
...
+ * Mapping of object keys to value filter
+ */
+export type ObjectFilter = Record<string, AnyValueFilter>
+
+/**
+ * Advanced query filters on a document fields
+ */
+export type QueryFilters =
+  | { doc: ObjectFilter }
+  | { and: Array<QueryFilters> }
```
dbcfd, line 71: Specifying `and` and `or ` this way allows creation of invalid query filters, such as `{and: []}` or even worse, queries like `{and: {... {and: []} ... }}`.

I'd like to see a strong type here like `{and: {lhs: QueryFilters, rhs: QueryFilters}}`. We could do something similar to prevent deep queries, but at least this prevents just passing empty arrays.

packages/common/src/index-api.ts
```diff
@@ -24,13 +24,75 @@ export type BackwardPagination = {
...
+
+/**
+ * Field sort order, 'ASC' for ascending, 'DESC' for descending
+ */
+export type SortOrder = 'ASC' | 'DESC'
+
+/**
+ * Mapping of object keys to value sort order
+ */
+export type OrderBy = Record<string, SortOrder>
```
dbcfd, line 83: Rather than exposing the database command here, can we call this `Sorting`?

packages/common/src/index-api.ts
```diff
@@ -24,13 +24,75 @@ export type BackwardPagination = {
...
 /**
  * Base query to the index. Disregards pagination.
  */
 export type BaseQuery = {
   model: StreamID | string
   account?: string
+  // Deprecated filter for relations support in ComposeDB <= 0.4
   filter?: Record<string, string>
+  // New query filters and order by for ComposeDB >= 0.5
+  filters?: QueryFilters
```
dbcfd, line 94: I'm not really a fan of `filter` and `filters`, since they are close enough to impact readability. At least typing will keep them from causing issues. Maybe just call them `queryFilters`?

Also, the comment mentions two pieces, which are handled by two fields. I'd suggest removing the comment and adding `@deprecated` to `filter` instead.

dbcfd (changes requested):

packages/common/src/index-api.ts
ukstv: +1 to that.

ukstv (changes requested):

packages/common/src/index-api.ts
PaulLeCam: When it comes to usability, I think it's much easier to provide and document the ability to do ex:
```js
{ doc: { date: { greaterThanOrEqualTo: '2023-06-01' }, status: { notEqualTo: 'ARCHIVED' } } }
```
than the alternative without support for these:
```js
{ and: [
  { or: [
    { doc: { date: equalTo: '2023-06-01' } },
    { doc: { date: greaterThan: '2023-06-01' },
  }]},
  { not: {
    doc: { status: { equalTo: 'ARCHIVED' } }
  }}]
}
```
This would be much more in line with what similar systems are providing.

ukstv: 2 cents. I agree _this_ reads nicer. As with other similar systems though, it is a pain (for me!) to express anything more complicated than a hello world example.

ukstv (approved):

packages/common/src/index-api.ts
dbcfd: Counterpoint is that this isn't what is presented to users. So even though composedb/sql/etc. may say `field != value`, our internal representation doesn't need to say `notEqualTo`. This is internal or at best, part of an sdk, so needs to be able to allow us to perform the functionality, but doesn't need to be a 1 to 1 mapping. 

I prefer not having the additional operator, since it's additional code for each type we have to deal with, rather than a generalized not.

packages/common/src/index-api.ts
PaulLeCam: Sure, I updated it

packages/common/src/index-api.ts
dbcfd: @PaulLeCam @ukstv See above comment. I think this comes down to how much we think these changes are user facing (I don't think they are) versus the maintenance overhead of similar fields across different objects. 

Normally I'd lean towards changes that are more inline with what the user would see, but I don't think this API is very usable directly by users.

ukstv: "the maintenance overhead" - this is the most complicated part for me. If we have a general `not` instead of `notEqualTo` we would have a hard time translating that into lower level operators used in a storage backend IMO. We might expect some weird cases of deoptimisation because of generalised `not`.

packages/common/src/index-api.ts
```diff
@@ -24,13 +24,76 @@ export type BackwardPagination = {
...
+  | { isNull: boolean }
+  | { equalTo: string }
+  | { notEqualTo: string }
+  | { in: Array<string> }
+  | { notIn: Array<string> }
+
+/**
+ * String or number field value filter
+ */
+export type ScalarValueFilter<T extends string | number = string | number> =
```
stbrody, line 45: sorry my typescript knowledge isn't good enough to understand what this is saying, can you break it down for me?

packages/common/src/index-api.ts
```diff
@@ -24,13 +24,76 @@ export type BackwardPagination = {
...
+/**
+ * Mapping of object keys to value filter
+ */
+export type ObjectFilter = Record<string, AnyValueFilter>
+
+/**
+ * Advanced query filters on a document fields
+ */
+export type QueryFilters =
+  | { doc: ObjectFilter }
```
stbrody, line 70: can you explain more how `doc` filters would work?

stbrody (commented):
this makes me a little nervous as we're basically building a query language but (AFAIK) none of us our query language experts.  There is a ton of prior art on designing query languages and there are a lot of pitfalls you can run into if not careful (I know MongoDB is still struggling with the inconsistencies of their JSON-document based query language that was designed >12 years ago)


Is there not some existing JSON-based query language we could use here instead of designing our own from scratch?

packages/common/src/index-api.ts
PaulLeCam: - `T extends string | number` restricts the valid types for `T`, so `ScalarValueFilter<string>` is valid, but `ScalarValueFilter<boolean>` is invalid.
- `T = string | number` defines a default value for `T` when not defined, so `ScalarValueFilter` can be used without explicitly providing the type for `T`, using `const filter: ScalarValueFilter` is the same as using `const filter: ScalarValueFilter<string | number>`.

packages/common/src/index-api.ts
PaulLeCam: I think they'd work by iterating over all the keys and matching value filters to create the SQL query, something like the following:

```ts
// Build a single condition filter string
function buildConditionFilter(condition: keyof AnyValueFilter, value: string): string {
  switch (condition) {
    case 'equalTo':
      return `= ${value}`
    case 'notEqualTo':
      return `<> ${value}`
    case 'in':
      return `IN (${value})`
    // ... other cases covering all supported conditions
  }
}

const RANGE_GREATER_CONDITIONS = ['greaterThan', 'greaterThanOrEqualTo']
const RANGE_LOWER_CONDITIONS = ['lowerThan', 'lowerThanOrEqualTo']

// Build the filter string for a single field
function buildValueFilter(field: string, filter: AnyValueFilter): string {
  const conditions = Object.keys(filter)
  // Check valid number of conditions
  if (conditions.length === 0 || conditions.length > 2) {
    throw new Error(`Invalid number of conditions in filter for field ${field}`)
  }

  if (conditions.length === 1) {
    const condition = conditions[0]
    const value = Array.isArray(filter[condition]) ? filter[condition].join(', ') : filter[condition].toString()
    return `${field} ${buildConditionFilter(condition, value)}`
  }

  // Check valid range when 2 conditions are provided
  const [conditionA, conditionB] = conditions
  if (RANGE_GREATER_CONDITIONS.includes(conditionA) && RANGE_LOWER_CONDITIONS.includes(conditionB) || RANGE_GREATER_CONDITIONS.includes(conditionB) && RANGE_LOWER_CONDITIONS.includes(conditionA)) {
    const filterA = buildConditionFilter(conditionA, filters[conditionA].toString())
    const filterB = buildConditionFilter(conditionB, filters[conditionB].toString())
    return `${field} ${filterA} AND ${field} ${filterB}`
  }
  throw new Error(`Invalid range filter for field ${field}`)
}

// Build the filter string for all fields in an object
function buildObjectFilter(objectFilter: ObjectFilter): string {
  const parts = Object.entries(objectFilter).map(([field, filter]) => buildValueFilter(field, filter))
  return parts.join(' AND ')
}
```

In our case we'll likely want to leverage Knex to build the queries rather than generating SQL directly as I did here, but the logic should be similar I think.

PaulLeCam:
> this makes me a little nervous as we're basically building a query language but (AFAIK) none of us our query language experts. There is a ton of prior art on designing query languages and there are a lot of pitfalls you can run into if not careful (I know MongoDB is still struggling with the inconsistencies of their JSON-document based query language that was designed >12 years ago)
> 
> Is there not some existing JSON-based query language we could use here instead of designing our own from scratch?

Yes that's a goal here, we're simply using a subset of what Graphile defines in https://github.com/graphile-contrib/postgraphile-plugin-connection-filter/blob/master/docs/operators.md
I picked Graphile as we have the same constraints on both ends: GraphQL inputs to describe the objects and SQL to generate the actual DB query, so no need to invent something new here.

packages/common/src/index-api.ts
stbrody: okay, so fields that are specified will have filters set, but fields not included will not be filtered on.  If you wanted to specify that the doc matched exactly a given set of fields *and no others*, I guess the only way to do that would be to know all possible optional fields in the schema and have a filter to explicitly check that they are `undefined`?  Would that even work?  Can you send `undefined` as a value for a filter?

stbrody:
> Yes that's a goal here, we're simply using a subset of what Graphile defines in [graphile-contrib/postgraphile-plugin-connection-filter@`master`/docs/operators.md](https://github.com/graphile-contrib/postgraphile-plugin-connection-filter/blob/master/docs/operators.md?rgh-link-date=2023-06-19T09%3A45%3A13Z)
> I picked Graphile as we have the same constraints on both ends: GraphQL inputs to describe the objects and SQL to generate the actual DB query, so no need to invent something new here.

Okay, I think that seems reasonable.  Still makes me a little nervous, but I'm not sure what else we can reasonably do to be safer/more future proof here at this time.

packages/common/src/index-api.ts
stbrody: Maybe instead of `doc` we can call this something else (`fields`, `filter`, `match`, `matchFields`, etc).  Then if in the future we wanted an "exact-match" operator for a full document structure we could add `doc` later.

stbrody: I think I like `matchFields` most?

Or do we need to call this `doc` to maintain compatibility with Graphile?

stbrody: Actually, if I'm reading this right, I think Graphile calls this `filter`?  Which I'd also be fine with.
https://github.com/graphile-contrib/postgraphile-plugin-connection-filter/blob/master/docs/examples.md

PaulLeCam: Yes we can use a different name than `doc` if you think it's confusing, the main thing is that we need a specific key to differentiate from the `and`, `or` and `not` operators.
Graphile uses `filter` as the input name, same as what we have as `queryFilters` in the `BaseQuery` objects so I'd rather use a different term so we don't have `queryFilters.filter`. `fields` sounds good to me, `where` could also be a possibility I think. `match` could be misleading if we ever supported regular expressions, but could be another option.

stbrody: `fields` or `where` both sound good to me

stbrody: I think I've seen `where` used more often than I've seen `fields`, so I guess I very very slightly lean that way?  But yeah its very slight, I think `fields` would be totally fine also

packages/common/src/index-api.ts
```diff
@@ -27,32 +27,54 @@ export type Pagination = ForwardPagination | BackwardPagination
...
+export type LessThanValueFilter<T extends string | number> =
+  | { lessThan: T }
+  | { lessThanOrEqualTo: T }
+
+/**
+ * Range value filter using a required greater than or less than value filter
+ * and optional other boundary
+ */
+export type RangeValueFilter<T extends string | number> =
+  | (GreaterThanValueFilter<T> & Partial<LessThanValueFilter<T>>)
```
stbrody, line 69: why does the RHS need to be a `Partial`?  Does `Partial` even do anything on a type with only one field?

packages/common/src/index-api.ts
stbrody: I guess there's no way with the type system to express that an array field must be of length greater than X?

stbrody (approved):
lgtm, though probably worth considering a bit more the tradeoffs between an array for `and` and `or ` expressions vs explicit `lhs` and `rhs` like danny proposed.  The arrays are nice because you can have multiple conditions combined in one statement with a big messy nested structure, but the ability to construct invalid queries like danny points out is certainly a downside.

Do we know what other similar systems do here?  How much complexity will it add to detect and error on malformed queries using the `and` and `or` operators?

packages/common/src/index-api.ts
dbcfd: You could introduce a type that is an array of length one or more. Can even use https://www.npmjs.com/package/@pacote/non-empty-array

dbcfd (approved):

packages/common/src/index-api.ts
PaulLeCam: `Partial` makes all the fields of the object optional, so here the type ensures that when setting a range boundary, only another valid boundary can be set, but it's optional.

PaulLeCam:
About arrays for `and` and `or` I don't think it's a big issue if they are empty (just ignore) or containing a single item (just apply the logic on it). I think we can warn about in the client but still handle these in the server?

We can create a type to ensure there are at least 2 items in the array but because the data comes from the client, we'd still need to do runtime validation and possibly type casting as I don't think the TS compiler handles these checks well. Ideally we should start using codecs for client/server interactions, but that's likely something we want to do for all the APIs at once considering there could be breaking changes.

PR #2843, opened 2023-06-22
add workflow for publishing follow-up RCs that bump the RC version

stbrody:
Lets say the current version of the ceramic core package is `2.35.0`.  When we do an RC, we want to bump the minor version (leaving the patch version available for potential hotfix releases) and go to `2.36.0-rc.0`.  That is what our current `rc` workflow does.  Then when we do the main release it goes to `2.36.0`.  All good so far.  The problem comes if we want to do a second RC before doing a main release.  With the current workflow, the second RC will get version `2.37.0-rc.0`.  This wastes a minor version and loses information about the fact that the two RCs are linked and the second is effectively replacing the first as the candidate to become the next full release.  Instead, it would be better for the second RC to get the version `2.36.0-rc.1`.  This PR gives us a workflow to do that.

I am not attached to the name `rc-followup` for the new workflow.  Also lmk if you think it would be better to re-use the `rc` workflow but add an argument to select whether this is the first RC or a follow-up.

### DIFF
--- a/.github/workflows/reusable-release.yml
+++ b/.github/workflows/reusable-release.yml
@@ -111,7 +111,7 @@ jobs:
 
   publish_rc:
     runs-on: ubuntu-latest
-    if: ${{ inputs.release_type == 'rc' }}
+    if: ${{ inputs.release_type == 'rc' || inputs.release_type == 'rc-followup' }}
     steps:
       -
         name: Setup Node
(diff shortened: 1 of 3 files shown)


.github/workflows/reusable-release.yml
```diff
@@ -186,6 +186,83 @@ jobs:
           git merge release-candidate
           git push
 
+  publish_rc_followup:
```
stbrody, line 189: this is a total copy of the `rc` workflow, the only difference is the one line where we run the NPM script to publish the release.  Maybe (hopefully) there's a way to de-dupe this code?

smrz2001: Thanks for pointing out the difference in the flows. I think we might be able to reuse the existing `rc` flow for both options.

Maybe something like this on line 114:
```
    if: ${{ inputs.release_type == 'rc' || inputs.release_type == 'rc-followup' }}
```
And something like this on line 168:
```
RELEASE_TYPE=${{ inputs.release_type }}
if [[ $RELEASE_TYPE == "rc" ]]; then
    npm run publish:release-candidate -- --yes
else
    npm run publish:release-candidate-followup -- --yes
fi
```

stbrody (commented):
okay, updated with your suggestions for combining the workflows.  Is there any way to test this?

smrz2001 (approved):
LGTM!

I usually run new/modified workflows out of the PR branch to make sure there aren't any weird syntax issues before merging.

Do you want to try to create a follow-up RC using this workflow? Won't have any changes since nothing new has gone to `develop`.

Running from a branch just means selecting the workflow and then selecting the branch to run the workflow from (defaulted to `develop`).

smrz2001:
> okay, updated with your suggestions for combining the workflows. Is there any way to test this?

Oh wait, just saw you just asked what I suggested 😄 Want me to run it from the branch?

stbrody:
> Want me to run it from the branch?

Yeah that would be great.  I tried to do it myself here but maybe I did it wrong?  It didn't find the new npm script: https://github.com/ceramicnetwork/js-ceramic/actions/runs/5350208195/jobs/9702510241

smrz2001:
> > Want me to run it from the branch?
> 
> Yeah that would be great. I tried to do it myself here but maybe I did it wrong? It didn't find the new npm script: [ceramicnetwork/js-ceramic/actions/runs/5350208195/jobs/9702510241](https://github.com/ceramicnetwork/js-ceramic/actions/runs/5350208195/jobs/9702510241)

No, you ran it correctly, it just didn't work because the workflow explicitly deals with just the `main`, `release-candidate`, or `develop` branches.

However, since the script got as far as trying to publish the follow-up RC, we know that it is working with the checks we added. So, I'd say it's safe to merge assuming that the follow-up RC NPM command works correctly.

PR #2844, opened 2023-06-22
feat: CDB-2493 Adjust API to support model index changes

dbcfd:
Adjusts AdminAPI to support new functionality around adding indices to models. New structure ModelData is added, and new methods are start, stop, and get indices related to ModelData are added.

This splits out part of the changes in https://github.com/ceramicnetwork/js-ceramic/pull/2829

### DIFF
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -6,8 +6,10 @@ import { ServiceMetrics as Metrics } from '@ceramicnetwork/observability'
 import { IpfsConnectionFactory } from './ipfs-connection-factory.js'
 import {
   DiagnosticsLogger,
+  FieldsIndex,
   LoggerProvider,
   LogStyle,
+  ModelData,
   MultiQuery,
   StreamUtils,
   SyncOptions,
@@ -202,23 +204,43 @@ function validatePort(inPort) {
 }
 
 /**
+ * ModelData contents in JWS
+ */
+type AdminAPIJWSModelData = {
+  streamID: string
+  indices: Array<FieldsIndex>
+}
+
+/**
  * Contents an authorization header signed by an admin DID
  */
 type AdminAPIJWSContents = {
   kid: string
   code: string
   requestPath: string
-  models: Array<string>
+  models?: Array<string>
+  modelData?: Array<AdminAPIJWSModelData>
 }
 
 type AdminApiJWSValidationResult = {
   kid?: string
   code?: string
   models?: Array<StreamID>
+  modelData?: Array<ModelData>
   error?: string
 }
 
-type AdminApiModelMutationMethod = (modelIDs: Array<StreamID>) => Promise<void>
+type AdminApiMutationMethod = AdminApiModelIdsMutationMethod | AdminApiModelDataMutationMethod
+
+type AdminApiModelDataMutationMethod = {
+  type: 'modelData'
+  method: (modelData: Array<ModelData>) => Promise<void>
+}
+
+type AdminApiModelIdsMutationMethod = {
+  type: 'modelIDs'
+  method: (modelsIDs: Array<StreamID>) => Promise<void>
+}
 
 /**
  * Ceramic daemon implementation
@@ -328,7 +350,9 @@ export class CeramicDaemon {
     const did = new DID(didOptions)
     if (provider) {
       await did.authenticate()
-      diagnosticsLogger.imp(`Node DID set to '${did.id}. This DID will be used to authenticate to the anchor service'`)
+      diagnosticsLogger.imp(
+        `Node DID set to '${did.id}. This DID will be used to authenticate to the anchor service'`
+      )
     }
     ceramic.did = did
     await ceramic._init(true)
@@ -349,6 +373,7 @@ export class CeramicDaemon {
     const collectionRouter = ErrorHandlingRouter(this.diagnosticsLogger)
     const adminCodesRouter = ErrorHandlingRouter(this.diagnosticsLogger)
     const adminModelRouter = ErrorHandlingRouter(this.diagnosticsLogger)
+    const adminModelDataRouter = ErrorHandlingRouter(this.diagnosticsLogger)
     const adminPinsRouter = ErrorHandlingRouter(this.diagnosticsLogger)
     const legacyPinsRouter = ErrorHandlingRouter(this.diagnosticsLogger)
     const adminNodeStatusRouter = ErrorHandlingRouter(this.diagnosticsLogger)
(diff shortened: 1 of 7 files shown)


packages/cli/src/ceramic-daemon.ts
```diff
@@ -754,10 +770,54 @@ export class CeramicDaemon {
     }
 
     // Process request
-    await successCallback(jwsValidation.models)
+    switch (successCallback.type) {
```
PaulLeCam, line 773: This logic looks weird to me, there's a callback function provided in the method call, but it needs specific logic in this handler?
Couldn't the branching logic be handled by the caller directly rather than being split between different methods?

packages/common/src/ceramic-api.ts
```diff
@@ -8,6 +8,30 @@ import type { IndexApi } from './index-api.js'
 import { NodeStatusResponse } from './node-status-interface.js'
 import type { AnchorStatus } from '@ceramicnetwork/common'
 
+/**
+ * Field definition for index
+ */
+export interface Field {
```
PaulLeCam, line 14:
```suggestion
export type Field = {
```
For consistency with the other types defined below

PaulLeCam (changes requested):
LGTM overall, just the logic with handling the models IDs or data that looks a bit strange to me

packages/cli/src/ceramic-daemon.ts
dbcfd: Can you describe what that would look like in the caller?

dbcfd: Since we're adjusting the response object in this method, trying to move this functionality out produces a lot of code duplication, and actually makes the code harder to read. Going to leave this as is.

packages/cli/src/ceramic-daemon.ts
```diff
@@ -393,6 +419,9 @@ export class CeramicDaemon {
     adminModelRouter.getAsync('/', this.getIndexedModels.bind(this))
     adminModelRouter.postAsync('/', this.startIndexingModels.bind(this))
     adminModelRouter.deleteAsync('/', this.stopIndexingModels.bind(this))
+    adminModelDataRouter.getAsync('/', this.getIndexedModelData.bind(this))
+    adminModelDataRouter.postAsync('/', this.startIndexingModelData.bind(this))
+    adminModelDataRouter.deleteAsync('/', this.stopIndexingModels.bind(this))
```
stbrody, line 424: this seems weird.  It breaks the pattern and symmetry with the other methods.  I think we should have a real stopIndexingModelData method that takes its input in the same nested object format as the other ModelData methods.  Even if only the `streamId` field of the object is used for now, it still sets up the object structure which would allow us to add new arguments in the future.  Perhaps one day we could use it to delete a specific field index, for example.

If we introduce this endpoint now but with the same API structure as the existing method, then we have two identical endpoints for no gain, plus if we wanted to extend the options available in the future we'd need to add a third endpoint to break the API.

packages/common/src/ceramic-api.ts
```diff
@@ -71,13 +95,24 @@ export interface AdminApi {
    */
   getIndexedModels(): Promise<Array<StreamID>>
 
+  /**
+   * List indexed model streams with data
```
stbrody, line 99:
```suggestion
   * List indexed model streams with additional model data (such as the defined field indices)
```

stbrody (changes requested):
looks good, with two small suggestions

packages/cli/src/ceramic-daemon.ts
```diff
@@ -935,6 +941,13 @@ export class CeramicDaemon {
     })
   }
 
+  async stopIndexingModelData(req: Request, res: Response): Promise<void> {
+    await this._processAdminModelsMutationRequest(req, res, {
+      type: 'modelIDs',
```
stbrody, line 946: the behavior of both endpoints is still the same, so splitting it out to a new method didn't really do anything.  This would need to be of type `modelData` to use the new object-based API structure and get the future-proofing that was the motivation for this change

packages/cli/src/ceramic-daemon.ts
```diff
@@ -683,13 +712,14 @@ export class CeramicDaemon {
...
       models: result.payload.requestBody ? result.payload.requestBody.models : undefined,
+      modelData: result.payload.requestBody ? result.payload.requestBody.modelData : undefined,
     }
   }
 
   private async _validateAdminApiJWS(
     basePath: string,
     jws: string | undefined,
-    shouldContainModels: boolean
+    successCallback?: AdminApiMutationMethod
```
stbrody, line 722: does this actually need the full successCallback?  Can't it just take the 'type' string?  I feel like it makes this method confusing and obfuscates its core purpose of validation by also having it pass around the successCallback

dbcfd: We need a way to know what arguments to pass to each method. Otherwise we end up back where we started, where the indexing methods have to validate that modelIds or modelData are present. Although you would expect the error to not occur, it would be a little weird to return an error to the user that is "JWS Validation passed back model id's when I was expecting modelData"

packages/cli/src/ceramic-daemon.ts
dbcfd: Although we only need stream id's for stop indexing, you want to pass it in as an object, in case there's additional information we may want to pass to stop indexing at a later time?

packages/cli/src/ceramic-daemon.ts
stbrody: don't we already know in the caller which argument we need to use because of the context from which api method was called?  In `startIndexingModelData` we can just always take the `modelData` field from the input and ignore `modelIds`.  If we passed the validation from this function (`_validateAdminApiJWS`), then we know the data will be present

packages/cli/src/ceramic-daemon.ts
stbrody: Yes exactly, see my earlier comment: https://github.com/ceramicnetwork/js-ceramic/pull/2844#discussion_r1242701934

We can still only pass model streamids, but we can do it in the object form (e.g. `[{streamId: "<modelAStreamID>"}, {streamId: "<modelBStreamID>"}]`) as opposed to the array of strings form (e.g. `["<modelAStreamID>", "<modelBStreamID>"]`).  That way we have the structure in place that is aligned with the other endpoints and is more extensible to be able to add additional per-model arguments in the future if we need them.

packages/cli/src/ceramic-daemon.ts
dbcfd: Yeah, if we want to assume that the fields will be present after passing validation, then we don't have to use the callback approach. I was mainly using that in case the code changes, so that the code will break at compile time rather than run time.

stbrody: I don't quite follow how this structure helps catch more errors at compile time?  What kinds of issues would this catch that separating validation from execution would not?  Could these be solved by having validation return a more restricted type with just the parsed and validated fields back to the caller to then be passed into the execution layer?

I wonder how much of this will all become moot once we use a codec here

stbrody: for the record, I don't feel that strongly about this thread of comments - this is internal code structure we can always clean up later.  My other comment thread about the actual API structure is the more important comment and the only one I would hold up merging this PR over.

stbrody (approved):
LGTM!

Although I realized that since we don't expose a stopIndexingModelData function in the http client, the http endpoint for it is unlikely to ever be used.  But whatever, that seems fine

PR #2845, opened 2023-06-28
feat(core): CDB 2492 Add database indices for model fields

dbcfd:


packages/cli/src/ceramic-daemon.ts
```diff
@@ -877,10 +877,11 @@ export class CeramicDaemon {
...
 
-    const indexedModelStreamIDs = await this.ceramic.admin.getIndexedModels()
-    res.json({
-      models: indexedModelStreamIDs.map((modelStreamID) => modelStreamID.toString()),
-    })
+    const indexedModels = await this.ceramic.admin.getIndexedModelData()
+    const body = {
+      models: indexedModels.map((idx) => idx.streamID.toString()),
+    }
+    res.header('Deprecation', 'Wed, 21 Jun 2023 23:59:59 GMT').json(body)
```
dbcfd, line 884: Need to update this time and add to other calls.

packages/common/src/ceramic-api.ts
```diff
@@ -11,7 +11,7 @@ import type { AnchorStatus } from '@ceramicnetwork/common'
 /**
  * Field definition for index
  */
-export type Field = {
+export interface Field {
```
dbcfd, line 14: Reverting this

packages/http-client/src/__tests__/remote-admin-api.test.ts
```diff
@@ -95,7 +86,7 @@ test('stopIndexingModels()', async () => {
   await adminApi.stopIndexingModels([MODEL])
 
   expect(fauxFetch.mock.calls[0][0]).toEqual(new URL(`https://example.com/admin/getCode`))
-  expect(fauxFetch.mock.calls[1][0]).toEqual(new URL(`https://example.com/admin/models`))
+  expect(fauxFetch.mock.calls[1][0]).toEqual(new URL(`https://example.com/admin/modelData`))
```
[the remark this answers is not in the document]
dbcfd: All the existing models methods now delegate to modelData. Between that and being deprecated, don't think tests are needed.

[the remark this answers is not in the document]
PaulLeCam: So is there still any test or check that ensures this endpoint doesn't get removed from the codebase until we release a v3?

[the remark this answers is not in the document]
dbcfd: If someone removed the endpoint, they'd remove the test as well. This is better handled with a note in the function documentation, with a link to a removal plan. Then code reviewers can see if it should be removed in a PR.

packages/cli/src/ceramic-daemon.ts
```diff
@@ -242,6 +242,8 @@ type AdminApiModelIdsMutationMethod = {
   method: (modelsIDs: Array<StreamID>) => Promise<void>
 }
 
+const MODELS_DEPRECATION_AT = 'Thu, 29 Jun 2023 23:59:59 GMT'
```
stbrody, line 245: why does the timestamp of when it was deprecated matter to anyone?

stbrody: I think a more informative message like "the /models HTTP endpoint has been deprecated in favor of the /modelData endpoint" would be more useful

packages/cli/src/ceramic-daemon.ts
```diff
@@ -927,6 +932,7 @@ export class CeramicDaemon {
   }
 
   async startIndexingModels(req: Request, res: Response): Promise<void> {
+    res.header('Deprecation', MODELS_DEPRECATION_AT)
     await this._processAdminModelsMutationRequest(req, res, {
       type: 'modelIDs',
       method: (modelIDs) => this.ceramic.admin.startIndexingModels(modelIDs),
```
stbrody, line 938: nitpick, but it seems odd to rewrite `getIndexedModels` in terms of `getIndexedModelData` but leave this and `stopIndexingModels` using dedicated methods on the underlying admin api.  I'd either leave `getIndexedModels` alone or change all 3, just so we're consistent

It's just code structure though so doesn't really matter, can always be changed later

packages/common/src/ceramic-api.ts
```diff
@@ -91,10 +99,16 @@ export interface AdminApi {
...
   /**
+   * @deprecated
    * List indexed model streams
    */
   getIndexedModels(): Promise<Array<StreamID>>
 
+  /**
+   * List indexed model streams with data
+   */
+  getIndexedModelData(): Promise<Array<ModelData>>
```
stbrody, line 110: wasn't this already added in the previous PR?  Is this PR not merged to latest develop?

stbrody (commented):
started reviewing then noticed that artifacts from the last PR are still included in this one.  To avoid having to re-read a bunch of code I'll hold off on reviewing further until this PR is brought in sync with the develop branch

packages/cli/src/ceramic-daemon.ts
dbcfd: This is the format  specified for the header. https://datatracker.ietf.org/doc/html/draft-ietf-httpapi-deprecation-header-02#page-3

packages/common/src/ceramic-api.ts
dbcfd: This is up to date with develop, but this wasn't in the last PR. Last PR only focused on the external API. This method is required for all internal calls.

packages/cli/src/ceramic-daemon.ts
dbcfd: The underlying calls do the forwarding, so I'll change these back for consistency.

packages/core/src/indexing/database-index-api.ts
```diff
@@ -113,6 +161,7 @@ export abstract class DatabaseIndexApi<DateType = Date | number> {
         models.map((indexModelArgs) => {
           return {
             model: indexModelArgs.model.toString(),
+            is_indexed: true,
```
willex, line 164: good catch!

packages/core/src/indexing/migrations/1-create-model-table.ts
```diff
@@ -38,44 +38,48 @@ export type TableIndices = {
   indices: Array<TableIndex>
 }
 
+export function indexNameFromTableName(tableName: string): string {
+  return `idx_${tableName.substring(tableName.length - 10)}`
+}
+
 export function indices(tableName: string): TableIndices {
   // create unique index name less than 64 chars that are still capable of being referenced to MID table.
```
willex, line 46: Nitpick - can you move the comment above the newly created helper function so we have the explanation captured at the right place.

packages/core/src/indexing/database-index-api.ts
```diff
@@ -29,6 +32,16 @@ export interface IndexStreamArgs {
...
 }
 
+/**
+ * Create a valid name for a fields index, if the index has a name specified, or generate a valid
+ * name if one is not specified
+ * @param idx Index to create a name for
+ * @param table Table to add index to
+ */
+export function fieldsIndexName(idx: FieldsIndex, table: string): string {
+  return `${indexNameFromTableName(table)}_${idx.fields.flatMap((f) => f.path).join('_')}`
```
dbcfd, line 42: Should truncate to 64 characters

packages/core/src/indexing/migrations/1-create-model-table.ts
```diff
@@ -38,44 +38,48 @@ export type TableIndices = {
   indices: Array<TableIndex>
 }
 
+export function indexNameFromTableName(tableName: string): string {
```
willex, line 41: As discussed, needs to be more robust so we can stay within 64chars limit with variable field names.

packages/cli/src/ceramic-daemon.ts
```diff
@@ -872,15 +874,17 @@ export class CeramicDaemon {
   }
 
   async getIndexedModels(req: Request, res: Response): Promise<void> {
+    res.header('Deprecation', MODELS_DEPRECATION_AT)
```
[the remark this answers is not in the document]
willex: I'm comfortable with this if we can confirm:
- outdated client really logs the message / warning and its clear for developers that they need to update the client
- we can update the composedb cli client and verify its working during RC phase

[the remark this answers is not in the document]
dbcfd: Added additional steps for RC verification https://www.notion.so/threebox/Field-Indexing-e8cb899e2c9649acb84eeddf4a68a32e?pvs=4#b807efbd87a2408ea8e0dd3fddd0ab85

packages/http-client/src/__tests__/remote-admin-api.test.ts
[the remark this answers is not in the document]
willex: If this is covered as part of the release process I think we can cover this sufficiently; deprecation warning is given and it should handle correctly with different client / server combinations.

@dbcfd added addt. notes to the Notion doc for this as well.

willex (approved):
@dbcfd Looks good with the minor left over work as discussed through walking through the open issues.

packages/cli/src/ceramic-daemon.ts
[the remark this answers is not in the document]
PaulLeCam: > 



> I'm comfortable with this if we can confirm:
> 
>     * outdated client really logs the message / warning and its clear for developers that they need to update the client
> 
>     * we can update the composedb cli client and verify its working during RC phase

AFAIK we don't have anything like this implemented in our clients.
I think that would be a good thing to have, but probably out of scope for this project? Even if we ship with this project, the deprecation message wouldn't show up with current versions.

packages/common/src/ceramic-api.ts
stbrody: look on line 115, looks like the method definition wound up duplicated and included twice.  Maybe from a merge issue?  Can we remove the duplicate in a follow-up?

packages/common/src/ceramic-api.ts
```diff
@@ -114,13 +130,27 @@ export interface AdminApi {
   startIndexingModelData(modelData: Array<ModelData>): Promise<void>
 
   /**
+   * Adds model streams to index as specified by ModelData
+   * @param modelData - array of model streams with field indices to index
+   */
+  startIndexingModelData(modelData: Array<ModelData>): Promise<void>
```
stbrody, line 136: another duplicate

packages/core/src/indexing/__tests__/database-index-api.test.ts
```diff
@@ -14,7 +14,7 @@ import {
   IndexModelArgs,
   PostgresIndexApi,
   SqliteIndexApi,
-  asTimestamp,
+  asTimestamp, fieldsIndexName
```
stbrody, line 17: formatting - need to run prettier?

packages/core/src/indexing/database-index-api.ts
```diff
@@ -29,6 +32,20 @@ export interface IndexStreamArgs {
   readonly firstAnchor: Date | null
 }
 
+/**
+ * Create a valid name for a fields index, if the index has a name specified, or generate a valid
```
stbrody, line 36: this comment is not accurate anymore as we no longer let users specify the index name

packages/core/src/indexing/local-index-api.ts
```diff
@@ -17,27 +19,34 @@ import { Model } from '@ceramicnetwork/stream-model'
...
+
   if (modelStreamId.type == Model.STREAM_TYPE_ID) {
     const modelState = await repository.load(modelStreamId, {})
     const content = modelState.state.next?.content ?? modelState.state.content
     Model.assertVersionValid(content, 'major')
     if (content.relations) {
-      return { model: modelStreamId, relations: content.relations }
+      opts.relations = content.relations
     }
+    opts.indices = req.indices ?? (await databaseIndexApi?.getFieldsIndex(modelStreamId))
```
stbrody, line 46: I know I asked this like 3 times on the earlier version of this PR but I just still don't get why this is necessary.  In the follow-up PR can you try adding a comment here explaining why this is necessary, and if it's still not clear to me maybe we can hop on quick call to discuss?

packages/core/src/indexing/migrations/1-create-model-table.ts
stbrody: where did we land on this?  Is there a ticket somewhere capturing what remains to be done here?

packages/http-client/src/__tests__/remote-admin-api.test.ts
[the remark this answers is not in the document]
stbrody: FWIW I agree with Paul that in general as long as some code still exists and can be hit, deprecated or not, the tests for that code should remain.  It's not to prevent the code from being removed prematurely, it's to prevent the functionality from being accidentally broken while it's in the deprecated-but-not-yet-removed state.  As long as we want to support old clients hitting the old endpoint, we should keep the tests that ensure the old endpoint works as expected.

That said, not a huge deal and not worth going back to do anything about it at this point.

packages/http-client/src/remote-admin-api.ts
```diff
@@ -82,23 +83,16 @@ export class RemoteAdminApi implements AdminApi {
...
   async startIndexingModels(modelsIDs: Array<StreamID>): Promise<void> {
-    const code = await this.generateCode()
-    const body = modelIDsAsRequestBody(modelsIDs)
-    await this._fetchJson(this.getModelsUrl(), {
-      method: 'post',
-      body: {
-        jws: await this.buildJWS(this._getDidFn(), code, this.getModelsUrl().pathname, body),
-      },
-    })
+    await this.startIndexingModelData(convertModelIdsToModelData(modelsIDs))
```
stbrody, line 86: this worries me a bit.  Now if someone updates their client before updating their node, things will break.  We've never defined a standard upgrade order between the client and server, so far we've always allowed devs to upgrade in either order and maintained backwards compatibility in both directions.  I don't think that approach is sustainable long term - I think we will eventually (ideally soon) need to standardize and document whether it is expected that devs update their client or their server first / more often.  But until we set that expectation with our community, this change feels risky to me.

CC @oed

stbrody (commented):
mostly LGTM, though I have a few small requests for cleanups to be addressed in a follow-up PR, and one bigger more important question about upgrade order requirements.

packages/http-client/src/remote-admin-api.ts
dbcfd: We can just change to use the existing method, and leave the proxying to the daemon.

PR #2848, opened 2023-07-02
feat: Send ipfs version in a keepalive message

ukstv:


stbrody (approved):
LGTM

@smrz2001 do you know if our version of kubo that has the validator will return a different string here for the version?  It would be great to be able to track the prevalence of our custom ipfs builds as they get adopted.

stbrody:
> We can update the image so that it has a different version string, but that won't help for nodes already out there running the current image.

if it's an easy change I'd say let's do it.  We haven't actually promoted the new image with the validator specifically, and if we ever do it will be nice to be able to track uptake of it, even if imperfectly

PR #2851, opened 2023-07-06
fix (core): remove stream from the anchorRequestStore if anchor failed

stephhuynh18:
Before: If the anchor was for a non tip we would keep the stream in the anchorRequestStore
After: We only keep the stream in the anchorRequestStore if the anchor was for a non tip, and the tip has an anchor request.

### DIFF
--- a/packages/core/src/__tests__/state-manager.test.ts
+++ b/packages/core/src/__tests__/state-manager.test.ts
@@ -645,6 +645,7 @@ describe('anchor', () => {
         streamId: tile.id,
         cid: tile.state.log[0].cid,
         message: 'CAS accepted the request',
+        id: 'id',
       })
       await expectAnchorStatus(tile, AnchorStatus.PENDING)
 
@@ -659,6 +660,7 @@ describe('anchor', () => {
         streamId: tile.id,
         cid: tile.state.log[0].cid,
         message: 'Replaced',
+        id: 'id',
       })
 
       // Polling for the 1st commit should stop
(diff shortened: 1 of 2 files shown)


packages/core/src/state-management/state-manager.ts
```diff
@@ -447,12 +447,27 @@ export class StateManager {
...
+
+              // if this is the anchor response for the tip update the state
+              if (asr.cid.equals(state$.tip)) {
+                state$.next({ ...state$.value, anchorStatus: AnchorStatus.FAILED })
+              }
+
+              // If state has become a terminal anchor state, or already was a terminal anchor state, remove it from the anchor store.
+              // if the anchor response was not for the tip, and there was no request to anchor the tip, remove any anchor requests for the stream from the store
+              if (
+                state$.state.anchorStatus === AnchorStatus.FAILED ||
```
ukstv, line 462:
```
const stillProcessing = [AnchorStatus.PENDING, AnchorStatus.PROCESSING].include(state$.value.anchorStatus)
if (!stillProcessing) {
//...
}`

File: packages/core/src/__tests__/state-manager.test.ts [hunk 2]
```diff
@@ -828,5 +830,248 @@ describe('anchor', () => {
       await TestUtils.delay(2000) // Wait a bit to confirm that the request is *not* deleted from the anchor request store
       await expect(anchorRequestStore.load(stream.id)).resolves.not.toBeNull()
     })
+
+    describe('handling of failed anchors', () => {
```
stbrody, line 834: I love how thorough and complete these test cases are.  nice job

stbrody (commented):
I don't actually understand what the problem was that this fixes.  We already remove the request from the persistent store whenever the request for the tip reaches a terminal state.  How would we wind up in a case where a request for an older commit is still polling and then fails, the tip request is already in a terminal state, but we still have the request in the persistent store?

stephhuynh18:
> I don't actually understand what the problem was that this fixes. We already remove the request from the persistent store whenever the request for the tip reaches a terminal state. How would we wind up in a case where a request for an older commit is still polling and then fails, the tip request is already in a terminal state, but we still have the request in the persistent store?

@stbrody if the anchor response was non terminal we did not stop the polling subscription (we did not call `stopSignal.next()`. The reason we didn't do this before was because `FAILED` was not a terminal state (because we retried failed requests). But with the changes @ukstv  has done to the CAS it is now a terminal state. 

As I was writing tests I also found the non tip but tip has no anchor request case and added that one in

stbrody:
> if the anchor response was non terminal we did not stop the polling subscription (we did not call `stopSignal.next()`. The reason we didn't do this before was because `FAILED` was not a terminal state (because we retried failed requests). But with the changes @ukstv has done to the CAS it is now a terminal state.

Sorry, I still don't really follow what the issue was.
We already did indeed call `stopSignal.next()` on the FAILED status so long as it was for the current tip.  It seems to me that this PR only changes behavior in the following scenario:

1. The tip is in a terminal state
2. we still have the stream in the persistent store
3. a non-tip request is still polling and then enters FAILED state

If (1) is true I don't see how (2) could be true, even without this change

stbrody:
Another way of saying this is: why should status changes to requests that are *not* for the tip ever make any changes to whether or not the stream is in the persistent store?  Shouldn't that be driven entirely by changes to the status of the request for the tip?

PR #2859, opened 2023-07-11
bug(core): CDB-2599 Migrate existing tables to have new columns

dbcfd:


ukstv (changes requested):
I confirm that it works. The only problem is failing tests.

dbcfd:
@ukstv reran the pipeline and tests passed. That test has kind of a short timeout (30 seconds) compared to our other tests.

ukstv (approved):

PR #2861, opened 2023-07-11
bug(core): CDB-2601 Handle issues with null queries

dbcfd:
Add additional tests for null queries not working. Additionall test string based queries.


PaulLeCam (approved):
Thanks!

willex (approved):
Thanks for the quick turnaround!

PR #2863, opened 2023-07-12
test(blockchain-utils-validation): Disable eos test

ukstv:
It fails due to 3rd party endpoint having an expired SSL certificate.


stbrody (approved):
lgtm.

We really need to figure out some sort of sustainable system for maintaining the integrations with 3rd party wallets.  Leaving them as "supported" but not actually tested at all is just asking for trouble.  CC @bigs

PR #2864, opened 2023-07-13
feat(core): query results ordering on custom fields

PaulLeCam:
~Tests pass in SQLite but not Postgres so I still need to figure out the issue, but I think the logic can be reviewed.~

This PR changes the cursor format: instead of only storing the `created_at` value to order results, the values of content fields can be set according to the `sorting` fields defined in the query. The stream ID is also stored in the cursor as a tie-breaker for ordering.

The main ordering logic is to check the `sorting` object in the query, or fallback on ordering by the `created_at` field when `sorting` is not set or empty, to keep the existing behavior.


packages/core/src/indexing/insertion-order.ts
```diff
@@ -28,45 +36,69 @@ type QueryFunc = (bldr: Knex.QueryBuilder<any, any>) => Knex.QueryBuilder<any, a
...
-  if (!input) return undefined
-  return { created_at: input.created_at }
+    let cursor: CursorData
+    if (orderByKeys.length === 0) {
+      // Use `created_at` field
+      cursor = { type: 'timestamp', id: input.stream_id, value: input.created_at }
+    } else {
+      // Use custom content fields
+      const content = JSON.parse(input.stream_content)
+      cursor = orderByKeys.reduce(
```
stbrody, line 68: NITPICK, feel free to ignore:

is this code equivalent to:
```
cursor = { type: 'content', id: input.stream_id, value: {} }
for (const key of orderByKeys) {
  if (content[key] != null) {
    cursor.value[key] = content[key]
  }
}
```
?

If so, I'd personally prefer the for-loop version as I think it's much more immediately obvious to a reader what is happening.  I generally try to err on the side of verbosity and straightforwardness over anything that comes close to "magic", even at the cost of "elegance".  
The reduce function is standard and well-understood enough to not really qualify as magic, so I think it's fine if you prefer to leave it as-written, but I at least had to think for a few seconds to understand what this was doing, and anytime a reader has to think for a few seconds to understand some code it is an increase in cognitive load for the reader and a signal that it should be simplified if possible, IMHO.

packages/core/src/indexing/insertion-order.ts
```diff
@@ -177,11 +197,45 @@ export class InsertionOrder {
         base = base.where(converted.where)
       }
 
-      if (isReverseOrder) {
-        base = base.orderBy(reverseOrder(INSERTION_ORDER))
+      // Handle cursor if present
```
stbrody, line 200: this function is getting very long and hard to read.  Can we break it up into some helpers to improve readability?

packages/core/src/indexing/insertion-order.ts
```diff
@@ -177,11 +197,45 @@ export class InsertionOrder {
...
+        } else {
+          // Paginate using previous values of custom fields
+          for (const [key, prevValue] of Object.entries(cursor.value)) {
+            const field = contentKey(key)
+            const sign = getComparisonSign(query.sorting?.[key], isReverseOrder)
+            const value = JSON.stringify(prevValue)
+            base = base.where((qb) => {
+              qb.whereRaw(`${field} ${sign} ${value}`) // strict next value
+                .orWhereRaw(`${field} = ${value}`) // or current value
+                .andWhere('stream_id', '>', cursor.id) // with stream ID tie-breaker
```
stbrody, line 218: should we still use created_at as the first tiebreaker for the explicit sort keys and only fall back to stream_id in the rare event of a tie on created_at?

packages/core/src/indexing/insertion-order.ts
```diff
@@ -177,11 +197,45 @@ export class InsertionOrder {
...
+        base = base.orderBy(isReverseOrder ? reverseOrder(INSERTION_ORDER) : INSERTION_ORDER)
       } else {
-        base = base.orderBy(INSERTION_ORDER)
+        // Order by custom fields
+        for (const [field, order] of sortingEntries) {
+          const orderBy = isReverseOrder ? REVERSE_ORDER[order] : order
+          base = base.orderByRaw(`${contentKey(field)} ${orderBy}`)
+        }
       }
+      // Always order by stream ID as tie-breaker
```
stbrody, line 236: + created_at first?

packages/stream-tests/src/__tests__/basic-indexing.test.ts
```diff
@@ -521,6 +519,224 @@ describe.each(envs)('Basic end-to-end indexing query test for $dbEngine', (env)
...
 
+  describe('queries with custom sorting', () => {
+    test('multiple documents - one page - ASC order with forward pagination', async () => {
+      const doc0 = await ModelInstanceDocument.create(ceramic, CONTENT0, midMetadata)
+      const doc1 = await ModelInstanceDocument.create(ceramic, CONTENT1, midMetadata)
+      const doc2 = await ModelInstanceDocument.create(ceramic, CONTENT2, midMetadata)
+
+      const resultObj = await ceramic.index.query({
+        model: model.id,
+        sorting: { myData: 'ASC' },
```
stbrody, line 530: I want to make sure we leave room for this to be extensible.  Like you might want to be able to specify more complex sorting behavior than just a set of content fields.  For example, you might want to sort on anchored timestamp or updated_at timestamp instead of created_at.  You also might want a combination of sorting on custom content fields PLUS some control over built-in fields like created_at, updated_at, etc.

So as to leave this more extensible for the future, I don't think we should put the custom fields at the top level of the sorting object.  But instead putting it behind a `byFields` parameter so it looks like `sorting: { byFields: { myData: 'ASC' } }` would leave us the option of adding other reserved top-level arguments to the sorting object in the future that could further fine-tune the behavior.

stbrody: now that I think about it, the same argument could be made for the filters themselves.  Like, do we have a way today to filter on `created_at` greater than some timestamp?  If not, we should think about how we might add such capability in the future and make sure our syntax/api is extensible enough to allow that.

stbrody (changes requested):
a few small nitpicks, but my main concern is around future-proofing the API to allow for expressing more complex sorting and filtering logic in the future

packages/core/src/indexing/insertion-order.ts
```diff
@@ -11,13 +11,21 @@ import {
...
 import { UnsupportedOrderingError } from './unsupported-ordering-error.js'
 import { addColumnPrefix } from './column-name.util.js'
-import { convertQueryFilter, DATA_FIELD } from './query-filter-converter.js'
+import { contentKey, convertQueryFilter, DATA_FIELD } from './query-filter-converter.js'
 import { parseQueryFilters } from './query-filter-parser.js'
 
-type SelectedRequired = { stream_id: string; last_anchored_at: number; created_at: number }
-type SelectedOptional = Record<string, boolean | number | string>
-type Selected = SelectedRequired & SelectedOptional
+type StreamContent = Record<string, boolean | number | string>
```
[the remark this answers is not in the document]
PaulLeCam: Yes good catch, I just renamed the existing `SelectedOptional` but I'll update the value to `unknown`

packages/core/src/indexing/insertion-order.ts
PaulLeCam: Yes no problem changing to a for loop

packages/core/src/indexing/insertion-order.ts
PaulLeCam: With the logic here the cursor always contains the stream `id`, but either contains the `created_at` value or the custom field(s) value(s).
We could always store the `created_at` value in the cursor, but we need to define the behavior more I think: do we also need to order by `created_at` in any way? It would add complexity to the logic and it's not a "fully reliable" tie-breaker as we'd still need to use the `stream_id` so I'd rather just use the `stream_id` if possible?
Also always adding the `created_at` field to the cursor makes the cursor longer. I don't think it's a big issue but seems a bit wasteful to make the payloads larger than they need to be.

packages/stream-tests/src/__tests__/basic-indexing.test.ts
PaulLeCam: I think we should leave the `created_at` and `updated_at` abstracted from the consumers. These values are defined in the DB, not across the network so the behavior could be confusing, I'd much rather make it explicit that these need to be defined in the contents if apps want to filter and/or order by them.

PaulLeCam: Regarding future-proofing, what we've done so far is adding fields to the `BaseQuery` object, for example we have the `filter` object that we're using for relations support, and the new `queryFilters` object with custom filters, so we could have the `sorting` field as-is for now and add a new object if we need to change the shape?
Eventually it can be abstracted at the GraphQL level anyway so I don't mind having a `byFields` object in `sorting` if you prefer, the logic wouldn't change much either way.

stbrody: yeah, I kind of hate that we have both `filter` and `queryFilters`.  I really wonder if we can't combine them into one?
By the same token I slightly prefer pushing down a byFields subobj into `sorting` so that `sorting` itself can be extended, but you're right that a new top-level option could accomplish the same goal so I ultimately don't feel strongly either way

packages/core/src/indexing/insertion-order.ts
stbrody: yeah, I was thinking we would add `created_at` to the sort as well, as it's a slightly more meaningful thing to sort on than `stream_id`.  Really ideally the default sort would be `last_anchored_at` first, then `created_at`, then `stream_id`.  But yes, that does keep inflating the size of the cursor we pass back and forth between the client and server.

But I don't feel super strongly about this, we can always change later if users ask for different behavior.

packages/core/src/indexing/insertion-order.ts
```diff
@@ -128,72 +165,116 @@ export class InsertionOrder {
...
+    if (query.queryFilters) {
+      const parsed = parseQueryFilters(query.queryFilters)
+      const converted = convertQueryFilter(parsed)
+      if (converted) {
+        builder = builder.where(converted.where)
       }
-      if (query.account) {
-        base = base.where({ controller_did: query.account })
+    } else if (query.filter) {
+      // Handle legacy `filter` object used for relations
```
stbrody, line 221: what if you query both on a relation and a custom indexed field?  Does it all go into queryFilters?  If so, could we just use the existing `filter` arg for both?  Is there a reason we need to introduce a new arg here?

stbrody (approved):
lgtm mod question around if we can get rid of `queryFilters` and just use `filters` for everything

packages/core/src/indexing/insertion-order.ts
PaulLeCam: With v0.5 ComposeDB uses only `queryFilters` for both relations and custom fields.
We can't use the existing `filter` arg without breaking backwards compatibility, but we should remove it in the next major version of Ceramic.

packages/core/src/indexing/insertion-order.ts
```diff
@@ -28,45 +40,69 @@ type QueryFunc = (bldr: Knex.QueryBuilder<any, any>) => Knex.QueryBuilder<any, a
...
-  if (!input) return undefined
-  return { created_at: input.created_at }
+    let cursor: CursorData
+    if (orderByKeys.length === 0) {
+      // Use `created_at` field
+      cursor = { type: 'timestamp', id: input.stream_id, value: input.created_at }
+    } else {
+      // Use custom content fields
+      const content =
+        typeof input.stream_content === 'string'
```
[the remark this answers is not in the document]
stbrody: ah yeah I encountered that in a recent PR of mine also.  I handled it by pushing the parsing into the db-specific code.  See https://github.com/ceramicnetwork/js-ceramic/blob/832b97a07c5e02de5776a5cbc444c2ce6b1c1ea9/packages/core/src/indexing/database-index-api.ts#L417-L419 vs https://github.com/ceramicnetwork/js-ceramic/blob/832b97a07c5e02de5776a5cbc444c2ce6b1c1ea9/packages/core/src/indexing/database-index-api.ts#L366-L368

packages/core/src/indexing/insertion-order.ts
stbrody: why can't we re-use filter?  Is the format of how the args within it are specified so different?

willex (approved):
LGTM!

packages/core/src/indexing/insertion-order.ts
PaulLeCam: Yes, cf the type definitions here: https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/common/src/index-api.ts#L116-L117

stbrody: I see.  So are queries on relations expressed as `where` queries now?

PaulLeCam: Yes currently they are expressed as `where` queries with exact values matches, as it's what we needed for relations (matching the wanted stream ID)

PR #2865, opened 2023-07-18
fix(core): fix anchor retry logic

stephhuynh18:
This addresses the following problems: 
* Not terminating polling for anchor request responses when we receive a FAILED anchor request response thats not the tip. We did not terminate polling before because FAILED requests were retried. But this has changed and now FAILED is a terminal state. 
* ~~Removing streams from the anchor request store when we receive anchor request responses for non tips. If the node were to restart we would not resume anchor response polling for those streams, even if we were waiting for a response for the tip~~
* We resume polling for stream TIPS, even if the tip does not have a requested anchor.


packages/core/src/state-management/state-manager.ts
```diff
@@ -442,17 +454,28 @@ export class StateManager {
...
               )
-              state$.next({ ...state$.value, anchorStatus: AnchorStatus.FAILED })
-              await this.anchorRequestStore.remove(state$.id)
+
+              // if this is the anchor response for the tip update the state
+              if (asr.cid.equals(state$.tip)) {
+                state$.next({ ...state$.value, anchorStatus: AnchorStatus.FAILED })
+              }
+
+              if (!this._tipHasProcessingAnchorRequest(state$)) {
```
stbrody, line 474: so is this just an extra safety check just in case?  Is there ever a scenario we would expect this to actually run?

stbrody (commented):
> Not terminating polling for anchor request responses when we receive a FAILED anchor request response. We did not terminate polling before because FAILED requests were retried. But this has changed and now FAILED is a terminal state.

Woah, that would definitely be a big oversight.  But I do see us remove the request from the AnchorRequestStore when we encounter a FAILED status, and I know from the issue @ukstv has been investigating that the in-memory polling stops when it encounters a FAILED status.  Was this behavior you actively observed ever?  I also don't see anything in this PR that looks like it would change this behavior?

I'm still not really sure I fully understand exactly what the problem with the current behavior here is that we're trying to solve, although the specific changes in this PR all seem reasonable

packages/core/src/state-management/state-manager.ts
```diff
@@ -442,17 +454,28 @@ export class StateManager {
...
+                await this.anchorRequestStore.remove(state$.id)
```
stbrody, line 475: do we also need to remove from the `anchorRequestStore` for the REPLACED status below?

stephhuynh18:
> Woah, that would definitely be a big oversight. But I do see us remove the request from the AnchorRequestStore when we encounter a FAILED status, and I know from the issue @ukstv has been investigating that the in-memory polling stops when it encounters a FAILED status. Was this behavior you actively observed ever? I also don't see anything in this PR that looks like it would change this behavior?

Oops miswrote this, we don't stop polling for failed anchor requests if they are not the tip. Ex. we create commitA and send an anchor request. We then create commitB on top of commitA and send an anchor request for that. If the anchor request for commitA fails we do not stop polling for it because it is not the tip, even if the ancho request for commitB comes back as completed or failed. 

I have not observed this but i was trying to come up with reasons to why Gitcoin's node was receiving anchor responses for old streams (https://linear.app/3boxlabs/issue/CDB-2578/investigate-anchor-retry-logic), and discovered this edge case while going through the possible cases.

packages/core/src/__tests__/state-manager.test.ts
```diff
@@ -828,5 +831,354 @@ describe('anchor', () => {
...
+          cid: tile.state.log[1].cid,
+          message: 'CAS accepted the request',
+          id: 'id',
+        })
+
+        // Polling for the 1st commit should stop
+        await expect(whenSubscriptionDone(firstAnchorResponseSub)).resolves.not.toThrow()
+        expect(firstAnchorResponseSub.closed).toBeTruthy()
+
+        // SERGEY: the stream should still be in the anchor request store so this is very wrong
```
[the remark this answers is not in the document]
stephhuynh18: not relevant. I was using this test to demonstrate this issue

packages/core/src/state-management/state-manager.ts
```diff
@@ -336,6 +336,14 @@ export class StateManager {
     const carFile = await this._buildAnchorRequestCARFile(state$.id, state$.tip)
     const genesisCID = state$.value.log[0].cid
     const genesisCommit = carFile.get(genesisCID)
+
+    // Set the anchor status to PENDING before it is added to the anchor request store.
+    // We do this to prevent the stream from being removed from the anchor request store by previous anchor requests.
```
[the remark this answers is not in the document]
stephhuynh18: Currently 

1. We request an anchor for CommitA for Stream
2. CommitA is successfully anchored and we receive the anchor cid from pubsub. We have yet to receive the response through the anchor response polling 
3. We create CommitB (anchor status NOT_REQUESTED) and anchor it. When we anchor a stream we add it to the request store before `_processAnchorResponse` is called. `_processAnchorResponse` is what updates the Stream's anchor status to PENDING 
4. We receive the anchor response for CommitA. We remove Stream from the anchor request store. There is nothing in the anchor request store for CommitB

We should have something in the anchor request store for CommitB so
4. We receive the anchor response for CommitA. Since CommitA is not the tip we do not remove Stream from the anchor request store.

Unfortunately we cannot rely on only the tip, because there could be a situation where the tip does not have an outstanding anchor request

4. We receive the anchor response for CommitA. Since its not the tip, and the tip has a requested anchor (the anchor status is PENDING) we do not remove the stream from the store

BUT because we could be receiving the anchor response after we have added the stream to the anchor request store but before the anchor status of the Stream has been updated to PENDING via `_processAnchorResponse` we could still erroneously remove the Stream from the anchor request store. 

We could only remove things from the store if the stream and cid matches, but our current stores do not support atomic transactions (a read + write in one go). 

So the solution I have ended with is updating the anchor status to PENDING before we add it to the anchor request store. We could get into a situation where the anchor status has been updated to PENDING but things crash before we add the stream to the anchor request store but i'm not sure what to do there.

packages/core/src/state-management/state-manager.ts
```diff
@@ -336,6 +336,14 @@ export class StateManager {
...
+    const next = {
+      ...state$.value,
+      anchorStatus: AnchorStatus.PENDING,
```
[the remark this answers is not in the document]
stephhuynh18: sure i don't mind =)

packages/core/src/state-management/state-manager.ts
stephhuynh18: The scenario could be:
We have CommitA requested. CommitA is PROCESSING. Before waiting for CommitA to complete or fail we request CommitB. If CommitA fails we don't want to remove the stream from the anchorRequestStore because CommitB might pass. 


If CommitA passes we would be fine removing the stream from the anchorRequestStore because the anchor for CommitB will be invalid anyway. But that will make the logic very confusing.

packages/core/src/state-management/state-manager.ts
stephhuynh18: I think we can safely assume if we received a REPLACED status, that means that we have another outstanding request for a commit for that stream

stephhuynh18:
Also @stbrody feel free to vote this PR unnecessary. This was a weird crazy tangent I followed with possible but unlikely edge cases.

packages/core/src/state-management/state-manager.ts
stephhuynh18: I can also add an extra check as above. It seems based on https://linear.app/3boxlabs/issue/CDB-2622/investigate-why-ceramic-node-is-polling-for-very-old-anchor-request we can get into very weird states.

packages/core/src/state-management/state-manager.ts
stbrody: I guess I meant but where `asr.cid.equals(state$.tip)` is false.  I'm wondering why we need any changes to the `FAILED` case at all.  It seems like the case where `this._tipHasProcessingAnchorRequest` is true would always require that `asr.cid.equals(state$.tip)` be false, in which case with the original logic we would have done nothing here anyway

packages/core/src/state-management/state-manager.ts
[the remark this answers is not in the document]
stbrody: okay, I *think* I get it.

If we swap the order of writing to the persistent anchor request store and updating the state to PENDING, does that address the race condition you mention at the end where the node crashes before the stream is added to the anchor request store?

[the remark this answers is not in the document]
stephhuynh18: No it does not. 

We really need some sort of db transaction here.

packages/core/src/state-management/state-manager.ts
stephhuynh18: > It seems like the case where this._tipHasProcessingAnchorRequest is true would always require that asr.cid.equals(state$.tip) be false,

Can state$ have been updated in between `asr.cid.equals(state$.tip)` and `this._tipHasProcessingAnchorRequest`.... Maybe but then this doesn't fix anything anyway. The more that i think about this the more I think any approach other than DB transaction won't work. 

I'mg oing to reduce the scope of this PR to things that can actually be fixed and file the ticket for the rest

ukstv (approved):

ukstv:
I think we should merge this.

PR #2867, opened 2023-07-19
Cleanups to indexing initialization and fieldIndex tracking

stbrody:
No real functionality changes, but cleans up and simplifies some of the logic, reduces the number of reads/writes to the database, and expands test coverage a bit.

### DIFF
--- a/packages/core/src/indexing/local-index-api.ts
+++ b/packages/core/src/indexing/local-index-api.ts
@@ -22,11 +22,7 @@ import { ISyncQueryApi } from '../sync/interfaces.js'
  * Takes a ModelData, loads it, and returns the IndexModelArgs necessary to prepare the
  * database for indexing that model.
  */
-async function _getIndexModelArgs(
-  req: ModelData,
-  repository: Repository,
-  databaseIndexApi?: DatabaseIndexApi
-): Promise<IndexModelArgs> {
+async function _getIndexModelArgs(req: ModelData, repository: Repository): Promise<IndexModelArgs> {
   const modelStreamId = req.streamID
   if (modelStreamId.type != Model.STREAM_TYPE_ID && !modelStreamId.equals(Model.MODEL)) {
     throw new Error(`Cannot index ${modelStreamId.toString()}, it is not a Model StreamID`)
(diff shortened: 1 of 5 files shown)


packages/core/src/indexing/local-index-api.ts
```diff
@@ -105,35 +101,7 @@ export class LocalIndexApi implements IndexApi {
...
-            node: node,
-          }
-        })
-      )
-      return {
-        edges: edges,
-        pageInfo: page.pageInfo,
-      }
-    } else {
+    if (!this.databaseIndexApi) {
```
stbrody, line 104: no functional changes to this function, just using an early return to remove an indentation layer and improve readability.

packages/core/src/indexing/tables-manager.ts
```diff
@@ -269,7 +269,7 @@ export class PostgresTablesManager extends TablesManager {
    * @param args
    */
   async hasMidIndices(tableName: string, args: IndexModelArgs): Promise<boolean> {
-    const expectedIndices = indices(tableName).indices.flatMap((index) => index.name)
+    const expectedIndices = defaultIndices(tableName).indices.flatMap((index) => index.name)
     if (args && args.indices) {
```
dbcfd, line 273: If we're removing the ability to save the args in the database, then we'll want to remove this check as well. As it is, if composedb sets indices for a model, ceramic is shutdown, then brought back up with the model specified in config, we will have a migration failure, since indices are present that shouldn't be.

dbcfd (changes requested):

packages/core/src/indexing/tables-manager.ts
stbrody: this PR doesn't remove tracking the args in the database.  I'm still exploring if there's a way to actually reconstruct the index definitions just from the native index metadata in the database or not.  If I go that route though that will be in a different PR.  This PR has no functionality change, just cleanup.

> As it is, if composedb sets indices for a model, ceramic is shutdown, then brought back up with the model specified in config

I'm not quite sure what you mean by "brought back up with the model specified in config"?

dbcfd: Thought the hooks still existed, but looks like config hooks are gone now. Should be good to go then.

dbcfd (approved):

PR #2873, opened 2023-07-26
fix(core): Fix CAS DID-auth

stbrody:


packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -229,9 +229,16 @@ export class AuthenticatedEthereumAnchorService
     auth: AnchorServiceAuth,
     readonly anchorServiceUrl: string,
     logger: DiagnosticsLogger,
-    pollInterval: number = DEFAULT_POLL_INTERVAL
+    pollInterval: number = DEFAULT_POLL_INTERVAL,
+    maxPollTime: number = MAX_POLL_TIME
```
stbrody, line 233: in https://github.com/ceramicnetwork/js-ceramic/commit/07ff4e5b0558fd3ae6069c2fe2bdff5c6de69306 we added the `maxPollTime` arg to `EthereumAnchorService` but didn't propagate the change up to `AuthenticatedEthereumAnchorService`.  It would have been nice if the type system had caught this for us - we were passing a function handle to a constructor that was expecting a `number` for that field, but everything compiled fine.

ukstv: Nice catch!

ukstv (approved):

PR #2875, opened 2023-07-28
Add logs to streams writes steps

PaulLeCam:
Tested locally by creating and updating a stream using the CLI


stbrody (commented):
these logs seem extremely verbose and will likely fill up the log files and make them hard to read.  Can we put these messages at the `trace` log level instead of `debug`?

PaulLeCam:
@stbrody I haven't seen a `trace` log level but it seems the one under `debug` is `verbose`, is it OK like this please?

stbrody (approved):
whoops yes that's what I meant.

lgtm

PR #2876, opened 2023-07-28
fix(http-client): Add a dummy pin api to not break old callers

ukstv:
Do `console.warn` with a message `You use a deprecated ceramic.pin.${operation} API. Please, reconsider your choices`

Suggestions for a better message are welcome.


stbrody (commented):
My preference would be to hold off on merging this until we talk to disco later today.  If they do in fact have a working fork of 3id that they can build, then maybe we can just have them delete the line that calls the pin api, and we don't need to put this in at all

ukstv:
Yup. I agree with the course of action.

stbrody (approved):
sounds like disco can't easily work around this, so let's go ahead and merge this, sad as it makes me

PR #2877, opened 2023-08-01
feat: add recon service api

zachferland:
Adds flag/config to support recon service api. Currently intended for testing/benchmarking env only (not backwards support, run with pubsub off, etc). Tries to minimize changes outside of recon service.

Add config 
```
{
  ...
  "node": {
    "recon-url": "http://127.0.0.1:6001"
    ...
  },
  ...
}
```

References: 
Spec - https://www.notion.so/threebox/Recon-js-ceramic-3964bec706434050ad6a80653cea990a
Rust Ceramic - https://github.com/3box/rust-ceramic/pull/60


oed:
Why are we using `recon-url` and not `enable-recon` with a boolean? I thought we wanted rust-ceramic to expose the same http server for ipfs (kubo) as well as recon?

Maybe this is not the case @nathanielc ?

nathanielc:
For technical reasons we were not able to expose both Kubo RPC and the Ceramic API on the same port. (They use different incompatible http frameworks) We can port Kubo RPC over to the framework Ceramic API  uses but it wasn't a priority to start.

oed:
Ok, thanks for the context @nathanielc. Seems like it might be worth to do that later when we want to migrate users to use rust-ceramic instead of kubo.

PR #2878, opened 2023-08-02
chore: use stricter TS config for 3id and blockchain-related packages

ukstv:


packages/3id-did-resolver/src/index.ts
```diff
@@ -109,9 +115,9 @@ function extractMetadata(
   const { timestamp: updated, cid: versionId } = lastAnchorOrGenesisEntry(requestedVersionState.log)
 
   const { timestamp: nextUpdate, cid: nextVersionId } =
-    latestVersionState.log.find(
-      ({ timestamp }) => timestamp > updated || (!updated && timestamp)
-    ) || {}
+    latestVersionState.log.find(({ timestamp }) => {
+      return (timestamp && updated && timestamp > updated) || (!updated && timestamp)
```
dbcfd, line 119: It's _ok_ like this, but might be more readable as `return (timestamp && ((updated && timestamp > updated) || !updated))`

dbcfd: As it is currently written, it just takes a little bit longer to understand.

packages/3id-did-resolver/src/legacyResolver.ts
```diff
@@ -63,14 +63,22 @@ const compressKey = (key: string) => {
...
+  didId: string,
+  ipfs = ipfsMock
+): Promise<LegacyResolverResult> {
   const doc = (await ipfs.dag.get(didId)).value
   let signingKey, encryptionKey
 
   try {
-    const keyEntrySigning = doc.publicKey.findIndex((e) => e.id.endsWith('signingKey'))
-    const keyEntryEncryption = doc.publicKey.findIndex((e) => e.id.endsWith('encryptionKey'))
+    const keyEntrySigning = doc.publicKey.findIndex((e: any) => e.id.endsWith('signingKey'))
```
dbcfd, line 80: This seems like it should have a type other than any.

packages/blockchain-utils-linking/src/ethereum.ts
```diff
@@ -133,16 +133,16 @@ function safeSend(provider: any, method: string, params?: Array<any>): Promise<a
 
   if (provider.request) {
     return provider.request({ method, params }).then(
-      (response) => response,
-      (error) => {
+      (response: any) => response,
```
dbcfd, line 136: Same as above, be nice if this had a type other than any.

dbcfd (changes requested):

packages/3id-did-resolver/src/index.ts
ukstv: Yup. Changed that to the suggested expression.

packages/blockchain-utils-linking/src/ethereum.ts
ukstv: I just realised that `.then` call here is a no-op :/ so removed it.

dbcfd (approved):

PR #2879, opened 2023-08-02
feat(cli): Don't allow running IPFS in bundled mode on mainnet

stbrody:
Note this is technically a breaking change for any deployments currently relying on bundled mode on mainnet.  If they upgrade to the new version their node will fail to start up unless they switch to a dedicated IPFS node. Do we need to make a major version bump then?  Or can we just communicate it on discord and to the partners we know are using 'bundled' mode?

FYI @jpham2023

### DIFF
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -13,6 +13,7 @@ import {
   LogStyle,
   ModelData,
   MultiQuery,
+  Networks,
   StreamUtils,
   SyncOptions,
 } from '@ceramicnetwork/common'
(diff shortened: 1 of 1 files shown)


stbrody:
ping @oed 

@PaulLeCam maybe this can be added to the list of things for the breaking Ceramic v3 release

oed:
What's the reason for this change?

stbrody:
> What's the reason for this change?

We recently noticed that multiple big parters were using "bundled" mode in production (Orbis for sure, and I think maybe also Disco or Soulbonds?  I can't remember who the second one was, maybe @smrz2001 remembers).  Bundled mode installs vanilla kubo without any of the tweaks we've made in the Docker image to ensure connectivity to our infra "bootstrap" nodes, disabling the proxy feature that was making connections unreliable, etc.

smrz2001:
> > What's the reason for this change?
> 
> We recently noticed that multiple big parters were using "bundled" mode in production (Orbis for sure, and I think maybe also Disco or Soulbonds? I can't remember who the second one was, maybe @smrz2001 remembers). Bundled mode installs vanilla kubo without any of the tweaks we've made in the Docker image to ensure connectivity to our infra "bootstrap" nodes, disabling the relay client feature that was making connections unreliable, etc.

Yes, it was Soulbonds.

PaulLeCam:
A major version bump seems safer but that can be up to product if we know all the affected partners and communicate with them, we could release as minor.
I'm wondering though why we provide our fork of Kubo only via docker and not directly via npm? This way the bundled mode would still be usable, no?

ukstv:
@stbrody Mind merging this to v3 progress branch please?

stbrody:
> @stbrody Mind merging this to v3 progress branch please?

Good call!  Changed the base branch, rebased to develop, and set you as the reviewer.

Unfortunately I can't really test this currently because of https://linear.app/3boxlabs/issue/CAT-2829/-gateway-flag-is-broken

ukstv (approved):

PR #2881, opened 2023-08-04
feat(core, anchor-utils): sync start block 24 hours before model anchor

stephhuynh18:
For each indexed model - load the model and check when it was anchored.  Take the Model with the earliest anchor time, and then start your sync at 24 hours before that time (given that we promise to always anchor a write within 24 hours).

Another circular dependency introduced in the local admin api. sad

### DIFF
--- a/packages/core/src/local-admin-api.ts
+++ b/packages/core/src/local-admin-api.ts
@@ -4,12 +4,16 @@ import {
   ModelData,
   NodeStatusResponse,
   PinApi,
+  LoadOpts,
 } from '@ceramicnetwork/common'
-import { StreamID } from '@ceramicnetwork/streamid'
+import { Model } from '@ceramicnetwork/stream-model'
+import { StreamID, CommitID } from '@ceramicnetwork/streamid'
 import { LocalIndexApi } from './indexing/local-index-api.js'
-import { SyncApi } from './sync/sync-api.js'
+import { ISyncApi, ModelSyncOptions } from './sync/interfaces.js'
+import { convertCidToEthHash } from '@ceramicnetwork/anchor-utils'
 
 type NodeStatusFn = () => Promise<NodeStatusResponse>
+type LoadStreamFn<T> = (streamId: StreamID | CommitID | string, opts?: LoadOpts) => Promise<T>
 
 /**
  * AdminApi for Ceramic core.
@@ -17,9 +21,10 @@ type NodeStatusFn = () => Promise<NodeStatusResponse>
 export class LocalAdminApi implements AdminApi {
   constructor(
     private readonly indexApi: LocalIndexApi,
-    private readonly syncApi: SyncApi,
+    private readonly syncApi: ISyncApi,
     private readonly nodeStatusFn: NodeStatusFn, // TODO(CDB-2293): circular dependency back into Ceramic
-    private readonly pinApi: PinApi
+    private readonly pinApi: PinApi,
+    private readonly loadStream: LoadStreamFn<Model>
   ) {}
 
   async nodeStatus(): Promise<NodeStatusResponse> {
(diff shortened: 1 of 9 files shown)


packages/core/src/local-admin-api.ts
```diff
@@ -32,7 +38,36 @@ export class LocalAdminApi implements AdminApi {
 
   async startIndexingModelData(modelData: Array<ModelData>): Promise<void> {
     await this.indexApi.indexModels(modelData)
-    await this.syncApi.startModelSync(modelData.map((idx) => idx.streamID.toString()))
+
+    if (this.syncApi.enabled) {
```
stbrody, line 42: early return if syncApi is disabled will let you remove an indentation layer and make this a bit more readable I think

packages/core/src/local-admin-api.ts
```diff
@@ -32,7 +38,36 @@ export class LocalAdminApi implements AdminApi {
...
+      const models = await Promise.all(modelData.map(({ streamID }) => this.loadStream(streamID)))
+
+      const oldestModel = models.reduce((oldestModel, currentModel) => {
```
stbrody, line 45: can we make a helper method for figuring out what should be the start block we start the sync at?

packages/core/src/local-admin-api.ts
```diff
@@ -4,22 +4,28 @@ import {
...
+import { StreamID, CommitID } from '@ceramicnetwork/streamid'
 import { LocalIndexApi } from './indexing/local-index-api.js'
-import { SyncApi } from './sync/sync-api.js'
+import { ISyncApi, ModelSyncOptions } from './sync/interfaces.js'
+import { convertCidToEthHash } from '@ceramicnetwork/anchor-utils'
 
 type NodeStatusFn = () => Promise<NodeStatusResponse>
+type LoadStreamFn<T> = (streamId: StreamID | CommitID | string, opts?: LoadOpts) => Promise<T>
 
+export const NUMBER_OF_BLOCKS_BEFORE_TX_TO_START_SYNC = 360 // >24 hours
```
stbrody, line 18: are you sure this is more than 24 hours?  Ethereum blocks are produced once every 12 seconds so by my math 360 blocks is only 1.2 hours?

packages/core/src/local-admin-api.ts
```diff
@@ -32,7 +38,36 @@ export class LocalAdminApi implements AdminApi {
 
   async startIndexingModelData(modelData: Array<ModelData>): Promise<void> {
     await this.indexApi.indexModels(modelData)
-    await this.syncApi.startModelSync(modelData.map((idx) => idx.streamID.toString()))
+
+    if (this.syncApi.enabled) {
+      const models = await Promise.all(modelData.map(({ streamID }) => this.loadStream(streamID)))
```
stbrody, line 43: gah I hate that we need to do this circular dependency.  Really I think the way this should work is that when you try to start indexing a set of models, the `core` code should load those models and pass in the necessary information to the indexing code so that the indexing code has everything it needs and doesn't need to load streams.  The reason you can't do that here is that LocalAdminApi and RemoteAdminApi are constrained to having the same interface.  I really think we need to break that requirement and let the `core` code have different APIs than the http client.

CC @PaulLeCam since this is relevant to the refactoring work you are working on.

packages/core/src/sync/sync-api.ts
```diff
@@ -393,14 +394,35 @@ export class SyncApi implements ISyncApi {
       .subscribe()
   }
 
+  async _getStartBlock(syncOptions: ModelSyncOptions): Promise<number> {
+    let startBlock = Math.max(this.initialIndexingBlock, syncOptions.startBlock || 0)
```
stbrody, line 398: what's the difference between `this.initialIndexingBlock` and `syncOptions.startBlock`?

packages/core/src/local-admin-api.ts
```diff
@@ -32,7 +38,36 @@ export class LocalAdminApi implements AdminApi {
...
+      if (oldestModel.state.anchorProof) {
+        syncOptions = {
+          startBeforeTx: {
+            txHash: convertCidToEthHash(oldestModel.state.anchorProof.txHash),
+            numberOfBlocksBeforeTx: NUMBER_OF_BLOCKS_BEFORE_TX_TO_START_SYNC,
+          },
+        }
+      }
+
+      await this.syncApi.startModelSync(
```
stbrody, line 66: how do we handle Models that aren't anchored?  Also what if there's a mix of some Models that are anchored and some that are not?

stbrody:
Adding @PaulLeCam to the review, both because he's familiar with the HDS code and because he's looking to break up the circular dependancies between the indexing code and the rest of the `core` ceramic package.

packages/core/src/local-admin-api.ts
PaulLeCam: Yeah I agree that having the models injected directly would be better, eventually once we have a proper ComposeDB node we shouldn't need these APIs abstractions/duplication anyway so I'm not too concerned about it.

PaulLeCam (commented):
Looks like a great change overall, good idea to identify when to start each model sync!
I haven't noticed anything in particular beyond Spencer's comments, I think it's OK to add a circular dependency for now as it's unlikely it can be avoided anyways.

PaulLeCam (approved):

PR #2884, opened 2023-08-09
fix(core): send anchor status request once every poll interval

stephhuynh18:
## Description
Currently: 
If our polling interval is 1 minute. 
and our DEFAULT_FETCH_TIMEOUT is 3 minutes
Then we can get into the following scenario:

I begin polling for an anchor response

Cas is unavailable. I send first request,  I wait 3 minutes, it times out, delay for 1 minute, I try again with 2nd request, wait 3 minutes, times out, delay for 1 minute, I send 3rd request, wait 2 minutes, and i get a response woohoo! CAS has become available again.

Then we immediately send the 4th request, response after 10 seconds, immediately send 5th request, response after 10 seconds, immediately send 6th request, response after 10 seconds, immediately send 7th request, wait for 3 minutes (CAS is unresponsive again sad), etc

So why are we immediately sending the 4th, 5th, and 6th request? RXJS interval emits every POLLING_INTERVAL meaning every minute, regardless of what work is being done. By the time the 1st, 2nd, and 3rd request is through, 10 values have been emitted by RXJS interval (because 10 minutes), with the next one immediately being processed (hence why the 4th request gets processed right away without a delay of POLLING_INTERVAL). 

So we are polling once a minute… sometimes =)

A simple fix so I decided to just do it


ukstv (approved):
!YES!

packages/core/src/anchor/ethereum/__tests__/ethereum-anchor-service.retry-failure.test.ts
```diff
@@ -3,8 +3,8 @@ import { whenSubscriptionDone } from '../../../__tests__/when-subscription-done.
 import { generateFakeCarFile, FAKE_STREAM_ID, FAKE_TIP_CID } from './generateFakeCarFile.js'
 
 const MAX_FAILED_ATTEMPTS = 2
-const POLL_INTERVAL = 100 // ms
-const MAX_POLL_TIME = 500 // ms - to test if polling stops after this threshold
+const POLL_INTERVAL = 1000 // ms
```
stbrody, line 6: do we need to increase these values?  Isn't this just going to slow down the test?

packages/core/src/anchor/ethereum/__tests__/steph-temp.test.ts
```diff
@@ -0,0 +1,179 @@
+import { jest, test, expect } from '@jest/globals'
```
stbrody, line 1: can we either delete this file or clean it up and rename it?

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -171,7 +171,7 @@ export class EthereumAnchorService implements AnchorService {
     const requestUrl = [this.requestsApiEndpoint, tip.toString()].join('/')
     const cidStream = { cid: tip, streamId }
```
stbrody, line 172: note I refactored this function in https://github.com/ceramicnetwork/js-ceramic/commit/ee1370a6181fd75391c3e47a86eacb4f79caff07.

Probably will want to double-check how the merge goes to make sure git handles it right

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -184,16 +184,99 @@ export class EthereumAnchorService implements AnchorService {
...
+          throw new Error('Exceeded max anchor polling time limit')
+        } else {
+          return timer(this.pollInterval).pipe(concatMap(() => requestWithError))
+        }
+      }),
+      map((response) => this.parseResponse(cidStream, response))
+    )
+  }
+
+  newPollForAnchorResponse(task: () => Promise<any>): Observable<CASResponse> {
```
stbrody, line 200: is this used anywhere?  Is this actually meant to be merged?

stbrody (changes requested):
its a little hard for me to review what are the actual changes here because it seems like there are still changes in this PR that aren't intended to be merged?

packages/core/src/anchor/ethereum/__tests__/ethereum-anchor-service.retry-failure.test.ts
stephhuynh18: Changed back (increased for the demo)

packages/core/src/anchor/ethereum/__tests__/steph-temp.test.ts
stephhuynh18: deleted

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
stephhuynh18: nope just for the demo

stbrody (approved):
nice and clean, lgtm

PR #2887, opened 2023-08-09
Temporarily allow disabling anchor polling retry logic

stbrody:
Adds an env var `CERAMIC_DISABLE_POLLING_RETRIES` which if set to true reverts the node's behavior to be what it was before https://github.com/ceramicnetwork/js-ceramic/commit/07ff4e5b0558fd3ae6069c2fe2bdff5c6de69306#diff-7920408ff457031025a51c6c64241f297dc91febb1a810e724475548c154c781.

This is not a good fix and should be reverted before we do the next release, but if we can have gitcoin run with this flag enabled it could help us confirm whether or not the polling fix is what caused the recent node lockup


stbrody:
FYI @gvelez17 @bigs

PR #2888, opened 2023-08-09
feat: Add config option to disable data fetching in IPFS

stbrody:
This could potentially be combined with disabling the DHT on the node to possibly improve performance and reduce load on IPFS


packages/cli/src/index.ts
```diff
@@ -2,3 +2,4 @@ export * from './ceramic-cli-utils.js'
 export * from './ceramic-daemon.js'
 export * from './daemon-config.js'
 export * from './s3-store.js'
+export * from './__tests__/make-did.js'
```
stbrody, line 5: this feels possibly like bad practice?  @ukstv let me know if you have ideas for a better way I can share this testing utility function between the `cli` and `stream-tests` packages.  Or maybe this approach is fine?

stbrody:
FYI @oed @gvelez17 @smrz2001

packages/core/src/dispatcher.ts
```diff
@@ -163,7 +170,8 @@ export class Dispatcher {
 
   async getIpfsBlock(cid: CID): Promise<Uint8Array> {
     return await this._shutdownSignal.abortable((signal) => {
-      return this._ipfs.block.get(cid, { signal })
+      // @ts-ignore
+      return this._ipfs.block.get(cid, { signal, offline: !this._enableSync })
```
oed, line 174: Where did you find info about this `offline` option? Maybe we should include a reference in a comment for context?

stbrody: From a message from Adin in the filecoin slack.  Also realized we've used it before: https://github.com/ceramicnetwork/js-ceramic/blob/81afa77edeedb8e69f5e74c44dfccd85e8052f35/packages/core/src/dispatcher.ts#L274

packages/core/src/dispatcher.ts
```diff
@@ -31,7 +31,8 @@ import { CAR, CARFactory, CarBlock } from 'cartonne'
 import all from 'it-all'
 
 const IPFS_GET_RETRIES = 3
-const DEFAULT_IPFS_GET_TIMEOUT = 30000 // 30 seconds per retry, 3 retries = 90 seconds total timeout
+const DEFAULT_IPFS_GET_SYNC_TIMEOUT = 30000 // 30 seconds per retry, 3 retries = 90 seconds total timeout
+const DEFAULT_IPFS_GET_LOCAL_TIMEOUT = 3000 // 3 seconds to get data from local ipfs store
```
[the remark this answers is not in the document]
stbrody: yeah, I was being conservative because I know ipfs gets overloaded and slows down under load.  Really this shouldn't take more than a few milliseconds, but I just don't really trust IPFS to actually perform that well.
Happy to drop it further though.  I originally considered 1 second.  We could even try 500 millis though that starts to make me nervous without testing the ipfs behavior here more.  WDYT?

CC @smrz2001

PaulLeCam:
Is there anything validating incompatible config options please?
If I understand correctly setting this option will disable using the IPFS network, which I expect would make HDS unusable?
If so, could we make sure the node fails to start if both this option and HDS are enabled?

packages/core/src/dispatcher.ts
oed: ok!

oed:
Good point @PaulLeCam

packages/common/src/utils/test-utils.ts
```diff
@@ -73,16 +73,18 @@ export class TestUtils {
     predicate: (state: StreamState) => boolean,
     onFailure: (state: StreamState) => void
   ): Promise<void> {
-    if (predicate(stream.state)) return
-    const timeoutPromise = new Promise((resolve) => setTimeout(resolve, timeoutMs))
-    // We do not expect this promise to return anything, so set `defaultValue` to `undefined`
-    const completionPromise = lastValueFrom(stream.pipe(filter((state) => predicate(state))), {
```
stbrody, line 79: as part of testing I noticed that `waitForState` was always taking as long as `timeoutMs` to complete, even when it completes successfully.  It turns out that the `completionPromise` here never fires.  Ideally it would be nice to fix this to use the rxjs machinery again, but I couldn't figure out how to do that, so I turned it to a simple polling-based approach that I knew would work.  This managed to significantly speed up several tests.

stbrody (commented):
updated based on review comments.  Also changed to not bother processing pubsub messages we know will fail because it can't load CIDs from other nodes.

packages/common/src/utils/test-utils.ts
PaulLeCam: Looks like the timeout in `timeoutPromise` is never cleared so it's expected it keeps the process active.
You can probably get the behavior you want using RxJS's `timeout` operator instead, ex:
```ts
const completionPromise = lastValueFrom(stream.pipe(timeout(timeoutMs), filter((state) => predicate(state))), {
  defaultValue: undefined,
})
```

packages/core/src/dispatcher.ts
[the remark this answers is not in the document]
stbrody: changed to 1 second for now

packages/common/src/utils/test-utils.ts
stbrody: that doesn't fix it, it just makes it always throw a timeout error.  Fundamentally the core logic of `completionPromise` is just never returning.  It's not the timeoutPromise keeping things around, if I remove the timeout functionality entirely then this function just hangs forever.

PR #2902, opened 2023-08-14
Add env var to disable peer data sync

stbrody:
follow-up to https://github.com/ceramicnetwork/js-ceramic/pull/2888

### DIFF
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -106,6 +106,9 @@ export function makeCeramicConfig(opts: DaemonConfig): CeramicConfig {
     )
   } // else do nothing, no metrics are configured.
 
+  const disablePeerDataSync =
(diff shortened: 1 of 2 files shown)


packages/cli/src/ceramic-daemon.ts
```diff
@@ -106,6 +106,9 @@ export function makeCeramicConfig(opts: DaemonConfig): CeramicConfig {
     )
   } // else do nothing, no metrics are configured.
 
+  const disablePeerDataSync =
+    opts.ipfs.disablePeerDataSync || process.env.CERAMIC_DISABLE_IPFS_PEER_DATA_SYNC == 'true'
```
gvelez17, line 110: all the other env variable overrides are in ./packages/cli/src/ceramic-cli-utils.ts in CreateDaemon i think we should put this there

stbrody: good call, done

gvelez17 (approved):

PR #2903, opened 2023-08-15
chore: Faster LRU implementation, and more aggressive caching of IPFS access

ukstv:
Note https://github.com/ceramicnetwork/js-ceramic/pull/2903/files#diff-3dc086303a234c0aa9cb5b7b4b675827819e66906135448b777b221da01e0d60R55

`applyCommit` reuses IPLD records it got.


packages/core/src/ancillary/ipld-records-cache.ts
```diff
@@ -0,0 +1,81 @@
...
+   */
+  get(cid: CID): CacheEntry | undefined {
+    return this.cache.get(cid.toString())
+  }
+
+  /**
+   * IPFS allows you to get an IPLD record by query like `CID/some/path`. We do not have traversal mechanism,
+   * so we just store an entry by query. See also `getR`.
+   */
+  setR(key: string, entry: CacheEntry): void {
```
dbcfd, line 70: Not a fan of this name, maybe `setIPLDRecord`?

packages/core/src/ancillary/ipld-records-cache.ts
```diff
@@ -0,0 +1,81 @@
...
+    this.cache.set(key, entry)
+  }
+
+  /**
+   * IPFS allows you to get an IPLD record by query like `CID/some/path`. We do not have traversal mechanism,
+   * so here we do our best to return an entry by query. See also `setR`.
+   */
+  getR(key: string): CacheEntry | undefined {
```
dbcfd, line 78: Same as above

packages/core/src/state-management/state-cache.ts
```diff
@@ -1,23 +1,23 @@
...
    */
   constructor(limit: number, private readonly onEvicted?: (item: T) => void) {
-    this.volatile = new lru.LRUMap(limit)
-    this.volatile.shift = function () {
-      const entry = lru.LRUMap.prototype.shift.call(this)
-      onEvicted?.(entry[1])
-      return entry
+    this.volatile = new LRUCache(limit)
+    if (onEvicted) {
+      this.volatile.on('evicted', (_key, value) => {
```
dbcfd, line 18: Can we add eviction metrics?

packages/core/src/store/pin-store-factory.ts
```diff
@@ -39,6 +41,8 @@ export class PinStoreFactory {
     const ipfs = this.ipfs
     const pinning = PinningAggregation.build(ipfs, this.pinningEndpoints, this.pinningBackends)
     const retrieve = async (cid: CID): Promise<any> => {
+      const fromCache = this.ipldRecordsCache.get(cid)
+      if (fromCache) return fromCache.record
```
dbcfd, line 45: Can we add metrics here?

packages/core/src/dispatcher.ts
```diff
@@ -167,16 +167,21 @@ export class Dispatcher {
...
   }
 
   async getIpfsBlock(cid: CID): Promise<Uint8Array> {
-    return await this._shutdownSignal.abortable((signal) => {
-      // @ts-ignore
-      return this._ipfs.block.get(cid, { signal, offline: !this.enableSync })
-    })
+    const found = this.ipldCache.get(cid)
+    if (found) {
+      return found.block
```
dbcfd, line 178: Can we add metrics here (think this is the right spot)?

dbcfd (changes requested):

packages/core/src/store/pin-store-factory.ts
ukstv: Uhums! Added

packages/core/src/dispatcher.ts
ukstv: Added.

dbcfd (approved):

packages/core/src/dispatcher.ts
```diff
@@ -306,8 +327,12 @@ export class Dispatcher {
 
     // Lookup CID in cache before looking it up IPFS
     const resolutionPath = `${asCid}${path || ''}`
-    const cachedDagNode = this.dagNodeCache.get(resolutionPath)
-    if (cachedDagNode) return cloneDeep(cachedDagNode)
+    const fromCache = this.ipldCache.getUnresolved(resolutionPath)
+    if (fromCache) {
+      Metrics.count(IPFS_CACHE_HIT, 1)
+      return cloneDeep(fromCache.record)
```
stbrody, line 333: do you think we really need to clone this here?  How likely is it that any callers are modifying the record?

packages/core/src/store/ipld-records-cache.ts
```diff
@@ -0,0 +1,84 @@
...
+   */
+  get(cid: CID): CacheEntry | undefined {
+    return this.cache.get(cid.toString())
+  }
+
+  /**
+   * IPFS allows you to get an IPLD record by a query like `CID/some/path`. We do not have resolution mechanism that would translate `cid/path` query into a CID,
+   * so we just store an entry by query. See also `getUnresolved`.
+   */
+  setUnresolved(query: string, entry: CacheEntry): void {
```
stbrody, line 73: nitpick: maybe `setWithResolutionPath` is clearer?

packages/core/src/store/pin-store-factory.ts
```diff
@@ -39,6 +43,12 @@ export class PinStoreFactory {
     const ipfs = this.ipfs
     const pinning = PinningAggregation.build(ipfs, this.pinningEndpoints, this.pinningBackends)
     const retrieve = async (cid: CID): Promise<any> => {
+      const fromCache = this.ipldRecordsCache.get(cid)
+      if (fromCache) {
+        Metrics.count(IPFS_CACHE_HIT, 1)
+        return fromCache.record
+      }
+      Metrics.count(IPFS_CACHE_MISS, 1)
       const blob = await ipfs.dag.get(cid, { timeout: IPFS_GET_TIMEOUT })
```
stbrody, line 52: Should we add the entry into the cache here?

packages/core/src/store/pin-store-factory.ts
```diff
@@ -39,6 +43,12 @@ export class PinStoreFactory {
     const ipfs = this.ipfs
     const pinning = PinningAggregation.build(ipfs, this.pinningEndpoints, this.pinningBackends)
     const retrieve = async (cid: CID): Promise<any> => {
+      const fromCache = this.ipldRecordsCache.get(cid)
```
stbrody, line 46: Ideally at some point I'd like to pull out the parts of the Dispatcher that just manage direct puts and gets into IPFS (and add the caching, timeout, and retry logic) into a separate IpfsService API and then use that here directly so we can pick have unified access to the caching logic and also pick up the other extras that the Dispatcher adds over raw IPFS access.  But that is certainly out of scope for this PR.

stbrody (approved):
nice

packages/core/src/store/ipld-records-cache.ts
ukstv: Oh! Finally a good name for this method! Thank you!

packages/core/src/store/pin-store-factory.ts
ukstv: Someday, yeah. That’s on my list too.

packages/core/src/dispatcher.ts
ukstv: Haha, I had the same question to myself. I thought you would be against removing it :)

stbrody: Ideally the type system would let us specify the return value as immutable, but alas

dbcfd: Just sprinkle some rust on it 😆

ukstv:
Nope, cloneDeep is here to stay. Absence of cloneDeep breaks the tests. IDK what's wrong, but it's another pain point better addressed separately.

PR #2907, opened 2023-08-16
feat: Add option to wait until anchor requests are durably created on the CAS

stbrody:
For now the flag is still false by default until we've tested it further, but the plan is to make it true by default in a later release.

This is also still missing the behavior where if the anchor fails the stream state is left unmodified as if the write never happened.  That requires some refactoring of the state manager and repository and will come in a future PR (test coverage will also come at that time as that will change behavior quite a bit).  But I wanted to get this in place first so we can test the latency implications.

### DIFF
--- a/packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
+++ b/packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
@@ -9,7 +9,7 @@ import {
   FetchRequest,
 } from '@ceramicnetwork/common'
 import { StreamID } from '@ceramicnetwork/streamid'
-import { Observable, concat, timer, of, defer, expand, interval } from 'rxjs'
+import { Observable, concat, timer, of, defer, expand, interval, lastValueFrom } from 'rxjs'
 import { concatMap, catchError, map, retry } from 'rxjs/operators'
 import { CAR } from 'cartonne'
 import { AnchorRequestCarFileReader } from '../anchor-request-car-file-reader.js'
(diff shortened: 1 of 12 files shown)


packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -87,26 +87,45 @@ export class EthereumAnchorService implements AnchorService {
...
+            cid: carFileReader.tip,
+            message: error.message,
+          })
+        )
+      )
+    }
+
+    if (waitForConfirmation) {
+      await lastValueFrom(requestCreated$)
+      return addErrorHandling(anchorCompleted$)
```
ukstv, line 125: Nitpick suggestion:

```
const streamId = carFileReader.streamId
const tip = carFileReader.tip
const handleErrors = catchError((error) => 
  of<CASResponse>({
            id: '',
            status: AnchorRequestStatusName.FAILED,
            streamId: streamId,
            cid: tip,
            message: error.message,
          })
)

return anchorCompleted$.pipe(handleErrors)
```

and few lines below:
```
      return concat(requestCreated$, anchorCompleted$).pipe(handleErrors)
```

ukstv (approved):
OK, mod nitpick wrt closure size.

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
stbrody: I got a compiler error trying to bind catchError to a variable, but I did something similar with the variable just holding the inner error handler and then adding the pipe to catcherror in both places

PR #2909, opened 2023-08-18
chore(core): remove Repository/StateManager cyclic dependencies - CDB-2713

PaulLeCam:
This PR introduces a new `RepositoryInternals` class that combines extracted methods from `Repository` and `StateManager` that depend on one-another.
With these changes, the `Repository` creates the `RepositoryInternals` instance and injects it to the `StateManager` instance, leading to the following separation:
- `Repository` can call `RepositoryInternals` and `StateManager` methods
- `StateManager` can call `RepositoryInternals` methods but has no access to `Repository`
- `RepositoryInternals` has no access to `Repository` and `StateManager`

This is just a first step to break this specific cycle, there are other problems such as the need for the `setDeps()` method on `Repository` but I don't think that can be fixed without breaking API changes, and I'd like to avoid long-running branches with large changes.
Eventually I think the `RepositoryInternals` should be refactored, maybe split into specific concerns, but before improving the logic there I think other cycles should be broken up to make sure the architecture changes fully work.


ukstv (approved):
OK, and a question/suggestion, not for this PR.

On CAS we use https://github.com/nicojs/typed-inject for DI. Maybe it makes sense to start using it here as well?

PaulLeCam:
@ukstv I definitely think we need a better system for dependencies between modules, and a DI framework like this seems interesting, but I'm wondering how much it will actually prevent risks of creating cycles?
It seems to me the main problems we have are caused by bad injections in a few specific places (ex `context.api = this` in Ceramic instance, bound methods on instance injected to sub-modules, etc.), any idea how we could catch/prevent these please?

stephhuynh18 (approved):

ukstv:
> @ukstv I definitely think we need a better system for dependencies between modules, and a DI framework like this seems interesting, but I'm wondering how much it will actually prevent risks of creating cycles?
> It seems to me the main problems we have are caused by bad injections in a few specific places (ex `context.api = this` in Ceramic instance, bound methods on instance injected to sub-modules, etc.), any idea how we could catch/prevent these please?

Hehe, my suggestion would do nothing to this kind of references, I admit.

PaulLeCam:
@stbrody looks like your changes in https://github.com/ceramicnetwork/js-ceramic/commit/a36ebf72b1701836a3360fb30ba935a265d0aeda created conflicts, do you mind updating this branch so we can merge the PR please?
I'm off next week so I'd like to merge it ASAP to avoid having too much to catch-up from develop, it's tricky to keep these changes up-to-date as the logic moved to other modules.

stbrody (approved):
looks great.  I'll work on merging my changes now

PR #2911, opened 2023-08-18
chore: Leverage generated version

ukstv:
Would not move a needle. Just a peculiar not leveraging of the existing stuff.


stbrody (approved):
seems fine

PR #2912, opened 2023-08-18
fix: Slow down rate of restoring polling for streams with pending anchors

stbrody:


packages/core/src/state-management/anchor-resuming-service.ts
```diff
@@ -4,8 +4,9 @@ import type { Repository } from './repository.js'
 import { LogStyle, type DiagnosticsLogger } from '@ceramicnetwork/common'
 import { TaskQueue } from '../ancillary/task-queue.js'
 
-const RESUME_QUEUE_CONCURRENCY = 30
+const RESUME_QUEUE_CONCURRENCY = 5
 const RESUME_BATCH_SIZE = RESUME_QUEUE_CONCURRENCY * 5
+const RATE_LIMIT_DELAY = 100
```
stbrody, line 9: with .1 sec delay between streams we can resume 10 streams a second.  That's 600 streams a minute or 36000 streams an hour.  I think that should be plenty fast enough, at least for now until we can optimize how the polling works.  No one currently is going to have 36k streams with pending anchors at once, and taking an hour to resume them is actually fine since CACAOs expire on an order of 24 hours and we strive to anchor any stream within 12 hours.  A 1 hour delay on detecting an anchor commit, while not ideal as it can lead to data forks, is still unlikely to cause data corruption and so is probably acceptable for now.

We should definitely consider increasing the concurrency and removing the rate limit if we get to a world where this only a single background task that polls for all pending anchors at once.

packages/core/src/state-management/anchor-resuming-service.ts
```diff
@@ -33,24 +37,37 @@ export class AnchorResumingService {
...
       batch = await repository.anchorRequestStore.list(RESUME_BATCH_SIZE, gt)
-      batch.forEach((listResult) => {
+      for (const listResult of batch) {
         this.resumeQ.add(async () => {
           await repository.fromMemoryOrStore(listResult.key)
           this.logger.verbose(`Resumed running state for stream id: ${listResult.key.toString()}`)
+          numRestoredStreams++
         })
-      })
+        await this.delay(RATE_LIMIT_DELAY)
```
dbcfd, line 62: We probably don't need to hardcode this to 100, but rather jitter it between like 10 and 50ms.

packages/core/src/state-management/anchor-resuming-service.ts
```diff
@@ -33,24 +37,37 @@ export class AnchorResumingService {
...
+    this.logger.imp(`Resuming polling for streams with pending anchors`)
+
+    let numRestoredStreams = 0
     let gt: StreamID | undefined = undefined
     let batch = new Array<AnchorRequestStoreListResult>()
     do {
       batch = await repository.anchorRequestStore.list(RESUME_BATCH_SIZE, gt)
-      batch.forEach((listResult) => {
+      for (const listResult of batch) {
         this.resumeQ.add(async () => {
```
dbcfd, line 57: We should jitter before adding to prevent thundering herd as well.

dbcfd (changes requested):

packages/core/src/state-management/anchor-resuming-service.ts
stbrody: done

packages/core/src/state-management/anchor-resuming-service.ts
stbrody: we sleep after adding each entry to the queue so no need to sleep before, that would just be redundant.

dbcfd: Ah, then we want the jitter at https://github.com/ceramicnetwork/js-ceramic/blob/c6906048236e9b101149bc189a88ecbe27fb9234/packages/core/src/state-management/anchor-resuming-service.ts#L60 so we interrupt before the next one kicks off.

stbrody: sorry I still don't understand how that isn't redundant.  We already sleep between each stream

dbcfd: we add all the requests for getting the streams to the queue, sleeping between that. We don't necessarily want to immediately readd to the queue when a request finishes, but rather give the event loop time to do other work.

dbcfd (approved):

ukstv (approved):

PR #2916, opened 2023-08-21
feat(core, stream-tests): negated ors/ands and multiple keys/values with where filters

stephhuynh18:
Added a bunch of tests
not(a & b)  and not a || b) fixed 
multiple keys/values with where filters

changed the approach if you looked at this branch before


bigs:
Fantastic, thank you! Let's plan on merging this mid-day tomorrow :)

dbcfd (approved):

PR #2922, opened 2023-08-25
chore: Extract anchor validation out of conflict_resolution.ts

stbrody:
Builds on https://github.com/ceramicnetwork/js-ceramic/pull/2921 (this is `#2` in the sequence)


packages/core/src/conflict-resolution.ts
```diff
@@ -274,7 +197,7 @@ export function commitAtTime(stateHolder: StreamStateHolder, timestamp: number):
 export class ConflictResolution {
   constructor(
     public logger: DiagnosticsLogger,
-    public anchorValidator: AnchorValidator,
+    public anchorTimestampExtractor: AnchorTimestampExtractor,
```
PaulLeCam, line 200: Could we make it private please? It's easier to avoid circular dependencies if we have clear module boundaries and modules can't access instances attached to other modules.

stbrody: done

PaulLeCam (approved):

ukstv (approved):

PR #2926, opened 2023-08-28
chore: Clarify directory names

stbrody:
I was worried that `sync` and `loading` are too similar and would be confusing.  So renamed `sync` to `history-sync` and `loading` to `stream-loading`.

This builds on https://github.com/ceramicnetwork/js-ceramic/pull/2922 (This is `#3` in the sequence)

CC @stephhuynh18


PaulLeCam (approved):

ukstv (changes requested):
Dash or underscore? o_O I guess, dash is more universal in our codebase. Hence, `log_syncer.ts` should become `log-synced.ts` I believe.

PR #2932, opened 2023-08-30
feat(core): Implement LogSyncer

stbrody:


PaulLeCam (approved):
LGTM but I'm wondering if we might need to add logic to handle cancellations, if nothing else the node shutting down? Or is it handled by the dispatcher directly?

ukstv (approved):

stbrody:
> LGTM but I'm wondering if we might need to add logic to handle cancellations, if nothing else the node shutting down? Or is it handled by the dispatcher directly?

Cancellation support is a good idea.  I filed https://linear.app/3boxlabs/issue/CDB-2734/add-cancellation-support-to-all-streamloader-components to circle back and add that to all the relevant new components

PR #2942, opened 2023-09-11
chore: Almost never-ending stream of stored anchors

ukstv:
One-line change to make it never-ending:

```typescript
gt = undefined // instead of return
```


packages/core/src/state-management/anchor-resuming-service.ts
```diff
@@ -81,7 +76,6 @@ export class AnchorResumingService {
   async close(): Promise<void> {
     this.logger.debug('Closing AnchorResumingService')
     this.#shouldBeClosed = true
-    this.resumeQ.clear()
```
stbrody, line 84: why remove this?

stbrody (approved):
lgtm

stephhuynh18 (approved):

packages/core/src/state-management/anchor-resuming-service.ts
ukstv: oops

PR #2945, opened 2023-09-11
chore: drop signature verification for in-memory-anchor-service

ukstv:
We do not use it anywhere, and it does not conform to what real CAS does.


dbcfd (approved):

stbrody (approved):
lgtm

PR #2946, opened 2023-09-11
feat: Implement StateManipulator with cleaner, more efficient conflict resolution

stbrody:


packages/core/src/stream-loading/__tests__/stream-loader.test.ts
```diff
@@ -0,0 +1,129 @@
+import { jest } from '@jest/globals'
```
stbrody, line 1: this test file is not intended to be exhaustive.  I just happened to write a simple test at this layer while trying to figure out how to write a test at the lower StateManipulator layer.  Figured it out, but figured I'd leave this test in because why not.  I expect to more fully flesh out the tests for the StreamLoader in a subsequent PR though

packages/core/src/stream-loading/__tests__/state-manipulator.test.ts
```diff
@@ -0,0 +1,521 @@
...
+      await expect(
+        stateManipulator.applyLogToState(currentState, makeAppliable(conflictingCommits), {
+          throwOnInvalidCommit: true,
+          throwIfStale: false,
+          throwOnConflict: true,
+        })
+      ).rejects.toThrow(/rejected by conflict resolution/)
+    })
+
+    describe('pickLogToAccept', () => {
```
stbrody, line 370: these tests are ported almost directly from `conflict-resolution.test.ts`.

packages/core/src/stream-loading/state-manipulator.ts
```diff
@@ -1,27 +1,266 @@
...
+
+    // If we got this far, that means that we don't have sufficient information to make a good
+    // decision about which log to choose.  The most common way this can happen is that neither log
+    // is anchored, although it can also happen if both are anchored but in the same blockNumber or
+    // blockTimestamp. At this point, the decision of which log to take is arbitrary, but we want it
+    // to still be deterministic. Therefore, we take the log whose last entry has the lowest CID.
+    return log1[log1.length - 1].cid.bytes < log2[log2.length - 1].cid.bytes ? log1 : log2
+  }
+
+  async _applyLogToState_noCacaoVerification(
```
PaulLeCam, line 182: Not a big deal but I don't really like mixing casing conventions like this, couldn't this method be named `_applyLogToStateWithoutCacaoVerification` for example please?

packages/core/src/stream-loading/state-manipulator.ts
```diff
@@ -1,27 +1,266 @@
...
+ * @param throwOnInvalidCommit - if true, throws if there is an error applying a commit, otherwise
+ *   returns the state that was built so far.
+ * @param throwIfStale - if true, throws if the log to apply does not build directly on top of
+ *   the existing state.
+ * @param throwOnConflict - if true, throws if the log to apply is rejected by conflict resolution.
+ */
+interface ApplyLogToStateOpts {
+  throwOnInvalidCommit: boolean
+  throwIfStale: boolean
+  throwOnConflict: boolean
```
PaulLeCam, line 34: Do these default to `true` or `false` when applying the logic please?
If they default to `true` and must be explicitly set to `false`, could we inverse the logic please? Ex:
```ts
allowInvalidCommit: boolean
allowStale: boolean
allowConflict: boolean
```
With these flags having to be set explicitly to `true` not to throw? I think that would make it clearer the default behavior is to be strict, unless explicitly set?

PaulLeCam (approved):
LGTM just noticed some small things that could be changed but aren't blockers

packages/core/src/stream-loading/state-manipulator.ts
stbrody: they don't actually default to anything, it is required for the caller to pass all 3 in explicitly.  There isn't really a clear case for which behavior should be the "default" - they are all set differently in different contexts.  E.g. `throwOnInvalidCommit` is true for user-initiated writes coming in via the http api, but false for writes that come in via pubsub.

PR #2948, opened 2023-09-12
chore(indexing): extract indexing package

PaulLeCam:
Extracts the indexing and history sync logic from `core` into the new `indexing` package.

To avoid circular dependencies, the `indexing` package doesn't depend on `core` but uses a new `CeramicCoreApi` type, that the Ceramic instance implements.
Some tests from history sync depending on the Ceramic instance are moved to the `stream-tests` package to avoid circular dependencies.

Overall there should be no change to the logic, mostly moving modules around and addressing TS and ESLint errors.


ukstv (approved):

packages/indexing/src/local-index-api.ts
```diff
@@ -6,22 +6,28 @@ import type {
...
+import { type CommitID, StreamID } from '@ceramicnetwork/streamid'
 import { IndexingConfig } from './build-indexing.js'
-import { makeIndexApi } from '../initialization/make-index-api.js'
+import { makeIndexApi } from './make-index-api.js'
 import { Networks } from '@ceramicnetwork/common'
 import { Model } from '@ceramicnetwork/stream-model'
-import { ISyncQueryApi } from '../history-sync/interfaces.js'
+import { ISyncQueryApi } from './history-sync/interfaces.js'
+
+export type CeramicCoreApi = {
```
stbrody, line 21: a comment on this type would be helpful to make clear why we need this instead of being able to just use `CeramicApi`

stbrody (approved):
lgtm

packages/indexing/src/local-index-api.ts
PaulLeCam: Yes good call, I'll add this when I move this type, probably to the common package

PR #2952, opened 2023-09-13
feat: Add StreamUpdater to handle applying incoming commits

stbrody:
Only the `StreamLoader` and `StreamUpdater` should be used by the `Repository`, all the components that make them up should be considered private implementation details.  The public methods of `StreamLoader` and `StreamUpdater` are the API presented to the `Repository` for handling all Stream operations.  In turn, nothing outside of the Repository should access the `StreamLoader` or `StreamUpdater` - the `Repository` should be interface to all Stream operations performed by any other part of the codebase.  The Repository will be responsible for handling concurrency control (the loading and execution queues) as well as pinning, indexing, publishing tips to pubsub, etc.


stbrody:
accidentally created against wrong ticket

PR #2954, opened 2023-09-13
chore: Make StateManager private within Repository

stbrody:


packages/core/src/__tests__/state-manager.test.ts
```diff
@@ -85,7 +85,7 @@ describe('anchor', () => {
       const stream1 = await TileDocument.create(ceramic, INITIAL_CONTENT, null, { anchor: false })
```
stbrody, line 85: its a little weird to still call this test `state-manager.test.ts` when it never actually accesses the StateManager directly anymore.  Conceptually its just an extension of `repository.test.ts` now.  I could combine the two of them into one file, but that would be a BIG file, these are already two of our largest test files.  I'm open to any suggestions about what to do with this test!

PaulLeCam: Could the `StateManager` be tested on its own, without involving the `Repository`? Maybe by mocking the necessary dependencies?
If not, I think it's fine to keep as-is.

PaulLeCam (approved):

PR #2955, opened 2023-09-14
chore: Use current polling to populate events, pt1

ukstv:


packages/core/src/__tests__/ceramic-anchor.test.ts
```diff
@@ -228,73 +228,6 @@ describe('Ceramic anchoring', () => {
     await ceramic2.close()
   })
 
-  it('test the same doc anchored twice (different Ceramic instances), first one wins)', async () => {
```
ukstv, line 231: It does not make sense anymore.

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -168,34 +218,22 @@ export class EthereumAnchorService implements AnchorService {
       sendRequest$ = sendRequest$.pipe(
         retry({
           delay: (error) => {
-            this._logger.err(
-              new Error(
-                `Error connecting to CAS while attempting to anchor ${carFileReader.streamId} at commit ${carFileReader.tip}: ${error.message}`
-              )
+            this.#logger.err(
```
dbcfd, line 221: Since we're retrying, this should probably be a warning saying it failed and we are retrying. 

Only log an error if we give up on trying to send it. 

We also might want to log and error when we shutdown for these pending requests.

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -204,29 +242,20 @@ export class EthereumAnchorService implements AnchorService {
...
+    const requestWithError = defer(() => this.#sendRequest(requestUrl)).pipe(
       retry({
         delay: (error) => {
-          this._logger.err(
-            new Error(
-              `Error connecting to CAS while polling for anchor ${streamId} at commit ${tip}: ${error.message}`
-            )
-          )
-          return timer(this.pollInterval)
+          this.#logger.err(new CasConnectionError(streamId, tip, error.message))
```
dbcfd, line 257: Similar comment here.

dbcfd (changes requested):

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
ukstv: okay

packages/core/src/__tests__/ceramic-anchor.test.ts
stbrody: How so? It seems like the idea of what the test is trying to test is still relevant no?  That the first anchored update wins?

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -123,9 +180,12 @@ export class EthereumAnchorService implements AnchorService {
 
     if (waitForConfirmation) {
       await lastValueFrom(requestCreated$)
-      return anchorCompleted$.pipe(catchError(errHandler))
+      return anchorCompleted$.pipe(catchError(errHandler), this._updateEvents())
     } else {
-      return concat(requestCreated$, anchorCompleted$).pipe(catchError(errHandler))
+      return concat(requestCreated$, anchorCompleted$).pipe(
+        catchError(errHandler),
+        this._updateEvents()
```
stbrody, line 187: given that `_updateEvents()` is also added to the pipeline in `_pollForAnchorResponse()` - does it still need to be added here as well?

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -204,29 +242,20 @@ export class EthereumAnchorService implements AnchorService {
...
-      return this._pollForAnchorResponseLegacy(streamId, tip)
-    } else {
-      return this._pollForAnchorResponse(streamId, tip)
-    }
+  pollForAnchorResponse(streamId: StreamID, tip: CID): Observable<AnchorEvent> {
+    return this._pollForAnchorResponse(streamId, tip)
   }
 
-  private _pollForAnchorResponse(streamId: StreamID, tip: CID): Observable<CASResponse> {
+  private _pollForAnchorResponse(streamId: StreamID, tip: CID): Observable<AnchorEvent> {
```
stbrody, line 249: nitpick: no need for this helper anymore, can fold it up into `pollforAnchorResponse`

packages/core/src/anchor/memory/in-memory-anchor-service.ts
```diff
@@ -1,43 +1,32 @@
 import { CID } from 'multiformats/cid'
 import { Observable, Subject, concat, of } from 'rxjs'
 import { filter } from 'rxjs/operators'
-import {
+import { TestUtils } from '@ceramicnetwork/common'
```
stbrody, line 4: this is an awesome cleanup.  Nice to see the InMemoryAnchorService finally reflect the way the CAS actually works these days again

stbrody (commented):
this is looking really great.  My only real concern is the loss of the test case for conflict resolution between nodes based on anchoring time.

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
ukstv: makes sense, ty

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
ukstv: yup

packages/core/src/__tests__/ceramic-anchor.test.ts
ukstv: "the first anchored update wins" - that would be a test for conflict resolution, not for in-memory anchoring

dbcfd (approved):

packages/core/src/__tests__/ceramic-anchor.test.ts
stbrody: > And we have such tests

Do we?  I know we have some targeted unit tests of the conflict resolution code itself, but do we have any tests of that logic end-to-end that operate with a full Ceramic node?

ukstv: The test here has nothing to do with Ceramic node behaviour. Two nodes connected to a single CAS instance. They send two anchor requests for the same stream. The assertion is for CAS to decide that one of them wins. That is, that InMemoryAnchorService conducts correct conflict resolution. Since removal of a Ceramic node from CAS side, that behaviour does not make sense.

stephhuynh18 (approved):

packages/core/src/__tests__/ceramic-anchor.test.ts
stbrody: OOOOH.  Yep, I get it now.  I thought the test was anchoring both branches and testing how Ceramic resolved the conflict.  I agree that testing the behavior of the InMemoryAnchorService here is silly

stbrody (approved):
lgtm

PR #2956, opened 2023-09-14
chore: Rewrite tests of how sync waits for responses

stbrody:
The implementation of how streams get synced during loadStream is about to change a lot, so these tests as-written in `state-manager.test.ts` won't work anymore.  They're testing functionality we want to preserve though, so I want to make sure we don't lose that test coverage, so I re-wrote them to happen at a higher level.

### DIFF
--- a/packages/core/src/__tests__/ceramic-query-response.test.ts
+++ b/packages/core/src/__tests__/ceramic-query-response.test.ts
@@ -1,0 +1,190 @@
+import { jest, expect, describe, test, afterEach, beforeEach } from '@jest/globals'
+import { type IpfsApi, TestUtils } from '@ceramicnetwork/common'
+import { TileDocument } from '@ceramicnetwork/stream-tile'
+import { StreamID } from '@ceramicnetwork/streamid'
+import { createIPFS } from '@ceramicnetwork/ipfs-daemon'
+import { createCeramic } from './create-ceramic.js'
+import { MsgType, PubsubMessage, QueryMessage, ResponseMessage } from '../pubsub/pubsub-message.js'
+import { MAX_RESPONSE_INTERVAL } from '../pubsub/message-bus.js'
+import { CID } from 'multiformats/cid'
+import { asIpfsMessage } from '../pubsub/__tests__/as-ipfs-message.js'
+import { Ceramic } from '../ceramic.js'
+
+const LONG_SYNC_TIME = 60 * 1000
+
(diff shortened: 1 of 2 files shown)


packages/core/src/__tests__/ceramic-query-response.test.ts
```diff
@@ -0,0 +1,123 @@
...
+  test('sync returns after only response', async () => {
+    const genesisCommit = await TileDocument.makeGenesis(ceramic, { foo: 'bar' }, null)
+    const genesisCID = await ceramic.dispatcher.storeCommit(genesisCommit)
+    const streamID = new StreamID(0, genesisCID)
+
+    const timeBeforeResponding = 100
+    await otherIpfs.pubsub.subscribe(ceramic.pubsubTopic, async (rawMessage: SignedMessage) => {
+      const message = deserialize(rawMessage)
+      if (message.typ == MsgType.QUERY && message.stream.equals(streamID)) {
+        await TestUtils.delay(timeBeforeResponding)
```
stbrody, line 42: I'd love to be able to write this test file without using a second real IPFS node but instead by injecting pubsub messages into the IncomingChannel at the time rate that I want.  I think it would make the tests lighter-weight and more reliable, instead of depending on a real pubsub connection.  But I couldn't figure out how to do that.  Please LMK if you think that is relatively straightforward to do.

ukstv: You could attach a jest spy to ceramicIpfs instance's `subscribe`, that calls subscription handler at any rate you want after it gets a query. The latter part again could be achieved by a jest spy.

ukstv (approved):

packages/core/src/__tests__/ceramic-query-response.test.ts
stbrody: I tried to get this to work but failed :(
You can see my attempt here if you like: https://github.com/ceramicnetwork/js-ceramic/blob/tipResponseTestRewriteWithJestMocks/packages/core/src/__tests__/ceramic-query-response.test.ts#L114.  If you have any suggestions on how to fix it I'm happy to try again, otherwise I'll probably just merge this PR as-is tomorrow.

stbrody: Ooh good call, thanks!  Got the tests all updated and switched over to injecting pubsub responses directly instead of actually using a real ipfs pubsub connection.  PTAL

ukstv (approved):

PR #2961, opened 2023-09-18
chore: Clean up anchoring initialization

ukstv:
- No anchorService in context,
- validator is a part of AnchorService,
- AnchorService creation through opaque functions.

### DIFF
--- a/packages/core/src/ceramic.ts
+++ b/packages/core/src/ceramic.ts
@@ -16,7 +16,6 @@ import {
   MultiQuery,
   PinningBackendStatic,
   LoggerProvider,
-  Networks,
   UpdateOpts,
   SyncOptions,
   AnchorValidator,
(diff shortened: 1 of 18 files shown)


stephhuynh18 (approved):
🎈

packages/core/src/ceramic.ts
```diff
@@ -240,12 +226,11 @@ export class Ceramic implements CeramicApi {
 
     this.context = {
       api: this,
-      anchorService: modules.anchorService,
       ipfs: modules.ipfs,
       loggerProvider: modules.loggerProvider,
     }
     if (!this._gateway) {
-      this.context.anchorService.ceramic = this
+      this.anchorService.ceramic = this
```
stbrody, line 233: 😢

stbrody (approved):
nice

packages/core/src/ceramic.ts
ukstv: A necessary evil now, that we'll change someday. In prod it is a no-op though, so not as harmful as it looks.

PR #2963, opened 2023-09-19
feat: Use StreamLoader for loading streams

stbrody:
This also addresses https://linear.app/3boxlabs/issue/CDB-2732/dont-throw-error-when-loading-a-stream-if-the-tip-isnt-accessible-via

### DIFF
--- a/packages/core/src/stream-loading/state-manipulator.ts
+++ b/packages/core/src/stream-loading/state-manipulator.ts
@@ -13,6 +13,7 @@ import {
 import { CID } from 'multiformats/cid'
 import { HandlersMap } from '../handlers-map.js'
 import { LogSyncer } from './log-syncer.js'
+import cloneDeep from 'lodash.clonedeep'
 
 /**
  * @param throwOnInvalidCommit - if true, throws if there is an error applying a commit, otherwise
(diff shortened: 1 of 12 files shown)


packages/core/src/stream-loading/state-manipulator.ts
```diff
@@ -48,12 +49,24 @@ export class StateManipulator {
...
+   * StateManipulator and so the input state that was given to the StateManipulator will not be
+   * modified.
+   */
   async _applyLog<T extends Stream>(
     handler: StreamHandler<T>,
     state: StreamState | null,
     log: AppliableStreamLog,
-    throwOnInvalidCommit: boolean
+    throwOnInvalidCommit: boolean,
+    cloneInput: boolean
```
PaulLeCam, line 64: That seems like a potential source of issues if relying of the caller to know if an object is safe to mutate or not. Could we instead make objects that get passed around immutable so the functions using them can't accidentally mutate them? I think that would also be more efficient than always cloning "just in case".

PaulLeCam (approved):
LGTM, just concerned about having to clone state based on caller knowledge but I guess it's probably not the only place we have logic like this?

packages/core/src/stream-loading/state-manipulator.ts
stbrody: Is there actually a way to make objects immutable in js?

PaulLeCam: There are full-feature libraries such as https://immutable-js.com/ that provide APIs to manipulate immutable structures, but if we don't want to leverage a dedicated library, at least we could use [`Object.freeze()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze) to leverage the guaranties that can be provided by the JS engine.
Considering our usage of cloning objects, I think we could instead freeze them on creation and use TS' [`Readonly` type](https://www.typescriptlang.org/docs/handbook/utility-types.html#readonlytype) in arguments when passing these objects along to ensure the functions consuming them don't try to mutate them.

stbrody: Ooh okay, I kind of like using the Readonly type, though that will mean changing the signature of all the StreamTypeHandlers.  I'm going to leave this PR as-is for now, but I have https://linear.app/3boxlabs/issue/CDB-2767/figure-out-best-way-to-ensure-streamtype-handlers-dont-mutate-input to follow-up on this.

PR #2965, opened 2023-09-20
feat: rust ceramic tests

dbcfd:


stbrody:
I'm going to hold off on reviewing this until tests are working, given that this PR is specifically about the tests.  When things are stable LMK and I'll take a look @dbcfd

packages/ipfs-daemon/package.json
```diff
@@ -44,6 +44,7 @@
   "dependencies": {
     "@ceramicnetwork/common": "^2.36.0-rc.0",
     "@ceramicnetwork/ipfs-topology": "^2.30.0-rc.0",
+    "dockerode": "^3.3.5",
```
smrz2001, line 47: Now that we have the ceramic-one binary included, we don't need the Docker approach anymore, right? It can all be switched to launching the ceramic-one process?

dbcfd: The docker approach provides a way to do an isolated embedded ceramic. If we don't think we need that functionality, can remove it.

packages/ipfs-daemon/src/rust-ipfs.ts
```diff
@@ -0,0 +1,275 @@
...
+  port: number | undefined
+}
+
+export type RustIpfsBinaryOptions = {
+  type: 'binary'
+  path: string | undefined
+  port: number | undefined
+}
+
+export type RustIpfsOptions = RustIpfsEmbeddedOptions | RustIpfsRemoteOptions | RustIpfsBinaryOptions
```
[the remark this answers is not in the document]
stbrody: it looks like CI will use the Binary version, is that right?

stbrody (changes requested):
most comments are small, my biggest concern is the new need to sleep after subscribing to pubsub and why that's necessary for rust-ceramic but wasn't for kubo

packages/core/src/state-management/__tests__/repository.test.ts
```diff
@@ -578,18 +578,26 @@ describe('validation', () => {
     await TestUtils.anchorUpdate(ceramic, schema)
     // Create invalid stream
     const ipfs2 = await createIPFS()
+    await swarmConnect(ipfs, ipfs2)
+
     const permissiveCeramic = await createCeramic(ipfs2)
+
+    // Wait for peers to discover they are both subscribed to same topic
```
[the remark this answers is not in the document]
nathanielc: You are right for this test, I must have gotten it mixed up, it doesn't use pubsub and passes fine without these changes. Reverted.

packages/ipfs-daemon/src/rust-ipfs.ts
```diff
@@ -0,0 +1,275 @@
...
+import { create as createIpfsClient } from 'ipfs-http-client'
+import { RunningIpfs } from './create-ipfs.js'
+import * as stream from 'stream'
+import { DiagnosticsLogger } from '@ceramicnetwork/common'
+import { spawn, ChildProcess } from 'child_process'
+import fs from 'node:fs'
+import path from 'node:path'
+
+
+export type RustIpfsEmbeddedOptions = {
```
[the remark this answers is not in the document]
nathanielc: There are only two know with clear names `remote` and `binary`.

packages/ipfs-daemon/src/rust-ipfs.ts
```diff
@@ -0,0 +1,275 @@
...
+    return this._api
+  }
+
+  isOnline(): boolean {
+    return this._api.isOnline()
+  }
+
+  async shutdown(): Promise<void> {
+    //no additional shutdown required
+    //calling api.stop() would attempt to shutdown the remote server
```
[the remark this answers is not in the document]
nathanielc: No, we didn't start the node so we should not shut it down. I updated the comment to be more clear.

packages/ipfs-daemon/src/rust-ipfs.ts
```diff
@@ -0,0 +1,275 @@
...
+      await this.container.stop()
+    } catch (e) {
+      if (logger) {
+        logger.err(`Failed to shutdown embedded Rust IPFS: ${e}`)
+      }
+    }
+  }
+}
+
+async function embedded(image?: string, port?: number): Promise<RunningIpfs> {
```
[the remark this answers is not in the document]
nathanielc: This option has been removed as it was made redundant by the `binary` option.

packages/ipfs-daemon/src/rust-ipfs.ts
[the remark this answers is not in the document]
nathanielc: Yes, CI uses the binary version and the remote version is handy for debugging because you can spawn the process yourself.

nathanielc:
@stbrody Re the pubsub delay. There is naturally a race between node A subscribing and node B publishing a message to a topic. We could poll the nodes to determine they are connected via gossipsub instead of just using a delay, however neither Kubo nor rust-ceramic provide an API to get the list of grafted gossipsub peers. I decided a simple delay was best. In practice the window of time the race can occur is on the order of 10s of milliseconds so a delay of 500ms was sufficient to never hit the race in repeated tests locally. I expect it might be a bit slower on CI hence the larger delay.

stbrody:
> Re the pubsub delay. There is naturally a race between node A subscribing and node B publishing a message to a topic

So is the reason this only shows up after switching to rust that the rust code is slower to do peer discovery and connection via pubsub, compared to the kubo code?

nathanielc:
@stbrody We discussed in person, but adding here for the record. The Kubo peers where using mdns to discover and peer with each other as well as also peered with the public mainnet nodes. This means the Kubo nodes were connected to long lived peers that can become gossipsub grafted peers and that Kubo nodes are connected sooner since they would discover each other on start instead of when the test sends the connect API calls. Both of these combined meant that Kubo nodes can do pubsub initially with less latency.

packages/cli/src/ipfs-connection-factory.ts
```diff
@@ -65,6 +70,40 @@ export class IpfsConnectionFactory {
...
+              ],
+              ID: 'QmXALVsXZwPWTUbsT8G6VVzzgTJaAWRUD7FWL5f7d5ubAL',
+            },
+            {
+              Addrs: [
+                '/dns4/go-ipfs-ceramic-private-cas-mainnet-external.3boxlabs.com/tcp/4011/ws/p2p/QmUvEKXuorR7YksrVgA7yKGbfjWHuCRisw2cH9iqRVM9P8',
+              ],
+              ID: 'QmUvEKXuorR7YksrVgA7yKGbfjWHuCRisw2cH9iqRVM9P8',
+            },
+            {
```
smrz2001, line 87: Would you mind taking out the two ELP multiaddrs? These nodes are no longer in service. We might reuse their peer IDs later but the DNS names will change so these entries will be stale regardless.

smrz2001: WDYT @stbrody? We _could_ also retain the ELP DNS names so that old Ceramic nodes with this list will just work with any future nodes reusing the ELP multiaddrs, even if they're not "ELP" nodes. That feels a little hacky though.

stbrody: yeah I actually feel like it's nice to leave them in and know we have multiaddrs in our back pocket we can use that will be strongly connected to even old nodes

packages/cli/src/ipfs-connection-factory.ts
```diff
@@ -65,6 +70,40 @@ export class IpfsConnectionFactory {
...
+              Addrs: [
+                '/dns4/go-ipfs-ceramic-elp-1-2-external.3boxlabs.com/tcp/4011/ws/p2p/QmRNw9ZimjSwujzS3euqSYxDW9EHDU5LB3NbLQ5vJ13hwJ',
+              ],
+              ID: 'QmRNw9ZimjSwujzS3euqSYxDW9EHDU5LB3NbLQ5vJ13hwJ',
+            },
+            {
+              Addrs: [
+                '/dns4/go-ipfs-ceramic-private-cas-clay-external.3boxlabs.com/tcp/4011/ws/p2p/QmbeBTzSccH8xYottaYeyVX8QsKyox1ExfRx7T1iBqRyCd',
```
stbrody: why a clay multiaddr here?

packages/core/src/state-management/__tests__/repository.test.ts
```diff
@@ -578,18 +578,26 @@ describe('validation', () => {
...
     )
     // eslint-disable-next-line @typescript-eslint/no-empty-function
-    validateSchemaSpy.mockImplementation(() => {})
+    validateSchemaSpy.mockImplementation(() => { })
     const invalidDoc = await TileDocument.create(
       permissiveCeramic,
       { stuff: 1 },
       { schema: schema.commitId }
     )
+
```
stbrody, line 599: make sure to always run prettier to clean up code formatting!

packages/core/src/stream-loading/__tests__/stream-loader.test.ts
```diff
@@ -65,8 +65,9 @@ describe('StreamLoader querying against real Ceramic node', () => {
 
     dispatcherIpfs = await createIPFS()
     dispatcher = await createDispatcher(dispatcherIpfs, TOPIC)
-    // speed up how quickly the dispatcher gives up on loading a non-existent commit from ipfs.
-    dispatcher._ipfsTimeout = 100
+    // We need this timeout to be short enough that the non-existent lookup timesout before
+    // the test case itself timesout.
+    dispatcher._ipfsTimeout = 1000
```
stbrody, line 70: why increasing this timeout?

packages/core/src/stream-loading/__tests__/stream-loader.test.ts
```diff
@@ -65,8 +65,9 @@ describe('StreamLoader querying against real Ceramic node', () => {
 
     dispatcherIpfs = await createIPFS()
     dispatcher = await createDispatcher(dispatcherIpfs, TOPIC)
-    // speed up how quickly the dispatcher gives up on loading a non-existent commit from ipfs.
-    dispatcher._ipfsTimeout = 100
+    // We need this timeout to be short enough that the non-existent lookup timesout before
```
stbrody, line 68: it's also just to make the tests run faster.  Any reason for updating this comment?

packages/cli/src/ipfs-connection-factory.ts
stbrody: hmm, I do see now that's how we were doing it before also.  I guess it is nice to have a bootstrap node for clay nodes too to keep them connected.  Really we should be setting this list dynamically based on which network you want to connect to.  Out of scope for this ticket though, I think this is fine for now.

packages/ipfs-daemon/src/create-ipfs.ts
```diff
@@ -118,54 +113,98 @@ async function createIpfsOptions(
...
+  api(): IpfsApi
+
+  shutdown(logger?: DiagnosticsLogger): Promise<void>
+}
+
 /**
  * Create an IPFS instance
- * @param overrideConfig - IFPS config for override
+ * @param flavor - IPFS flavor to create
+ * @param disposable - ???
```
smrz2001, line 162: Should this be filled in?

packages/cli/src/ipfs-connection-factory.ts
nathanielc: Yep, I tried to not change production behavior, only the test code should use a different list of peers.

packages/core/src/state-management/__tests__/repository.test.ts
nathanielc: I did, maybe it doesn't catch this change?

packages/core/src/stream-loading/__tests__/stream-loader.test.ts
nathanielc: A 100ms timeout was too short for the rest of the test cases. This timeout applies to about 7 tests that all share the dispatcher. It wasn't long enough to fetch blocks over bitswap.

packages/core/src/stream-loading/__tests__/stream-loader.test.ts
nathanielc: Because its not just about making the test faster. If you remove that line the test fails because the ipfsTimeout by default is longer that the test timeout. I changed the comment to make that clear.

packages/core/src/stream-loading/__tests__/stream-loader.test.ts
stbrody: oh, that's surprising.  100ms feels like a long time for well-connected nodes.  This also has never been a problem with kubo.  Is rust bitswap a lot slower?

packages/ipfs-daemon/src/create-ipfs.ts
nathanielc: Not sure what this is, @dbcfd Can you add some clarity? I remember looking at what disposable did and now I don't remember.

packages/core/src/state-management/__tests__/repository.test.ts
stbrody: there are a bunch of small formatting issues throughout this PR, I only pointed out this one though.  Maybe make sure your IDE is set up to run it on save?  You should also be able to run `npm run format` in the js-ceramic root dir to format the whole repo.

nathanielc: My IDE is setup to run it on save, however it does change various existing formatting. Maybe there is a missing config my IDE needs to find? 

I tried `npm run format` but the `format` script is not defined in package.json (toplevel).

I ran `npx prettier -w packages/**/src/**/*.ts`  before pushing as it was the closest command I could find in the package.json.

stbrody: Oh yeah, you're right, my bad.  It was the ceramic-integration-tests repo where I saw a "format" script: https://github.com/3box/ceramic-integration-tests/blob/213925d06e546ef7ad6db4ff78215eb84269dabb/package.json#L24

stbrody: > My IDE is setup to run it on save, however it does change various existing formatting.

hmm interesting.  Maybe also double-check what version of prettier you are using?

packages/core/src/stream-loading/__tests__/stream-loader.test.ts
nathanielc: Honestly I don't concretely know why rust is sensitive to 100ms timeouts, but I do have a few theories. However my perspective is these are correctness tests not performance tests and so increasing the timeout to get stable tests was best. 

I did however test one of my theories that its TCP thats slow compared to QUIC. I changed rust-ceramic to use QUIC exclusively for these tests and it passes consistently with 200ms timeout. At 100ms it still had ~5% failure rate. I can make the QUIC change and use 200ms thoughts?

packages/ipfs-daemon/src/create-ipfs.ts
dbcfd: I think it means when done with the ipfs to shut it down. If not disposable, it leaves it up. I think it does this with proxy to stop or not.

stbrody (approved):
lgtm

PR #2969, opened 2023-09-25
Update Repository.applyCommit to use the new StreamUpdater

stbrody:


packages/core/src/state-management/repository.ts
```diff
@@ -267,10 +271,25 @@ export class Repository {
...
+      const originalState = state$.state
+      const updatedState = await this.streamUpdater.applyCommitFromUser(originalState, commit)
+      if (StreamUtils.tipFromState(updatedState).equals(StreamUtils.tipFromState(originalState))) {
+        return state$ // nothing changed
+      }
+
+      state$.next(updatedState) // emit the new state
+
+      await this._internals._updateStateIfPinned(state$)
+      await this.applyWriteOpts(state$, opts, OperationType.UPDATE)
```
stbrody, line 288: the only functional change in this PR is moving this call to `applyWriteOpts` to be inside the execution queue.  This seems to me the easiest way to be able to decide whether or not to update the persistent streamstate in the state store based on whether or not creating the anchor request against the CAS is successful.

stbrody (commented):
This still needs work to rework the test coverage before its ready for final review, but I wanted to run the change of moving more of the extra steps of commit application (e.g. requesting the anchor, publishing the tip to pubsub) into the execution queue context by you @ukstv.  Please let me know if you think that will cause any problems for us.

ukstv (approved):
I hope it works :) I have no objections for the approach.

stbrody:
This is ready for real review now!

packages/common/src/utils/stream-utils.ts
```diff
@@ -32,6 +32,10 @@ export class StreamUtils {
     return new StreamID(state.type, state.log[0].cid)
   }
 
+  static tipFromState(state: StreamState): CID {
+    return state.log[state.log.length - 1].cid
```
PaulLeCam, line 36: Is there any guaranty the state log has at least 1 item please?

PaulLeCam (approved):

packages/common/src/utils/stream-utils.ts
stbrody: Yep.  To construct a StreamState there needs to be a genesis commit

PR #2971, opened 2023-10-03
Off-memory anchoring

ukstv:
Probably, some tests will fail.


packages/core/src/ceramic.ts
```diff
@@ -455,7 +459,19 @@ export class Ceramic implements CeramicApi {
...
+        const carBuilder = this.repository._internals.anchorRequestCarBuilder
+        const repository = this.repository
+        const handler: AnchorLoopHandler = {
+          buildRequestCar(streamId: StreamID, tip: CID): Promise<CAR> {
+            return carBuilder.build(streamId, tip)
+          },
+          async handle(event: AnchorEvent): Promise<boolean> {
+            const state$ = await repository.fromMemoryOrStore(event.streamId)
+            if (!state$) return true
+            return repository._internals.handleAnchorResponse2(state$, event)
```
stbrody, line 471: this is introducing another circular dependency, just as I'm trying to hunt down and remove the last of these from the Repository.  Instead of the AnchorService calling back into the Repository as it gets anchor responses, can there just be an Observable that spits out anchor commits in need of handling, and the Repository can subscribe to that observable and handle the anchor commits as they come out?

As in, basically similar to how it works today, but instead of one Observable per stream there is one global Observable of all streams

packages/cli/src/__tests__/ceramic-daemon.test.ts
```diff
@@ -153,23 +153,27 @@ describe('Ceramic interop: core <> http-client', () => {
 
   it('gets anchor commit updates', async () => {
     const doc1 = await TileDocument.create(core, { test: 123 })
+    await TestUtils.hasAcceptedAnchorRequest(core, doc1.tip)
```
stbrody, line 156: I really don't like seeing these scattered throughout all of our tests.  Why is this necessary?

stbrody (changes requested):
this PR is massive and very difficult to review.  Can we please break this up into smaller pieces?  I also really want to avoid introducing more circular dependencies into the Repository

packages/core/src/ceramic.ts
ukstv: Oh, the circular dependencies gave me so much pain! There might be an option to clean that up, but for this particular case, I see no other way. Maybe after all the other stuff from here gets unbundled.

Observables are not an option. We need to know on the loop side, that an event is terminal or not. There is no sane way one could signal that back to the loop. Streams (NodeJS and Web Streams alike) go the same route for the same reason. Iterables/Generators maybe, yet it's a fucking joke to control their lifecycle.

dbcfd (changes requested):

packages/core/src/anchor/anchor-service.ts
```diff
@@ -0,0 +1,144 @@
...
+   * Performs whatever initialization work is required by the specific anchor service implementation
+   */
+  init(store: AnchorRequestStore, eventHandler: AnchorLoopHandler): Promise<void>
+
+  /**
+   * Set Ceramic API instance
+   *
+   * @param ceramic - Ceramic API used for various purposes
+   */
+  ceramic: CeramicApi
```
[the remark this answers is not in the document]
ukstv: Definitely. That's a remnant of what has been here prior. It could be even bigger PR if this thing has been changed too. Added an issue to fix that.

packages/core/src/__tests__/ceramic.test.ts
```diff
@@ -17,7 +17,7 @@ import { createCeramic as vanillaCreateCeramic } from './create-ceramic.js'
 import { AnchorResumingService } from '../state-management/anchor-resuming-service.js'
 
 const MOCK_WAS_CALLED_DELAY = 3 * 1000
-const TEST_TIMEOUT = 1000 * 60 * 12 // 12 minutes
+const TEST_TIMEOUT = 1000 * 30 // 12 minutes
```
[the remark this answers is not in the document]
ukstv: or adjust the timeout. done.

packages/cli/src/__tests__/ceramic-daemon.test.ts
ukstv: And, it's gone.

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -122,46 +77,25 @@ export class EthereumAnchorService implements AnchorService {
     // Do Nothing
   }
 
-  async init(): Promise<void> {
+  async init(store: AnchorRequestStore, eventHandler: AnchorLoopHandler): Promise<void> {
+    this.#store = store
+    // Get the chainIds supported by our anchor service
     const supportedChains = await this.#cas.supportedChains()
+    if (supportedChains.length > 1) {
```
dbcfd, line 84: Thinking about this more, we could just do

```
if (supportedChains.length != 1) {
  throw new NotSingleChainError()
}```

File: packages/core/src/anchor/ethereum/ethereum-anchor-service.ts [hunk 6]
```diff
@@ -244,15 +158,17 @@ export class AuthenticatedEthereumAnchorService
     ethereumRpcUrl: string | undefined,
     logger: DiagnosticsLogger,
     pollInterval: number = DEFAULT_POLL_INTERVAL,
-    maxPollTime: number = MAX_POLL_TIME
+    maxPollTime: number = MAX_POLL_TIME,
+    enableLoop = false
```
dbcfd, line 162: EthereumAnchorService defaults this to true, while AuthenticatedEthereumAnchorService and AnchorService defaults this to false. Why the difference?

packages/core/src/anchor/processing-loop.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  constructor(source: ProcessingLoop<T>['source'], onValue: ProcessingLoop<T>['handleValue']) {
+    this.source = source
+    this.handleValue = onValue
+    this.#processing = undefined
+    this.#deferred = new Deferred()
+    this.#abortController = new AbortController()
+  }
+
+  start() {
```
dbcfd, line 37: What are the cases where this is not just started right away? Looking through the code, it seems like this should just always be called in the constructor.

dbcfd (changes requested):

packages/core/src/state-management/repository-internals.ts
```diff
@@ -244,6 +239,14 @@ export class RepositoryInternals {
     return this.fromStreamStateStore(streamId)
   }
 
+  async fromMemoryOrStoreSafe(streamId: StreamID): Promise<RunningState | undefined> {
```
stbrody, line 242: please, comment what makes this function "safe" compared to regular `fromMemoryOrStore`.

Also, can this be implemented in terms of `fromMemoryOrStore` instead of calling _fromMemory and fromStreamStateStore directly?

packages/core/src/state-management/repository.ts
```diff
@@ -541,6 +543,21 @@ export class Repository {
     })
   }
 
+  anchorLoopHandler(): AnchorLoopHandler {
```
stbrody, line 546: nice.  It's not a perfect avoidance of the circular relationship between the Repository and the AnchorService, but I think it's the best we can do for now and does a good job keeping each module focused on its core concerns.

packages/stream-tests/src/__tests__/node-status.test.ts
```diff
@@ -93,40 +78,4 @@ describe('Node Status Endpoint tests', () => {
       expect(typeof addr).toEqual('string')
     }
   })
-
-  test('Pending anchor counter test', async () => {
```
stbrody, line 97: does this change break the `pendingAnchors` reporting in the `nodeStatus` endpoint?  If so, should we remove that field from the nodeStatus output rather than just have it always return 0?

stbrody (changes requested):
please don't be scared by the large number of comments, the vast majority are quite trivial, just requests for comments or variable renames for clarity.  Overall I think this PR looks really good.  My only one real concern is deferring anchor request creation against the CAS.  I'd like us to still create the requests greedily against the CAS, and have the processing loop only responsible for polling, not creation.

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
ukstv: Okay. Changed.

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
ukstv: That's a mistake. Thank for noticing it.

packages/core/src/anchor/processing-loop.ts
ukstv: ...unless we explicitly disable it (enableLoop = false) to avoid interference with some tests. Also, it is not quite compatible with the way other stateful parts of Ceramic work, like storage.

packages/core/src/__tests__/state-manager.test.ts
```diff
@@ -209,71 +204,15 @@ describe('anchor', () => {
...
-      controllers = [ceramic.did.id]
-      inMemoryAnchorService = ceramic.repository.anchorService as InMemoryAnchorService
+      inMemoryAnchorService = ceramic.anchorService as InMemoryAnchorService
     })
 
     afterAll(async () => {
       await ceramic.close()
     })
 
-    test('stop on REPLACED', async () => {
```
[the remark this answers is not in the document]
ukstv: Because we do not have polling per anchor request anymore.

packages/core/src/__tests__/state-manager.test.ts
```diff
@@ -287,11 +226,10 @@ describe('anchor', () => {
...
       expect(stream$.value.anchorStatus).toEqual(AnchorStatus.ANCHORED)
 
-      // dag.import is called 4 times.  Once when the genesis commit is created, twice with the
-      // AnchorCommit itself (once when the InMemoryAnchorService publishes the AnchorCommit and
-      // once when Ceramic applies it), and once with the CAR file with the merkle witness
-      // received from the CAS.
-      expect(dagImportSpy).toHaveBeenCalledTimes(4)
+      // dag.import is called 2 times:
+      // 1) when the genesis commit is created on Ceramic side,
+      // 2) when Ceramic applies it
```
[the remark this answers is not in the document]
ukstv: Yup. Changed to "when Ceramic applies an anchor commit received from CAS"

packages/core/src/anchor/anchor-processing-loop.ts
```diff
@@ -0,0 +1,45 @@
...
+import type { CASClient } from './anchor-service.js'
+import type {
+  AnchorRequestStore,
+  AnchorRequestStoreListResult,
+} from '../store/anchor-request-store.js'
+import { AnchorRequestCarFileReader } from './anchor-request-car-file-reader.js'
+import type { AnchorLoopHandler } from './anchor-service.js'
+import type { DiagnosticsLogger } from '@ceramicnetwork/common'
+
+export class AnchorProcessingLoop {
```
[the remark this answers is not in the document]
ukstv: Done.

packages/core/src/__tests__/state-manager.test.ts
```diff
@@ -431,142 +373,96 @@ describe('anchor', () => {
         // @ts-ignore anchorRequestStore is private
```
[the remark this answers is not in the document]
ukstv: Yep. This test suite is worthless now. Moved majority of the relevant tests to `repository.test.ts`.

packages/core/src/anchor/anchor-processing-loop.ts
```diff
@@ -0,0 +1,45 @@
...
+    batchSize: number,
+    cas: CASClient,
+    store: AnchorRequestStore,
+    logger: DiagnosticsLogger,
+    eventHandler: AnchorLoopHandler
+  ) {
+    this.#loop = new ProcessingLoop(store.infiniteList(batchSize), async (entry) => {
+      try {
+        const event = await cas.get(entry.key, entry.value.cid).catch(async () => {
+          const requestCAR = await eventHandler.buildRequestCar(entry.key, entry.value.cid)
```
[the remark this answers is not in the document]
ukstv: 1. CAS creation is idempotent. That is if you create a request with the same streamId, tip, it does nothing. Even if it is a sort of connectivity error, creation does nothing bad.
2. Here we operate under assumption, that if something is in AnchorRequestStore it should be anchored. Hence, even if CAS says "no such request", we try to create it to maintain the assumption.

packages/core/src/anchor/anchor-processing-loop.ts
```diff
@@ -0,0 +1,45 @@
...
+          return cas.create(new AnchorRequestCarFileReader(requestCAR), false)
+        })
+        const isTerminal = await eventHandler.handle(event)
+        if (isTerminal) {
+          await store.remove(entry.key)
+        }
+      } catch (e) {
+        logger.err(e)
+      }
+      await new Promise((resolve) => setTimeout(resolve, pollInterval))
```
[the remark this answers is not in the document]
ukstv: Actually, there is already a delay in storage, right as you described.

packages/core/src/anchor/ethereum/remote-cas.ts
```diff
@@ -82,8 +82,17 @@ export class RemoteCAS implements CASClient {
     carFileReader: AnchorRequestCarFileReader,
```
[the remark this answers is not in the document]
ukstv: Added a doc comment.

I see that issue on Linear. It is assigned to you. I do not dare to interfere.

packages/core/src/anchor/memory/in-memory-anchor-service.ts
```diff
@@ -1,186 +1,99 @@
...
 
   /**
    * Set Ceramic API instance
    *
    * @param ceramic - Ceramic API used for various purposes
    */
-  set ceramic(ceramic: Ceramic) {
-    this.#ceramic = ceramic
-    this.#dispatcher = this.#ceramic.dispatcher
+  set ceramic(ceramic: CeramicApi) {
```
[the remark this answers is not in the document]
ukstv: Not yet. It is used in authenticated AnchorService to pass DID instance, used for signature verification.

packages/core/src/anchor/processing-loop.ts
```diff
@@ -0,0 +1,74 @@
+export class Deferred<T = void> {
```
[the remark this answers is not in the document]
ukstv: Done.

packages/core/src/anchor/processing-loop.ts
```diff
@@ -0,0 +1,74 @@
...
+  constructor(source: ProcessingLoop<T>['source'], onValue: ProcessingLoop<T>['handleValue']) {
+    this.source = source
+    this.handleValue = onValue
+    this.#processing = undefined
+    this.#deferred = new Deferred()
+    this.#abortController = new AbortController()
+  }
+
+  start() {
+    const abortion = new Promise<IteratorResult<T>>((resolve) => {
```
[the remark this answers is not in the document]
ukstv: Okay, now it should be more appropriate, I believe.

packages/core/src/anchor/processing-loop.ts
```diff
@@ -0,0 +1,74 @@
...
+      try {
+        let isDone = false
+        do {
+          const next = await Promise.race([this.source.next(), abortion])
+          isDone = next.done
+          if (isDone) break
+          const value = next.value
+          await Promise.race([this.handleValue(value), abortion])
+        } while (!isDone)
+        this.#deferred.resolve()
```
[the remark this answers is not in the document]
ukstv: Let it be `#whenComplete` ?

packages/stream-tests/src/__tests__/node-status.test.ts
ukstv: There is a ticket to handle it separately. This PR would be tooo big to review if `nodeStatus` is also taken care of.

packages/core/src/__tests__/state-manager.test.ts
```diff
@@ -431,142 +373,96 @@ describe('anchor', () => {
...
-          streamId: tile.id,
-          cid: tile.state.log[0].cid,
-          message: 'CAS is processing the request',
-        })
-        await expectAnchorStatus(tile, AnchorStatus.PROCESSING)
+        await TestUtils.expectAnchorStatus(tile, AnchorStatus.PENDING)
 
         await inMemoryAnchorService.anchor()
 
         // We have received the anchor commit through pubsub, we have not received the anchor response yet
```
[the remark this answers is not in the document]
ukstv: Probably not. Moved relevant tests to repository.test.ts.

packages/core/src/anchor/anchor-processing-loop.ts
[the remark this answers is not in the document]
ukstv: Removed from here.

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -174,58 +108,38 @@ export class EthereumAnchorService implements AnchorService {
...
-  }
+  async requestAnchor(carFile: CAR, waitForConfirmation: boolean): Promise<AnchorEvent> {
+    const carFileReader = new AnchorRequestCarFileReader(carFile)
+    const streamId = carFileReader.streamId
+    const tip = carFileReader.tip
 
-  private _parseResponse(streamId: StreamID, tip: CID): OperatorFunction<unknown, AnchorEvent> {
-    return map((response) => parseResponse(streamId, tip, response))
-  }
+    await this.#store.save(streamId, {
```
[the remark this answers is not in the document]
ukstv: > I see that issue on Linear. It is assigned to you. I do not dare to interfere.

Better copy it here as well from https://github.com/ceramicnetwork/js-ceramic/pull/2971#discussion_r1361094213

packages/core/src/anchor/processing-loop.ts
```diff
@@ -0,0 +1,74 @@
...
+    this.reject = dReject
+    this.then = promise.then.bind(promise)
+    this.catch = promise.catch.bind(promise)
+  }
+}
+
+export class ProcessingLoop<T> {
+  private readonly source: AsyncGenerator<T>
+  private readonly handleValue: (value: T) => Promise<void>
+  #processing: Promise<void> | undefined
```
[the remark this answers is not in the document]
ukstv: Done.

packages/core/src/state-management/repository-internals.ts
ukstv: Actually, yes. Done.

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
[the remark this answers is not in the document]
stbrody: My work for anchoring robustness only changes the behavior when `waitForConfirmation` is true. I'm talking about the behavior when `waitForConfirmation` is false (ie the current behavior and what will still be the default behavior for a while yet).

This PR already does what I'm asking when `waitForConfirmation` is true, but I think even if `waitForConfirmation` is false we should proactively create the request against the CAS (but not wait for that creation to complete before returning to the client)

packages/core/src/anchor/ethereum/remote-cas.ts
```diff
@@ -78,12 +75,27 @@ export class RemoteCAS implements CASClient {
...
+   * has already saved an anchor request to AnchorRequestStore.
+   */
   async create(
     carFileReader: AnchorRequestCarFileReader,
     waitForConfirmation: boolean
   ): Promise<AnchorEvent> {
-    const response = await firstValueFrom(this.create$(carFileReader, waitForConfirmation))
-    return parseResponse(carFileReader.streamId, carFileReader.tip, response)
+    if (waitForConfirmation) {
+      const response = await firstValueFrom(this.create$(carFileReader, waitForConfirmation))
```
stbrody, line 89: I'm a little confused about the relationship between `waitForConfirmation` and the `shouldRetry` arg to `create$`.  It looks like in the code that exists in develop today, `shouldRetry` is set to the *opposite* of `waitForConfirmation`, whereas in this PR its set to the same thing.  But I actually don't really understand why `waitForConfirmation` should have any affect on the retry behavior?

stbrody: Maybe now that we have retries by way of the anchor processing loop, there's no need for retries at this layer at all?  Or maybe a fixed number of retries here, one or two retries, and then stop, instead of retrying indefinitely as seems to be the behavior currently.

packages/core/src/anchor/processing-loop.ts
```diff
@@ -0,0 +1,74 @@
...
+        resolve({ done: true, value: undefined })
+      }
+      this.#abortController.signal.addEventListener('abort', done)
+    })
+    const processing = async (): Promise<void> => {
+      try {
+        let isDone = false
+        do {
+          const next = await Promise.race([this.source.next(), abortion])
+          isDone = next.done
```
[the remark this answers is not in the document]
stbrody: Ah, I see.  I thought AsyncGenerator<T> would just emit Ts, I didn't realize it emitted a structured IteratorResult<T>

packages/core/src/anchor/anchor-processing-loop.ts
```diff
@@ -0,0 +1,60 @@
...
+      try {
+        const event = await cas.getStatusForRequest(entry.key, entry.value.cid).catch(async () => {
+          const requestCAR = await eventHandler.buildRequestCar(entry.key, entry.value.cid)
+          return cas.create(new AnchorRequestCarFileReader(requestCAR), false)
+        })
+        const isTerminal = await eventHandler.handle(event)
+        if (isTerminal) {
+          // We might store a new entry during the processing, so we better check if the current entry is indeed terminal.
+          const current = await store.load(entry.key)
+          if (current.cid.equals(entry.value.cid)) {
```
stbrody, line 37: won't this check be racy?  What if there is a write to the request store in flight already at this point?  I think all reads/writes to the anchor request store for a stream need to be serialized in the execution queue to be safe

stbrody: Maybe if we can push this down into the eventHandler itself and do it all within the context of the execution queue?

packages/core/src/state-management/__tests__/anchor-resuming.test.ts
```diff
@@ -0,0 +1,93 @@
...
+afterEach(async () => {
+  jest.resetAllMocks()
+})
+
+test('resume anchors from AnchorRequestStore', async () => {
+  const numberOfStreams = 3
+
+  const ceramic = await createCeramic(ipfs, {
+    stateStoreDirectory: stateStoreDirectoryName,
+    enableLoop: false,
```
stbrody, line 35: this looks weird at this level to just see `enableLoop` without context.  Maybe can we call it `enableAnchorPollingLoop`?

packages/core/src/state-management/__tests__/repository.test.ts
```diff
@@ -637,3 +649,105 @@ describe('applyWriteOpts', () => {
...
+          ...state$.value,
+          anchorStatus: AnchorStatus.ANCHORED,
+        })
+      })
+      const shouldRemove = await repository.handleAnchorEvent(state$, event)
+      expect(shouldRemove).toBeTruthy()
+      expect(handleAnchorCommitSpy).toBeCalled()
+    })
+
+    test('Anchor REPLACED for non tip should remove the request from the store', async () => {
```
stbrody, line 721: shouldn't this NOT remove it from from the store if the REPLACED status is coming for a commit that isn't the current tip?

packages/core/src/state-management/__tests__/repository.test.ts
```diff
@@ -637,3 +649,105 @@ describe('applyWriteOpts', () => {
...
+      const state$ = await repository.load(tile.id)
+      const event: AnchorEvent = {
+        status: AnchorRequestStatusName.REPLACED,
+        message: 'First REPLACED',
+        streamId: tile.id,
+        cid: tile.state.log[0].cid,
+      }
+      const nextSpy = jest.spyOn(state$, 'next')
+      const shouldRemove = await repository.handleAnchorEvent(state$, event)
+      expect(shouldRemove).toBeTruthy()
```
stbrody, line 733: this should be false no?  We still need to keep polling for the tip, the fact that the non-tip previous commit was REPLACED doesn't change that.

packages/core/src/state-management/__tests__/repository.test.ts
stbrody: actually ditto for the other two tests in this "non-tip" suite.  Any changes to the state of the request for commits other than the tip shouldn't change the fact that we need to keep polling for the tip ever, right?

stbrody (changes requested):

gvelez17:
hey @ukstv we think this PR may resolve an ongoing incident https://linear.app/3boxlabs/issue/CORE-112/knex-issues-in-clay-cas - is it possible to prioritize it ?  Or can we ask for that from someone upstream?

packages/core/src/state-management/__tests__/repository.test.ts
ukstv: "keep polling for the tip" = we do not know about the tip, so REPLACED is a terminal state ceteris paribus.

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -1,116 +1,68 @@
...
 } from '../anchor-service.js'
-import { CasConnectionError, MaxAnchorPollingError } from '../anchor-service.js'
+import type { AnchorRequestStore } from '../../store/anchor-request-store.js'
+import { NotSingleChainError, type AnchorLoopHandler } from '../anchor-service.js'
+import { AnchorProcessingLoop } from '../anchor-processing-loop.js'
 import { RemoteCAS } from './remote-cas.js'
+import { doNotWait } from '../../ancillary/do-not-wait.js'
+import { NamedTaskQueue } from '../../state-management/named-task-queue.js'
 
 const DEFAULT_POLL_INTERVAL = 60_000 // 60 seconds
```
stbrody, line 27: is this used anymore?

packages/core/src/store/anchor-request-store.ts
```diff
@@ -73,7 +75,34 @@ export class AnchorRequestStore extends ObjectStore<StreamID, AnchorRequestData>
     } while (true)
   }
 
-  async close(): Promise<void> {
-    await this.store.close(this.useCaseName)
+  async *infiniteList(
+    batchSize = 1,
+    batchDelay = 100
```
stbrody, line 80: this delay doesn't actually happen between each batch, it happens between each full loop when you get to the end of the entire result set and are about to loop back around to the beginning.  Can we rename it so that's clearer?  Maybe `restartDelay` or something like that?

packages/core/src/anchor/anchor-processing-loop.ts
```diff
@@ -0,0 +1,64 @@
...
+    batchSize: number,
+    cas: CASClient,
+    store: AnchorRequestStore,
+    logger: DiagnosticsLogger,
+    eventHandler: AnchorLoopHandler,
+    queue: NamedTaskQueue
+  ) {
+    this.#queue = queue
+    this.#loop = new ProcessingLoop(store.infiniteList(batchSize), (entry) =>
+      this.#queue.run(entry.key.toString(), async () => {
```
stbrody, line 33: what's the advantage of using the NamedTaskQueue here?  It just serializes handling for a single streamid, but we're already guaranteed that everything coming out of the AnchorRequestStore is going to be for a unique StreamID since StreamID is the key in the store.

Can we just get rid of `#queue` entirely?

packages/core/src/anchor/processing-loop.ts
```diff
@@ -0,0 +1,124 @@
...
+      if (this.#abortController.signal.aborted) {
+        resolve({ done: true, value: undefined })
+      }
+      const done = () => {
+        this.#abortController.signal.removeEventListener('abort', done)
+        resolve({ done: true, value: undefined })
+      }
+      this.#abortController.signal.addEventListener('abort', done)
+    })
+    const processing = async (): Promise<void> => {
```
stbrody, line 90: I'm trying to understand the degree of concurrency that happens with the new polling system.  Let me see if I'm understanding correctly.  I see that when we query entries from the AnchorRequestStore, we use a batch size of 10.  So we fetch 10 entries to process from the store at once, but if I'm reading this properly, we still process them one at a time since this processing loop waits for each value to be handled before starting processing on the next value, is that right?

I think we're likely going to need more concurrency here to keep up with the incoming write load, especially with our planned performance targets for this season, but I'm fine with leaving that for follow-up work, just trying to make sure I understand the current state of things

packages/core/src/anchor/anchor-processing-loop.ts
stbrody: Aha!  I now see we use the same queue in the EthereumAnchorService to serialize writes to the AnchorRequestStore there.  So now I see that the goal of this is to serialize all writes to the AnchorRequestStore, instead of using the ExecutionQueue for that.  Nice.  Can we just add some comments to the queue both here and in the EthereumAnchorService to make the clearer?  And maybe also rename it in both places to something like `anchorStoreQueue` or something?

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -1,116 +1,68 @@
...
-  /**
-   * Retry a request to CAS every +pollInterval+ milliseconds.
-   */
-  readonly #pollInterval: number
-  readonly #maxPollTime: number
+  #loop: AnchorProcessingLoop
+  readonly #enableAnchorPollingLoop: boolean
   readonly #events: Subject<AnchorEvent>
-  readonly #cas: RemoteCAS
+  readonly #queue: NamedTaskQueue
```
stbrody, line 38: Can we name this `#anchorStoreQueue` or something like that, and add comments explaining why we need it?

packages/core/src/state-management/__tests__/anchor-resuming.test.ts
```diff
@@ -0,0 +1,93 @@
+import { expect, jest } from '@jest/globals'
```
stbrody, line 1: ugh sorry, I just added some tests to `anchor-resuming-service.test.ts` as part of the fixes I just put in.  LMK if you want any help porting them over, but we should make sure not to lose the coverage

packages/core/src/anchor/anchor-processing-loop.ts
stbrody: wow, the deeper I look at things like `Repository.handleAnchorEvent` the more I realize how awesome and elegant this approach is.  It totally lets us work around not having a transactional state store while *almost* giving us proper transactional read-modify-write semantics, all without slowing down regular write processing since it all happens outside of the ExecutionQueue.

It's so close to being a perfect solution, but there is still one small race condition I believe.  The actual query to the AnchorRequestStore doesn't happen within the queue, so it's possible that by the time the `onValue` handler in the ProcessingLoop is run, the entry in the AnchorRequestStore could have changed from what we read out and polled the CAS for.  That could cause us to delete an entry for a newer tip from the store because we hit a terminal state with an older tip that we read out.

Can you think of any ways to close that final gap?  The only way I can think of would be to change the `store.infiniteList` batch query to just get the keys from the store but not the values, and then within the actual `onValue` processor load the value from the AnchorRequestStore so that you're already within the queue and are guaranteed that the value can't then change out from under you.  It's more i/o to the store which I don't love, but it would fully solve all the race conditions around managing the store once and for all.  What do you think?

stbrody (commented):
alright this is looking veeeeery close.  There are really only 2 remaining issues that I actually care about.  One is closing the last remaining race condition around how the AnchorRequestStore is maintained.  The other is making sure we don't lose the test coverage for the bugs in AnchorRequestStore maintenance that I just added in https://github.com/ceramicnetwork/js-ceramic/commit/85f8a790df1916971e4bfafd9120132e64bf0899#diff-379d49076ceda06ac206ea7ab13ae8f78827a456349aef73236d6d53dbe01060 and https://github.com/ceramicnetwork/js-ceramic/commit/c3dc51fd712b213687559e5ebdbd2c49ec527ca1#diff-379d49076ceda06ac206ea7ab13ae8f78827a456349aef73236d6d53dbe01060.

LMK if you want any help of thinking of how to rework those tests to work in the new model without losing coverage.

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
ukstv: True. Removed.

packages/core/src/anchor/anchor-service.ts
```diff
@@ -120,16 +110,29 @@ export interface AnchorValidator {
...
+   */
   supportedChains(): Promise<Array<string>>
-  create(
-    carFileReader: AnchorRequestCarFileReader,
-    waitForConfirmation: boolean
-  ): Promise<AnchorEvent>
-  get(streamId: StreamID, tip: CID): Promise<AnchorEvent>
+
+  /**
+   * Create an anchor request on CAS through `fetch`. If `waitForConfirmation` is `true`,
```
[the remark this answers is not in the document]
ukstv: Nice catch! Removed.

packages/core/src/store/anchor-request-store.ts
ukstv: let it be `restartDelay`

packages/core/src/anchor/anchor-processing-loop.ts
ukstv: A game of tradeoffs again. Considering AnchorRequestStore is most probably a local thing, I guess, IO overhead is not that big a deal.

ukstv: Yeah, I think it is done.

packages/core/src/anchor/processing-loop.ts
ukstv: Yes, it is one streamid at a time. You are correct. We should explore how to make it more concurrent definitely. We have talked about it before. If I remember correctly, the decision was to postpone more advanced polling to the next iteration, after we confirm the off-chain memory mechanism works as a whole.

packages/core/src/state-management/__tests__/anchor-resuming.test.ts
ukstv: Please! If you could port the changes, that would be great!

ukstv:
@stbrody I believe, I handle https://github.com/ceramicnetwork/js-ceramic/pull/3055 and https://github.com/ceramicnetwork/js-ceramic/pull/3053 kind of automatically here. I mean, these cases are handled by a general control flow.

stbrody (commented):
okay I think this basically looks good to go other than porting the tests over.  Going to take a crack at that now

packages/core/src/state-management/__tests__/anchor-resuming.test.ts
stbrody: Done!

packages/core/src/anchor/anchor-processing-loop.ts
```diff
@@ -17,7 +15,7 @@ import type { NamedTaskQueue } from '../state-management/named-task-queue.js'
  * We create a request on CAS then.
  */
 export class AnchorProcessingLoop {
-  readonly #loop: ProcessingLoop<AnchorRequestStoreListResult>
+  readonly #loop: ProcessingLoop<StreamID>
   readonly #queue: NamedTaskQueue
```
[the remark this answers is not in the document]
stbrody: I went ahead and did this and added it to the branch

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -35,7 +34,10 @@ export class EthereumAnchorService implements AnchorService {
   #loop: AnchorProcessingLoop
   readonly #enableAnchorPollingLoop: boolean
   readonly #events: Subject<AnchorEvent>
-  readonly #queue: NamedTaskQueue
+  /**
+   * Linearizes requests to AnchorStoreQueue by stream id.
```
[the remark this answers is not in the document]
stbrody: I went ahead and did this and added it to the branch

stbrody (approved):
LGTM!!!!

PR #2974, opened 2023-10-04
Make StateManipulator solely responsible for preserving immutability of StreamStates

stbrody:
Builds on https://github.com/ceramicnetwork/js-ceramic/pull/2982

I tried to use Object.freeze and the Readonly type to prevent StreamHandlers from modifying their input, but that failed to catch any of the places where the state is modified today.  Given that it doesn't seem like typescript can reliably prevent mutations to objects, I decided to go the other direction and just let the StreamHandlers mutate their inputs as much as they want, but have the StateManipulator make sure to clone the input state before calling into the StreamHandlers.

### DIFF
--- a/packages/core/src/stream-loading/state-manipulator.ts
+++ b/packages/core/src/stream-loading/state-manipulator.ts
@@ -73,7 +73,7 @@ export class StateManipulator {
 
     for (const commit of log.commits) {
       try {
-        state = await handler.applyCommit(commit, this.context, Object.freeze(state))
+        state = await handler.applyCommit(commit, this.context, state)
       } catch (err) {
         if (throwOnInvalidCommit || state == null) {
           throw err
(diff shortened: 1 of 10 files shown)


packages/core/src/stream-loading/state-manipulator.ts
```diff
@@ -54,22 +54,16 @@ export class StateManipulator {
...
+   * Applies log to state and returns a new state.  Note that the input state may be modified,
+   * if you don't want to change the input state, callers should clone it before passing it into
+   * this function.
    */
   async _applyLog<T extends Stream>(
     handler: StreamHandler<T>,
     state: StreamState | null,
     log: AppliableStreamLog,
     throwOnInvalidCommit: boolean,
-    cloneInput: boolean
```
stbrody, line 68: decided to get rid of this argument and make the cloneDeep explicit at the call site.  That seemed clearer and more straightforward to understand to me.

PaulLeCam (approved):

PR #2976, opened 2023-10-05
chore: Extract inmemory validator

ukstv:
...due to a difference for what `init` expects on the inmemory anchor service and the inmemory validator.

Previously `InMemoryAnchorService` implemented `AnchorValidator` interface. The PR extracts `InMemoryAnchorValidator` class to free `InMemoryAnchorService` from validator responsibility, and make unconvoluted `init`.

Positive side-effect: InMemoryAnchorService is more inline with EthereumAnchorService.

That's part of off-memory anchoring PR unbundling.

### DIFF
--- a/packages/core/src/anchor/memory/in-memory-anchor-service.ts
+++ b/packages/core/src/anchor/memory/in-memory-anchor-service.ts
@@ -6,12 +6,12 @@ import type { AnchorProof, AnchorCommit, AnchorEvent } from '@ceramicnetwork/com
 import type { Dispatcher } from '../../dispatcher.js'
 import type { Ceramic } from '../../ceramic.js'
 import type { StreamID } from '@ceramicnetwork/streamid'
-import { LRUCache } from 'least-recent'
 import { CARFactory, type CAR } from 'cartonne'
 import * as DAG_JOSE from 'dag-jose'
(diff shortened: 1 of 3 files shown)


packages/core/src/anchor/memory/in-memory-anchor-service.ts
```diff
@@ -278,7 +277,7 @@ export class InMemoryAnchorService implements AnchorService, AnchorValidator {
       //TODO (NET-1657): Update the InMemoryAnchorService to mirror the behavior of the contract-based anchoring system
       txType: V1_PROOF_TYPE,
     }
-    txnCache.set(txHashCid.toString(), timestamp)
+    TRANSACTION_CACHE.set(txHashCid.toString(), timestamp)
```
dbcfd, line 280: Can we inject rather than import this?

dbcfd (changes requested):

packages/core/src/anchor/memory/in-memory-anchor-service.ts
ukstv: Would love to, but again, please not for now. "Too many changes." Let me track that separately.

dbcfd (approved):

PR #2978, opened 2023-10-06
chore: Add RemoteCAS

ukstv:
Abstract over fetch to CAS instance for EthereumAnchorService.

### DIFF
--- a/packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
+++ b/packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
@@ -25,32 +25,17 @@ import { AnchorRequestCarFileReader } from '../anchor-request-car-file-reader.js
 import { CASResponseOrError, ErrorResponse, AnchorRequestStatusName } from '@ceramicnetwork/codecs'
 import { decode } from 'codeco'
 import { EthereumAnchorValidator } from './ethereum-anchor-validator.js'
-import type { AnchorService, AnchorServiceAuth, AnchorValidator, AuthenticatedAnchorService } from '../anchor-service.js'
+import type {
+  AnchorService,
+  AnchorServiceAuth,
+  AnchorValidator,
+  AuthenticatedAnchorService,
+} from '../anchor-service.js'
+import { CasConnectionError, MaxAnchorPollingError } from '../anchor-service.js'
+import { RemoteCAS } from './remote-cas.js'
 
 const DEFAULT_POLL_INTERVAL = 60_000 // 60 seconds
 const MAX_POLL_TIME = 86_400_000 // 24 hours
-
-class MultipleChainsError extends Error {
-  constructor() {
-    super(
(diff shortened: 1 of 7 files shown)


packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -142,12 +123,8 @@ export class EthereumAnchorService implements AnchorService {
   }
 
   async init(): Promise<void> {
-    // Get the chainIds supported by our anchor service
-    const response = await this.#sendRequest(this.#chainIdApiEndpoint)
-    if (response.supportedChains.length > 1) {
-      throw new MultipleChainsError()
-    }
-    this.#chainId = response.supportedChains[0]
+    const supportedChains = await this.#cas.supportedChains()
```
dbcfd, line 126: Should we handle no chains being returned here, or assume chains will always be returned?

dbcfd (approved):

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
ukstv: There was previously an assumption, that the chains are always returned. Decided not to fix it to focus on the bigger thing. TBH, it is better handled by explicitly parsing the response, and trigger an error if the response is malformed.  Created a ticket to fix it later.

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -25,33 +25,18 @@ import { AnchorRequestCarFileReader } from '../anchor-request-car-file-reader.js
...
-}
-
-class MaxAnchorPollingError extends Error {
-  constructor() {
-    super('Exceeded max anchor polling time limit')
-  }
-}
-
 /**
  * Parse JSON that CAS returns.
```
stbrody, line 41: Can we dedupe the `parseResponse` in this file with the one in `remote-cas.ts`?

packages/core/src/anchor/ethereum/ethereum-anchor-service.ts
```diff
@@ -203,35 +180,7 @@ export class EthereumAnchorService implements AnchorService {
...
-    } else {
-      sendRequest$ = sendRequest$.pipe(
-        catchError((error) => {
-          throw new CasConnectionError(carFileReader.streamId, carFileReader.tip, error.message)
-        })
-      )
-    }
-
+    const sendRequest$ = this.#cas.create$(carFileReader, shouldRetry)
     return sendRequest$.pipe(this._parseResponse(carFileReader.streamId, carFileReader.tip))
```
stbrody, line 184: should we push this part down into `RemoteCAS.create$`?

stbrody (commented):
LGTM with a request to de-dupe some code in a future PR

PR #2981, opened 2023-10-10
feat(stream-model-handler): Support model interfaces creation

PaulLeCam:
Builds on https://github.com/ceramicnetwork/js-ceramic/pull/2972
Checks models correctly implement the interfaces they declare with matching or more strict schema constraints.


packages/stream-model-handler/src/interfaces-utils.ts
```diff
@@ -0,0 +1,641 @@
+import type { Context } from '@ceramicnetwork/common'
```
dbcfd, line 1: This looks like a lot of code to do schema comparison. It seems like we would do better to store interfaces as essentially a meta schema, and just use schema validation.

dbcfd (changes requested):

packages/stream-model-handler/src/interfaces-utils.ts
PaulLeCam: Interfaces mostly work like regular models, so storing them as meta-schemas would likely require significant changes to multiple parts of the stack that would increase the scope of this project.

dbcfd (approved):

PR #2985, opened 2023-10-11
chore: Remove StateManager

stbrody:
The last remaining functionality of StateManager gets folded up into Repository.


packages/core/src/state-management/repository.ts
```diff
@@ -318,18 +303,46 @@ export class Repository {
...
    *
    * @param state$ - Running State
    * @param opts - Initialization options (request anchor, publish to pubsub, etc.)
+   * @param opType - If we load, create or update a stream
    * @private
    */
   async applyWriteOpts(state$: RunningState, opts: CreateOpts | UpdateOpts, opType: OperationType) {
-    await this.stateManager.applyWriteOpts(state$, opts, opType)
+    const anchor = (opts as any).anchor
+    const publish = (opts as any).publish
```
PaulLeCam, line 339: Do `opts` really need to be cast to `any`? Seems the properties should be property typed, no?

PaulLeCam (approved):

packages/core/src/state-management/repository.ts
stbrody: yep, seems fine to remove.  will do in next PR

PR #2986, opened 2023-10-11
chore: Move load functionality out of RepositoryInternals

stbrody:
Builds on https://github.com/ceramicnetwork/js-ceramic/pull/2985

Step one of killing RepositoryInternals

### DIFF
--- a/packages/core/src/state-management/repository.ts
+++ b/packages/core/src/state-management/repository.ts
@@ -116,155 +116,6 @@ export class Repository {
   _internals: RepositoryInternals
 
   /**
-   * A collection of private helper functions used by the stream loading code
-   * @private
-   */
-  readonly _loadHelpers = new (class {
-    constructor(readonly parentRepository: Repository) {}
-
-    _fromMemory(streamId: StreamID): RunningState | undefined {
-      const state = this.parentRepository.inmemory.get(streamId.toString())
-      if (state) {
-        Metrics.count(CACHE_HIT_MEMORY, 1)
-      }
-      return state
-    }
-
-    async _fromStreamStateStore(streamId: StreamID): Promise<RunningState | undefined> {
-      const streamState = await this.parentRepository.pinStore.stateStore.load(streamId)
-      if (streamState) {
-        Metrics.count(CACHE_HIT_LOCAL, 1)
-        const runningState = new RunningState(streamState, true)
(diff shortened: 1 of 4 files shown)


packages/core/src/state-management/repository.ts
```diff
@@ -94,11 +102,168 @@ export class Repository {
...
   /**
    * Internal APIs
    */
   _internals: RepositoryInternals
 
+  /**
+   * A collection of private helper functions used by the stream loading code
+   * @private
+   */
+  readonly #loadHelpers = new (class {
```
stbrody, line 122: I'm curious what yall think about this approach.  I want to remove RepositoryInternals and fold its functionality up into Repository.  But that makes Repository very large.  All of this functionality is fairly tightly focused around stream loading, and seems like it should be possible to pull out into a self-contained module, but it isn't really actually possible to put into an entirely separate component without making a lot of the internal members of Repository public, since a lot of these functions need access to members of Repository.  Conceptually this functionality really is a part of Repository - managing the caches, state store, and network access is one of the core jobs of the Repository in my mind.  So this approach let this remain the domain of the Repository, but still have some sort of logical grouping and separation of this functionality from the rest of the public API of Repository.  It's not my favorite, but was the best I could come up with.  The alternative is just remove this subclass and put all of these functions as regular methods on Repository.

Definitely open to input or ideas here.

stbrody:
FYI @ukstv, the next step is moving the anchoring-related functionality out of RepositoryInternals.  Please let me know if you have thoughts on where it should be moved to

packages/core/src/state-management/repository.ts
PaulLeCam: If the functionality is part of the Repository, I think it's fine just having these as private methods.
The logic here seems more like a hack to me, considering there are circular references between the Repository and these internal helpers it doesn't seem to add any value over simply having these methods directly on the Repository class?

stbrody: Okay done.  I don't love how big the Repository is getting, but I'd still rather have that then the arbitrary and circular split between Repository/StateManager/RepositoryInternals.  Hopefully we can figure out some good abstractions to break Repository up at some point in the future

stbrody:
Closing in favor of https://github.com/ceramicnetwork/js-ceramic/pull/2987 which is more comprehensive

PR #2996, opened 2023-10-16
feat: Remove InternalOpts

stbrody:


packages/core/src/state-management/repository.ts
```diff
@@ -445,14 +447,10 @@ export class Repository {
...
-  async applyCommit(
-    streamId: StreamID,
-    commit: any,
-    opts: CreateOpts | UpdateOpts
-  ): Promise<RunningState> {
+  async applyCommit(streamId: StreamID, commit: any, opts: UpdateOpts): Promise<RunningState> {
     this.logger.verbose(`Repository apply commit to stream ${streamId.toString()}`)
 
-    const state$ = await this.load(streamId, opts)
+    const state$ = await this.load(streamId)
```
stbrody, line 453: there's technically some small amount of risk from this change as it is technically a behavior change.  But I don't see any cases in our http-client or any of our stream updating libraries where we would actually pass any LoadOpts as part of a stream update, and I can't really imagine any real use case for doing so.  I think it was kind of a quirk/bug of the way the types were structured before that this was possible at all.

But if either of you are really worried about it I can put this back for now

PaulLeCam (approved):
Nice!

PR #3001, opened 2023-10-17
feat: Add admin-DID protected HTTP endpoint for shutting down a node

stbrody:
tested manually

### DIFF
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -831,11 +831,11 @@ export class CeramicDaemon {
             onSuccess = () => successCallback.method(force)
             break
           }
-          // default:
-          //   throw new UnreachableCaseError(
-          //     successCallback.type,
-          //     'Unsupported admin API success callback type'
-          //   )
+          default:
+            throw new UnreachableCaseError(
+              callbackType,
+              'Unsupported admin API success callback type'
+            )
         }
       }
       return {
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -15,6 +15,7 @@ import {
   MultiQuery,
   StreamUtils,
   SyncOptions,
+  UnreachableCaseError,
 } from '@ceramicnetwork/common'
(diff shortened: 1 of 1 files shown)


packages/cli/src/ceramic-daemon.ts
```diff
@@ -796,6 +826,16 @@ export class CeramicDaemon {
             onSuccess = () => successCallback.method(models)
             break
           }
+          case AdminApiSuccessCallbackType.shutdownServer: {
+            const force = parsedJWS.force
+            onSuccess = () => successCallback.method(force)
+            break
+          }
+          // default:
```
stbrody, line 834: @PaulLeCam @ukstv any idea why whenever I uncomment this it fails to compile?  The error seems to indicate it thinks  I haven't exhaustively covered every case in the switch statement, but I totally have...

 `TS2345: Argument of type 'AdminApiSuccessCallbackType' is not assignable to parameter of type 'never'.   Type 'AdminApiSuccessCallbackType' is not assignable to type 'never'.`

nathanielc (approved):

packages/cli/src/ceramic-daemon.ts
PaulLeCam: I think it's the opposite, TS detects the `default` case can "never" be reached because the other conditions already cover all the possible cases, so `default` case is unreachable code.
In practice I think it's safer to keep the runtime check, we can probably ignore the TS error here.

stbrody: Except that we don't get this TS compilation error in any of the other places we use UnreachableCaseError.  For example: https://github.com/ceramicnetwork/js-ceramic/blob/18bb1ce3c61bd05a697099d70d367b3b9df0a80d/packages/core/src/state-management/repository.ts#L251

ukstv: TS could be stupid sometimes. Here it fails to do exhaustiveness check because the first parameter to `UnreachableCaseError` differs from what is tested in `switch`.
Changing it to:
```
default:
            throw new UnreachableCaseError(
              callbackType,
              'Unsupported admin API success callback type'
            )
```
makes it work.

PR #3002, opened 2023-10-18
chore: Make WitnessCAR optional again

ukstv:
We still get requests for old anchors.


stbrody (approved):
😭 It's driving me crazy that we can't figure out why nodes are still polling for such old anchor requests.

In any case LGTM to address the current user error

stbrody:
Merging so this can get into the RC I'm about to make

PR #3007, opened 2023-10-19
chore(common): cat-2800 remove lowercase http method support

dbcfd:


dbcfd:
@ukstv @stephhuynh18 Only other instances of lower case methods were found in test code, dealing with external services.

ukstv (approved):

PR #3012, opened 2023-10-23
Patch for http-client URL constructor 

mzkrasner:
## Description

This patch includes a single-line change to the client URL constructor, allowing for a full URL extension instead of host isolation.

## How Has This Been Tested?

- with-go-ipfs - passes
- dispatch-event - passes
- build-and-test - passes
- image build - passes

## References:

This relates to this forum post: https://forum.ceramic.network/t/configuring-composeclient-with-a-ceramic-url-string-that-contains-a-prefix/1307/6


packages/http-client/src/ceramic-http-client.ts
```diff
@@ -78,7 +78,7 @@ export class CeramicClient implements CeramicApi {
   constructor(apiHost: string = CERAMIC_HOST, config: Partial<CeramicClientConfig> = {}) {
     this._config = { ...DEFAULT_CLIENT_CONFIG, ...config }
 
-    this._apiUrl = new URL(API_PATH, apiHost)
+    this._apiUrl = new URL(apiHost.replace(/\/$/, '') + API_PATH)
```
stbrody, line 81: I personally find regexes pretty inscrutable.  Can you maybe add a constant for this regex with a comment on what it does?

stbrody: This also feels like the kind of thing that the URL class should be able to handle and abstract away for us without us needing to do raw string manipulation.  For one thing I'd be worried about how cross-platform the string-manipulation-based approach will be - have you tested this on windows?

stbrody (changes requested):
Can we please add a unit test for this?  You should be able to mock out the "fetch" function and then test calling different functions on the http client with different input urls and then test what url the client tries to actually hit.  You can probably put the new tests into ceramic-http-client.test.ts, but check out remote-admin-api.test.ts to see how you can mock fetch and test what urls the client tries to hit

packages/http-client/src/ceramic-http-client.ts
ukstv: `apiHost` could be provided with a slash or without a slash, or maybe with multiple slashes at the end (highly unlikely). The patch handles the first two cases by effectively replacing the single slash. Then `API_PATH` provides a slash symbol. That works.

A single thing really missing here to my taste is a unit test that goes through scenarios of zero, one slashes.

ukstv (dismissed):
Unit tests, please. Otherwise fine by me.

packages/http-client/src/ceramic-http-client.ts
mzkrasner: Just circling back to this - thank you @stbrody and @ukstv for your feedback! I'll start by adding in the unit tests. As for the regex - are we fine with this?

mzkrasner: @ukstv  - went in a slightly different direction compared to the original patch. Let me know if this looks good to you!

packages/http-client/src/ceramic-http-client.ts
```diff
@@ -78,6 +80,7 @@ export class CeramicClient implements CeramicApi {
   constructor(apiHost: string = CERAMIC_HOST, config: Partial<CeramicClientConfig> = {}) {
     this._config = { ...DEFAULT_CLIENT_CONFIG, ...config }
 
+    // API_PATCH contains leading dot-slash, so preserves the full path
```
stbrody, line 83:
```suggestion
    // API_PATH contains leading dot-slash, so preserves the full path
```

packages/http-client/src/__tests__/ceramic-http-client.test.ts
```diff
@@ -15,11 +30,81 @@ const initial = {
...
+let did: DID
+let getDidFn
+let client: CeramicClient
+
+beforeEach(() => {
+  jest.resetAllMocks()
+})
+
 describe('buildStreamFromState', () => {
   test('build instance of Streamtype', async () => {
```
stbrody, line 47: this test feels redundant now of the 	`build instances of Streamtype from URLs with extensions` test you added below.  You can probably just delete this test (and the API_URL constant which then won't be used anymore)

packages/http-client/src/__tests__/ceramic-http-client.test.ts
```diff
@@ -15,11 +30,81 @@ const initial = {
...
 } as unknown as StreamState
 
+const GET_RESPONSE = {
+  ...initial,
+  state: initial,
+}
+
+let did: DID
+let getDidFn
+let client: CeramicClient
```
stbrody, line 40: any reason this needs to be a global?  I think you can probably define it locally-scoped within every test

packages/http-client/src/__tests__/ceramic-http-client.test.ts
```diff
@@ -15,11 +30,81 @@ const initial = {
...
 })
+
+describe('URL constructor', () => {
+  beforeAll(async () => {
+    const seed = randomBytes(32)
+    const provider = new Ed25519Provider(seed)
+    const actingDid = new DID({ provider, resolver: KeyResolver.getResolver() })
+    await actingDid.authenticate()
+    did = actingDid
+    getDidFn = () => {
```
stbrody, line 62: `did` is never used in any of the tests, and `getDidFn` is only used in one test, the `setDID` test.  Instead of putting this into the test fixture setup, could you put this into a helper function that just returns a DID and use that function only in the `setDID` test?

stbrody (commented):
look great!  A few small comments but they're all pretty nitpicky to be honest, just some small code cleanup to the test.

stbrody (approved):
LGTM!

PR #3014, opened 2023-10-23
chore(3id-did-resolver, common, core, stream-tests): Remove deprecated multiquery atTime

stephhuynh18:


PaulLeCam (approved):

packages/stream-tests/src/__tests__/multiquery.test.ts
```diff
@@ -54,66 +54,6 @@ describe('multiquery API http-client tests', () => {
     expect(Math.abs(expectedTimestamp - givenTimestamp)).toBeLessThan(5)
   }
 
-  it('loads the same stream at multiple points in time using atTime', async () => {
```
oed, line 57: Shouldn't we just change this test to use `opts.atTime`? Would be good to know that this still works as 3ID relies on being able to query the same stream at multiple timestamps.

oed (changes requested):

packages/stream-tests/src/__tests__/multiquery.test.ts
stephhuynh18: there is already a tests for this right under this test. It is called "loads the same stream at multiple points in time using opts.atTime"

oed: Ah ok, my bad!

oed (approved):

PR #3016, opened 2023-10-26
feat(3id-did-resolver, blockchain-utils-validation, cli, common, ipfs-topology, pinning-crust-backend, stream-tests): Remove cross-fetch

stephhuynh18:
The biggest test change was to common's fetch-json. I decided not to mock anything, as we aren't testing native fetch if we do so. Instead I spin up a server.


ukstv (approved):
Woah! Nicety!

PR #3017, opened 2023-10-27
chore: Add npm script for running prettier to format code

stbrody:
also runs it and corrects some formatting that went in in https://github.com/ceramicnetwork/js-ceramic/commit/e6bbb11fe124d71ddfc5aabb73bf260a437ee707


stbrody:
if you just copy the "format" script and run it against current develop, does anything change for you @nathanielc?  For me it made all these changes

package.json
```diff
@@ -19,6 +19,7 @@
     "publish:hotfix": "npx lerna publish --dist-tag latest --no-verify-access --conventional-graduate bump patch",
     "format": "npx prettier -w packages/**/src/**/*.ts",
```
nathanielc, line 20: If I run this script I do not get a diff. Turns out the single quotes around the pattern match are critical. I wouldn't expect npm run scripts to do glob argument expansion but I guess they do.

nathanielc (approved):
When I run `./node_modules/.bin/prettier --write 'packages/**/src/**/*{.ts,.tsx,.js}'` verbatum I get the same diff you got for this PR. 


These formatting changes look good too. Must have been the single quote issue that was preventing the formatting from getting fixed. Thanks for looking into this.

package.json
stbrody: oh wow how'd I miss that there was already a format script. Updated the PR to remove the existing non-working version

PR #3018, opened 2023-10-28
feat (CAT-2791): Worker Threads

dbcfd:
Use worker threads for signing and verifying.


packages/http-client/src/ceramic-http-client.ts
```diff
@@ -99,16 +101,18 @@ export class CeramicClient implements CeramicApi {
     }
   }
 
-  get did(): DID | undefined {
+  get didVerifier(): DidVerifier | undefined {
+    return this.context.didVerifier
+  }
+
+  get did(): ThreadedDid | undefined {
```
stbrody, line 108: I don't think we want to change the http client like this.  This has implications for app devs and is a breaking change to our API.  It also now requires app developers to provide a ThreadedDid just to sign the messages for creating commits, which is probably overkill added complexity they don't need.

I think if we need a ThreadedDid in the `Ceramic` object in the `core` package, we need to put it in a place that isn't part of the shared contract with the http client

packages/common/src/ceramic-api.ts
```diff
@@ -75,7 +74,8 @@ interface CeramicCommon {
  * available users can provide any object containing an authenticated DID instance.
  */
 export interface CeramicSigner extends CeramicCommon {
-  did: DID | undefined
+  did: ThreadedDid | undefined
+  didVerifier: DidVerifier | undefined
```
stbrody, line 78: I'm guessing a `DidVerifier` is used to verify did signatures?  The http-client doesn't need to ever verify DID signatures, it only needs to create them, so this probably shouldn't be part of the common interface that is shared between the core and http client implementations

packages/stream-model-instance/src/model-instance-document.ts
```diff
@@ -306,7 +306,7 @@ export class ModelInstanceDocument<T = Record<string, any>> extends Stream {
         await _ensureAuthenticated(signer)
         // When did has a parent, it has a capability, and the did issuer (parent) of the capability
         // is the stream controller
-        controller = signer.did.hasParent ? signer.did.parent : signer.did.id
+        controller = signer.did.did.hasParent ? signer.did.did.parent : signer.did.did.id
```
stbrody, line 309: I'm not a big fan of this `did.did` thing.  Can we make the names clearer or in same way structure this so that it's clear what the difference between these two objects is?

packages/http-client/src/ceramic-http-client.ts
dbcfd: We could change the http client to not inherit from Ceramic, and instead just have it as a dependency.

packages/common/src/ceramic-api.ts
dbcfd: I need a better name for that. It's responsible for managing the threaded did instances, so that data can flow properly. 

It is needed, since if someone does a late authentication for a did, the caching needs to be updated.

packages/stream-model-instance/src/model-instance-document.ts
dbcfd: Yeah, we can expose id, parent, and hasParent in ThreadedDid.

packages/common/src/ceramic-api.ts
stbrody: It's needed in the http-client?  Why?  It doesn't use it now and seems to be fine, we don't need to use threads in the client.

packages/http-client/src/ceramic-http-client.ts
stbrody: I'm definitely interested in breaking up the http client from core and allowing them to diverge more.

But I don't think we need to do that just to deliver this feature?  We could just take the `did` field out of the common interface and let each implementation define it how it needs.

PR #3019, opened 2023-10-30
Add quickstart doc links

gvelez17:
## Description

A developer should be able to start and upgrade a node in < 5 min

Also DX usability testers need quick documentation to complete their tasks in < 90 min

see https://linear.app/3boxlabs/issue/MET-1622/quickstart-and-upgrade-readmes

## How Has This Been Tested?

- [ ] Run manually

## PR checklist

- [ x] I have made corresponding changes to the documentation

## References:


docs-dev/UPGRADING.md
```diff
@@ -0,0 +1,25 @@
...
+Note that although this references the CLI package, the dependencies will cause the core and other server packages to update.
+
+3) restart the ceramic daemon
+
+Depending on how you have configured your server, this may involve a command such as
+
+`systemctl stop ceramic && systemctl start ceramic`
+
+Note this presumes that you have used systemctl startup scripts
+
```
gvelez17, line 24: hey @kammerdiener can we get a copy of one of your systemctl startup scripts or where did those come from? 

 @3benbox what would be the one line command to upgrade the digitalocean kubernetes based install?  will it automatically upgrade like every time it restarts or?

docs-dev/QUICKSTART.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Configuration
+
+The default configuration file is located at 
+
+`$HOME/.ceramic/daemon.config.json`
+
+For more information about the configuration options see [Server Configurations](https://composedb.js.org/docs/0.5.x/guides/composedb-server/server-configurations)
+
+For an example how to run in the cloud using Digital Ocean and Kubernetes see [Running in the Cloud](https://composedb.js.org/docs/0.4.x/guides/composedb-server/running-in-the-cloud)
```
JustinaPetr, line 28: I seems like we are pointing to the 0.5x version of the ComposeDB here. Can we use the latest on here and point to the 0.5.x just like int the link right above this one?

JustinaPetr (changes requested):

docs-dev/UPGRADING.md
gvelez17: cool would it work with `npx @ceramicnetwork/cli daemon` without the version pin?  That way they could update without editing the script?

is npx installed with node or its a separate install?  ha sorry for lazy questions but just wanted to give docs for how to set up like you have it on hirenodes!

Maybe we should mention the ulimit too now that i think of it - that might be a separate doc for server-config-example ...

docs-dev/QUICKSTART.md
gvelez17: fixed!

docs-dev/UPGRADING.md
3benbox: When running with systemd, we should not use the root account. Any security issues with the ports we expose could escalate to a host compromise.

My ugly unit file is templated with Ansible in the simpledeploy repo
https://github.com/ceramicstudio/simpledeploy/blob/main/ansible/roles/ceramic/templates/ceramic.service.j2

gvelez17: ok cool - linked them to SimpleDeploy and pointed out that our upgrade example is just a simplified example.

gvelez17:
hey @JustinaPetr i think we have addressed the changes now!

JustinaPetr (approved):

PR #3020, opened 2023-10-30
feat(cli): deprecate gateway flag and add read only flag

stephhuynh18:


packages/cli/src/ceramic-cli-utils.ts
```diff
@@ -210,8 +213,8 @@ export class CeramicCliUtils {
       if (pubsubTopic) {
         config.network.pubsubTopic = pubsubTopic
       }
-      if (gateway) {
-        config.node.gateway = gateway
+      if (gateway || readOnly) {
+        config.node.gateway = gateway || readOnly
```
stbrody, line 217: can we rename the value within ceramic core too while we're at it?

packages/cli/src/bin/ceramic.ts
```diff
@@ -24,7 +24,10 @@ program
     '--state-store-s3-bucket <string>',
     `The S3 bucket name to use for state store. If not provided data will only be saved locally but not to S3. Deprecated.`
   )
-  .option('--gateway', 'Makes read only endpoints available. It is disabled by default')
+  .option(
```
stbrody, line 27: Given that this is a breaking release, do you think we need to keep `--gateway` around at all?

packages/cli/src/bin/ceramic.ts
```diff
@@ -48,6 +51,7 @@ program
     'Global forced mode for syncing all streams. One of: "prefer-cache", "sync-always", or "never-sync". Defaults to "prefer-cache". Deprecated.'
   )
   .option('--disable-composedb', 'Run node without ComposeDB indexing enabled.')
+  .option('--read-only', 'Makes read only endpoints available. It is disabled by default')
```
stbrody, line 54:
```suggestion
  .option('--read-only', 'Disables http endpoints that perform writes, making the node effective for reading data from the network but not for writing to it.  It is disabled by default')
```

stbrody (changes requested):

packages/cli/src/ceramic-cli-utils.ts
stephhuynh18: gotcha

packages/cli/src/bin/ceramic.ts
stephhuynh18: I'm in favour of removing it

stbrody (approved):
lgtm

PR #3021, opened 2023-10-31
feat: Remove experimental DID resolvers for did:nft and did:safe

stbrody:


packages/common/src/ceramic-api.ts
```diff
@@ -6,7 +6,7 @@ import type { LoggerProvider } from './logger-provider.js'
 import type { GenesisCommit } from './index.js'
 import type { IndexApi } from './index-api.js'
 import { NodeStatusResponse } from './node-status-interface.js'
-import type { AnchorStatus } from '@ceramicnetwork/common'
```
stbrody, line 9: I'm not really sure why removing the other resolvers from the cli package caused these imports to suddenly start failing. But it was probably always a mistake to import from `@ceramicnetwork/common` from NPM from within the common package itself, so this is a good fix to make even if it's not clear why it suddenly became necessary

avious00:
i miss did:nft already

dbcfd (approved):

oed (approved):

PR #3024, opened 2023-11-02
Update DEVELOPMENT file

JulissaDantes:
## Description

When cloning and setting up the project for the first time I noticed the uint tests were failing even after following all the `Project setup` instructions, turns out I needed to spin up a Docker container in order for the uint tests to be succesful.

In this PR I'm including the directions to ensure all tests run as they should, and also the commands to run, in case the user is not familiar with Docker terminology.

Is it worth mentioning that the `DEVELOPMENT_LOCAL` does mention how to build the image and run the container, but is use in the context of running your own local ceramic instance, and from that it is not clear this is an actual requirement when running the test.

## How Has This Been Tested?

N/A

## PR checklist

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -2,7 +2,7 @@
 > Getting started with development on js-ceramic
 
 ## Prerequisites
-This project requires Docker to be installed and actively running in your development environment.
+This project requires Docker to be installed and actively running in your development environment for the unit tests to run.
 
 ## Project setup
 First clone the repo:
@@ -31,10 +31,6 @@ npx prettier --write "packages/**/src/**/*.ts"
 First build a docker image of the project, 
 ```
 docker build -t js-ceramic -f Dockerfile.daemon .
-```
-Run the docker container,
-```
-docker run -it --rm js-ceramic
 ```
 Then you can run all tests at the top level,
 ```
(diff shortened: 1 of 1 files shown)


JulissaDantes:
@JustinaPetr This is the same as #3022 but that one was from a forked repo.

DEVELOPMENT.md
```diff
@@ -25,7 +28,15 @@ npx prettier --write "packages/**/src/**/*.ts"
 ```
 
 ## Run tests
-You can run all tests at the top level,
+First build a docker image of the project,
```
stbrody, line 31: hmm, that's interesting, I've never actually needed to do this.  I do need to have docker installed and running, since the tests use pg-test, which in turn uses docker to spin up a postgres node.  But I haven't needed to manually run any docker images to run the tests locally, just having docker running was enough.

DEVELOPMENT.md
```diff
@@ -1,6 +1,9 @@
 # Development
 > Getting started with development on js-ceramic
 
+## Prerequisites
+This project requires Docker to be installed and actively running in your development environment.
```
stbrody, line 5: might want to clarify that this is only necessary for running the tests locally

DEVELOPMENT.md
JulissaDantes: You are right, just made the update

stbrody (approved):
lgtm, thanks!

PR #3029, opened 2023-11-06
fix(ci): update rust-ceramic repo path

smrz2001:
Update the `rust-ceramic` release path to point to the [new location](https://github.com/ceramicnetwork/rust-ceramic).


stbrody (approved):
lgtm.  I'm going to revert the last commit that broke the build in a different way

dbcfd (approved):

PR #3030, opened 2023-11-06
Revert "Use v2 codec for relations definition validation (#3026)"

stbrody:
Unbreak the build


stbrody:
This commit doesn't seem to actually be the culprit of the build breakage, it seems to be caused by an update to one of our dependencies, likely `pg` or `pg-boss`

PR #3036, opened 2023-11-10
chore(core): Add logging when creating new leveldb store

stephhuynh18:
## Description

* Noticed that the anchor request store wasn't being closed, its a no op for this store, but can be for other types of stores.
* This is not a solution but can provide better insight into whats happening with little impact. I have a solution on deck, but I'm not too fond of it as the packages it needs are "deprecated". 
* Should we be upgrading our level packages (level-ts uses deprecated level packages) 

## How Has This Been Tested?

Unit tests are running

### DIFF
--- a/packages/core/src/store/level-db-store.ts
+++ b/packages/core/src/store/level-db-store.ts
@@ -31,7 +31,7 @@ class LevelDBStoreMap {
   readonly networkName
   readonly #map: Map<string, Level>
 
-  constructor(storeRoot: string, networkName: string) {
+  constructor(storeRoot: string, networkName: string, readonly logger: DiagnosticsLogger) {
     this.networkName = networkName
     this.#storeRoot = storeRoot
     this.#map = new Map<string, Level>()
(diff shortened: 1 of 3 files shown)


packages/core/src/store/level-db-store.ts
```diff
@@ -45,7 +45,18 @@ class LevelDBStoreMap {
...
       fs.mkdirSync(storePath, { recursive: true }) // create dir if it doesn't exist
     }
-    this.#map.set(fullLocation, new LevelC(storePath))
+
+    const levelDb = new LevelC(storePath)
+    // level-ts does not have an error handling exposed for opening errors. I access the private DB variable to add callbacks for logging.
+    // @ts-ignore private field
+    levelDb.DB.on('error', (err) => {
+      this.logger.warn(
+        `Received error when starting up leveldb at ${storePath} using level-ts: ${err}`
```
dbcfd, line 54: Will error only occur on startup, or could this occur at any time, and should be indicated as a runtime error?

stephhuynh18: I want to say only on startup when i look at the libraries, but I'm not 100% on that. From my crude tests, I haven't seen this get triggered outside of startup errors. Level-ts has "Promise forcing" so there should be no callback sused and therefore error events should not be used.

dbcfd (approved):

stbrody (approved):

PR #3039, opened 2023-11-13
chore(cli,common,http-client,indexing): remove legacy query fields

PaulLeCam:
Makes the `models` field required instead of having both `model` and `models` optional, and removes the `filters` field.


stbrody (changes requested):
Given that we only *just* added support for "models" to the http client, I am not in favor of merging this PR at this time, since it would require all apps to upgrade their client in their apps at the same time as they upgrade their node or else their apps would break.

Ultimately though this is a decision for product.  @jpham2023

PaulLeCam:
Closing to keep supporting supporting 2.x clients as discussed in the chat.

PR #3041, opened 2023-11-15
feat(docs): restore workflow for publishing docs

smrz2001:


stbrody (approved):
lgtm assuming the workflow is passing!

smrz2001:
> lgtm assuming the workflow is passing!

Yep! I made sure it passed fully in the PR.

PR #3042, opened 2023-11-17
fix: Use envelope not payload when building StreamID for applying genesis commits

stbrody:
This mostly manifests as logging the wrong StreamID.
In theory, the StreamID we pass to `SignatureUtils.verifyCommitSignature` can matter, because it's possible to issue a CACAO that is authorized only to write to a specific StreamID.  This definitely matters for updates.  But this PR only changes code for applying genesisCommits.  It's pretty hard to create a CACAO that will let you only create a specific stream with a specific StreamID, because you usually don't know the StreamID until after the Stream is created.  So practically speaking its very hard for this to ever be an actual issue.  In theory if you were using a deterministic stream you could predict the StreamID before creating it and then issue a CACAO that allows someone to create it.  But that's a pretty niche and unlikely use case.  Also, while CACAOs in theory can be specified to grant access to only a single StreamID, I don't think the did-session library actually exposes this very easily (it mostly exposes granting access to specific models).  So it's extremely unlikely this bug has actually affected any real users.

### DIFF
--- a/packages/stream-model-handler/src/model-handler.ts
+++ b/packages/stream-model-handler/src/model-handler.ts
@@ -106,7 +106,7 @@ export class ModelHandler implements StreamHandler<Model> {
       throw new Error('Exactly one controller must be specified')
     }
 
-    const streamId = await StreamID.fromGenesis('model', commitData.commit)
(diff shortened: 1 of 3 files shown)


smrz2001 (approved):
🚀

PR #3043, opened 2023-11-17
fix: initialize s3 state store after setting up did

smrz2001:
The S3 state store was being initialized before the Ceramic DID was set up. Normally, this worked fine but if the state store path was changed and indexed model streams previously accessible were no longer available in the store, the node would attempt to load them from the network. Without the DID set up correctly, this would cause the startup process to fail.


dbcfd (approved):

nathanielc (approved):
LGTM

PR #3045, opened 2023-11-20
fix: Don't initialize Repository before Ceramic

stbrody:


packages/core/src/state-management/repository.ts
```diff
@@ -148,12 +148,16 @@ export class Repository {
     this.updates$ = this.updates$.bind(this)
   }
 
+  /**
+   * Sets the StateStore to sue.
```
smrz2001, line 152:
```suggestion
   * Sets the StateStore to use.
```
😅

smrz2001 (approved):
LGTM! CI and all unit tests also passed so the `init` must not have been necessary there.

I'm trying to recreate the issue with the old code and then see what your fix changes. Will keep you posted.

smrz2001:
> LGTM! CI and all unit tests also passed so the `init` must not have been necessary there.
> 
> I'm trying to recreate the issue with the old code and then see what your fix changes. Will keep you posted.

FYI, taking the `await this.init()` call out from the old code worked without any extra changes.

PR #3046, opened 2023-11-21
Check if ELP bucket exists for S3 state store

PaulLeCam:
From my understanding of the ticket this is the kind of logic required, though I'm wondering about the following:
- Is the `storeRoot` path logic correct?
  - If the `s3Bucket` option is not provided, is there a default value?  
  - Is `elp` the right name for the network?
- Is the `storeRoot` value the right one to provide for the S3 `Bucket` check? 
- Is the `S3Store` internal logic OK if I replace the network name by `elp` or are there other changes needed there?
- Is there any way to test the behavior in unit tests or manually locally?

Thanks!

### DIFF
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -22,7 +22,6 @@ import { StreamID } from '@ceramicnetwork/streamid'
 import * as ThreeIdResolver from '@ceramicnetwork/3id-did-resolver'
 import * as KeyDidResolver from 'key-did-resolver'
 import * as PkhDidResolver from 'pkh-did-resolver'
-import AWSSDK from 'aws-sdk'
 
 import { DID, DIDOptions, DIDProvider } from 'dids'
 import cors from 'cors'
@@ -369,6 +368,7 @@ export class CeramicDaemon {
         opts.stateStore?.s3Bucket,
         opts.stateStore?.s3Endpoint
       )
+
       await ceramic.repository.injectKeyValueStore(s3Store)
     }
 
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -22,6 +22,7 @@ import { StreamID } from '@ceramicnetwork/streamid'
 import * as ThreeIdResolver from '@ceramicnetwork/3id-did-resolver'
 import * as KeyDidResolver from 'key-did-resolver'
 import * as PkhDidResolver from 'pkh-did-resolver'
+import AWSSDK from 'aws-sdk'
 
 import { DID, DIDOptions, DIDProvider } from 'dids'
 import cors from 'cors'
@@ -364,8 +365,28 @@ export class CeramicDaemon {
 
     // This needs to happen after the DID is set up so that any stream operations have a valid DID to use
     if (opts.stateStore?.mode == StateStoreMode.S3) {
+      let networkName = params.networkOptions.name
+
+      // Check if ELP bucket is used
+      if (networkName === Networks.MAINNET) {
+        const s3 = new AWSSDK.S3()
+        const storeRoot = opts.stateStore?.s3Bucket
+          ? `${opts.stateStore?.s3Bucket}/ceramic/elp`
+          : 'ceramic/elp'
+        try {
+          await s3.headBucket({ Bucket: storeRoot }).promise()
+          // Bucket exists and needs to be used
+          console.warn(
+            `S3 bucket found with ELP location, using it instead of default mainnet location for state store`
+          )
+          networkName = 'elp' as Networks
+        } catch (error) {
+          // Ignore error
+        }
+      }
+
       const s3Store = new S3Store(
-        params.networkOptions.name,
+        networkName,
         opts.stateStore?.s3Bucket,
         opts.stateStore?.s3Endpoint
       )
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -365,32 +365,11 @@ export class CeramicDaemon {
 
     // This needs to happen after the DID is set up so that any stream operations have a valid DID to use
     if (opts.stateStore?.mode == StateStoreMode.S3) {
-      let networkName = params.networkOptions.name
-
-      // Check if ELP bucket is used
-      if (networkName === Networks.MAINNET) {
-        const s3 = new AWSSDK.S3()
-        const storeRoot = opts.stateStore?.s3Bucket
-          ? `${opts.stateStore?.s3Bucket}/ceramic/elp`
-          : 'ceramic/elp'
-        try {
-          await s3.headBucket({ Bucket: storeRoot }).promise()
-          // Bucket exists and needs to be used
-          console.warn(
-            `S3 bucket found with ELP location, using it instead of default mainnet location for state store`
-          )
-          networkName = 'elp' as Networks
-        } catch (error) {
-          // Ignore error
-        }
-      }
-
       const s3Store = new S3Store(
-        networkName,
+        params.networkOptions.name,
         opts.stateStore?.s3Bucket,
         opts.stateStore?.s3Endpoint
       )
-
       await ceramic.repository.injectKeyValueStore(s3Store)
     }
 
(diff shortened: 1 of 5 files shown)


packages/cli/src/ceramic-daemon.ts
```diff
@@ -364,8 +365,28 @@ export class CeramicDaemon {
 
     // This needs to happen after the DID is set up so that any stream operations have a valid DID to use
     if (opts.stateStore?.mode == StateStoreMode.S3) {
+      let networkName = params.networkOptions.name
+
+      // Check if ELP bucket is used
```
stbrody, line 370: I think this is generally the right idea, but I think it'd be better done inside of the S3Store's initialization instead of up here.

Also, I know this is a slight scope increase from what we originally talked about, but it looks like the behavior for leveldb just *always* uses "elp" as the storage location for state store data (https://github.com/ceramicnetwork/js-ceramic/blob/c61623e9761872070fa86768137d9286ba54cf2b/packages/core/src/store/level-db-store.ts#L65), instead of only using it if there's existing data there and preferring "mainnet" for new deploys.  Ideally both state store implementations would have the same logic for preferring "mainnet" as the location but using "elp" if data is detected there.

PaulLeCam: I looked into putting the logic in the `S3Store` class directly but it seems all the initialization logic is in the constructor and there is no async initialization method?
So I guess we can either have this async logic out of the store or change the store logic to support async initialization, but that seems like it would add complexity to handle as currently all the stores creation seems synchronous?

I think the alternative for S3 could be to add the ELP check in the `createStore()` method of `S3StoreMap` that is handled as if it was async even though it only runs sync logic? 
I wonder if we need to check the bucket existence for every `createStore()` call though, or running the check only once would be enough to detect the bucket we need to use?

For the LevelDB store I think that can work either way as we can use Node's synchronous FS APIs if needed?

stbrody: I think adding an async init() should be pretty straightforward.  There's already a place we're initializing the state store already (`Repository.init()` calls `pinStore.open()` which calls `stateStore.open()`), it just doesn't get threaded all the way down to the underlying S3Store or LevelDbStore, but it's be easy to plug them in.  Probably can just add a call to store.init() into the open method of the ObjectStore: https://github.com/ceramicnetwork/js-ceramic/blob/c61623e9761872070fa86768137d9286ba54cf2b/packages/core/src/store/object-store.ts#L29

PaulLeCam: OK thanks!
I moved the logic to the S3Store and updated the logic in the LevelDB store.

There are existing tests for the ELP/mainnet change for the LevelDB store so I think it's covered, but any idea how to test the S3Store behavior please? Is the bucket existence check correct?

packages/cli/src/s3-store.ts
```diff
@@ -75,13 +78,33 @@ export class S3Store implements IKVStore {
...
   }
 
+  async init(): Promise<void> {
+    // Check if ELP bucket is used
+    if (this.networkName === Networks.MAINNET) {
+      const s3 = new AWSSDK.S3()
+      try {
+        await s3.headBucket({ Bucket: `${this.#bucketName}/ceramic/elp` }).promise()
+        // Bucket exists and needs to be used
+        console.warn(
```
stbrody, line 97: Can we thread the DiagnosticsLogger into this class and use that instead of using the console directly?

packages/core/src/state-management/repository.ts
```diff
@@ -161,6 +161,7 @@ export class Repository {
   }
 
   async init(): Promise<void> {
+    await this.#deps.keyValueStore.init()
```
stbrody, line 164: nitpick: I'd probably just move this into `pinStore.open()`.  Not important though

packages/core/src/store/level-db-store.ts
```diff
@@ -61,23 +60,33 @@ class LevelDBStoreMap {
...
   private getFullLocation(useCaseName = DEFAULT_LEVELDB_STORE_USE_CASE_NAME): string {
-    if (useCaseName === DEFAULT_LEVELDB_STORE_USE_CASE_NAME) {
-      return this.getDefaultLocation()
-    } else {
-      return `${this.getDefaultLocation()}-${useCaseName}`
+    // Check if store exists at legacy ELP location
+    if (this.networkName === Networks.MAINNET) {
+      const elpLocation = this.getStoreLocation(useCaseName, OLD_ELP_DEFAULT_LOCATION)
+      const storePath = path.join(this.#storeRoot, elpLocation)
+      if (fs.existsSync(storePath)) {
```
stbrody, line 74: I believe as written this will wind up running this check every time the leveldb state store is accessed, which will both be inefficient and flood the logs with the warning message.  Can we do this check just once at startup and remember the results?

packages/core/src/store/level-db-store.ts
```diff
@@ -61,23 +60,33 @@ class LevelDBStoreMap {
...
+        console.warn(
```
stbrody, line 75: Can we use the DiagnosticsLogger please?

stbrody (changes requested):
I'm not really sure how to write automated tests of the S3 logic.  Maybe there's some js library you can find that provides a mock S3?  Otherwise we might have be content with testing this manually.  @smrz2001 can likely provide guidance on how to set up the S3-related env vars so you can test this manually

packages/core/src/state-management/repository.ts
PaulLeCam: No problem changing this but the `anchorRequestStore` below also uses the `keyValueStore`, so it seems a bit fragile to me to rely on the `pinStore` running the initialization for both?

stbrody: oh good call.  Yeah I didn't realize that the PinStore doesn't fully encapsulate the StateStore anymore.  Leaving this here then does indeed seems better

stbrody (approved):
everything looks good, assuming manual testing works!  I have no idea if the S3 code for checking for the folder is right or not, so we definitely want to make sure this is tested before being merged.  But the general structure and approach lgtm

PaulLeCam:
I updated and checked the S3 bucket logic as follows:
1. run a Ceramic v2 deamon with `elp` network in a new S3 bucket
2. check a `ceramic/elp` object has been added the S3 bucket
3. run the Ceramic daemon with the code from this branch using `mainnet`
4. check warning message about using the ELP bucket is logged
5. delete the `ceramic/elp` object from S3
6. restart the Ceramic daemon - no warning about using the ELP bucket
7. check the `ceramic/mainnet` object is used in S3

@stbrody @smrz2001 does this look OK to you or is there any additional check you could think of please?

packages/cli/src/s3-store.ts
```diff
@@ -66,22 +67,51 @@ class S3StoreMap {
...
   readonly #loadingLimit = new PQueue({
     intervalCap: MAX_LOAD_RPS,
     interval: 1000,
     carryoverConcurrencyCount: true,
   })
 
-  constructor(networkName: string, bucketName: string, customEndpoint?: string) {
+  constructor(
+    networkName: string,
+    diagnosticsLogger: DiagnosticsLogger,
```
gvelez17, line 83: minor point but if we put this at the end and made it optional we could be backwards compatible

packages/core/src/store/level-db-store.ts
```diff
@@ -30,17 +30,17 @@ class LevelDBStoreMap {
...
+  #fullLocations: Record<string, string> = {}
 
   constructor(storeRoot: string, networkName: string, readonly logger: DiagnosticsLogger) {
     this.networkName = networkName
     this.#storeRoot = storeRoot
     this.#map = new Map<string, Level>()
   }
 
-  createStore(useCaseName = DEFAULT_LEVELDB_STORE_USE_CASE_NAME): Promise<void> {
+  createStore(fullLocation: string): Promise<void> {
```
gvelez17, line 41: isn't this used as relative location not full location?

gvelez17 (approved):
lets get this out the door - i had one small comment on order of parameters but i don't think its a blocker

packages/cli/src/s3-store.ts
PaulLeCam: AFAIK it's internal to the CLI so I don't think it affects external consumers, can you confirm please @stbrody?

stbrody: yep, no one should be depending on this, seems fine to change

PaulLeCam:
The state store is only created when creating a first stream so the first check should be covered, I'll do the update check tomorrow 👍

PaulLeCam:
I just checked that updates are correctly applied to the state store using the ELP path when it exists, should be all good.

PR #3048, opened 2023-11-27
Fix Uint8Array inputs for concatenation calls

PaulLeCam:
This should be a fix for https://forum.ceramic.network/t/from-discord-user-elcharitas-please-help-i-am-running-a-local-ceramic-node-and-i-keep-getting-this-error/1374/15

If I understand the issue correctly, [`uint8arrays` v4.0.9 released 3 days ago](https://github.com/achingbrain/uint8arrays/releases/tag/v4.0.9) is more strict in the input values it accepts for `concat()`, it only accepts `Uint8Array` and no longer plain `Array` that can be generated by using `varint.encode()`, so this PR ensures these are explicitly cast.

### DIFF
--- a/packages/streamid/package.json
+++ b/packages/streamid/package.json
@@ -35,7 +35,7 @@
     "cborg": "^1.10.2",
     "mapmoize": "^1.2.1",
     "multiformats": "^11.0.1",
-    "uint8arrays": "^4.0.3",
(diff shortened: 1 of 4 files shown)


stbrody (approved):
seems reasonable as best I can tell

PR #3049, opened 2023-11-29
chore(core): cat 2855 add lint rule to detect promises that never get awaited on core

samika98:
## Description

This PR introduces an ESLint rule in the js-ceramic codebase to prevent floating promises in the core package. Floating promises occur when the outcome of a promise is not explicitly managed. To address this, we apply one of three strategies:

1) Handling Positive and Negative Outcomes: This involves adding a .catch() to manage errors when promises fail.
2) Using void Keyword: This is used when the result of the promise is intentionally ignored.
3) Silencing the Linter: This is a last resort and should be used sparingly.

## How Has This Been Tested?

To ensure the effectiveness of this rule, the following test was conducted:

Test 1: Executed npm lint to confirm no errors related to @typescript-eslint/no-floating-promises.

- Test 1 : ran npm lint, ensured no errors regarding to the rule @typescript-eslint/no-floating-promises

## PR checklist

- [ Yes] I have tagged the relevant reviewers and interested parties
- [ Not Needed] I have updated the READMEs of affected packages
- [ Not required] I have made corresponding changes to the documentation

## References:
https://typescript-eslint.io/rules/no-floating-promises/ : rule for no floating promises.

### DIFF
--- a/packages/core/.eslintrc.json
+++ b/packages/core/.eslintrc.json
@@ -4,7 +4,7 @@
   "plugins": ["@typescript-eslint", "import"],
   // required for no-floating -promises lint rule
   "parserOptions": {
-    "project": "tsconfig.json" 
+    "project": "tsconfig.linter.json" 
   },
   "rules": {
     "@typescript-eslint/no-explicit-any": "off",
--- a/packages/core/.eslintrc.json
+++ b/packages/core/.eslintrc.json
@@ -2,12 +2,17 @@
   "root": true,
   "parser": "@typescript-eslint/parser",
   "plugins": ["@typescript-eslint", "import"],
+  // required for no-floating -promises lint rule
+  "parserOptions": {
+    "project": "tsconfig.json" 
+  },
   "rules": {
     "@typescript-eslint/no-explicit-any": "off",
     "@typescript-eslint/ban-ts-ignore": "off",
     "@typescript-eslint/ban-ts-comment": "off",
     "import/no-unresolved": "error",
-    "import/default": "off"
+    "import/default": "off",
+    "@typescript-eslint/no-floating-promises": "error"
   },
   "settings": {
     "import/parsers": {
--- a/packages/core/src/__tests__/shutdown-signal.test.ts
+++ b/packages/core/src/__tests__/shutdown-signal.test.ts
@@ -6,13 +6,9 @@ const fn = (abortSignal: AbortSignal) => {
     abortSignal.addEventListener('abort', () => {
       reject(new Error(`aborted after run`))
     })
-    new Promise((resolve1) => setTimeout(resolve1, 500))
-      .then(() => {
-        resolve('ok')
-      })
-      .catch((error) => {
-        reject(error)
-      })
+    setTimeout(() => {
+      resolve('ok')
+    }, 500)
   })
 }
 
--- a/packages/core/src/__tests__/shutdown-signal.test.ts
+++ b/packages/core/src/__tests__/shutdown-signal.test.ts
@@ -6,7 +6,13 @@ const fn = (abortSignal: AbortSignal) => {
     abortSignal.addEventListener('abort', () => {
       reject(new Error(`aborted after run`))
     })
-    new Promise((resolve1) => setTimeout(resolve1, 500)).then(() => resolve('ok'))
+    new Promise((resolve1) => setTimeout(resolve1, 500))
+      .then(() => {
+        resolve('ok')
+      })
+      .catch((error) => {
+        reject(error)
+      })
   })
 }
 
--- a/packages/core/src/state-management/repository.ts
+++ b/packages/core/src/state-management/repository.ts
@@ -996,7 +996,7 @@ export class Repository {
           })
         })
         .catch((error) => {
-          this.logger.err(`An error occurred in updates$ for StreamID ${id}: ${error.message}`)
+          this.logger.err(`An error occurred in updates$ for StreamID ${id}: ${error}`)
         })
     })
   }
--- a/packages/core/src/state-management/repository.ts
+++ b/packages/core/src/state-management/repository.ts
@@ -997,6 +997,8 @@ export class Repository {
         })
         .catch((error) => {
           this.logger.err(`An error occurred in updates$ for StreamID ${id}: ${error}`)
+          // propogate the error to the subscriber
+          subscriber.error(error)
         })
     })
   }
--- a/packages/core/src/state-management/repository.ts
+++ b/packages/core/src/state-management/repository.ts
@@ -985,15 +985,19 @@ export class Repository {
   updates$(init: StreamState): Observable<StreamState> {
     return new Observable<StreamState>((subscriber) => {
       const id = new StreamID(init.type, init.log[0].cid)
-      this.fromMemoryOrStore(id).then((found) => {
-        const state$ = found || new RunningState(init, false)
-        this.inmemory.endure(id.toString(), state$)
-        state$.subscribe(subscriber).add(() => {
-          if (state$.observers.length === 0) {
-            this.inmemory.free(id.toString())
-          }
+      this.fromMemoryOrStore(id)
+        .then((found) => {
+          const state$ = found || new RunningState(init, false)
+          this.inmemory.endure(id.toString(), state$)
+          state$.subscribe(subscriber).add(() => {
+            if (state$.observers.length === 0) {
+              this.inmemory.free(id.toString())
+            }
+          })
         })
-      })
+        .catch((error) => {
+          this.logger.err(`An error occurred in updates$ for StreamID ${id}: ${error.message}`)
+        })
     })
   }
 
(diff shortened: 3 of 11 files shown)


packages/core/src/__tests__/shutdown-signal.test.ts
```diff
@@ -6,7 +6,13 @@ const fn = (abortSignal: AbortSignal) => {
...
       reject(new Error(`aborted after run`))
     })
-    new Promise((resolve1) => setTimeout(resolve1, 500)).then(() => resolve('ok'))
+    new Promise((resolve1) => setTimeout(resolve1, 500))
+      .then(() => {
+        resolve('ok')
+      })
+      .catch((error) => {
+        reject(error)
+      })
```
PaulLeCam, line 15: Seems unnecessary to have a promise inside a promise here, I think it could be simplified as

```suggestion
    setTimeout(() => {
      resolve('ok')
    }, 500)
```

packages/core/src/state-management/repository.ts
```diff
@@ -985,15 +985,19 @@ export class Repository {
...
+          const state$ = found || new RunningState(init, false)
+          this.inmemory.endure(id.toString(), state$)
+          state$.subscribe(subscriber).add(() => {
+            if (state$.observers.length === 0) {
+              this.inmemory.free(id.toString())
+            }
+          })
+        })
+        .catch((error) => {
+          this.logger.err(`An error occurred in updates$ for StreamID ${id}: ${error}`)
```
PaulLeCam, line 999: Shouldn't this update propagate to the Observable @stbrody?

PaulLeCam (approved):
Nice to get these additional checks, changes LGTM I'm just wondering about the potential issues they surface.

samika98:
> Nice to get these additional checks, changes LGTM I'm just wondering about the potential issues they surface.

@PaulLeCam are there any specific potential issues that we can tackle in this PR?

packages/core/src/state-management/repository.ts
PaulLeCam: Sorry I meant propagating the error to the observable consumer, otherwise here it is swallowed by the `catch()` handler.

PaulLeCam:
> > Nice to get these additional checks, changes LGTM I'm just wondering about the potential issues they surface.
> 
> @PaulLeCam are there any specific potential issues that we can tackle in this PR?

I just noticed the repository updates tracking logic that doesn't seem to propagate errors to the consumer, it could be the expected behavior but I'm not sure

packages/core/.eslintrc.json
```diff
@@ -2,12 +2,17 @@
...
+    "project": "tsconfig.linter.json" 
+  },
   "rules": {
     "@typescript-eslint/no-explicit-any": "off",
     "@typescript-eslint/ban-ts-ignore": "off",
     "@typescript-eslint/ban-ts-comment": "off",
     "import/no-unresolved": "error",
-    "import/default": "off"
+    "import/default": "off",
+    "@typescript-eslint/no-floating-promises": "error"
```
stbrody, line 15: What about the other non-core packages?  Do we need to add this rule to the `.eslintrc.json` for each of the packages?

packages/core/src/state-management/repository.ts
stbrody: yeah this definitely looks like a behavior change at the least.  I think we probably do indeed want the error to propagate to the consumer.  I'll tag in @ukstv for his take as well though.

@samika98, if you don't add the .catch here, does the new lint rule complain?  That would be surprising to me.

packages/core/src/stream-loading/__tests__/tip-fetcher.test.ts
```diff
@@ -20,6 +20,7 @@ describe('TipFetcher test', () => {
     ipfsClient = await createIPFS()
 
     dispatcher = await createDispatcher(ipfsClient, TOPIC)
+    const logger = new LoggerProvider().getDiagnosticsLogger()
```
stbrody, line 23: this seems to be a no-op?  No need to define this here given its never used and then immediately goes out of scope

packages/core/src/stream-loading/__tests__/tip-fetcher.test.ts
```diff
@@ -44,7 +46,11 @@ describe('TipFetcher test', () => {
       if (message.typ == MsgType.QUERY && message.stream.equals(streamID)) {
         const tipMap = new Map().set(streamID.toString(), tip)
         const response: PubsubMessage = { typ: MsgType.RESPONSE, id: message.id, tips: tipMap }
-        ipfs2.pubsub.publish(TOPIC, serialize(response))
+        ipfs2.pubsub.publish(TOPIC, serialize(response)).catch((error) => {
+          logger.err(
```
stbrody, line 50: hmm... I don't think it really makes sense to use the logger like this in test code.  If any of this fails we want the test to fail, not for the error to be swallowed with a message to the logs.  Can we remove the catch statement and leave the test code as it was?

stbrody (changes requested):
This will be a nice extra safety check for us, I'm excited to get this in.

packages/core/src/state-management/repository.ts
samika98: Yes the new lint rule does complain if I don't add the catch here. However, if we do want to silent it we can.

packages/core/.eslintrc.json
samika98: My plan os to create a set of three PRs ( easy reviewing purposes), they will cover all the packages. 1 is for core which is this one, 1 will be for packages that do err when linter is on with 1 change, 1 will be for 4 packages with smaller errors on adding the linter.

samika98: To answer your first question we need to this rule for all the eslint.json files in all the packages

packages/core/src/stream-loading/__tests__/tip-fetcher.test.ts
samika98: catching the error, re-throwing it and adding a comment to explain

packages/core/src/state-management/repository.ts
ukstv: @samika98 Indeed this is a change. If `catch` is required, you could add something like:
```
promise.catch(error => { subscriber.error(error) })
```

packages/core/src/stream-loading/__tests__/tip-fetcher.test.ts
```diff
@@ -44,7 +44,10 @@ describe('TipFetcher test', () => {
       if (message.typ == MsgType.QUERY && message.stream.equals(streamID)) {
         const tipMap = new Map().set(streamID.toString(), tip)
         const response: PubsubMessage = { typ: MsgType.RESPONSE, id: message.id, tips: tipMap }
-        ipfs2.pubsub.publish(TOPIC, serialize(response))
+        ipfs2.pubsub.publish(TOPIC, serialize(response)).catch((error) => {
```
ukstv, line 47: This is more like a harness, so `void ipfs2.pubsub.publish` should be enough IMO.

ukstv (changes requested):
Mostly ok. Subscription logic though should be changed, as Spencer and I proposed.

packages/core/src/stream-loading/__tests__/tip-fetcher.test.ts
samika98: In the context of this test, ignoring a promise's result and potential errors with void can indeed lead to misleading test outcomes. If ipfs2.pubsub.publish() fails and throws an error, the promise is ignored due to void, the test might incorrectly pass because the error is never caught or handled.
Thus I added logic to catch the error and rethrow it to ensure the test fails in a scenario where this throws an error.

samika98:
@stbrody @ukstv thanks for the review :) Added the repository.ts change for subscription logic, let me know if it is good to merge!

packages/core/src/state-management/repository.ts
samika98: Added the fix suggested

packages/core/src/stream-loading/__tests__/tip-fetcher.test.ts
ukstv: Okay, going to treat it as a minor stylistic difference.

> the promise is ignored due to void

Result of an expression that goes into `void` gets replaced with `undefined`. The expression is still evaluated, i.e. `void foo()` executes `foo()` first. Here `ipfs.pubsub.publish` still gets executed and a Promise gets created. If the promise gets rejected, the fact it is ignored as a value does not affect its lifecycle: Node.js would still try to find  an error handler for it.

`promise.catch(e => throw e)` results in a `Promise`. `catch` clause re-throws an error. It is a no-op. Could be omitted.

So, both `void ipfs2.pubsub.publish` and `ipfs2.pubsub.publish.catch()` are equivalent. The difference is one suggests Promise is ignored, and another persuades the error is handled for real, while it is not. I'd prefer just the shorter version :)

ukstv: Hey, @samika98 mind removing `catch(() => throw)` clause please? It is a semantical no-op :)

samika98: Added the fix, thanks for the suggestion :)

PR #3051, opened 2023-12-04
chore(indexing, ipfs-daemon, stream-model-handler, stream-tests) : cat added no-floating promise linter to pkgs and fixed errors related

samika98:
## Description

This PR introduces an ESLint rule in the js-ceramic codebase to prevent floating promises in the core package. Floating promises occur when the outcome of a promise is not explicitly managed. Packages touched are indexing, ipfs-daemon, stream-model-handler and stream-tests
To address this, we apply one of three strategies:

Handling Positive and Negative Outcomes: This involves adding a .catch() to manage errors when promises fail.
Using void Keyword: This is used when the result of the promise is intentionally ignored.
Silencing the Linter: This is a last resort and should be used sparingly. This is used when the linter is highlighting code that we do not think should be highlighted

This PR is an extension of : https://github.com/ceramicnetwork/js-ceramic/pull/3049

## How Has This Been Tested?

To ensure the effectiveness of this rule, the following test was conducted:

Test 1: Executed npm lint to confirm no errors related to @typescript-eslint/no-floating-promises.

## PR checklist

- [Yes ] I have tagged the relevant reviewers and interested parties
- [ Not needed] I have updated the READMEs of affected packages
- [ Not needed] I have made corresponding changes to the documentation

## References:

https://typescript-eslint.io/rules/no-floating-promises/ : rule for no floating promises.
For ignoring lint rules for Knex query builder code : https://github.com/knex/knex/blob/6243f6cbd5421c9db046979b60fe8661b523581f/lib/query/querybuilder.js#L545

### DIFF
--- a/packages/indexing/src/insertion-order.ts
+++ b/packages/indexing/src/insertion-order.ts
@@ -258,6 +258,7 @@ export class InsertionOrder {
     if (cursor.type === 'timestamp') {
       // Paginate using the `created_at` field when no custom field ordering is provided
       builder = builder.where((qb) => {
+        // eslint-disable-next-line @typescript-eslint/no-floating-promises
         qb.where('created_at', isReverseOrder ? '<' : '>', cursor.value) // strict next value
           .orWhere('created_at', '=', cursor.value) // or current value
           .andWhere('stream_id', '>', cursor.id) // with stream ID tie-breaker
@@ -268,6 +269,7 @@ export class InsertionOrder {
         const field = contentKey(key)
         const sign = getComparisonSign(sorting[key], isReverseOrder)
         builder = builder.where((qb) => {
+          // eslint-disable-next-line @typescript-eslint/no-floating-promises
           qb.whereRaw(`${field} ${sign} ?`, [value]) // strict next value
             .orWhereRaw(`${field} = ?`, [value]) // or current value
             .andWhere('stream_id', '>', cursor.id) // with stream ID tie-breaker
--- a/packages/indexing/src/query-filter-converter.ts
+++ b/packages/indexing/src/query-filter-converter.ts
@@ -98,6 +98,7 @@ function handleIn<T extends number | string>(
       op = ` not${op}`
     }
     const raw = bldr.client.raw(`cast(${key} as ${cast})${op}(${arrValue})`)
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
     bldr.whereRaw(raw)
   }
   if (first) {
--- a/packages/stream-tests/src/__tests__/invalid_tip.test.ts
+++ b/packages/stream-tests/src/__tests__/invalid_tip.test.ts
@@ -62,7 +62,10 @@ describe('Test loading a stream when pubsub replies with an invalid tip', () => 
       if (message.typ == MsgType.QUERY && message.stream.equals(streamID)) {
         const tipMap = new Map().set(streamID.toString(), nonExistentTip)
         const response = { typ: MsgType.RESPONSE, id: message.id, tips: tipMap }
-        ipfs2.pubsub.publish(ceramic.pubsubTopic, serialize(response))
+        ipfs2.pubsub.publish(ceramic.pubsubTopic, serialize(response)).catch((error) => {
+          // we want the test to fail if an error occurs thus we are rethrowing the error
+          throw error
+        })
       }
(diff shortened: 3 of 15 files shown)


samika98:
In this pull request, I've included three specific instances where I've used the ESLint ignore directive (clearly marked with comments for easy review). These instances are necessary due to the linter incorrectly expecting a QueryBuilder from the Knex library to return a promise. Since QueryBuilder in these contexts does not inherently return a promise, the most appropriate course of action was to override the linter for these particular lines. This approach ensures that our code adheres to the intended use of QueryBuilder without being constrained by a misinterpretation from the linter.

Deep dive on why this was happening (can ignore) : 

In Knex, a QueryBuilder instance can be considered "thenable," meaning it can be used with .then(), .catch(), and await, in certain contexts. This is because QueryBuilder is designed to build and execute SQL queries, and when it comes time to execute these queries, it returns a Promise.

When is QueryBuilder a Thenable?
- After Terminating Methods: QueryBuilder becomes a thenable after you call a terminating method. Terminating methods are those that actually execute the query and return a promise. Examples of such methods include select(), insert(), update(), delete(), first(), etc. After you call any of these methods, you can use .then(), .catch(), and await on the resulting object. 

Hypothesis : where / whereRaw is not a terminating method --- linter does not have this understanding. Thus it is throwing an error. 

Proof to accept hypothesis : 
Before Terminating Methods: Before calling a terminating method, the QueryBuilder is just a chainable object that's used to build up a query. In this state, it does not represent a promise and thus is not thenable. 
This reasoning suggests that the linter does not have this understanding and is therefore expecting a promise return.

packages/indexing/src/insertion-order.ts
```diff
@@ -258,6 +258,7 @@ export class InsertionOrder {
     if (cursor.type === 'timestamp') {
       // Paginate using the `created_at` field when no custom field ordering is provided
       builder = builder.where((qb) => {
+        // eslint-disable-next-line @typescript-eslint/no-floating-promises
```
samika98, line 261: Using ignore here

packages/indexing/src/insertion-order.ts
```diff
@@ -268,6 +269,7 @@ export class InsertionOrder {
         const field = contentKey(key)
         const sign = getComparisonSign(sorting[key], isReverseOrder)
         builder = builder.where((qb) => {
+          // eslint-disable-next-line @typescript-eslint/no-floating-promises
```
samika98, line 272: using ignore here

packages/indexing/src/query-filter-converter.ts
```diff
@@ -98,6 +98,7 @@ function handleIn<T extends number | string>(
       op = ` not${op}`
     }
     const raw = bldr.client.raw(`cast(${key} as ${cast})${op}(${arrValue})`)
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
```
samika98, line 101: using ignore here

stephhuynh18 (approved):
The missing test awaits are a surprise

packages/stream-tests/src/__tests__/invalid_tip.test.ts
```diff
@@ -62,7 +62,10 @@ describe('Test loading a stream when pubsub replies with an invalid tip', () =>
       if (message.typ == MsgType.QUERY && message.stream.equals(streamID)) {
         const tipMap = new Map().set(streamID.toString(), nonExistentTip)
         const response = { typ: MsgType.RESPONSE, id: message.id, tips: tipMap }
-        ipfs2.pubsub.publish(ceramic.pubsubTopic, serialize(response))
+        ipfs2.pubsub.publish(ceramic.pubsubTopic, serialize(response)).catch((error) => {
```
dbcfd, line 65: For a test, seems like this might be better with `await expect(...).resolves.not.toThrow()`

stbrody: agreed, but the problem is that the callback passed to `pubsub.subscribe` is expected to be a serial, non-async function, so we can't use `await` here.

stbrody: I don't love it either, wish there was a better way

stbrody (approved):
lgtm

PR #3052, opened 2023-12-04
fix: Update state store after applying update

stbrody:
This is the bug that causes anchored updates to be lost after a node restart.  It was introduced in https://github.com/ceramicnetwork/js-ceramic/commit/62bc7cbdef356e32768282d09c1c96409142b16a, almost 2 months ago.

### DIFF
--- a/packages/core/src/state-management/repository.ts
+++ b/packages/core/src/state-management/repository.ts
@@ -533,7 +533,7 @@ export class Repository {
       if (next) {
         state$.next(next)
         await this._updateStateIfPinned(state$)
-        this.logger.imp(`Stream ${state$.id} successfully updated to tip ${cid}`)
+        this.logger.verbose(`Stream ${state$.id} successfully updated to tip ${cid}`)
         return true
       } else {
         return false
--- a/packages/core/src/state-management/repository.ts
+++ b/packages/core/src/state-management/repository.ts
@@ -741,7 +741,9 @@ export class Repository {
             // If we failed to apply the commit at least once, then it's worth logging when
             // we are able to do so successfully on the retry.
             this.logger.imp(
-              `Successfully applied anchor commit ${anchorCommitCID} for stream ${state$.id}`
+              `Successfully applied anchor commit ${anchorCommitCID} for stream ${
+                state$.id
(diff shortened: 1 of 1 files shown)


packages/core/src/state-management/repository.ts
```diff
@@ -722,8 +723,10 @@ export class Repository {
       remainingRetries--
     ) {
       try {
+        this.logger.verbose(`Handling anchor commit for ${streamId}`)
```
dbcfd, line 726: Should this move out of the retry block?

packages/core/src/state-management/repository.ts
```diff
@@ -738,6 +743,10 @@ export class Repository {
             this.logger.imp(
               `Successfully applied anchor commit ${anchorCommitCID} for stream ${state$.id}`
```
AaronGoldman, line 744: Should we log the number of trys it took?

packages/core/src/state-management/repository.ts
```diff
@@ -738,6 +743,10 @@ export class Repository {
...
             )
+          } else {
+            this.logger.verbose(
+              `Successfully applied anchor commit ${anchorCommitCID} for stream ${state$.id}`
```
gvelez17, line 748: this doesn't make sense why log the same thing in both cases?  if there are two cases we should have two different log things?

packages/core/src/state-management/repository.ts
```diff
@@ -531,8 +531,8 @@ export class Repository {
       this.logger.verbose(`Learned of new tip ${cid} for stream ${state$.id}`)
       const next = await this.streamUpdater.applyTipFromNetwork(state$.state, cid)
       if (next) {
-        await this._updateStateIfPinned(state$)
         state$.next(next)
+        await this._updateStateIfPinned(state$)
```
gvelez17, line 535: huh how does this fix the problem we saw, I'm not deep in enough to understand exactly - is it that we didn't advance the state before pinning it?

stbrody: we write the old, pre-anchor-commit state to the state store.

This code is applying a new commit, but writing the old state that doesn't contain the new commit to the state store.  This fixes the code to apply the commit to the state and THEN update the state store with the new state

ukstv (approved):

PR #3053, opened 2023-12-04
fix: Clean up anchor request store even if stream is already anchored

stbrody:


packages/core/src/state-management/repository.ts
```diff
@@ -614,10 +614,16 @@ export class Repository {
         return false
       }
       case AnchorRequestStatusName.COMPLETED: {
-        if (anchorEvent.cid.equals(state$.tip)) {
+        const anchorCommitCID = anchorEvent.witnessCar.roots[0]
+        await this._handleAnchorCommit(state$, anchorEvent.cid, anchorEvent.witnessCar)
+        if (state$.tip.equals(anchorCommitCID)) {
```
AaronGoldman, line 619: Do we have a `if anchorHistory.includes(anchorCommitCID)`?

stbrody: depends on what you mean by "anchorHistory".  We can check if the anchor commit CID exists anywhere in the log of the state, but if we use that as the criteria to delete the entry from the anchorRequestStore, we risk deleting a valid entry for a newer update

packages/core/src/state-management/repository.ts
```diff
@@ -614,10 +614,16 @@ export class Repository {
         return false
       }
       case AnchorRequestStatusName.COMPLETED: {
-        if (anchorEvent.cid.equals(state$.tip)) {
+        const anchorCommitCID = anchorEvent.witnessCar.roots[0]
+        await this._handleAnchorCommit(state$, anchorEvent.cid, anchorEvent.witnessCar)
```
stephhuynh18, line 618: do we expect `anchorEvent.witnessCar.roots[0]` and  `anchorEvent.cid` to be different?

stephhuynh18: AH i see thanks

stephhuynh18 (approved):

PR #3058, opened 2023-12-07
Feature/cat 2882 mvp of firehose

JulissaDantes:
## Description

Creates a `feed` property to the ceramic class. Said property will be populated by state changes

## How Has This Been Tested?

Unit tests


ukstv:
Nice. Let me play a bit with a different, less manual way of pushing the events, and then prob we can merge it.

ukstv (approved):

packages/core/src/feed.ts
```diff
@@ -0,0 +1,57 @@
...
+
+class DocumentsSubject extends Observable<Document> {
+  readonly subject: Subject<StreamState>
+  constructor() {
+    super((subscriber) => {
+      const subscription = this.subject
+        .pipe(
+          // transform each incoming StreamState to a Document
+          map((streamState: StreamState) => {
+            const tipCID = streamState.log[streamState.log.length - 1].cid
```
dbcfd, line 19: Do we want any error handling here around these accesses? The log should never be empty, but if we do error here, it might do weird stuff to the observable and the sse.

dbcfd (approved):

ukstv:
Oh, found few issues during testing.
1. We do not stop an SSE feed that's been subscribed to after Ctrl+C. It hangs.
2. Anchoring status is not reflected in Document structure we emit.

@JulissaDantes

packages/core/src/feed.ts
```diff
@@ -1,10 +1,12 @@
 import { Subject, Observable, map } from 'rxjs'
 import type { StreamMetadata, StreamState } from '@ceramicnetwork/common'
 import { CommitID } from '@ceramicnetwork/streamid'
+import { AnchorStatus } from '@ceramicnetwork/common'
```
ukstv, line 4: `import type` pls

packages/core/src/feed.ts
```diff
@@ -0,0 +1,60 @@
...
+  metadata: StreamMetadata
+}
+
+class DocumentsSubject extends Observable<Document> {
+  readonly subject: Subject<StreamState>
+  constructor() {
+    super((subscriber) => {
+      const subscription = this.subject
+        .pipe(
+          // transform each incoming StreamState to a Document
```
PaulLeCam, line 19: Don't we already have a method or utility function that handles this somewhere please @ukstv @stbrody?
Looks like the kind of logic we want to have consistent over the entire codebase.

packages/core/src/feed.ts
```diff
@@ -0,0 +1,60 @@
+import { Subject, Observable, map } from 'rxjs'
+import type { StreamMetadata, StreamState } from '@ceramicnetwork/common'
+import { CommitID } from '@ceramicnetwork/streamid'
+import type { AnchorStatus } from '@ceramicnetwork/common'
+
+export type Document = {
+  id: CommitID
```
PaulLeCam, line 7: Could we make it more explicit the ID exposed here is a `CommitID` please?
I think it could be confusing otherwise for consumers that are used to having the `StreamID` provided.
```suggestion
  commitID: CommitID
```

packages/core/src/feed.ts
```diff
@@ -0,0 +1,60 @@
...
+  content: any
+  anchorStatus: AnchorStatus
+  metadata: StreamMetadata
+}
+
+class DocumentsSubject extends Observable<Document> {
```
PaulLeCam, line 13: This could make it easier to pipe an observable of `StreamState` directly to this subject:
```suggestion
class DocumentsSubject extends Observable<Document> implements NextObserver<StreamState> {
```

PaulLeCam (commented):
Logic LGTM and seems there's a good test coverage, I'm only wondering about a few things here.

packages/core/src/feed.ts
JulissaDantes: Ready!

packages/cli/src/daemon/sse-feed.ts
```diff
@@ -0,0 +1,42 @@
+import type { Observable } from 'rxjs'
+import type { Response } from 'express'
+import type { DiagnosticsLogger } from '@ceramicnetwork/common'
+
+export class SseFeed<TInput> {
```
stbrody, line 5: Can we add a class comment describing what this does?  To someone who isn't familiar with Server Sent Events they might not even know what "Sse" means.

packages/core/src/__tests__/ceramic-feed.test.ts
```diff
@@ -0,0 +1,158 @@
...
+import { TestUtils, type StreamState, type IpfsApi } from '@ceramicnetwork/common'
+import { TileDocument } from '@ceramicnetwork/stream-tile'
+import { createIPFS, swarmConnect, withFleet } from '@ceramicnetwork/ipfs-daemon'
+import type { Ceramic } from '../ceramic.js'
+import { createCeramic as vanillaCreateCeramic } from './create-ceramic.js'
+import { Model, ModelDefinition } from '@ceramicnetwork/stream-model'
+import { take } from 'rxjs'
+import { Document } from '../feed.js'
+
+function createCeramic(
```
stbrody, line 11: do we need this special createCeramic function?  Doesn't seem like we use the extra args anywhere in this test, can we just use the createCeramic from `create-ceramic.js` directly?

packages/core/src/__tests__/ceramic-feed.test.ts
```diff
@@ -0,0 +1,158 @@
...
+    streamCacheLimit,
+  })
+}
+
+describe('Ceramic feed', () => {
+  let ipfs: IpfsApi
+  let ceramic: Ceramic
+  beforeEach(async () => {
+    ipfs = await createIPFS()
+    ceramic = await createCeramic(ipfs)
```
stbrody, line 27: some tests don't use this ceramic instance at all since they make their own.  Can we do one of:
1. *always* leave the tests to create their own Ceramic
2.  change this setup to happen `beforeAll` instead of `beforeEach` so we aren't paying the cost of setting this up and tearing it down between each test?
3. set up 2 ceramic nodes in the test harness and use those in the tests, do no node setup within the body of any tests.

My personal preference is probably for 3, but I'm fine with any approach

packages/core/src/__tests__/ceramic-feed.test.ts
```diff
@@ -0,0 +1,158 @@
...
+        anchor: false,
+        publish: true,
+      })
+      // load stream on ceramic node 2
+      await TileDocument.load(ceramic2, stream1.id)
+
+      await stream1.update(updatedContent, null, { publish: true })
+      await TestUtils.delay(500)
+
+      expect(feed1.length).toEqual(3) // create + anchor + update
```
stbrody, line 83: why is there an anchor since the stream is created with `anchor:false`?

packages/core/src/feed.ts
```diff
@@ -0,0 +1,58 @@
+import { Subject, Observable, map, NextObserver } from 'rxjs'
+import { type StreamMetadata, type StreamState } from '@ceramicnetwork/common'
+import { CommitID } from '@ceramicnetwork/streamid'
+import { StreamUtils } from '../../common/lib/utils/stream-utils.js' //TODO: once commitIdFromStreamState is part of the package remove this line, and update the @ceramicnetwork/common import
+import type { AnchorStatus } from '@ceramicnetwork/common'
+
+export type Document = {
```
stbrody, line 7: to avoid confusion with the existing `Document` class in the http-client (which we should also probably rename), can we call this `FeedDocument` or `FeedEvent` or something like that?

packages/core/src/feed.ts
```diff
@@ -0,0 +1,58 @@
...
+  commitId: CommitID
+  content: any
+  anchorStatus: AnchorStatus
```
stbrody, line 10: are we sure we need to include this at all?  This was still an open question in the design doc, we had a note to follow up with Orbis and see if they need this.  I'd prefer to err on the side of leaving this out unless it's specifically asked for.

I think if users care about information from anchoring, they're more likely to care about the anchor timestamp than the AnchorStatus anyway, so I'd like to wait and think about how we expose anchor-related information more holistically if and when it becomes a requirement.

packages/core/src/feed.ts
stbrody: I don't think so.  This `Document` class used here in the feed is a brand new thing.

packages/core/src/state-management/repository.ts
```diff
@@ -819,6 +824,7 @@ export class Repository {
    * Adds the stream's RunningState to the in-memory cache and subscribes the Repository's global feed$ to receive changes emitted by that RunningState
    */
   private _registerRunningState(state$: RunningState): void {
+    state$.subscribe(this.feed.aggregation.documents)
```
stbrody, line 827: :chefs kiss:  So simple and clean.

packages/core/src/state-management/repository.ts
```diff
@@ -819,6 +824,7 @@ export class Repository {
    * Adds the stream's RunningState to the in-memory cache and subscribes the Repository's global feed$ to receive changes emitted by that RunningState
```
stbrody, line 824: Haha I have no idea why this comment is here already but it seems to have perfectly predicted this future 😂

packages/core/src/state-management/repository.ts
```diff
@@ -943,15 +949,17 @@ export class Repository {
         .then((found) => {
           const state$ = found || new RunningState(init, false)
           this.inmemory.endure(id.toString(), state$)
-          state$.subscribe(subscriber).add(() => {
-            if (state$.observers.length === 0) {
+          const subscription = state$.subscribe(subscriber)
+          state$.add(subscription)
+          subscription.add(() => {
+            if (state$.subscriptionSet.size === 0) {
```
stbrody, line 955: Can you explain to me the significance of this change from checking `observers.length` to `subscriptionSet.size`?

stbrody (changes requested):
This looks really great.  A few small nitpicks here and there but overall I'm amazed how simple and clean this change is.

I'd like to see some end-to-end tests going through the HTTP api.  I assume that will come in a follow-up PR that also adds functionality for accessing the feed to the ceramic-http-client?

packages/core/src/__tests__/ceramic-feed.test.ts
JulissaDantes: After the update the anchor status changes from 0 to 1

stbrody: aha, that exposes an interesting property of this feed - there's an event emitted every time the anchor status changes, even if there is no corresponding commit or change to the actual log/data of the stream.  This kind of relates to my comment elsewhere about whether or not we want anchorStatus to show up in the feed at all.  That's really a question for product though.

I suspect if you set `anchor:false` in the args for the update then the feed length will go to 2 here.

packages/core/src/feed.ts
stbrody: Looping in @rohhan for this question.  This corresponds to this open question from the Notion document: https://www.notion.so/Firehose-Requirements-and-API-Stub-1c8db095218d4e2db9b0ee1db9dce861?d=cb6573e840944a53b5f819be2292b6c1&pvs=4#16107f227a4a48f9b7fd216e98ff3a8c

packages/core/src/feed.ts
JulissaDantes: I agree it can be confusing, I'll go for `FeedDocument`.

packages/core/src/__tests__/ceramic-feed.test.ts
JulissaDantes: I see, I guess we'll revisit this when we figure out if needed or not first.

packages/core/src/feed.ts
dbcfd: @JulissaDantes Per discussion, we should remove this field, adding later as needed.

packages/core/src/state-management/repository.ts
dbcfd: @JulissaDantes Additional comment to respond to.

packages/core/src/__tests__/ceramic-feed.test.ts
dbcfd: @JulissaDantes We should add tests for both anchor:true and anchor:false, just to record behavior.

packages/cli/src/daemon/sse-feed.ts
JulissaDantes: I don't know why github isn't showing it here, but if you go to the file you'll see a comment

packages/core/src/__tests__/ceramic-feed.test.ts
JulissaDantes: I included the `anchor:false` in the update to test behavior without anchor, and if it gets an event when anchoring is true gets tested at the `add entry after anchoring stream` test.

packages/core/src/feed.ts
JulissaDantes: Ready!

packages/core/src/state-management/repository.ts
JulissaDantes: @stbrody @dbcfd The feed is going to be a subscriber to the state, and we dont want that to prevent the memory from being freed, so it is checking the `subscriptionSet.size` instead. 

Also, [according to the code](https://github.com/ReactiveX/rxjs/blob/fdfc8acf74acc314be7184fbfbc2089294c9253d/src/internal/Subject.ts#L19) `observers` is soon to be deprecated in an upcomming version.

dbcfd (approved):

packages/core/src/__tests__/ceramic-feed.test.ts
```diff
@@ -1,44 +1,40 @@
...
+import { createCeramic } from './create-ceramic.js'
 
 describe('Ceramic feed', () => {
-  let ipfs: IpfsApi
-  let ceramic: Ceramic
+  let ipfs1: IpfsApi
+  let ipfs2: IpfsApi
+  let ceramic1: Ceramic
+  let ceramic2: Ceramic
   beforeEach(async () => {
```
stbrody, line 16: can we make this beforeAll/afterAll?  Do we need the nodes spun up and down between each test?  Not a big deal either way, but if we can re-use the same nodes for the whole test it will improve test execution speed.

stbrody (commented):
looks good!  My last remaining concern is that now that we don't include anchorStatus in the FeedDocuments we return, we probably shouldn't emit a new update every time the anchor status changes.  Right now we can wind up emitting multiple identical FeedDocuments with the exact same content, metadata and even commitid just because anchor status changed.  An interesting question is what we should do when a stream applies an anchor commit and transitions to Anchored status, because in that case while the content and metadata haven't changed, the CommitID has since there's now a new commit in the log.  So maybe in that case it would still be a good idea to raise a new event?  I could kind of see it going either way, but I do feel like if indexers are tracking the tip of the stream we would want them to have the anchor commit as the tip and not always be stuck at the last data event.  So I think I lean towards saying we should include an event in that case.  So I think my suggestion is that we should add a filter to the pipeline to only emit new FeedDocuments the tip CID changed vs the previous event, which will filter out unimportant changes to AnchorStatus (like from `NOT_REQUESTED` to `PENDING`, or from `PENDING` to `PROCESSING`), but will leave in an extra update when a stream actually gets fully anchored.

packages/core/src/__tests__/ceramic-feed.test.ts
JulissaDantes: Good point, ready!

packages/core/src/state-management/repository.ts
```diff
@@ -943,15 +949,17 @@ export class Repository {
         .then((found) => {
           const state$ = found || new RunningState(init, false)
```
stbrody, line 950: I was looking at this code for a different PR I'm working on and it made me wonder... if we hit the case here where the state isn't found in the cache or state store, but then we wind up creating a new RunningState object here, we don't ever all _registerRunningState on it, so I think it misses getting wired into the feed.  Do we need to add a call to _registerRunningState here when making a new RunningState?

JulissaDantes: You are right, I've created #3074 to address that.

ukstv:
@stbrody Would it be all right by you if we release the PR as it is, and amend the code wrt deduplication later? To me it feels to be a choice between "ship and iterate based on feedback" and "stew in own fantasies". I do not question deduplication, it is more like, I would like to give a working feature to the (friendly and well-aware) partner for a feedback about general approach.

ukstv:
Oh, it is the holiday season. No-one will review that. Anyway, @stbrody here is a PR that addresses your deduplication concern: https://github.com/ceramicnetwork/js-ceramic/pull/3075

stbrody (approved):
yep, happy to merge this and address the filtering in the follow-up PR you started.

PR #3062, opened 2023-12-08
chore: ceramic context cleanup

dbcfd:
Break up inheritance chains by introducing injection of context functionality.
Split context functionality into parts for signing, stream manipulation, and
stream loading. Http client is no longer a ceramic api.


packages/3id-did-resolver/src/__tests__/index.test.ts
```diff
@@ -48,7 +49,7 @@ afterAll(() => {
 
 describe('3ID DID Resolver', () => {
   jest.setTimeout(20000)
-  let ceramic
+  let ceramic: CeramicClient
 
   beforeAll(async () => {
     const { CeramicClient } = await import('@ceramicnetwork/http-client')
```
JulissaDantes, line 55: We should delete this line because its getting imported at ln 5

packages/common/src/ceramic-api.ts
```diff
@@ -146,78 +181,29 @@ export interface AdminApi {
...
+export interface CeramicApi {
+  readonly loggerProvider: LoggerProvider
 
   readonly index: IndexApi
 
   readonly admin: AdminApi
 
+  // readonly streamManipulator: StreamManipulator
+  //
+  // readonly signer: CeramicSigner
```
JulissaDantes, line 193: I think this was meant to be deleted

packages/common/src/utils/test-utils.ts
JulissaDantes, line 1: I think if we are creating a new `test-utils` package we dont need this specific file, we just need to put what is left here there. It might be confusing for a new contributor when they want to add a new test utility and there are 2 places they can put it.

packages/core/src/__tests__/ceramic-api.test.ts
JulissaDantes, line 1: We can delete the `TestUtils` import from `@ceramicnetwork/common` here since it's not being used anymore.

packages/stream-model-handler/src/__tests__/model-handler.test.ts
```diff
@@ -127,20 +130,50 @@ const CONTENT_WITH_INVALID_SCHEMA = {
   },
 }
 
-const ThreeIdResolver = {
-  '3': async (did) => ({
+function resolutionResult(didUrl: string, optRotateDate?: string): DIDResolutionResult {
```
JulissaDantes, line 133: I think this should import the new `resolutionResult` from `test-utils/src/index.ts` it's the same code

packages/stream-tile/src/tile-document.ts
```diff
@@ -131,16 +130,16 @@ function headerFromMetadata(
...
-  if (!signer.did) throw new Error(`No DID provided`)
-  if (!signer.did.authenticated) {
-    await signer.did.authenticate()
-    if (signer.loggerProvider) {
-      signer.loggerProvider.getDiagnosticsLogger().imp(`Now authenticated as DID ${signer.did.id}`)
-    }
-  }
-  return signer.did
-}
+// async function getAuthenticatedDID(signer: CeramicSigner): Promise<DID> {
```
JulissaDantes, line 133: Why are we keeping this?

JulissaDantes (commented):
Left a few comments to have a better understanding, and asked for some changes, but other than that this is looking good!

packages/common/src/ceramic-api.ts
```diff
@@ -65,19 +64,55 @@ export interface PinApi {
...
+  ensureAuthenticated(): Promise<void>
+  createJWS<T extends string | Record<string, any>>(
+    payload: T,
+    options?: CreateJWSOptions
+  ): Promise<DagJWS>
+  createDagJWS(payload: Record<string, any>, options?: CreateJWSOptions): Promise<DagJWSResult>
+  verifyJWS(jws: string | DagJWS, options?: VerifyJWSOptions): Promise<VerifyJWSResult>
+  asController(): Promise<string>
+
+  // [index: string]: any // allow arbitrary properties
```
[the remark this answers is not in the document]
dbcfd: It should be, Thanks for the catch.

packages/common/src/ceramic-api.ts
dbcfd: It was. I'm actually wondering if I can delete this whole file now. Might give that a shot.

packages/common/src/utils/test-utils.ts
dbcfd: Yeah, I'd like not to see it in the common package, will try to move it out.

packages/common/src/stream.ts
```diff
@@ -291,9 +293,9 @@ export interface StreamConstructor<T extends Stream> {
   /**
    * Constructor signature
    * @param state$ - Stream state
-   * @param context - Ceramic context
+   * @param manipulator - Object for stream manipulation
```
dbcfd, line 296: Rename

packages/core/src/anchor/anchor-service.ts
```diff
@@ -22,11 +23,11 @@ export interface AnchorService {
   init(store: AnchorRequestStore, eventHandler: AnchorLoopHandler): Promise<void>
 
   /**
-   * Set Ceramic API instance
+   * Set Ceramic signer instance
    *
-   * @param ceramic - Ceramic API used for various purposes
+   * @param ceramic - Ceramic signer
```
dbcfd, line 28: rename

packages/stream-caip10-link-handler/src/__tests__/caip10-link-handler.test.ts
```diff
@@ -93,16 +101,12 @@ describe('Caip10LinkHandler', () => {
...
-      getSupportedChains: jest.fn(async () => {
-        return ['fakechain:123']
-      }),
-    }
-    context = {
-      ipfs: ipfs,
-      anchorService: null,
-      api: api as unknown as CeramicApi,
-    }
+    // const api = {
```
dbcfd, line 104: remove

packages/stream-model-instance/src/model-instance-document.ts
```diff
@@ -111,22 +100,22 @@ export class ModelInstanceDocument<T = Record<string, any>> extends Stream {
...
    * @param metadata - Genesis metadata, including the model that this document belongs to
    * @param opts - Additional options
    */
   static async create<T>(
-    ceramic: CeramicApi,
+    writer: StreamWriter,
     content: T | null,
     metadata: ModelInstanceDocumentMetadataArgs,
-    opts: CreateOpts = {}
+    opts?: CreateOpts
```
dbcfd, line 112: Might need to make signer nullable rather than option nullable.

dav1do (commented):
Some small comments/questions. 

I had some trouble with tests, mostly it was due to imports and the missing setter error:

``` FAIL  src/__tests__/model-instance-document-handler.test.ts
  ● Test suite failed to run

    ReferenceError: DID_ID is not defined

      64 | const CONTENT1 = { myData: 1 }
      65 | const CONTENT2 = { myData: 2 }
    > 66 | const METADATA = { controller: DID_ID, model: FAKE_MODEL_ID }
 FAIL  src/__tests__/model-handler.test.ts^
  ● ModelHandler › is constructed correctly

    TypeError: Cannot set property id of #<DID> which has only a getter

      181 |     const did = new DID({})
      182 |     //@ts-ignore
    > 183 |     did.id = DID_ID
          |     ^
      184 |     this.setDidToNotRotatedState(did)
      185 |     return did
      186 |   }

      at TestUtils.generateDID (../test-utils/src/index.ts:183:5)
      at Object.generateDID (src/__tests__/model-handler.test.ts:133:21)
```

packages/common/src/ceramic-api.ts
```diff
@@ -65,19 +64,53 @@ export interface PinApi {
...
-
 /**
- * Interface for an object that contains a DID that can be used to sign Ceramic commits.
+ * Interface for an object that can sign ceramic commits
  * Any implementation of CeramicAPI will match this interface, though if no CeramicAPI instance is
  * available users can provide any object containing an authenticated DID instance.
  */
-export interface CeramicSigner extends CeramicCommon {
-  did: DID | undefined
+export interface CeramicSignerRequirements {
```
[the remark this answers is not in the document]
ukstv: Oh, I get it. The suggestion does not make sense. It is more similar to [`underlyingSource` of a `ReadableStream`](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/ReadableStream#underlyingsource). Maybe `UnderlyingCeramicSigner`? Or, leave it as it is.

packages/core/src/ceramic.ts
```diff
@@ -247,13 +264,14 @@ export class Ceramic implements CeramicApi {
       this._logger,
       this.dispatcher,
       modules.anchorService.validator,
-      this._streamHandlers,
-      this.context
+      this,
```
ukstv, line 267: Full `this`? Nothing more concrete?

packages/core/src/ceramic.ts
```diff
@@ -247,13 +264,14 @@ export class Ceramic implements CeramicApi {
...
+      this._streamHandlers
     )
     const pinStore = modules.pinStoreFactory.createPinStore()
     const localIndex = new LocalIndexApi(
       params.indexingConfig,
-      this, // Circular dependency while core provided indexing
+      this,
+      this,
```
ukstv, line 274: Wow :) No need to fix it here. Just wow.

packages/core/src/ceramic.ts
```diff
@@ -530,13 +551,6 @@ export class Ceramic implements CeramicApi {
     }
   }
 
-  /**
-   * @deprecated - use the Ceramic.did setter instead
-   */
-  async setDID(did: DID): Promise<void> {
```
ukstv, line 536: Farewell, old friend.

packages/core/src/index.ts
```diff
@@ -6,3 +6,4 @@ export { Ceramic } from './ceramic.js'
 export * from './store/ikv-store.js'
 export * from './ceramic.js'
 export * from './pubsub/pubsub-message.js'
+export { InMemoryAnchorService } from './anchor/memory/in-memory-anchor-service.js'
```
ukstv, line 9: Why?

packages/core/src/initialization/stream-loading.ts
```diff
@@ -1,6 +1,6 @@
 import { StreamUpdater } from '../stream-loading/stream-updater.js'
 import { StreamLoader } from '../stream-loading/stream-loader.js'
-import { Context, DiagnosticsLogger } from '@ceramicnetwork/common'
+import { DiagnosticsLogger, StreamReaderWriter } from '@ceramicnetwork/common'
```
ukstv, line 3: import type maybe

packages/core/src/state-management/repository.ts
```diff
@@ -4,7 +4,7 @@ import {
   AnchorOpts,
   AnchorStatus,
   CommitType,
-  Context,
+  CeramicContext,
```
ukstv, line 7: Is it used even?

packages/core/src/stream-loading/__tests__/stream-loader.test.ts
```diff
@@ -49,7 +49,7 @@ function expectStatesEqualWithPendingAnchor(
   )
 }
 
-describe('StreamLoader querying against real Ceramic node', () => {
+describe('StreamStateLoader querying against real Ceramic node', () => {
```
ukstv, line 52: Why? It is still called `StreamLoader` AFAIK

packages/http-client/src/ceramic-http-client.ts
```diff
@@ -60,10 +66,39 @@ export interface CeramicClientConfig {
   syncInterval: number
 }
 
+class CeramicClientSignerRequirements implements CeramicSignerRequirements {
```
ukstv, line 69: Not quite uniform with how Core signer gets initialised.

packages/http-client/src/ceramic-http-client.ts
```diff
@@ -223,11 +263,7 @@ export class CeramicClient implements CeramicApi {
     const type = stream.state.type
     const streamConstructor = this._streamConstructors[type]
     if (!streamConstructor) throw new Error(`Failed to find constructor for stream ${type}`)
-    return new streamConstructor(stream, this.context) as T
-  }
-
-  async setDID(did: DID): Promise<void> {
-    this.context.did = did
+    return new streamConstructor(stream, this) as T
```
ukstv, line 266: I do not quite like passing full `this` around the code.

packages/stream-model-handler/src/model-handler.ts
```diff
@@ -73,13 +73,13 @@ export class ModelHandler implements StreamHandler<Model> {
...
   async applyCommit(
+    api: StreamReaderWriter,
     commitData: CommitData,
-    context: Context,
     state?: StreamState
   ): Promise<StreamState> {
     if (state == null) {
       // apply genesis
-      return this._applyGenesis(commitData, context)
+      return this._applyGenesis(api, commitData)
```
ukstv, line 82: Why do you need to change order here? To make TS notice the change?

packages/stream-model-instance-handler/src/__tests__/model-instance-document-handler.test.ts
```diff
@@ -121,20 +89,50 @@ const jwsForVersion1 = {
...
+  }
+
+  return {
     didResolutionMetadata: { contentType: 'application/did+json' },
     didDocument: wrapDocument(
       {
         publicKeys: {
           signing: 'zQ3shwsCgFanBax6UiaLu1oGvM7vhuqoW88VBUiUTCeHbTeTV',
-          encryption: 'z6LSfQabSbJzX8WAm1qdQcHCHTzVv8a2u6F7kmzdodfvUCo9',
+          encryption: 'z6MkjKeH8SgVAYCvTBoyxx7uRJFGM2a9HUeFwfJfd6ctuA3X',
```
ukstv, line 121: That's a bit concerning. Why has the encryption key changed?

packages/stream-tile/src/tile-document.ts
```diff
@@ -175,13 +160,13 @@ export class TileDocument<T = Record<string, any>> extends Stream {
...
   /**
    * Creates a Tile document.
-   * @param ceramic - Instance of CeramicAPI used to communicate with the Ceramic network
+   * @param writer - Interface to write to ceramic network
    * @param content - Genesis contents. If 'null', then no signature is required to make the genesis commit
    * @param metadata - Genesis metadata
    * @param opts - Additional options
    */
   static override async create<T>(
-    ceramic: CeramicApi,
```
ukstv, line 184: Would like to confirm. This does not break compatibility, right?

packages/stream-tile/src/tile-document.ts
```diff
@@ -289,7 +274,7 @@ export class TileDocument<T = Record<string, any>> extends Stream {
     opts: UpdateOpts = {}
   ): Promise<void> {
     opts = { ...DEFAULT_UPDATE_OPTS, ...opts }
-    const signer: CeramicSigner = opts.asDID ? { did: opts.asDID } : this.api
```
ukstv, line 292: Or maybe it breaks compatibility here? The API seems to be different.

ukstv (changes requested):
I see here humongous amount of good work. I am worried about two things in the PR:
1. Quite frivolous use of imports:
  - `import ... from './index.js` is quite dangerous
  - Would like to see more `import type ...` where they could be applied
 2. The change seems to break compatibility in a dev-facing code like TileDocument.

For (1) I would like to see the imports sorted out. Re (2): It might be me who misunderstands the code, but if I am right, we better figure out how to retain compatibility.

packages/core/src/ceramic.ts
dbcfd: Not in this PR. Was getting pretty big.

I can do another PR where I can pull out functionality so we're a little less circular.

packages/core/src/ceramic.ts
dbcfd: Yep, was a good indicator about how much functionality was in this class.

packages/core/src/index.ts
dbcfd: the anchor checking code needs to cast to this type for it to work. Rather than having core functionality in common and hoping the cast works, now it's explicit. But it did take an export.

packages/http-client/src/ceramic-http-client.ts
dbcfd: same as above. can split it out in this PR, it just becomes a lot bigger.

packages/stream-model-handler/src/model-handler.ts
dbcfd: More for uniformity. We can revert it, but given introduction of the new interfaces, seemed like a good time to change it.

packages/stream-model-instance-handler/src/__tests__/model-instance-document-handler.test.ts
dbcfd: I don't think it did, I think it's just weird with git. This is the encryption key for v1, while `z6LS` is for v0.

dbcfd: Here's the line it should be matched with https://github.com/ceramicnetwork/js-ceramic/pull/3062/files/8c235b419ac580a3ec623322972796a3c973dccb#diff-1d5070d6815fcf983adea9a7b864fc0fe112d98ac616adac272bd9d860758f5cL163

packages/stream-tile/src/tile-document.ts
dbcfd: Ah yeah, this would possibly break for someone, if they did opts that did something to `asDID`. Although that is mostly an internal method, someone could have a different implementation for it. I'm going to readd this method, then add a way to construct a signer from a did.

packages/common/src/ceramic-api.ts
```diff
@@ -159,65 +192,12 @@ export interface CeramicApi extends CeramicSigner {
...
-   * @param streamId - Stream ID
-   * @param opts - Initialization options
-   */
-  loadStream<T extends Stream>(streamId: StreamID | CommitID | string, opts?: LoadOpts): Promise<T>
-
   /**
    * Load all stream commits by stream ID
    * @param streamId - Stream ID
    */
   loadStreamCommits(streamId: StreamID | string): Promise<Array<Record<string, any>>>
```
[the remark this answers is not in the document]
dbcfd: Probably should be, but leaving it out of this PR.

packages/common/src/ceramic-api.ts
```diff
@@ -146,8 +130,8 @@ export interface AdminApi {
 /**
  * Describes Ceramic node API
  */
-export interface CeramicApi extends CeramicSigner {
-  // loggerProvider: LoggerProvider; // TODO uncomment once logger is available on http-client
+export interface CeramicApi {
```
stbrody, line 133: it's not really immediately clear to me why the methods that remain in this interface are still here while so many of the others have been removed.  Can you add some more comments explaining what the purpose of this interface is now going forward, and how it differs from StreamReaderWriter?

packages/common/src/stream.ts
```diff
@@ -249,7 +251,7 @@ export abstract class Stream extends Observable<StreamState> implements StreamSt
   }
 
   get state(): StreamState {
-    return cloneDeep(this.state$.value)
```
stbrody, line 252: this is a potentially a quite consequential change.  I'm open to making it, but can we separate it out of this PR into its own more targeted PR where we can discuss the implications in more depth without holding up the rest of the changes here?

packages/core/src/ceramic.ts
```diff
@@ -184,8 +185,10 @@ const tryStreamId = (id: string): StreamID | null => {
  * To install this library:<br/>
  * `$ npm install --save @ceramicnetwork/core`
  */
-export class Ceramic implements CeramicApi {
-  public readonly context: Context
+export class Ceramic implements StreamReaderWriter, StreamStateLoader {
+  _did?: DID
+  _ipfs?: IpfsApi
+  private _signer: CeramicSigner
```
stbrody, line 191: Can we add some comments explaining the relationship between `_did` and `_signer` and when you should use one or the other?

packages/core/src/state-management/stream-from-state.ts
```diff
@@ -1,19 +1,19 @@
...
- * @param context - Ceramic context
+ * @param context - Interface for reading and writing to ceramic network
  * @param handlersMap - available stream handlers
  * @param state - current state of the stream
  * @param update$ - On-demand feed of updates for the stream. If not provided then the returned
  *   Stream object is marked read-only and cannot be used to update the stream.
  */
 export function streamFromState<T extends Stream>(
-  context: Context,
+  context: StreamReaderWriter,
```
stbrody, line 16: should we rename this arg now?  Same for the other places we take a StreamReader or StreamReaderWriter and call it context

packages/common/src/context.ts
```diff
@@ -1,14 +1,12 @@
...
+import type { LoggerProvider } from './logger-provider.js'
+import type { IpfsApi } from './index.js'
+import type { CeramicSigner } from './ceramic-signer'
 
 /**
  * Encapsulates Ceramic context
  */
-export interface Context {
-  did?: DID
+export interface CeramicContext {
```
stbrody, line 8: is this interface actually even used anywhere anymore?  Can it just be deleted?

packages/core/src/__tests__/ceramic-query-response.test.ts
```diff
@@ -74,14 +74,14 @@ describe('Response to pubsub queries handling', () => {
 
     // Make sure we have some random valid CIDs to respond to pubsub queries with
     for (let i = 0; i < 10; i++) {
-      const commit = await TileDocument.makeGenesis(ceramic, { foo: i }, null)
+      const commit = await TileDocument.makeGenesis(ceramic.signer(), { foo: i }, null)
```
stbrody, line 77: this change shouldn't be necessary and indicates that we've made a breaking change.  I believe it's because `TileDocument.makeGenesis` takes the `CeramicSigner` as it's first arg, but this PR redefines what `CeramicSigner` is.  `CeramicSigner` used to extend `CeramicApi` and so passing a `Ceramic` instance was fine.  Now `CeramicSigner` is more a more specific type.  I think we need to update `TileDocument.makeGenesis` (and any other public client-facing functions that used to take the old `CeramicSigner` interface) to take a `CeramicWriter` instead.

packages/core/src/__tests__/state-manager.test.ts
```diff
@@ -146,39 +147,39 @@ describe('anchor', () => {
...
       const schemaDoc = await TileDocument.create(ceramic, STRING_MAP_SCHEMA)
-      await TestUtils.anchorUpdate(ceramic, schemaDoc)
+      await anchorUpdate(ceramic, schemaDoc)
 
       const stream = await TileDocument.create(ceramic, { stuff: 1 })
       const streamState = await ceramic.repository.load(stream.id, {})
-      await TestUtils.anchorUpdate(ceramic, stream)
-      const updateRec = await stream.makeCommit(ceramic, null, { schema: schemaDoc.commitId })
+      await anchorUpdate(ceramic, stream)
+      const updateRec = await stream.makeCommit(ceramic.signer(), null, {
```
stbrody, line 155: ditto my comment above about `TileDocument.makeGenesis`.  The fact that this change here is needed is a sign that we've made a breaking change to our public API.

packages/core/src/__tests__/ceramic-query-response.test.ts
stbrody: I think this same issue likely exists with all the other Stream clients, not just TileDocument, it's just most obvious with TileDocument because that's what our tests use the most.

packages/common/src/ceramic-api.ts
stbrody: Actually, is this interface even used anywhere anymore?  Can it be deleted?

packages/cli/src/__tests__/ceramic-error.test.ts
```diff
@@ -61,8 +61,8 @@ beforeAll(async () => {
   const apiUrl = `http://localhost:${daemonPort}`
   client = new CeramicClient(apiUrl, { syncInterval: 500 })
 
-  await core.setDID(makeDID(core))
-  await client.setDID(makeDID(client, SEED))
+  core.did = makeDID(core)
```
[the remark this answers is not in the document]
dbcfd: Given that the DID setter has functionality for signer, I've changed it in this PR, but have added a deprecated setDID method to the http client for compatibility.

packages/core/src/state-management/stream-from-state.ts
dbcfd: Yes, but not in this PR. Minimizing renaming in this PR to reduce scope.

packages/core/src/__tests__/ceramic-query-response.test.ts
```diff
@@ -75,14 +75,14 @@ describe('Response to pubsub queries handling', () => {
...
-      const commit = await TileDocument.makeGenesis(ceramic, { foo: i }, null)
+      const commit = await TileDocument.makeGenesis(ceramic.signer, { foo: i }, null)
       const cid = await ceramic.dispatcher.storeCommit(commit)
       cids.push(cid)
     }
   })
 
   beforeEach(async () => {
-    const genesisCommit = await TileDocument.makeGenesis(ceramic, { foo: 'bar' }, null)
+    const genesisCommit = await TileDocument.makeGenesis(ceramic.signer, { foo: 'bar' }, null)
```
stbrody, line 85: this still indicates there was a breaking change to our public API.  I think you updated the *most* of the methods of TileDocument, but not his one.  No test code calling methods on TileDocument should need to change as a result of this PR, if anywhere in this PR there is a change to code calling into TileDocument, that's a flag that we broke something.

dbcfd: It actually works both ways now. Some tests use `ceramic` and some use `ceramic.signer` to verify that behavior.

packages/core/src/ceramic.ts
```diff
@@ -184,8 +185,13 @@ const tryStreamId = (id: string): StreamID | null => {
  * To install this library:<br/>
  * `$ npm install --save @ceramicnetwork/core`
  */
-export class Ceramic implements CeramicApi {
-  public readonly context: Context
+export class Ceramic implements StreamReaderWriter, StreamStateLoader {
+  // Primarily for backwards compatibility around the get did call, any usages for signing
+  // or verification should be done using `_signer`. See `get did` and `get signer` for more
+  // information
+  private _did?: DID
```
stbrody, line 192: can we just use a getter function instead of storing two members?

packages/core/src/ceramic.ts
```diff
@@ -301,18 +301,29 @@ export class Ceramic implements CeramicApi {
...
 
   /**
-   * Get DID
+   * Get DID. This should only be used if you need to interrogate the DID that was set for this
+   * ceramic. For most instances, prefer `signer`, as it has the necessary information for creating
+   * and verifying signatures.
    */
   get did(): DID | undefined {
-    return this.context.did
+    return this._did
```
stbrody, line 326: can we get this from the signer and remove the need to keep the `_did` member?

packages/http-client/src/ceramic-http-client.ts
```diff
@@ -72,17 +78,18 @@ export class CeramicClient implements CeramicApi {
   public readonly pin: PinApi
   public readonly admin: AdminApi
   public readonly index: IndexApi
-  public readonly context: Context
+  private _signer: CeramicSigner
+  private _did?: DID
```
stbrody, line 82: same as for `Ceramic`.  Can we remove the need to store both the `_signer` AND the `_did`?  If not, let's at least add comments here too as to why we have both and when developers should use which one.

packages/common/src/stream.ts
```diff
@@ -321,8 +323,12 @@ export interface StreamHandler<T extends Stream> {
...
    * state modified.
    * @param commitData - Commit data
-   * @param context - Ceramic context
+   * @param api - Interface for reading from and writing streams to ceramic network
    * @param state - The existing state to apply the commit to.
    */
-  applyCommit(commitData: CommitData, context: Context, state?: StreamState): Promise<StreamState>
+  applyCommit(
+    commitData: CommitData,
+    api: StreamReaderWriter,
```
stbrody, line 331: I just realized that this should really be a `StreamReader`, and I think this is a great opportunity to tighten the interface here.  It's always made me nervous the fact that this handler had access to a full Ceramic API.  But since this PR is so big and changing this here requires changing it in all the stream handler packages as well, feel free to just add a TODO comment right now and then make the actual change in a follow-up PR.

stbrody (approved):
lgtm

ukstv (approved):

PR #3066, opened 2023-12-12
fix: Make AnchorRequestStore processing loop more robust

stbrody:
Adds logs and exception handling.  Also increases restart delay from 100ms to 1 second to avoid hammering S3 too often.


ukstv (approved):

ukstv:
Seems useful and innocent.

PR #3067, opened 2023-12-13
fix(indexing): support adding indices to existing models

PaulLeCam:
Should solve the issue reported by multiple users attempting to add indices to existing models, by checking existing indices and only creating the necessary ones.
This also unifies the indexing logic with relations to avoid having two different systems and possibly a duplication of the indexed data for handling relations and custom fields.


packages/indexing/src/tables-manager.ts
```diff
@@ -162,6 +154,35 @@ export class TablesManager {
...
+  }
+
+  /**
+   * Migrate a MID table
+   * @param tableName
+   */
+  async _migrateMidTable(tableName: string): Promise<void> {
+    const columns = await this.dataSource.table(tableName).columnInfo()
+    const relationColumns = Object.keys(columns).filter((name) => {
+      return name.startsWith(CUSTOM_COLUMN_PREFIX)
```
dbcfd, line 176: I see where we check for this, but not anything that indicates that a column in the table should not start with this prefix, otherwise it will get removed.

packages/indexing/src/tables-manager.ts
```diff
@@ -325,41 +332,38 @@ export class SqliteTablesManager extends TablesManager {
...
-    this.logger.imp(`Creating ComposeDB Indexing table for model: ${tableName}`)
-    await createSqliteModelTable(this.dataSource, tableName, relationColumns)
-    if (modelIndexArgs.indices) {
-      await createSqliteIndices(this.dataSource, tableName, modelIndexArgs.indices)
+    if (!existingTables.includes(tableName)) {
+      this.logger.imp(`Creating ComposeDB Indexing table for model: ${tableName}`)
+      await createSqliteModelTable(this.dataSource, tableName)
     }
+
+    await createSqliteIndices(this.dataSource, tableName, createIndices)
```
dbcfd, line 366: Given there's a promise involved here, it would be better to have this be only called if there's indices to create.

packages/indexing/src/tables-manager.ts
```diff
@@ -238,44 +254,35 @@ export class PostgresTablesManager extends TablesManager {
...
-          if (column.type === ColumnType.STRING) {
-            const columnName = addColumnPrefix(column.name)
-            table.string(columnName, 1024).nullable().alter()
-          }
-        }
-      })
+      await createPostgresModelTable(this.dataSource, tableName)
     }
+
+    await createPostgresIndices(this.dataSource, tableName, createIndices)
```
dbcfd, line 285: Given there's a promise involved here, it would be better to have this be only called if there's indices to create.

dbcfd (changes requested):

packages/indexing/src/database-index-api.ts
```diff
@@ -115,49 +144,81 @@ export abstract class DatabaseIndexApi<DateType = Date | number> {
...
+    const indexedModels: Record<string, ModelData> = cloneDeep(this.indexedModelsRecord)
+    // Models to index with merged indices
+    const modelsToIndex: Array<IndexModelArgs> = []
+    // Record of indices to create
+    const createIndices: CreateIndicesRecord = {}
+    // Checks for models in sync
+    const checkSyncPromises: Array<Promise<void>> = []
+
+    for (const modelData of models) {
+      // Check model is not currently going through historical sync
```
gvelez17, line 157: what will we need to do when it does?  is there a todo here ?

packages/indexing/src/database-index-api.ts
```diff
@@ -115,49 +144,81 @@ export abstract class DatabaseIndexApi<DateType = Date | number> {
...
+          model: indexModelArgs.model.toString(),
+          ...(indexModelArgs.indices && { indices: JSON.stringify(indexModelArgs.indices) }),
           is_indexed: true,
           updated_by: '0', // TODO: FIXME: CDB-1866 - <FIXME: PUT ADMIN DID WHEN AUTH IS IMPLEMENTED>',
-        })
+          updated_at: this.now(),
+        }
+        return await trx(INDEXED_MODEL_CONFIG_TABLE_NAME)
+          .insert(data)
+          .onConflict('model')
```
gvelez17, line 218: constant for this string?

packages/indexing/src/tables-manager.ts
```diff
@@ -32,17 +31,7 @@ type ConfigTable = {
...
- */
-function relationsDefinitionsToColumnInfo(relations?: ModelRelationsDefinition): Array<ColumnInfo> {
-  if (!relations) {
-    return []
-  }
-  return Object.keys(relations).map((keyName) => {
-    return { name: addColumnPrefix(keyName), type: ColumnType.STRING }
-  })
-}
+const CUSTOM_COLUMN_PREFIX = 'custom_'
```
gvelez17, line 34: what do we mean by 'custom' is this related to autogenerated columns?  are they private to us?

gvelez17:
It looks ok to me but I'm not sure I have enough context to confidently approve - could walk thru it live?

packages/indexing/src/database-index-api.ts
```diff
@@ -115,49 +144,81 @@ export abstract class DatabaseIndexApi<DateType = Date | number> {
...
+      const modelIndices = Object.values(modelIndicesRecord)
+
+      const id = modelData.model.toString()
+      if (indexedModels[id] == null) {
+        // Update indexed models record
+        indexedModels[id] = { streamID: modelData.model, indices: modelIndices }
+        // Add indices to create
+        createIndices[id] = modelIndices
+      } else {
+        const existingIndices = toIndicesRecord(indexedModels[id]?.indices)
```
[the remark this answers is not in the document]
PaulLeCam: Can you create a ticket for this please?

packages/indexing/src/tables-manager.ts
PaulLeCam: Only relation fields use this prefix, we don't have any other column that uses it.

packages/indexing/src/database-index-api.ts
PaulLeCam: If there is any sync in progress, the indices creation will fail, that is covered by https://github.com/ceramicnetwork/js-ceramic/pull/3067/files#diff-bfb31d44a4dee8c373a193980ab3484df1782c7d3ab07c5c5b54e47886ef7aeaR192 that ensures the attempt to sync

packages/indexing/src/database-index-api.ts
PaulLeCam: It's based on the structure above so I think it's clear as-is?
It's not logic I changed, only moved around.

packages/indexing/src/database-index-api.ts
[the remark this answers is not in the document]
dbcfd: Please add to this PR, and if needed in js-composedb, create a ticket.

packages/indexing/src/database-index-api.ts
gvelez17: ok should we add a comment about what happens whether the indexes are created later or they need to be manually created, like will this be ok or does someone need to do something?

packages/indexing/src/tables-manager.ts
gvelez17: ok - i would have prefered like _INTERNAL or something but sure :-)

packages/indexing/src/database-index-api.ts
PaulLeCam: It should throw an error if the sync is in progress, so the whole operation will fail and the person should try again later or stop the sync before trying again.

gvelez17 (approved):
From my perspective, lgtm

packages/indexing/src/database-index-api.ts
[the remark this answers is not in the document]
dbcfd: We don't have to do the force flag in this PR, but please add a log about updating indices.

dbcfd (approved):

PR #3068, opened 2023-12-14
chore: Link HTTP-internal feed

ukstv:


packages/core/src/feed.ts
```diff
@@ -42,12 +46,12 @@ class DocumentsSubject extends Observable<Document> {
  */
 export interface PublicFeed {
   aggregation: {
-    streamStates: Observable<Document>
+    documents: Observable<Document>
```
JulissaDantes, line 49: `documents` its a much better name, thanks for catching that!

packages/cli/src/ceramic-daemon.ts
```diff
@@ -120,50 +123,6 @@ export function makeCeramicConfig(opts: DaemonConfig): CeramicConfig {
...
-  return typedOpts
-}
-
-/**
- * Converts 'sync' option sent as a bool by outdated http-clients to the current format of an enum.
- * The old behaviors don't map directly to the new behaviors, so we take the best approximation.
- * TODO remove this once we no longer need to support clients older than v1.0.0
- * @param opts
- */
-function upconvertLegacySyncOption(opts: Record<string, any> | undefined) {
```
JulissaDantes, line 157: Just as a double check: Are we sure we don't have any clients running a version older than v1.0.0?

packages/cli/package.json
```diff
@@ -85,6 +85,7 @@
     "@types/levelup": "^5.1.0",
     "@types/node": "^18.0.3",
     "@types/stream-to-array": "^2.3.0",
+    "eventsource": "^2.0.2",
```
JulissaDantes, line 88: Where is this library being used?

packages/cli/src/ceramic-daemon.ts
ukstv: We are! Few months ago we waited until 90% of the nodes run 2.37 or something.

packages/core/src/feed.ts
ukstv: It is what the spec says :)

PR #3070, opened 2023-12-19
feat(stream-model-instance,stream-model-instance-handler): Allow custom values for deterministic MIDs

PaulLeCam:
@stbrody @ukstv following the discussions we had about supporting the SET account relation, here is a POC of the way I think it could work:

1. The MID logic allows setting a `unique` field in the metadata, which is an array of strings
2. When the `deterministic` flag is set on the metadata, the `unique` array of string is converted to a `Uint8Array` and used as the `unique` header value
3. The MID handler adds the `unique` value provided in header to the stream metadata so it can be checked later
4. When using a SET account relation, the MID handler validates that the `unique` metadata value matches the content fields defined, which ensures that the JSON schema validation of the content also indirectly applies to the deterministic value

The main reason for using an array of strings for the `unique` value set by the client is to ensure a consistent ordering in case multiple field values are used, and always using UTF-8 strings before converting to `Uint8Array` to avoid possible confusion about encodings.
The changes also make the MID content `null` instead of setting an empty object to make the behavior more consistent, for example avoiding loading a doc with expected required fields that would not be present in the object. Having a `null` content makes it explicit to the consumer there has been no content set yet.

What do you think please? Any concern about the behavior?


oed:
> The changes also make the MID content null instead of setting an empty object to make the behavior more consistent

Wouldn't this break determinism for existing deterministic MIDs with the SINGLE relation?

PaulLeCam:
> > The changes also make the MID content null instead of setting an empty object to make the behavior more consistent
> 
> Wouldn't this break determinism for existing deterministic MIDs with the SINGLE relation?

As far as I understand, the [genesis commit creation uses `null` as the `data` value](https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/stream-model-instance/src/model-instance-document.ts#L324), considering [the `single()` call exiplicitly sets the content to `null` when calling `_makeGenesis()`](https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/stream-model-instance/src/model-instance-document.ts#L151), it's only the [MID handler that uses an empty object when creating the `StreamState`](https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/stream-model-instance-handler/src/model-instance-document-handler.ts#L124).
Do you think it's OK or should we have explicit tests for this?

oed:
@PaulLeCam Ok makes sense. Yes we should definitely have a unit test for this if we don't have already!

packages/stream-model-instance-handler/src/__tests__/model-instance-document-handler.test.ts
```diff
@@ -258,6 +261,29 @@ const MODEL_DEFINITION_SINGLE: ModelDefinition = {
   },
 }
 
+const MODEL_DEFINITION_SET: ModelDefinition = {
+  name: 'MyModel',
+  version: '2.0',
+  interface: false,
+  implements: [],
+  accountRelation: { type: 'set', fields: ['one', 'two'] },
```
stbrody, line 269: Hmm, I don't actually see where in this PR we update the ModelDefinition to support this.  I believe the RFC only called for supporting a single field and calling it `accountRelationField`.  I'm down to change this to take an array of fields, that seems better and more flexible.  But A) we should update the RFC and B) I'm wondering if we can make the field name more descriptive.  `accountRelationFields` would be the logical change from what's in the RFC.  I also kind of like `uniqueFields`.

FYI @rohhan

packages/stream-tests/src/__tests__/deterministic-model-instance-document.test.ts
```diff
@@ -130,6 +154,60 @@ describe('ModelInstanceDocument API http-client tests', () => {
...
+      })
+      await expect(() => doc.replace({ unique: 'a', other: 'test' })).rejects.toThrow(
+        /Validation Error: data\/unique must NOT have fewer than 3 characters/
+      )
+    })
+
+    test('Can create, load and update unique documents', async () => {
+      const model = await Model.create(ceramic, MODEL_DEFINITION_SET)
+
+      const doc1 = await ModelInstanceDocument.single(ceramic, {
```
stbrody, line 190: hmm, I don't like that this function is called `single` now that it's being used for more than just the SINGLE account relation.  Can we add a different function, `set`, which always takes a `unique` field and remove the optional `unique` arg from the `single` function?

PaulLeCam:
@ukstv @stbrody should be OK to merge now if you can have a look please?

packages/stream-model-instance-handler/src/__tests__/model-instance-document-handler.test.ts
```diff
@@ -797,6 +857,107 @@ describe('ModelInstanceDocumentHandler', () => {
     expect(state).toMatchSnapshot()
   })
 
+  it('MIDs with SET account relation validate signed commit fields', async () => {
+    const genesisCommit = (await ModelInstanceDocument._makeGenesis(context.api, null, {
+      ...DETERMINISTIC_METADATA,
+      model: FAKE_MODEL_SET_ID,
+      unique: ['one', 'two'],
```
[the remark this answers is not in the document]
PaulLeCam: Yes it's only the values, I can change that in the test if it's confusing.
And yes the values must be provided in the order defined in the model.

packages/stream-model-instance-handler/src/__tests__/model-instance-document-handler.test.ts
```diff
@@ -797,6 +857,107 @@ describe('ModelInstanceDocumentHandler', () => {
...
+    await context.ipfs.dag.put(genesisCommit, FAKE_CID_1)
+
+    // apply genesis
+    const genesisCommitData = {
+      cid: FAKE_CID_1,
+      type: CommitType.GENESIS,
+      commit: genesisCommit,
+    }
+
+    // Genesis creation should always work independently of content validation
```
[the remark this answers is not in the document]
PaulLeCam: I don't think it's really a quirk, it's really about how deterministic generation works: we don't really "create" the stream but rather generate the header so we can derive the stream ID to access. When the genesis commit is created/accessed, it can't have content anyway, it's the same behavior with the SINGLE account relation.

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
```diff
@@ -308,4 +317,37 @@ export class ModelInstanceDocumentHandler implements StreamHandler<ModelInstance
...
+        throw new Error('Missing content')
+      }
+
+      const unique = model.content.accountRelation.fields
+        .map((field) => {
+          const value = content[field]
+          return value ? value.toString() : ''
+        })
+        .join('|')
+      if (unique !== toString(metadata.unique)) {
```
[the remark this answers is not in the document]
PaulLeCam: Yes that's why all the values are cast to string 3 lines above, and the `unique` metadata value is an array of strings.

I don't think it's a problem that `1` (integer) is converted to `"1"` (string) as the fields types are enforced by the JSON schema anyway, as long as the conversion to string is stable. Maybe we can have an additional check in the model creation to ensure the SET fields are only defined on scalar fields?

packages/stream-model-instance/src/model-instance-document.ts
```diff
@@ -40,6 +41,12 @@ export interface ModelInstanceDocumentMetadataArgs {
    * ModelInstanceDocuments whose Model has an accountRelation of 'SINGLE'.
    */
   deterministic?: boolean
+
+  /**
+   * Custom values to use for the stream determinism, only used if `deterministic` is set to `true`.  Should only be used for
+   * ModelInstanceDocuments whose Model has an accountRelation of 'SET'.
+   */
+  unique?: Array<string>
```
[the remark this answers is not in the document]
PaulLeCam: It's possible as long as they get converted to UTF-8 strings first, to ensure there is no ambiguity in the encoding.

packages/stream-model-instance-handler/src/__tests__/model-instance-document-handler.test.ts
PaulLeCam: The model definition was already updated as part of the work on interfaces, v2.0 supports the SET account relation as used here.

packages/stream-model-instance/src/model-instance-document.ts
```diff
@@ -317,7 +329,14 @@ export class ModelInstanceDocument<T = Record<string, any>> extends Stream {
       model: metadata.model.bytes,
       sep: 'model', // See CIP-120 for more details on this field
     }
-    if (!metadata.deterministic) {
+    if (metadata.deterministic) {
+      // Convert and use unique values for the deterministic bytes if provided (SET account relation)
```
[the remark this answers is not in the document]
PaulLeCam: Yeah I was wondering about this as well, but I don't think it's really an issue? If the client for some reason decides to provide the `unique` value, it still doesn't prevent the expected behavior of the LIST account relation to allow the creation of any number of MIDs.

packages/stream-model-instance/src/model-instance-document.ts
```diff
@@ -168,6 +162,31 @@ export class ModelInstanceDocument<T = Record<string, any>> extends Stream {
...
+   * Creates a deterministic ModelInstanceDocument with a 'set' accountRelation.
+   * @param ceramic - Instance of CeramicAPI used to communicate with the Ceramic network
+   * @param metadata - Genesis metadata
+   * @param unique - Unique fields values
+   * @param opts - Additional options
+   */
+  static async set<T>(
+    ceramic: CeramicApi,
+    metadata: ModelInstanceDocumentMetadataArgs,
+    unique: Array<string>,
```
stbrody, line 175: hmm I'm a bit worried about the arguments being ordering-dependent here.  I'm afraid it will be error-prone to users, and requires them to keep track of what order the fields were specified in the Model definition.  Can we instead make this a dictionary where the user has to specify the field names of the fields that get the uniqueness properties?  I'm fine if it still get encoded in an ordering-dependent way into the commit itself, but for the user-facing APIs I'd like to make them a bit harder to misuse by accident.

packages/stream-model-handler/src/account-relations-utils.ts
```diff
@@ -0,0 +1,48 @@
+import { JsonReference } from 'json-ptr'
+import type { JSONSchema } from 'json-schema-typed/draft-2020-12'
+
+type FieldSchema = Exclude<JSONSchema, boolean>
+
+const SUPPORTED_FIELD_TYPES = ['boolean', 'integer', 'number', 'string']
+
+export function validateSetFields(fields: Array<string>, modelSchema: JSONSchema.Object): void {
```
stbrody, line 8: Can we add a quick comment on what the "fields" array represents and what this function is doing?

packages/stream-model-handler/src/account-relations-utils.ts
```diff
@@ -0,0 +1,48 @@
...
+    throw new Error(`At least one field must be defined for the SET account relation`)
+  }
+
+  const requiredFields = modelSchema.required ?? []
+
+  const errors: Array<Error> = []
+  for (const field of fields) {
+    let fieldSchema = modelSchema.properties[field] as FieldSchema | undefined
+    if (fieldSchema?.$ref != null) {
+      const ref = new JsonReference(fieldSchema.$ref)
```
stbrody, line 19: I don't really understand what this block is doing?  Can you maybe add some comments here?

packages/stream-model-instance/src/model-instance-document.ts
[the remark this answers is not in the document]
stbrody: Is that how users use StreamIDs today?  How about for relations?  It seems like wanting SET semantics on a StreamID or DID relation will probably be a common use case.

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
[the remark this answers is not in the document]
stbrody: I think if instead of the `unique` field in the metadata always being a single string, it was allowed to be an array where each entry corresponded to one of the unique fields, then we could preserve the type information of each field and also allow things like StreamIDs to be used as unique fields

packages/stream-model-instance-handler/src/__tests__/model-instance-document-handler.test.ts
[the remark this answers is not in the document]
PaulLeCam: 👍 yes makes sense, I'll update the comment

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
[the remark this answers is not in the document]
PaulLeCam: The `unique` field in the metadata is a `Uint8Array`, here the logic converts all values to UTF-8 strings and then convert the concatenated string to `Uint8Array` to ensure the encoding is deterministic, but the purpose of the `unique` field is only to enforce determinism, no? How would it be needed or useful to have any type information about the values there?

packages/stream-model-instance/src/model-instance-document.ts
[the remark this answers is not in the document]
PaulLeCam: StreamIDs are encoded as string in the content, it has to be valid JSON. We can apply SET semantics on any scalar field, whether they are used for relations or not shouldn't be relevant to the SET logic.

packages/stream-model-instance/src/model-instance-document.ts
PaulLeCam: I don't mind changing it, but fundamentally the logic should be based on the `accountRelation.fields` array of the model anyway, for example as defined currently:

```ts
const unique = model.accountRelation.fields.map(field => doc.content[field] ?? '')
```

Then the handler side just needs to do something like:

```ts
const bytes = fromString(unique.join('|'))
```

Changing it to a dictionary would require the client to still use the `accountRelation.fields` as input anyway, for example:

```ts
const unique = model.accountRelation.fields.reduce((acc, field) => {
  acc[field] = doc.content[field] ?? ''
  return acc
}, {})
```

Then the handler side would need to convert it back to an array based on the model:

```ts
const uniqueParts = model.accountRelation.fields.map(field => {
  const value = unique[field]
  if (value == null) {
    throw new Error(`Missing set field value: ${field}`)
  }
  return value
})
const bytes = fromString(uniqueParts.join('|'))
```

The way I see it, this `set` method is going to be abstracted away by ComposeDB so the risk here is more a bad client implementation than a user/developer mistake and it's pretty limited, but if you feel strongly about it I can make the change?

stbrody: Yeah, I agree that for users going through ComposeDB this will be abstracted away for them, but I do think it's worth trying to design these APIs so they'll be easy to use for others building their own indexing solutions or trying to use the stream clients directly.

For a custom indexer, I guess it's probably true that they would likely have access to an instance of the corresponding Model and will be using the Model definition to figure out which fields to extract into the "unique" metadata anyway.  But if someone were building an app with the ModelInstanceDocument client directly without going through some other ComposeDB-like middleware, I could see the app just *knowing* which fields are the unique fields and providing them manually.

I'm conflicted, I do think most users are going to be going through some middleware like ComposeDB, OrbisDB, etc. Using the Stream clients directly is going to be rare.  That said, I do still feel nervous about client-facing APIs that are ordering-dependent based on some external context not necessarily available at the time that the API is being called...

I think I still lean slightly towards using a dictionary for this, but I do see your point and think the choice isn't as clear-cut as I first thought.  Maybe I'll tap in @ukstv and @dbcfd to see if either of them want to break the tie on this one.

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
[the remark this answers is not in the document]
stbrody: Well I'm thinking from the perspective of enforcing that the unique fields can't be changed.  Those fields should be immutable (essentially there's an implicit `@immutable` tag on them like we discussed in the field locking spec).  If the user can change the value from the number `1` to the string `"1"`, now that field isn't guaranteed immutable anymore.

That actually makes me think, do we actually have checks that these aren't changed?  We check that they match the corresponding values in the content, but what if there's a commit that updates the metadata and the content at the same time?  We should make sure we have a test for that and I don't believe the code will catch that case currently, unless I missed it.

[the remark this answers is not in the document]
PaulLeCam: Immutability is more a side-effect of the deterministic behavior, the validation here is for the `unique` data, not so much the field types or other constraints.
Considering the work on immutable fields happens in parallel, what do you think if we make it part of the immutable fields logic to implicitly enforce the validation for the SET account relation fields?

PaulLeCam:
@stbrody how can we reach a decision on the next steps for this PR please? I don't mind making changes if there's consensus, but I'd like to make sure that would get this PR closer to getting merged rather than leading to more uncertainties.

packages/stream-model-instance/src/model-instance-document.ts
stbrody: @dbcfd there's no way for our library to automatically sort or enforce sort ordering here, since our tooling doesn't have the context to know the app-specific meaning of the values being passed in.  That's why I suggested using a dictionary instead of an array where sort order matters, because then the developer is specifying the meaning of each argument explicitly, instead of implicitly via sort order

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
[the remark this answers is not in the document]
stbrody: Um, I do feel like the immutability is pretty core to the feature.  If you can change the value of the field you are using for the set semantics, the correct behavior of the system gets pretty hard to reason about.  I'm fine with doing a follow-up PR to add enforcement of that immutability, but I wouldn't really want to release SET account relation without it.

packages/stream-model-handler/src/account-relations-utils.ts
stbrody: I still don't entirely understand, but I'm just going to chalk this up to some jsonSchema shenanigans and move on 😅

packages/stream-model-instance-handler/src/__tests__/model-instance-document-handler.test.ts
stbrody: in our call last week we decided to call this `accountRelationFields` I believe, instead of just `fields`. Okay if you want to do this rename in a follow-up

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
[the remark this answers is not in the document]
stbrody: Actually, looks like we are already protected from this: https://github.com/ceramicnetwork/js-ceramic/blob/fa3e28c63b268359af5349a58dbdfea1d8992c53/packages/stream-model-instance-handler/src/model-instance-document-handler.ts#L155-L161.  So no concern here after all then.

stbrody (commented):
okay, other than the rename from `fields` -> `accountRelationFields`, which can go in a follow-up commit, the only remaining comment from me is the one around changing the client library to take the fields as a dictionary instead of an array where sort order matters.  I still prefer dictionary, but if I haven't managed to convince you I'm okay with going forward with the array approach.

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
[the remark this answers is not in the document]
PaulLeCam: You mean mutability of the `unique` field? If so yes for sure it should be immutable, I thought you were talking about the fields from which the `unique` field is derived that would need stronger validation than checking their string values, should be all good then 👍

packages/stream-model-instance-handler/src/__tests__/model-instance-document-handler.test.ts
PaulLeCam: I think there are 2 things here: the model structure definition (codec) and the GraphQL schema definition.
The model structure is already implemented to support `accountRelation: { type: 'set', fields: Array<string> }`, that was done along with the interfaces change to update the model definition version from v1.0 to v2.0.
Regarding the GraphQL schema, we have the `@createIndex(description: String!, accountRelation: ModelAccountRelation)` that needs to be updated to `@createIndex(description: String!, accountRelation: ModelAccountRelation, accountRelationFields: [String!])`, but that's on the ComposeDB tooling side, that itself creates the models from the GraphQL schema.
So basically at a high level yes we'll have `accountRelationFields`, but internally for the model structure these fields are specified along with the relation `type` in the `accountRelation` object.

packages/stream-model-instance/src/model-instance-document.ts
PaulLeCam: @stbrody to be clear, I agree that at a high level we should use an object with explicit key/value pairs for the unique fields, that's what ComposeDB will expose in the runtime GraphQL schema, but there is a need to convert to an array of strings at some point.
So to me the question is more if we need to have structured object -> intermediary dictionary -> array of strings or if we can have structured object -> array of strings directly, which simplifies the implementation?

packages/stream-model-handler/src/account-relations-utils.ts
PaulLeCam: Here is an example schema generated by the ComposeDB tools:

```json
{
  "$defs": {
    "ImageMetadata": {
      "additionalProperties": false,
      "properties": {
        "height": {
          "minimum": 1,
          "type": "integer",
        },
        "mimeType": {
          "maxLength": 50,
          "type": "string",
        },
        "size": {
          "minimum": 1,
          "type": "integer",
        },
        "src": {
          "maxLength": 150,
          "type": "string",
        },
        "width": {
          "minimum": 1,
          "type": "integer",
        },
      },
      "required": [
        "src",
        "mimeType",
        "width",
        "height",
      ],
      "title": "ImageMetadata",
      "type": "object",
    },
    "ImageSources": {
      "additionalProperties": false,
      "properties": {
        "alternatives": {
          "items": {
            "$ref": "#/$defs/ImageMetadata",
          },
          "maxItems": 20,
          "type": "array",
        },
        "original": {
          "$ref": "#/$defs/ImageMetadata",
        },
      },
      "required": [
        "original",
      ],
      "title": "ImageSources",
      "type": "object",
    },
  },
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "additionalProperties": false,
  "properties": {
    "background": {
      "$ref": "#/$defs/ImageSources",
    },
    "description": {
      "maxLength": 420,
      "type": "string",
    },
    "emoji": {
      "maxLength": 2,
      "type": "string",
    },
    "url": {
      "maxLength": 240,
      "type": "string",
    },
  },
  "type": "object",
}
```

In the `$defs` object we can have shared types that can be used (referenced) by other types in the schema, for example `ImageSources` uses `ImageMetadata` in 2 places, and the `background` property of the schema uses `ImageSources`.
This notably allows to avoid duplication in the JSON schema and therefore also in the runtime GraphQL schema.

So here regarding the logic itself, we use `JsonReference` to resolve the reference in order to get the actual structure of the field so we can check its type.

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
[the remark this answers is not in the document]
stbrody: oh you're right, I did mean that 😅.  I forgot what my concern was 😛 

There is still the case where if the json schema allows string or integer, the user could change the value from one to the other without any error.  That said, this approach we're using here would already make it so that if the user had a set account relation on such a field, the string value MID and number value MID would resolve to the same stream because the unique metadata for them would look identical.  So that seems like actually the bigger problem and extra validation on commit application wouldn't help with that.  I'm still not thrilled about it, and would prefer we use a metadata encoding with stronger typing.

But also, type safety in javascript is kind of a joke already anyway.  I think in the future we're going to need streamtypes that aren't based on jsonPatch, to not use json-to-string serialization/deserialization in the http api, and use something with much stronger type guarantees (like Protocol Buffers) as the underlying data encodings throughout.

As long as we remain in the javascript-centric world we are in right now, I think few of our users are likely to have real problems as a result of this.  So I think I'm fine moving forward with this as-is for now, even though it makes me a bit sad.

packages/stream-model-instance-handler/src/__tests__/model-instance-document-handler.test.ts
stbrody: The model streamtypes are still part of the public API.  Especially with things like OrbisDB coming online, not everyone is going to be going through the ComposeDB tooling.

Should be fine to change the model definition right?  Not like anyone can be dependent on it yet.

PaulLeCam: Yes we can change it but it's a pretty heavy change, it requires updating lots of types and handlers in ComposeDB to support the format change as a new major version of the model definition.

Having the `fields` along the `type` in the `accountRelation` object was intended from the beginning, that's why we changed `accountRelation` to be an object rather than a string when we started using codeco, so we can strongly type any additional information related to the account relation type.

Basically, with the current format we have something like:
```ts
type AccountRelationV2 = { type: 'list' } | { type: 'single' } | { type: 'set', fields: Array<string> }

type ModelDefinitionV2 = {
  version: '2.0',
  description: string,
  accountRelation: AccountRelationV2,
  // ... other fields
}
```
If we moved the `accountRelationFields` to the root, we'd need something like:
```ts
type ModelRelationV2 =
| {
  version: '2.0',
  description: string,
  accountRelation: { type: 'list' | 'single' },
  // ... other fields
}
| {
  version: '2.0',
  description: string,
  accountRelation: { type: 'set' },
  accountRelationFields: Array<string>,
  // ... other fields
}
```
It doesn't make sense to me to extract the `accountRelationFields` from the `accountRelation` object, it adds complexity without bringing any benefit IMO.

stbrody: Yep, that makes sense.  Sorry, I didn't notice that this was bundled within the accountRelation object.  Given that, from context it's clear that "fields" applies to the account relation, no need to rename it.  And definitely agree we don't want to move it up to the top level.  I think I wasn't fully processing the distinction between the `@createModel` directive semantics from the RFC and the Model definition structure.  Thanks for clarifying.

gvelez17:
hey the long delays in here caused conflicts. @bigs @dbcfd @stbrody can we please make a norm that if there are discussions like this we schedule a live sync and review and resolve quickly

bigs:
> hey the long delays in here caused conflicts. @bigs @dbcfd @stbrody can we please make a norm that if there are discussions like this we schedule a live sync and review and resolve quickly

Yes, thanks so much for calling this out. @stbrody @dbcfd @PaulLeCam y'all should meet TODAY to hash this out live. Goal should be to leave the session with commitment (not consensus!) on the way forward so Paul can close this out.

bigs:
@stbrody Awesome, appreciate that!

packages/stream-model-handler/src/account-relations-utils.ts
```diff
@@ -0,0 +1,54 @@
+import { JsonReference } from 'json-ptr'
+import type { JSONSchema } from 'json-schema-typed/draft-2020-12'
```
gvelez17, line 2: is this a standard just curious

packages/stream-model-instance-handler/src/__tests__/model-instance-document-handler.test.ts
```diff
@@ -739,6 +803,112 @@ describe('ModelInstanceDocumentHandler', () => {
...
+    await ipfs.dag.put(payloadFail, signedCommitFail.jws.link)
+
+    const signedCommitDataFail = {
+      cid: FAKE_CID_2,
+      type: CommitType.SIGNED,
+      commit: payloadFail,
+      envelope: signedCommitFail.jws,
+    }
+    await expect(handler.applyCommit(signedCommitDataFail, context, state)).rejects.toThrow(
+      'Unique content fields value does not match metadata. If you are trying to change the value of these fields, this is causing this error: these fields values are not mutable.'
```
gvelez17, line 844: i would like the error to include the field name in question this would be helpful to developers

packages/stream-model-instance-handler/src/__tests__/model-instance-document-handler.test.ts
```diff
@@ -739,6 +803,112 @@ describe('ModelInstanceDocumentHandler', () => {
...
+
+    const signedCommitData = {
+      cid: FAKE_CID_2,
+      type: CommitType.SIGNED,
+      commit: payload,
+      envelope: signedCommit.jws,
+    }
+    // MID can't be updated with content that does not pass schema validation
+    await expect(handler.applyCommit(signedCommitData, context, state)).rejects.toThrow(
+      'Validation Error: data/one must NOT have fewer than 2 characters, data/two must NOT have fewer than 2 characters'
```
gvelez17, line 908: what is data/one and data/two ?  I think this would be a confusing error message to me as a developer

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
```diff
@@ -114,7 +114,7 @@ export class ModelInstanceDocumentHandler implements StreamHandler<ModelInstance
         modelStreamID,
         streamId
       )
-    } else if (payload.data || payload.header.unique) {
```
gvelez17, line 117: curious what payload.header.unique did before, how does this relate to set/unset?

packages/stream-model-instance/src/model-instance-document.ts
gvelez17: lets put this into a ticket for future improvement i think.

packages/stream-model-instance/src/model-instance-document.ts
```diff
@@ -273,8 +305,12 @@ export class ModelInstanceDocument<T = Record<string, any>> extends Stream {
...
    * Helper function for makeUpdateCommit() to allow unit tests to update the commit before it is signed.
    */
-  private static _makeRawCommit<T>(prev: CommitID, oldContent: T, newContent: T | null): RawCommit {
-    const patch = jsonpatch.compare(oldContent, newContent || {})
+  private static _makeRawCommit<T>(
+    prev: CommitID,
+    oldContent: T | null,
+    newContent: T | null
+  ): RawCommit {
+    const patch = jsonpatch.compare(oldContent ?? {}, newContent ?? {})
```
gvelez17, line 313: is the use of jsonpatch entirely new or just moved here?

packages/stream-model-instance/src/model-instance-document.ts
stbrody: If we were to do this later it would be a breaking change.  I think if we want this to be the API we would need to change it now.

gvelez17 (approved):
Approved with caveat that we should create a ticket for suggested improvements (to discuss and/or implement them).  I think this is good to move forward.

The tests look good, i don't see where the link for test coverage is.  Assuming that coverage of new code is 100% I'm good with this.

gvelez17:
> We talked about this on discord already, I registered my opinion but am okay with merging this as-is. We're just waiting on another person to review

ok but that was totally unclear from the threads above in this PR, and i was not aware this was waiting on another person - who was that other person and when were they pinged?  i dunno just the number of delays above really raised flags imho.  i'd like us to do more live sync reviews.  that's just my opinion we can take it up elsewhere :-)

gvelez17:
@PaulLeCam this auto-merged, so if you don't mind following up with ticket for improvements that would be great or i can

not sure where code cov link is

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
PaulLeCam: For the "single" account relation the `unique` field must not be set, but now with the "set" account relation it should be

packages/stream-model-instance/src/model-instance-document.ts
PaulLeCam: Just moved, we were already using it

packages/common/src/utils/stream-utils.ts
```diff
@@ -113,6 +117,10 @@ export class StreamUtils {
       cloned.header.model = uint8arrays.fromString(cloned.header.model, 'base64')
     }
```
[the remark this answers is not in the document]
PaulLeCam: it's just for encoding to JSON, we do the same thing above for the `model` value

packages/stream-model-instance-handler/src/__tests__/model-instance-document-handler.test.ts
PaulLeCam: This is the error message thrown by AJV (the JSON schema validation library), we can customize them I think but that would likely need a dedicated project.

packages/stream-model-instance-handler/src/__tests__/model-instance-document-handler.test.ts
PaulLeCam: I guess we could extract the information from the existing `unique` field to detect what field(s) are affected, but hopefully it's not an error that would happen much, as ComposeDB would catch the issue before doing the mutation.

PR #3071, opened 2023-12-20
chore: Close connections on Ctrl+C

ukstv:
Now even if SSE feed is connected, the daemon could be closed.


JulissaDantes (approved):
LGTM!

PR #3072, opened 2023-12-21
fix: make Repository.fromMemoryOrStore safe by default

stbrody:
There are three codepaths that are calling fromMemoryOrStore today without using the LoadingQueue.  This PR fixes 2 of them: stream unpin, and the `update$` function that is used to subscribe to updates on a Stream handle.  The third path is when checking if an incoming pubsub message is for a stream we care about or not.  Since fixing that one would be a bit trickier to do in a way that doesn't degrade performance, and would complicate the (already complicated) code in the Repository, and given that pubsub is about to removed in Ceramic v4, I opted to leave the current behavior for now.  I could be persuaded to bite the bullet and fix this right now though if we think it's worth it to do so.


ukstv (approved):
The same restraint applies as in https://github.com/ceramicnetwork/js-ceramic/pull/3073#discussion_r1435578039 : would like to have it consistent.

PR #3073, opened 2023-12-21
fix: Always update the StateStore from within the ExecutionQueue

stbrody:
This ensures that two different tasks can't clobber each other's writes to the state store

### DIFF
--- a/packages/core/src/state-management/repository.ts
+++ b/packages/core/src/state-management/repository.ts
@@ -729,7 +729,9 @@ export class Repository {
   }
 
   /**
-   * Apply options relating to authoring a new commit
+   * Apply options relating to authoring a new commit.
+   *
+   * Must be called within the ExecutionQueue to be safe.
    *
    * @param state$ - Running State
    * @param opts - Initialization options (request anchor, publish to pubsub, etc.)
--- a/packages/core/src/state-management/repository.ts
+++ b/packages/core/src/state-management/repository.ts
@@ -575,7 +575,8 @@ export class Repository {
 
     const carFile = await this.#deps.anchorRequestCarBuilder.build(state$.id, state$.tip)
     const anchorEvent = await this.anchorService.requestAnchor(carFile)
-    await this.handleAnchorEvent(state$, anchorEvent)
+    // Don't wait on handling the anchor event, let that happen in the background.
+    void this.handleAnchorEvent(state$, anchorEvent)
   }
 
   /**
--- a/packages/core/src/state-management/repository.ts
+++ b/packages/core/src/state-management/repository.ts
@@ -97,11 +97,13 @@ enum SyncStatus {
 export class Repository {
   /**
    * Serialize loading operations per streamId.
+   * All writes to the 'inmemory' cache must happen within the context of this queue.
    */
   readonly loadingQ: ExecutionQueue
 
   /**
    * Serialize operations on state per streamId.
+   * All writes to the StateStore must happen within the context of this queue.
    */
   readonly executionQ: ExecutionQueue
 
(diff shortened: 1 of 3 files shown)


packages/core/src/state-management/repository.ts
```diff
@@ -560,7 +575,8 @@ export class Repository {
 
     const carFile = await this.#deps.anchorRequestCarBuilder.build(state$.id, state$.tip)
     const anchorEvent = await this.anchorService.requestAnchor(carFile)
-    await this.handleAnchorEvent(state$, anchorEvent)
+    // Don't wait on handling the anchor event, let that happen in the background.
+    void this.handleAnchorEvent(state$, anchorEvent)
```
stbrody, line 579: I don't love the fact that I needed to make this change, but without it we get a deadlock since handleAnchorEvent calls `_updateStateIfPinned_safe` which uses the execution queue, and this function is already being called from within the context of the ExecutionQueue.

The problem is that handleAnchorEvent is called in two places.  One of them is here, where we're always in the context of the ExecutionQueue, the other is from the AnchorProcessingLoop, which is not calling it from within the ExecutionQueue.  Since the AnchorProcessingLoop is actually the primary way that AnchorEvents are handled, `handleAnchorEvent` assumes it is not called within the ExecutionQueue, and so uses the queue explicitly any time it needs to update StreamState.

An alternative to this change would be to change the AnchorProcessingLoop to call all of handleAnchorEvent from within the ExecutionQueue, and change handleAnchorEvent to assume its always within the ExecutionQueue already and never use the queue itself.

That isn't a crazy change and may in fact be the right move.  Most of handleAnchorEvent is pretty light weight, the exceptions are actually applying the anchor commit (which already uses the ExecutionQueue) and importing the CAR file from the CAS with the anchor commit and witness proof data.  So the big downside of taking the alternate approach laid out in this comment would be that now we'd be taking the time to import that CAR file from within the context of the ExecutionQueue.  If you think that's preferable to the change presented in this PR, let me know and I'll happy switch over.

ukstv:
@stbrody I am sorry, I totally do not understand what this tries to achieve. It definitely is more complicated than before to my taste. What does get better/faster thanks to the change?

stbrody:
> @stbrody I am sorry, I totally do not understand what this tries to achieve. It definitely is more complicated than before to my taste. What does get better/faster thanks to the change?

@ukstv So this change is all about correctness.  There are race conditions in the code today where two different tasks could be running concurrently that both try to write to the state store.  This can cause them to override each other's writes.  This could lead to situations like the bug from the other week, where data could be lost on process restart if a newer StreamState got overridden with an older StreamState in the state store.  The ExecutionQueue exists specifically to serialize writes to a single stream, and so is the concurrency control mechanism best suited to serialize and protect all modifications to the state store.

packages/core/src/state-management/repository.ts
ukstv: Oh my God. I see now. All right, the change on its own (if we assume this is the final form of the code), is quite complicated for a reader, especially for a freshly onboarded person. What plain is there (if any) to make it less complicated, more congruent? Or rather, do you have one? If not, I am open to a sync up after the holidays to prepare one. I think we might be able to extract all the "unsafe" parts, i.e. free from the queues, into a separate entity.

For this particular change, I was going to ask you at make the suffixes uniform, but apparently, it is all right. So, approve, even if it feels sad.

ukstv (approved):

packages/core/src/state-management/repository.ts
stbrody: Yeah, I agree this makes the code a bit ugly, and we should do more to make it easier for devs to reason about how and when the queues are/must be used.  I think the idea of some separate classes that encapsulate the logic that must run within the different queues could work and be a nice cleanup.  I also had an idea based on something we did at my last company that might let us utilize the compiler a bit to catch errors here.  I'll put up a little draft PR to demonstrate the idea and see what you think

PR #3074, opened 2023-12-22
fix: register new running state

JulissaDantes:
## Description

If we hit the case here where the state isn't found in the cache or state store, but then we wind up creating a new RunningState object here, we don't ever call _registerRunningState on it.


stbrody (approved):
lgtm

PR #3075, opened 2023-12-25
chore: Deduplicate emissions

ukstv:
Emit if only tip changes.


packages/core/src/__tests__/ceramic-feed.test.ts
```diff
@@ -131,15 +124,11 @@ describe('Ceramic feed', () => {
     // process anchor
     await TestUtils.anchorUpdate(ceramic1, stream)
 
-    expect(feed.length).toEqual(3) // create + anchor request + anchor update
+    expect(feed.length).toEqual(2) // genesis + anchor commits
     // between and request anchor
```
stbrody, line 128: what does this comment mean? @JulissaDantes

packages/core/src/state-management/__tests__/repository.test.ts
```diff
@@ -758,3 +759,35 @@ describe('handleAnchorEvent', () => {
...
+    const emittedDocuments: Array<FeedDocument> = []
+    const subscription = ceramic.feed.aggregation.documents.subscribe((document) => {
+      emittedDocuments.push(document)
+    })
+    ;(ceramic.repository as any)._registerRunningState(state$)
+    state$.next({
+      ...state$.value,
+      anchorStatus: AnchorStatus.PROCESSING,
+    })
+    expect(emittedDocuments.length).toEqual(1)
```
stbrody, line 790: sorry, it's not clear to me how this test is testing the deduplication.  Here I see us emitting one state change and seeing one update to the feed.  I think we'd need to emit two state changes that only affect anchoring status and still see that that only results in one update to the feed.

stbrody: Unless the first update is from the stream creation?  If if you check `emittedDocuments` after registering the running state but before emitting the PROCESSING anchor status, do you also get an emittedDocuments length of 1?  If so, I think duplicating this check to be both before and after we emit the PROCESSING status will make the test clearer.

packages/core/src/state-management/repository.ts
```diff
@@ -824,7 +824,12 @@ export class Repository {
    * Adds the stream's RunningState to the in-memory cache and subscribes the Repository's global feed$ to receive changes emitted by that RunningState
    */
   private _registerRunningState(state$: RunningState): void {
-    state$.subscribe(this.feed.aggregation.documents)
+    state$
+      .pipe(
+        distinct((s) => s.log[s.log.length - 1].cid.toString()), // Distinct if the tip changes
```
stbrody, line 829: Hmm, looking at the documentation for the `distinct` function (https://rxjs.dev/api/index/function/distinct), it looks like it uses a Set internally to keep track of *all* previous values emitted.  We actually only really need to keep track of the single most recently emitted value.  If the stream state got reset to some older value in the log (which can happen if the CACAOs expire on newer log entries and the stream is reloaded with the ALWAYS_SYNC flag), we probably would want to emit an event for that.  I'm also worried that maintaining this hash set for all streams could be a meaningful amount of additional memory consumption.  Can we do this in a way that there's a constant amount of memory consumption needed, since it only needs to remember the previous tip, not *all* previous tips?

packages/core/src/feed.ts
```diff
@@ -1,44 +1,26 @@
...
+    readonly content: any,
+    readonly metadata: StreamMetadata
+  ) {}
 
-    this.subject = new Subject<StreamState>()
-  }
-
-  next(s: StreamState) {
-    this.subject.next(s)
+  static fromStreamState(streamState: StreamState): FeedDocument {
```
stbrody, line 13: nice code cleanup

stbrody (commented):
Mostly looks good, main comment is around the fact that the RXJS `distinct` operator keeps track of all previous values, not just the most recent one.

packages/core/src/__tests__/ceramic-feed.test.ts
JulissaDantes: The 2 elements inside the feed are `2 commits = 1 genesis commit + 1 anchor commit`

ukstv: changed to `2 commits = 1 genesis commit + 1 anchor commit`

packages/core/src/state-management/repository.ts
ukstv: True. Changed to `distinctUntilKeyChanged` which does not suffer like this.

packages/core/src/state-management/__tests__/repository.test.ts
ukstv: Okay, added more checks to make it clear.

packages/core/src/__tests__/ceramic-feed.test.ts
stbrody: nice, improved comment on the previous line.  Can we just delete the comment on this line?

stbrody (approved):
lgtm

PR #3077, opened 2023-12-29
feat: Respect context when creating MIDs

oed:
Enables adding [context to genesis headers](https://forum.ceramic.network/t/asking-for-support-of-the-context-field-in-the-metadata-object-for-streams/1425) in MIDs.


dbcfd:
@oed would it make more sense to call this `appData` or `userData`?

oed:
@dbcfd I don't have a strong stance, looks like Orbis is already using context. Maybe you can ask in the forum post linked above?

stbrody:
I really want us to make sure we've thought about how this will interact with Recon and get fed down to lower levels (for example, for separator key we have an extra indirection layer where we have one field that says that name of the actual separator key field name is.  Might we want something like this for context?  Might it make sense to be able to take a field from the actual content of the document as the context?), as well as what, if any, constraints we want on this field.  One of the problems we had with "family" was that it was so open-ended, every dev wound up using it differently, so it was hard to take advantage of in the protocol.  Are we sure we're not going to wind up with the same problem here?  I know Orbis has said they use it with a StreamID that references a different document that has information and details about the "context".  Might we want to enforce that as the way this is always used?

I think it's very possible this approach is totally fine, I just want to make sure we've thought it through before jamming it in just because this is what one partner asked for.

Adding Nathaniel and Aaron for their takes

packages/stream-model-instance/src/model-instance-document.ts
```diff
@@ -320,6 +320,9 @@ export class ModelInstanceDocument<T = Record<string, any>> extends Stream {
     if (!metadata.deterministic) {
       header.unique = randomBytes(12)
     }
+    if (metadata.context) {
```
stbrody, line 323: You'll need to add this as an optional field in `ModelInstanceDocumentMetadataArgs`

stbrody: And probably to `GenesisHeader` as well

packages/stream-model-instance-handler/src/__tests__/model-instance-document-handler.test.ts
```diff
@@ -70,6 +70,7 @@ const CONTENT0 = { myData: 0 }
 const CONTENT1 = { myData: 1 }
 const CONTENT2 = { myData: 2 }
 const METADATA = { controller: DID_ID, model: FAKE_MODEL_ID }
+const METADATA1 = { controller: DID_ID, model: FAKE_MODEL_ID, context: FAKE_MID_ID }
```
stbrody, line 73: nitpick: maybe call this `METADATA_WITH_CONTEXT`?

packages/stream-model-instance-handler/src/__tests__/model-instance-document-handler.test.ts
```diff
@@ -530,6 +548,30 @@ describe('ModelInstanceDocumentHandler', () => {
...
+
+    const commitData = {
+      cid: FAKE_CID_1,
+      type: CommitType.GENESIS,
+      commit: payload,
+      envelope: commit.jws,
+    }
+    const streamState = await handler.applyCommit(commitData, context)
+    delete streamState.metadata.unique
+    expect(streamState.metadata.context).toBeInstanceOf(StreamID)
```
stbrody, line 571:
```suggestion
    expect(streamState.metadata.context).toEqual(METADATA_WITH_CONTEXT.context)
```

stbrody (approved):
lgtm

PR #3080, opened 2024-01-02
feat: Fail incoming writes if the CAS has been consistently unavailable

stbrody:


ukstv (approved):

stbrody:
I'm going to hold off on merging this until the next release candidate is made.  I don't want this change AND the off-memory anchoring change to go out in the same release, I'd like to spread them out so it's easier to detect if there are any issues in either one

PR #3081, opened 2024-01-02
Update database-index-api.ts

JulissaDantes:
## Description

Remove double space in function documentation.


stbrody (approved):
lgtm, though if you start trying to change every place there's a double-space after a period you're going to have to update a LOT of comments 😅.  This is something I do in all my typing.  I know it's out of favor these days but my muscle memory on that one was set a long time ago 😛

PR #3083, opened 2024-01-03
feat: Timeout fetching batches from the AnchorRequestStore

stbrody:
This is kind of me taking a stab in the dark because I still don't have any really good hypothesis for why the anchor polling keeps stopping.  But since the processing loop is serial, I had the thought that if the query from S3 hung for any reason, it would lock up the whole loop.  So this patch addresses that, as well as adding some increased logging to ProcessingLoop.

### DIFF
--- a/packages/core/src/store/anchor-request-store.ts
+++ b/packages/core/src/store/anchor-request-store.ts
@@ -106,8 +106,9 @@ export class AnchorRequestStore extends ObjectStore<StreamID, AnchorRequestData>
     let numEntries = 0
     do {
       try {
+        let timeout
         const timeoutPromise = new Promise<null>((resolve) => {
-          setTimeout(() => {
+          timeout = setTimeout(() => {
             this.#logger.warn(`Timed out while waiting for AnchorRequestStore to fetch a batch`)
             resolve(null)
           }, this.#infiniteListBatchTimeoutMs)
@@ -119,6 +120,7 @@ export class AnchorRequestStore extends ObjectStore<StreamID, AnchorRequestData>
           gt: generateKey(gt),
         })
         const batch = await Promise.race([batchPromise, timeoutPromise])
+        clearTimeout(timeout)
         if (batch && batch.length > 0) {
           gt = StreamID.fromString(batch[batch.length - 1].key)
           for (const item of batch) {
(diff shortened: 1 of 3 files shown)


packages/core/src/store/anchor-request-store.ts
```diff
@@ -98,12 +106,22 @@ export class AnchorRequestStore extends ObjectStore<StreamID, AnchorRequestData>
...
       try {
-        const batch = await this.store.find({
+        let timeout
+        const timeoutPromise = new Promise<null>((resolve) => {
+          timeout = setTimeout(() => {
+            this.#logger.warn(`Timed out while waiting for AnchorRequestStore to fetch a batch`)
+            resolve(null)
+          }, this.#infiniteListBatchTimeoutMs)
+        })
+
```
samika98, line 116: Can we add a debug logs here? before initiating a find operation, this log might be beneficial in cases where we think the find operation hangs, we should add a subsequent completion log as well when the operations completes. Mostly when we check batch is non null and length is greater than 0
 
this.#logger.debug(`Initiating 'find' operation with batchSize: ${batchSize} and gt: ${gt}`);
```suggestion
```suggestion
this.#logger.debug(`Initiating 'find' operation in with batchSize: ${batchSize} and gt: ${gt}`);
```
```

File: packages/core/src/store/anchor-request-store.ts [hunk 2]
```diff
@@ -98,12 +106,22 @@ export class AnchorRequestStore extends ObjectStore<StreamID, AnchorRequestData>
...
+        const batchPromise = this.store.find({
           limit: batchSize,
           useCaseName: this.useCaseName,
           gt: generateKey(gt),
         })
-        if (batch.length > 0) {
+        const batch = await Promise.race([batchPromise, timeoutPromise])
+        clearTimeout(timeout)
```
samika98, line 123: Adding a log here as well after this line to indicate the completion of the find.

``` 
 if (batch !== null) {
  this.#logger.debug(`Completed 'find' operation with batchSize: ${batchSize}, retrieved ${batch.length} items`);
}
```

samika98 (approved):
Suggested some more log additions that can give us a better indication if a find operation hangs with details of the request like batchsize.

ukstv (approved):

packages/core/src/store/anchor-request-store.ts
stbrody: I'm worried that will be very verbose and will fill up our logs.  I'd like to try this one as-is first, and if we still can't figure it out after this PR goes in, then we can try a build with super verbose logging tracking every single part of the polling process, and revert it after we figure out what's wrong.

PR #3084, opened 2024-01-04
chore: update multiformats and other deps

dav1do:
## Description

All our tests started failing due to multiformats CIDs dropping a getter during clone. We updated a number of our packages to the newest multiformats and uint8array versions, and in some cases updated all dependencies. We incorporate those changes here.

Some notes:
- I considered updating `@ipld/dag-cbor` as we updated that in [js-dag-jose-utils](https://github.com/ceramicnetwork/js-dag-jose-utils/pull/6/files) and it's mostly breaking changes due to multiformats version updates (9.0.7 is on v13 like we are). It passed locally in tests but I didn't want to add any more changes. Happy to add that.
- `ipfs-core-types` is deprecated in favor of helia and uses multiformats v11.  I used `toCID(oldCid.toString())` to make sure we have the updated js object with a readonly value instead of the getter, but I'm not sure if it's necessary. I considered simply casting `oldCid as CID` but I believe that would only fool the compiler and keep current behavior. It didn't seem to make a difference for any tests.
 - I modified the LevelDB anchor tests to use `Utils.getCommitData` rather than talking to the pubsub dispatcher directly. That code handles converting types so we end up with the new CIDs and avoid getting any failures due to the ipfs-core-types version, but otherwise should do the same thing.

## How Has This Been Tested?

So far I've only run the unit tests.

## PR checklist


packages/core/src/store/__tests__/level-anchor-request-store.test.ts
```diff
@@ -22,6 +23,8 @@ import { createCeramic } from '../../__tests__/create-ceramic.js'
 import first from 'it-first'
 import all from 'it-all'
 import { OLD_ELP_DEFAULT_LOCATION } from '../level-db-store.js'
+import { Utils } from '../../utils.js'
+import { Dispatcher } from '../../dispatcher.js'
```
dav1do, line 27: bleh, remove

packages/core/src/dispatcher.ts
```diff
@@ -360,7 +362,10 @@ export class Dispatcher {
               signal: signal,
             })
           )
-          blockCid = resolution.cid
+          // HACK around ipfs-core-types multiformats version mismatch
+          // don't just as CID because we want to new version with the readonly field rather than getter
+          // that is dropped during a deepClone... assuming it actually matters...
+          blockCid = toCID(resolution.cid.toString())
```
stbrody, line 368: this makes me pretty sad.  Your comment implies it might not be necessary?  Can we test that and remove the hack if it's not needed?

stbrody: Or can we replace our use of ipfs-core-types with using the multiformats version of cids instead?

packages/core/src/store/__tests__/level-anchor-request-store.test.ts
```diff
@@ -88,37 +91,38 @@ describe('LevelDB-backed AnchorRequestStore state store', () => {
...
+  // use Utils to load the genesis commit for a stream so it converts CIDs for us
+  // and we avoid any issues with one having `Symbol(Symbol.toStringTag): "CID"` and one not
+  // because the getter function isn't copied by _.cloneDeep
+  async function loadGenesisCommit(
+    ceramic: CeramicApi,
+    streamId: StreamID
+  ): Promise<GenesisCommit> {
+    const commits = await ceramic.loadStreamCommits(streamId)
+    expect(commits).toBeDefined()
+    const first = commits[0].cid
```
stbrody, line 103: if all you need is the genesis commit CID you don't need to use `loadStreamCommits` for that.  The StreamID encodes the genesis commit CID.  You can just use `streamId.cid` to get it.

stbrody: Also for future reference, `loadStreamCommits` is a very expensive call.  It not just loads the stream but then also loads each individual commit's data from IPFS.  So if you just need the CID of a specific commit, it would be better to call `loadStream` and get the CID from the log in the state object.

packages/core/src/dispatcher.ts
dav1do: yeah, I agree it's awful. I tried to cover this in the summary cause it's weird. And the comment (and multiple versions don't read _super_ clean).

So it didn't seem to matter when I just used `old as CID` to make typescript happy. I _think_ it's technically different, as it's 2 js classes, one with a getter and one with a readonly field, but otherwise they should be about the same (v11 vs v13). That's what I was trying to get at with "actually matters". Cause even the mismatch equals errors we get don't _actually_ matter, as everything is identical except a field definition, but node still seems to find the string info correctly.

ipfs-core-types is deprecated in favor of helia and uses multiformats v11. I didn't want to update to helia (assumed it'd be hard without looking). We could provide a thin wrapper around the interface with our own that uses a v13 CID since we know they should be compatible, but I'm not sure if it's worth it. 

anyway, let me know what you think.

stbrody: Yeah, my understanding of all the work that's happened in the last 2 days has been around trying to make sure we only use v13 CIDs and remove every place that's pulling in older versions of the multiformats package.  So it seems like this is the last remaining holdout.  So it does seem like the right long term fix is to move away from `ipfs-core-types`.  But I also don't want to hold up this PR.  So I guess I'd say let's file a CORE ticket in linear to switch to helia for our ipfs http client, and add a `TODO(CORE-XXXX):` comment to each of the places we do this hack so we know to remove this when we make that upgrade.

dav1do: Comments updated to `// TODO(CORE-137) - Replace ipfs-core-types to get consistent multiformats version`

stbrody (approved):
lgtm assuming tests pass

PR #3086, opened 2024-01-06
chore: test utilities

dbcfd:
Move testing code into their own packages. Consolidate code when possible


packages/common-test-utils/src/index.ts
```diff
@@ -0,0 +1,129 @@
+import type { CID } from 'multiformats/cid'
```
stbrody, line 1: Could this just be a file inside of the `common` package instead of its own new package?  Each new package we introduce adds a noticeable increase in the js-ceramic build time and I'm worried about introducing so many new packages here.

packages/base-test-utils/src/index.ts
```diff
@@ -0,0 +1,64 @@
+import type { CID } from 'multiformats/cid'
+import { randomCID, StreamID } from '@ceramicnetwork/streamid'
+
+export class BaseTestUtils {
```
stbrody, line 4: Could this just be a type we export from the StreamIDs package instead of adding a whole new package for this?

packages/did-test-utils/src/index.ts
```diff
@@ -0,0 +1,227 @@
...
+import { wrapDocument } from '@ceramicnetwork/3id-did-resolver'
+import { parse as parseDidUrl } from 'did-resolver'
+import * as KeyDidResolver from 'key-did-resolver'
+import * as sha256 from '@stablelib/sha256'
+import { decode as decodeMultiHash } from 'multiformats/hashes/digest'
+import * as uint8arrays from 'uint8arrays'
+import { CeramicApi } from '@ceramicnetwork/common'
+import { jest } from '@jest/globals'
+
+export const FAKE_CID_1 = CID.parse('bafybeig6xv5nwphfmvcnektpnojts33jqcuam7bmye2pb54adnrtccjlsu')
```
stbrody, line 15: all these exported constants seem extremely specific to be exported from a package like this.  This feels like they should belong to the specific tests that are using them.

packages/did-test-utils/src/index.ts
```diff
@@ -0,0 +1,227 @@
...
+      },
+      did
+    ),
+    didDocumentMetadata: {
+      updated: rotateDate,
+    },
+  }
+}
+
+export class DidTestUtils {
```
stbrody, line 146: is this used for anything other than 3id-specific testing?  Can this be moved into the `3id-did-resolver` package?

stbrody (commented):
I'm getting increasingly concerned about how much the number of packages we have is affecting build times.

packages/did-test-utils/src/index.ts
dbcfd: It's used for 3id, tile, and stream model testing.

packages/common-test-utils/src/index.ts
dbcfd: Main thing is trying to get test code out of packages that we release. If we're worried about build times, we should switch from npm.

dbcfd: One of the things that has happened by having testing utilities in our packages is bleed between packages. Like anchoring related code in common.

packages/base-test-utils/src/index.ts
dbcfd: This one could possibly just be in streamid's package. It's more utility code than testing code, so if we're good having it used by people using our libraries, I'd be glad to get rid of this package.

packages/did-test-utils/src/index.ts
dbcfd: They were actually reused by multiple packages tests, hence consolidating them, rather than duplicating to tests.

stbrody (approved):
alright, confirmed that there's no compile-time overhead for additional packages, so LGTM

PR #3091, opened 2024-01-12
fix: load model should work under postgres

dbcfd:
Postgres was not preserving casing on field names, so index validation was
failing on MID table creation. Added additional test file to verify PG
behavior.

Addresses WS2-2960 and WS2-2937

### DIFF
--- a/packages/indexing/src/tables-manager.ts
+++ b/packages/indexing/src/tables-manager.ts
@@ -31,6 +31,15 @@ type ConfigTable = {
   readonly validSchema: object
 }
 
+type IndexRow = {
+  readonly tablename: string
+  readonly indexname: string
+}
+
+type IndexResult = {
+  readonly rows: Array<IndexRow>
+}
+
 const CUSTOM_COLUMN_PREFIX = 'custom_'
 
 export class TablesManager {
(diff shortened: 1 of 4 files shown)


packages/indexing/src/tables-manager.ts
```diff
@@ -292,28 +309,36 @@ export class PostgresTablesManager extends TablesManager {
...
+      await this.dataSource.raw<PgIndexResult>(`
 select *
 from pg_indexes
 where tablename like '${tableName}'
-and indexname in (${sqlIndices});
-  `)
-    return expectedIndices.length == actualIndices.rowCount
+    `)
+    )
+    const actualIndices = indicesResult ? indicesResult.rows.map((row) => row.indexname) : []
```
stbrody, line 326: should we also `toLowerCase()` here as well?

stbrody (approved):
YES

Good find.  Finally an explanation to what users have been reporting that actually makes sense.

packages/indexing/src/tables-manager.ts
dbcfd: pg already has them lowercased.

dbcfd: Yeah, we could future proof by just moving the lower casing into the check function.

PR #3098, opened 2024-01-17
fix(core): Use the anchor proofs chain id to verify the anchor proof

stephhuynh18:
## Description

Previously: In readOnly mode cannot load streams fully. Logs show the following error
```
[2024-01-17T22:22:25.350Z] WARNING: 'Error when validating anchor commit: Error: Any anchor proofs created after block undefined must include the txType field. Anchor txn blockNumber: 15325737'
[2024-01-17T22:22:25.352Z] WARNING: 'Error when validating anchor commit: Error: Any anchor proofs created after block undefined must include the txType field. Anchor txn blockNumber: 15325737'
[2024-01-17T22:22:25.354Z] WARNING: 'Error when validating anchor commit: Error: Any anchor proofs created after block undefined must include the txType field. Anchor txn blockNumber: 15325737'
```

Now: In readOnly mode can load streams fully. No errors in logs 

## How Has This Been Tested?

Run in readOnly mode. Successfully load streams fully (should have same state as a non read only node)

### DIFF
--- a/packages/core/src/anchor/ethereum/ethereum-anchor-validator.ts
+++ b/packages/core/src/anchor/ethereum/ethereum-anchor-validator.ts
@@ -214,7 +214,8 @@ export class EthereumAnchorValidator implements AnchorValidator {
       throw new Error(`The root CID ${anchorProof.root} is not in the transaction`)
     }
 
-    if (txResponse.blockNumber <= BLOCK_THRESHHOLDS[this._chainId]) {
+    // we use the anchor proof chain id instead of the this._chainId, as this._chainId is not set if the ceramic node is being used in readOnly mode
+    if (txResponse.blockNumber <= BLOCK_THRESHHOLDS[anchorProof.chainId]) {
       return block.timestamp
     }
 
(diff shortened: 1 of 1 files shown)


3benbox (approved):
LGTM

packages/core/src/anchor/ethereum/ethereum-anchor-validator.ts
```diff
@@ -214,15 +214,16 @@ export class EthereumAnchorValidator implements AnchorValidator {
       throw new Error(`The root CID ${anchorProof.root} is not in the transaction`)
     }
 
-    if (txResponse.blockNumber <= BLOCK_THRESHHOLDS[this._chainId]) {
+    // we use the anchor proof chain id instead of the this._chainId, as this._chainId is not set if the ceramic node is being used in readOnly mode
+    if (txResponse.blockNumber <= BLOCK_THRESHHOLDS[anchorProof.chainId]) {
```
stbrody, line 218: can we add an extra check so if the chainId is one we don't understand we throw a nice error message?

stbrody (approved):
lgtm

packages/core/src/anchor/ethereum/ethereum-anchor-validator.ts
stephhuynh18: that will happen right here: https://github.com/ceramicnetwork/js-ceramic/blob/765e89295ee94d745d52359022fbbe83fc77fc07/packages/core/src/anchor/ethereum/ethereum-anchor-validator.ts#L249

Is this error message sufficient?

stbrody: ah yep, that works

PR #3100, opened 2024-01-18
Fix codec import

stbrody:


packages/cli/src/ceramic-daemon.ts
```diff
@@ -43,7 +43,7 @@ import { S3Store } from './s3-store.js'
 import { commitHash } from './commitHash.js'
 import { parseQueryObject } from './daemon/parse-query-object.js'
 import { SseFeed } from './daemon/sse-feed.js'
-import { JsonAsString, AggregationDocument } from '../../codecs/lib/feed.js'
+import { JsonAsString, AggregationDocument } from '@ceramicnetwork/codecs'
```
ukstv, line 46: Damn. I fixed that thing twice. Missed during a rebase.

PR #3101, opened 2024-01-18
feat: Add static method for updating a MID without access to the StreamState

stbrody:
This will be useful to users of the Firehose who won't have access to the StreamState object but might still want to be able to issue updates to documents that they are maintaining via the Firehose.

### DIFF
--- a/packages/stream-model-instance/src/model-instance-document.ts
+++ b/packages/stream-model-instance/src/model-instance-document.ts
@@ -20,6 +20,7 @@ import {
   GenesisHeader,
 } from '@ceramicnetwork/common'
 import { CommitID, StreamID, StreamRef } from '@ceramicnetwork/streamid'
+import type { CID } from 'multiformats/cid'
 
 /**
  * Arguments used to generate the metadata for Model Instance Documents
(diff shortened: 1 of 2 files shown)


dbcfd:
Can you describe how you'd maintain a MID while receiving updates from Firehose, such that this function works?

stbrody:
> Can you describe how you'd maintain a MID while receiving updates from Firehose, such that this function works?

The firehose returns the stream content, metadata, and the CommitID of the update.  If you then want to issue an update back to Ceramic, you can use this new static method, passing in the most recent data received for that stream from the firehose (the content, and you can get the StreamID and tip from the CommitID) to construct a commit, then you can send that commit to your ceramic node using the http-client's applyCommit method.

packages/stream-model-instance/src/model-instance-document.ts
```diff
@@ -243,25 +250,39 @@ export class ModelInstanceDocument<T = Record<string, any>> extends Stream {
...
+   * @param prev - The current tip of the Stream that the update should be applied on top of.
+   * @param oldContent - The current content of the Stream.
+   * @param newContent - The new content to update the Stream with.
    */
-  private _makeCommit(signer: CeramicSigner, newContent: T | null): Promise<CeramicCommit> {
-    const commit = this._makeRawCommit(newContent)
+  static makeUpdateCommit<T>(
+    signer: CeramicSigner,
+    streamId: StreamID,
+    prev: CID,
```
ukstv, line 264: Suggestion: `prev: CommitID`.

ukstv (approved):
1. Maybe it makes sense to have `prev` as CommitID instead of having separate streamid and prev CID.
2. I'd say, we shall see in ~4 months if the method is actually used by Orbis or other folks who consume the Firehose. If not, better remove it - maintenance takes resources.

PR #3102, opened 2024-01-18
fix: Limit scope of what's in a queue

ukstv:


packages/core/src/anchor/anchor-processing-loop.ts
```diff
@@ -30,51 +30,54 @@ export class AnchorProcessingLoop {
     anchorStoreQueue: NamedTaskQueue
   ) {
     this.#anchorStoreQueue = anchorStoreQueue
-    this.#loop = new ProcessingLoop(logger, store.infiniteList(batchSize), (streamId) =>
-      this.#anchorStoreQueue.run(streamId.toString(), async () => {
```
ukstv, line 34: The big mutex gets downsized to a poor mans' version of Compare-and-Swap.

packages/core/src/anchor/anchor-processing-loop.ts
```diff
@@ -30,51 +30,54 @@ export class AnchorProcessingLoop {
...
+        )
+        const isTerminal = await eventHandler.handle(event)
+        logger.debug(
+          `Anchor event with status ${event.status} for commit CID ${entry.cid} of Stream ${streamId} handled successfully`
+        )
+        if (isTerminal) {
+          logger.debug(`Removing entry from AnchorRequestStore for Stream ${streamId}`)
+          // Remove iff tip stored equals to the tip we processed
+          // Sort of Compare-and-Swap
+          await this.#anchorStoreQueue.run(streamId.toString(), async () => {
```
ukstv, line 66: We only need to guard a removal here.

stephhuynh18 (approved):
yahoo

packages/core/src/anchor/anchor-processing-loop.ts
```diff
@@ -30,51 +30,54 @@ export class AnchorProcessingLoop {
...
+        const isTerminal = await eventHandler.handle(event)
+        logger.debug(
+          `Anchor event with status ${event.status} for commit CID ${entry.cid} of Stream ${streamId} handled successfully`
+        )
+        if (isTerminal) {
+          logger.debug(`Removing entry from AnchorRequestStore for Stream ${streamId}`)
+          // Remove iff tip stored equals to the tip we processed
+          // Sort of Compare-and-Swap.
+          await this.#anchorStoreQueue.run(streamId.toString(), async () => {
+            const loaded = await store.load(streamId)
```
stbrody, line 67: bummer about the extra state store read, but agreed this is the simplest fix without a transactional state store.  LGTM

ukstv: It is actually negligibly fast on qa-private node. Bummer, but a tolerable one.

PR #3103, opened 2024-01-19
feat(indexing): Automatically index fields defined in the SET account relation

PaulLeCam:
Ensures the fields defined as part of the set when the model account relation uses the `set` type are indexed even when indices are not explicitly added.


dbcfd:
Why do set account relation fields need to be indexed?

PaulLeCam:
It needs to be indexed so it can be efficiently queried, similar to what we do for document relations.

packages/indexing/src/local-index-api.ts
```diff
@@ -38,23 +39,35 @@ async function _getIndexModelArgs(
...
     model: modelStreamId,
   }
+  let relationFields = new Set<string>()
 
   if (modelStreamId.type == Model.STREAM_TYPE_ID) {
     const modelState = await core.loadStream(modelStreamId, {})
-    const content = modelState.state.next?.content ?? modelState.state.content
+    const content: ModelDefinition = modelState.state.next?.content ?? modelState.state.content
     Model.assertVersionValid(content, 'major')
-    if (content.interface) {
```
JulissaDantes, line 46: Can you explain why this is not needed here anymore? `ModelDefinition` can still have the interface property, wouldnt this cause interfaces to be indexed now?

PaulLeCam: Yes, the check is still there, just moved here: https://github.com/ceramicnetwork/js-ceramic/pull/3103/files#diff-2884de7343db780f37821fd5a33c5baf2dda390b70387db7c0e610550e9c769dR53

stbrody:
I have the same question as Danny.  Why do we assume users will need to query on these fields?

PaulLeCam:
> I have the same question as Danny. Why do we assume users will need to query on these fields?

Many queries are automatically generated by ComposeDB, that's how all the account and document relations already work. With the set account relation we need to query not only on the `model` and `controller`, but also on the set fields values, so I expect these need to be indexed as well, no?

stbrody:
@PaulLeCam if you are trying to load a specific document with the SET fields, wouldn't you just use the determinism to find the StreamID, then query by StreamID?  Why do we need extra queries at the postgres level to support set account relation?

PaulLeCam:
> @PaulLeCam if you are trying to load a specific document with the SET fields, wouldn't you just use the determinism to find the StreamID, then query by StreamID? Why do we need extra queries at the postgres level to support set account relation?

Because deterministic documents lookup is terribly slow, that's one of the reasons we created ComposeDB in the first place, no?
Instead of relying on deterministic lookups, we query the indexer to have efficient reads, that's what we've always done with the SINGLE account relation.

stbrody:
Deterministic is slow when you're actually querying for the current stream state over the network (ie pubsub).  I'm talking about deterministically generating the StreamID, then asking the **index** for the current state by streamid.  Like I think there's already a way to use GraphQL to load a stream by StreamID (is that what the `Node` type is for?).  Basically doing the same thing as that.

PaulLeCam:
The index query APIs don't provide query by stream ID, we use and have always used the multiquery API for this. The `Node` type is an interface, I don't see how it's related to this?

stbrody:
> The index query APIs don't provide query by stream ID, we use and have always used the multiquery API for this.
Interesting, I didn't realize that.  Maybe we should expose a way to use the normal query endpoint that ComposeDB uses to query the index by StreamID then.  In any case, my main point here is we don't need more indices at the database level when we already have all the information we need in the client to calculate the StreamID, and we already have an index on StreamID.

> The `Node` type is an interface, I don't see how it's related to this?
Maybe I misremembered what `Node` means or how it's used.  I just meant whatever is the way that users can use GraphQL to query for a specific stream by StreamID.

stbrody:
So then can we implement loading SET account relation streams with multiquery for now, and then add a way to get those from the index later?  I'd still prefer not to add more required indices to the postgres database than are strictly necessary

stbrody:
> I don't think that's an option, loads are so slow it's going to make it unusable.

They're only slow when actually querying pubsub.  With ComposeDB you don't need to ever query pubsub because you know that the data should already be in the index. So you could send the loadStream request with `SyncOptions.NEVER_SYNC` and `syncTimeoutSeconds:0` which should make it as fast as other composedb queries (actually slightly faster since it only goes to the state store and not the index, instead of both like normal CDB queries)

PaulLeCam:
> So loading a Stream by StreamID should be very fast, but it will only work if the stream is already indexed.

Yes, that's the whole problem and that's why we need to rely on indexing for the SINGLE and SET account relation queries: if the streams are not indexed, the indexer simply returns empty values without needing to go over the network.

For mutations I think it's better to still go over the network possibly to resolve the stream, but the performance impact is much more limited.

stbrody:
To clarify, when I said "but it will only work if the stream is already indexed", I meant that the Model the stream belongs to has been configured to be indexed by the node - ie that the node will sync and apply all data for that stream.  Loading the Stream by StreamID does not require that the stream be in the ComposeDB postgres database, and certainly doesn't require any specific database-level indices to be built in the Postgres database

PaulLeCam:
That seems to be the theory but in practice we get a higher latency for queries every time we lookup deterministic streams.
I created a branch in https://github.com/ceramicstudio/js-composedb/pull/207 to change the multiquery loading logic to [include the genesis commit when loading deterministic streams](https://github.com/ceramicstudio/js-composedb/pull/207/files#diff-0af4f6bd66935ecbd445ba22995dd687aeeee5468afdd46d060f8543a4a9a195R43) and [use the loader instead of the index query APIs](https://github.com/ceramicstudio/js-composedb/pull/207/files#diff-d546908c725df5a3ac18385d2ed8774a8a84916e6d458872565235f62810139aR387) only to see the execution time jump from [23s](https://github.com/ceramicstudio/js-composedb/actions/runs/7790319671/job/21243920649#step:10:297) to [33s](https://github.com/ceramicstudio/js-composedb/actions/runs/7802211175/job/21279108409#step:10:360) in the relevant test.

When running tests locally, I have a consistent 3s latency on any multiquery call that includes a deterministic stream, even setting the options of `sync: SyncOptions.NEVER_SYNC` and `syncTimeoutSeconds: 0`.
These are only very simple tests with few streams, so I expect it would get way worse for real use-cases, but even then 3s latency per load even without any HTTP latency is likely to seriously affect the UX.

Please have a look at the logic in https://github.com/ceramicstudio/js-composedb/pull/207 to let me know if I'm missing anything, but I don't see how the options I'm setting would cause this 3s delay?

stbrody:
@PaulLeCam I pushed a change to the test in your branch to set the sync options properly in all 3 tests.  This fixed the issue and brought the latency way down in the `using ModelInstanceDocument.set()` test.  The `using loader.loadSet()` is still slow, and I believe it's because the sync options don't get passed through to the lower levels properly.  I'm suspicious of this line for combining the key and opts into a single object: https://github.com/ceramicstudio/js-composedb/blob/8bb251f09c416292644a24fa7f2cceb010f019ee/packages/loader/src/loader.ts#L169

For the test using multiQuery, it seems multiQuery has an issue where the sync options don't get passed all the way down through all the layers.  I can put together a quick fix for that now.

stbrody:
Okay, in my testing this PR brings the latency of a multiquery with genesis contents down from consistently over 3 seconds to a bit over 100ms, when using SyncOptions.NEVER_SYNC: https://github.com/ceramicnetwork/js-ceramic/pull/3150

PaulLeCam:
Closing as the deterministic queries are fixed now, this PR in ComposeDB updates the loading logic to use deterministic queries instead of index ones: https://github.com/ceramicstudio/js-composedb/pull/207

PR #3105, opened 2024-01-19
feat: add streamid to feed document

JulissaDantes:
## Description
As requested after feedback, here we include streamId to the feed document.


stbrody (commented):
We already have CommitID in the feed, I think we just need to explain to Orbis how to use it to get the StreamID

ukstv:
@JulissaDantes What about the consumer? I believe, the codec part is missing.

ukstv:
Or, yeah, @stbrody is absolutely right here.

PR #3107, opened 2024-01-19
feat(streamid): update event id

stephhuynh18:
Now matches the rust-ceramic spec 
Ported tests over

### DIFF
--- a/packages/streamid/src/event-id.ts
+++ b/packages/streamid/src/event-id.ts
@@ -6,8 +6,15 @@ import * as cbor from 'cborg'
 import * as u8a from 'uint8arrays'
 import { STREAMID_CODEC } from './constants.js'
 import { Memoize } from 'mapmoize'
+import { randomCID } from './commit-id.js'
 
 const TAG = Symbol.for('@ceramicnetwork/streamid/EventID')
+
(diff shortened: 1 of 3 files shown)


packages/streamid/src/event-id.ts
```diff
@@ -65,19 +82,57 @@ export class EventID {
...
+      init,
+      eventHeight,
+      event
+    )
+  }
+
   /**
    * Create a new EventID.
    */
   static create(
```
dav1do, line 115: I was wondering if we cared about breaking changes but I'm pretty sure the answer is no, this is unused. Along those lines, do we want to export `EventID` from the index.js?

stephhuynh18: yea this is unused (not even exported HAHA) i'll export it tho so it can be used soon

dav1do (approved):
LGTM. We should definitely update the rust comments as they are out of date and [seeing things](https://github.com/ceramicnetwork/rust-ceramic/blob/0e5fbd30e66942bfc8df75a590be8e411fa3effe/core/src/event_id.rs#L6) like `last8Bytes(val) // 16 bytes` really threw me 😅

PR #3109, opened 2024-01-22

ukstv:
Heavily based on Paul's initial PR.

### DIFF
--- a/packages/indexing/src/database-index-api.ts
+++ b/packages/indexing/src/database-index-api.ts
@@ -36,6 +36,7 @@ export interface IndexStreamArgs {
   readonly tip: CID
   readonly lastAnchor: Date | null
   readonly firstAnchor: Date | null
+  readonly index?: boolean
 }
 
 export type IndicesRecord = Record<string, FieldsIndex>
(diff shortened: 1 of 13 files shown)


PaulLeCam (approved):
Thanks!

packages/indexing/src/database-index-api.ts
```diff
@@ -276,18 +277,25 @@ export abstract class DatabaseIndexApi<DateType = Date | number> {
...
    */
   async indexStream(
     indexingArgs: IndexStreamArgs & { createdAt?: Date; updatedAt?: Date }
   ): Promise<void> {
     const tableName = asTableName(indexingArgs.model)
-    const indexedData = this.getIndexedData(indexingArgs)
-    const toMerge = cloneDeep(indexedData)
-    delete toMerge['created_at']
-    await this.dbConnection(tableName).insert(indexedData).onConflict('stream_id').merge(toMerge)
+    if (indexingArgs.shouldIndex === false) {
```
dbcfd, line 289: This feels weird here. Maybe just needs a rename now, like `performStreamIndexing`.

ukstv: Is it about `shouldIndex` name?

packages/stream-model-instance/src/model-instance-document.ts
```diff
@@ -222,6 +242,32 @@ export class ModelInstanceDocument<T = Record<string, any>> extends Stream {
...
+   * @param shouldIndex - Whether the stream should be indexed or not
+   * @param opts - Additional options
+   */
+  async shouldIndex(shouldIndex: boolean, opts: CommonUpdateOpts = {}): Promise<void> {
+    // No-op if the wanted value is the current one
+    if (this.metadata.shouldIndex === shouldIndex) {
+      return
+    }
+
+    const rawCommit: RawCommit = {
```
stbrody, line 265: nitpick, can implement `shouldIndex()` in terms of `replace()` or `patch()` and avoid some code duplication

packages/stream-model-instance/src/model-instance-document.ts
```diff
@@ -186,11 +206,11 @@ export class ModelInstanceDocument<T = Record<string, any>> extends Stream {
    * @param opts - Additional options
    */
   async replace(content: T | null, opts: UpdateOpts = {}): Promise<void> {
-    opts = { ...DEFAULT_UPDATE_OPTS, ...opts }
+    const { header, ...options } = { ...DEFAULT_UPDATE_OPTS, ...opts }
```
stbrody, line 209: Can we add an arg to this function for any metadata updates, rather than jamming the header into the `opts` (which is violating the type definition of `UpdateOpts`)?

stbrody (changes requested):
sticking `header` into the `opts` arg feels like an abuse of that arg.  For all other streamtype update methods we've had dedicated args for metadata changes.  The `opts` arg is more for controlling behaviors of the node (like syncTimeout, anchoring behavior, etc) rather than for changing the actual data behind the update itself.

packages/stream-model-instance/src/model-instance-document.ts
stbrody: Yeah, that is a bit of a bummer.  Ideally we would insert the `metadata` arg between `content` and `opts`, but that would be a breaking change to the API.  Maybe we should consider a major version bump for the ModelInstanceDocument package?  I really don't like putting data into the `UpdateOpts`, that is fundamentally changing the concept of the `UpdateOpts` type.  At that point, we might as well just have a single argument to all these functions that is a dictionary of all the functions arguments, and never use multiple arguments for anything.

PR #3110, opened 2024-01-22
chore: Add context to errors publishing to pubsub

stbrody:


ukstv (approved):

packages/core/src/pubsub/pubsub.ts
```diff
@@ -120,7 +120,7 @@ export class Pubsub extends Observable<PubsubMessage> {
               topic: this.topic,
               message: logMessage,
             })
-            this.logger.err(error)
+            this.logger.err(`Error received while publishing message to pubsub: ${error}`)
```
ukstv, line 123: Probably we should log a more meaningful error message like "An IPFS node bounced a duplicate message" if it is "validation ignored".

stbrody: Updated

ukstv (approved):

PR #3111, opened 2024-01-23
Feature/ws2 2945 locking locked in model definition

JulissaDantes:
## Description

There is a requirement to enable users to create fields in the model that cannot be altered. This pull request introduces the logic to manage and identify such immutable fields, and it rejects any attempt to modify or update the value of these fields during a commit.


PaulLeCam (commented):
LGTM except the model definition change, I don't think it can be a required field unless the definition uses a new major version.

packages/stream-model/src/codecs.ts
```diff
@@ -274,6 +274,7 @@ export const ModelDefinitionV2 = sparse(
     interface: boolean,
     implements: array(streamIdString),
     schema: ObjectSchema,
+    immutableFields: array(string),
```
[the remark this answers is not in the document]
JulissaDantes: Since not every Model will have locked fields this can be optional

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
```diff
@@ -168,6 +167,7 @@ export class ModelInstanceDocumentHandler implements StreamHandler<ModelInstance
     const newContent = jsonpatch.applyPatch(oldContent, payload.data).newDocument
     const modelStream = await context.api.loadStream<Model>(metadata.model)
     await this._validateContent(context.api, modelStream, newContent, false)
+    await this._validateLockedFieldsUpdate(modelStream, payload)
```
stbrody, line 170: nitpick, but can we move this *into* `_validateContent`?

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
```diff
@@ -312,4 +312,23 @@ export class ModelInstanceDocumentHandler implements StreamHandler<ModelInstance
       }
     }
   }
+
+  /**
+   *  Helper function to validate if locked fields are being mutated
+   */
+  async _validateLockedFieldsUpdate(model: Model, payload: any): Promise<void> {
+    if ('immutableFields' in model.content) {
```
stbrody, line 320: nitpick: you can early return like you do if the array is empty, and save an indentation layer.

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
```diff
@@ -312,4 +312,23 @@ export class ModelInstanceDocumentHandler implements StreamHandler<ModelInstance
...
+      // No locked fields
+      if (model.content.immutableFields.length == 0) return
+
+      for (const lockedField of model.content.immutableFields) {
+        const mutated = payload.data.some(
+          (entry) => entry.op === 'replace' && entry.path.split('/').pop() === lockedField
```
stbrody, line 326: hmm... this makes me nervous.  Are we sure that "replace" is the only way jsonPatch could update a field?  Like, what if the locked field is an array and we try to push a new value into that array?  Can we just error if the field shows up in the payload at all, regardless of "op" type?  Ideally I'd prefer if we didn't have to understand and parse the jsonPatch encoding format at all, but I don't know of a way to do a general-purpose type-aware equality check in javascript, so this might be the best we can do for now.

This also makes me realize we should also have some additional unit tests for different types of updates rather than just full replacement of a string. E.g: add a field to an array, add a field to a sub-object, remove the field, change the type of a value (like from the number `1` to the string "1", on a field where the schema from the Model allows the field to be a number or a string), etc.

stbrody (changes requested):

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
JulissaDantes: You are correct; we do not genuinely care about the type of mutation being performed. If it is present in the payload, we assume that an attempt to make a change is being made.

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
```diff
@@ -218,7 +217,8 @@ export class ModelInstanceDocumentHandler implements StreamHandler<ModelInstance
     ceramic: StreamReader,
     model: Model,
     content: any,
-    genesis: boolean
+    genesis: boolean,
+    payload?: any
```
PaulLeCam, line 221: Could you type the `payload` here please? I think it should be an array of `Operation` from https://github.com/Starcounter-Jack/JSON-Patch/blob/master/src/core.ts#L13

PaulLeCam (changes requested):

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
JulissaDantes: jsonpatch doesnt export the Operation type so i created an equivalent one to be able to type the payload argument.

PaulLeCam (approved):
Thanks for the change!

stbrody (changes requested):
looks like we're still missing the extra test coverage I requested here: https://github.com/ceramicnetwork/js-ceramic/pull/3111#discussion_r1463960613

I'm still worried about how this will work with embedded objects and arrays.

For example, if you have doc like:
```
{
  name: 'spencer'
  address: {
    street: 'main',
    city: 'Boston',
    state: 'MA',
  }
}
```

And you have the "address" field marked as immutable, but you try to do an update that adds a "country" field into the address sub-object, how does that encoded by jsonPatch?  Will that show up as an update to the "address" field, or will the field name be "address.country", in which case it will fail the string equality check with the field name of the locked field, and so the update will go through?

I'd also like to see tests where a field in the sub-object is the one that is specified as locked.  And similar concerns for updates to arrays.

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
```diff
@@ -162,7 +173,7 @@ export class ModelInstanceDocumentHandler implements StreamHandler<ModelInstance
       streamId
     )
 
-    if (payload.header) {
+    if (payload.header) {ß
```
PaulLeCam, line 176: there's an extra character here
```suggestion
    if (payload.header) {
```

JulissaDantes: Thanks for catching this

ukstv:
@JulissaDantes I have an idea. What if we compare current content and a content after doing a JSON patch, and see if the immutable values have changed?

stbrody:
> @JulissaDantes I have an idea. What if we compare current content and a content after doing a JSON patch, and see if the immutable values have changed?

I had that thought too, but I was worried about catching type differences that way.  So long as we're careful to treat `1` (the number) different than `"1"` (the string), and other similar forms of values-that-look-the-same-but-actually-aren't-because-of-type-differences differently, then that could be a valid approach

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
```diff
@@ -342,10 +354,47 @@ export class ModelInstanceDocumentHandler implements StreamHandler<ModelInstance
     if (!('immutableFields' in model.content) || model.content.immutableFields.length == 0) return
 
     for (const lockedField of model.content.immutableFields) {
-      const mutated = payload.data.some((entry) => entry.path.split('/').pop() === lockedField)
+      const mutated = payload.data.some(
+        (entry) => entry.path.slice(1).split('/').shift() === lockedField
```
stbrody, line 358: this still feels pretty brittle to me, so if we can do @ukstv's suggestion of comparing the before and after content but make sure that we do it in a way that catches type changes, then I think that would be better.  But I'll defer to @ukstv on what he thinks is best.

packages/stream-model-instance-handler/src/__tests__/model-instance-document-handler.test.ts
```diff
@@ -1007,50 +1021,133 @@ describe('ModelInstanceDocumentHandler', () => {
...
-    // apply genesis
-    const genesisCommitData = {
-      cid: FAKE_CID_1,
-      type: CommitType.GENESIS,
-      commit: payload,
-      envelope: genesisCommit.jws,
-    }
-
-    const state = await handler.applyCommit(genesisCommitData, context)
+  describe('Immutable checks', () => {
```
stbrody, line 1024: can we also add a test that tries to change the type of a field where the schema allows multiple types, but where the value will look similar?  E.g. a field that allows a string or a number, where you try to change it from `1` (number) to `"1"` (string)?

stbrody (commented):
a few final small suggestions, but mostly looking good

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
PaulLeCam: I disagree, it seems much better to me to check the changes payload rather than having custom verification, especially for the cases you described with changes in nested objects and type changes: having custom logic to compare the values might be prone to edge cases or other bugs, while just checking if there is any change operation seems much safer as only keys (which are always strings) need to be compared.

stbrody: Okay fair.  I don't like us having to understand and parse the jsonPatch format, but I do see that that's a more targeted check to see if anything could possibly have failed with the field we care about.  I'm happy either way

stbrody (approved):
LGTM!  Thank you for all the work on fleshing out the exhaustiveness of the tests!

PR #3114, opened 2024-01-24
WIP feat: Add CERAMIC_ENABLE_V4_MODE env var that disables v3-specific features

stbrody:


packages/core/src/__tests__/ceramic.test.ts
```diff
@@ -33,7 +33,9 @@ function expectEqualStates(a: StreamState, b: StreamState) {
   expect(StreamUtils.serializeState(a)).toEqual(StreamUtils.serializeState(b))
 }
 
-describe('IPFS caching', () => {
+const describeIfV3 = process.env.CERAMIC_ENABLE_V4_MODE ? describe.skip : describe
```
stbrody, line 36: here's an example of how to selectively enable/disable a suite of tests based on which mode we're in.  A very similar thing could be done to make a `testIfV3` to selectively disable just a single test instead of a whole suite.

packages/core/src/dispatcher.ts
```diff
@@ -117,32 +123,36 @@ export class Dispatcher {
...
-      new PubsubRateLimit(
-        new PubsubKeepalive(
-          pubsub,
-          _ipfs,
-          MAX_PUBSUB_PUBLISH_INTERVAL,
-          MAX_INTERVAL_WITHOUT_KEEPALIVE
-        ),
+    this.enableSync = process.env.CERAMIC_ENABLE_V4_MODE ? false : enableSync
+
+    if (!process.env.CERAMIC_ENABLE_V4_MODE) {
```
dav1do, line 132: is having `this.messageBus = undefined` acceptable when it's not null?

packages/stream-caip10-link-handler/src/caip10-link-handler.ts
```diff
@@ -38,6 +38,10 @@ export class Caip10LinkHandler implements StreamHandler<Caip10Link> {
     context: StreamReaderWriter,
     state?: StreamState
   ): Promise<StreamState> {
+    if (process.env.CERAMIC_ENABLE_V4_MODE) {
```
dav1do, line 41: would it be worthwhile to throw errors when creating these stream types, not just when applying a commit? a lot of the tests _may_ pass because they use tile documents but don't actually load/apply a commit, but we should probably skip them and update them to use MIDs still (?)

stbrody: There's no such thing as creating a stream without applying a commit.  A stream is created by applying the genesis commit.

stbrody:
Closing in favor of https://github.com/ceramicnetwork/js-ceramic/pull/3121

PR #3118, opened 2024-01-25
fix: Don't attempt to close Ceramic if it never finished initializing

stbrody:
Otherwise we can get a different error while trying to close half-initialized state, and wind up masking whatever the actual real error was that caused initialization to fail in the first place.


smrz2001 (approved):
🥳

PR #3121, opened 2024-01-26
feat: Add CERAMIC_RECON_MODE env var that disables specific features

dav1do:
## Description

Based on #3114. Disables tests that don't pass yet (or ever). I set the env var in the circle CI pipeline for the rust side only, so the kubo based pipeline will still run all the tests.

Anything that includes TileDocuments or CAIP10 is skipped. I moved a handful of tests around into `describe` blocks to make it easier to skip them (particularly if they were all tile document mixed in with MID tests in some file). There are a number of tests that fail currently that should pass with recon enabled, mostly related to the current IPFS reliance and recon no longer supporting the Kubo API. I tried to mark those with a `describeIfV3ShouldPass` designation, but I'm sure there are some I missed. Many tests will pass locally if you don't specify the `IPFS_FLAVOR` and `CERAMIC_ONE_PATH`. 

To replicate what CI does: use something like this (updating the path accordingly): `CERAMIC_RECON_MODE=true IPFS_FLAVOR=rust CERAMIC_ONE_PATH=/path/to/rust-ceramic/target/debug/ceramic-one npm run test`


packages/core/src/__tests__/ceramic-api.test.ts
```diff
@@ -37,6 +37,12 @@ const generateStringOfSize = (size): string => {
   return random_data.join('')
 }
 
+// Wondering if these should pass but I'm not sure how to fix/change atm.
+// _indexStreamIfNeeded is not called because load() returns SyncStatus.ALREADY_SYNCED
+// we could update the repository.load to make sure indexing happens with v4 but it feels
+// like maybe it should happen when saving the stream instead of loading
+const testIfV3 = process.env.CERAMIC_ENABLE_V4_MODE ? test.skip : test
```
dav1do, line 44: maybe load should just be updated to call `_indexStreamIfNeeded` when in V4 mode (in addition to sync state). I think `load` does make sense as it catches both saved and received things since sync may not be enabled, and when something new comes in you don't really need to index it until someone asks for it. but leaving open for input

packages/core/src/__tests__/ceramic-feed.test.ts
```diff
@@ -30,63 +34,92 @@ describe('Ceramic feed', () => {
...
-    expect(feed2[1].commitId).toStrictEqual(feed1[1].commitId)
-    expect(feed2[1].metadata).toStrictEqual(feed1[1].metadata)
-    expect(feed2[1].content).toStrictEqual(updatedContent)
-
-    s1.unsubscribe()
-    s2.unsubscribe()
-  }, 20000)
-
-  test('add entry after creating/loading indexed model', async () => {
+  skipIfV4ShouldPass('add entry after creating/loading indexed model', async () => {
```
dav1do, line 122: so we short circuit and we never call `_indexStreamIfNeeded` when in V4 mode. I'm not sure if we should call `_updateStateIfPinned_safe` in the repository when in v4 mode, or if the process of storing/recv from recon should update this and it will work when it's implemented.

packages/core/src/dispatcher.ts
```diff
@@ -117,32 +123,36 @@ export class Dispatcher {
...
+        IPFS_NO_MESSAGE_INTERVAL,
+        _pubsubLogger,
         _logger,
-        maxQueriesPerSecond
-      ),
-      !this.enableSync
-    )
+        tasks
+      )
+      this.messageBus = new MessageBus(
```
dav1do, line 142: `this.messageBus` is not set with `CERAMIC_ENABLE_V4_MODE` but it's not nullable... is that acceptable?

packages/core/src/dispatcher.ts
```diff
@@ -157,6 +167,9 @@ export class Dispatcher {
   }
 
   async init() {
+    if (process.env.CERAMIC_ENABLE_V4_MODE) {
```
dav1do, line 170: is it necessary to have `CERAMIC_ENABLE_V4_MODE` as a separate variable? could we just rely on `IPFS_FLAVOR='rust'` as we don't really support rust with v3 (it's possible if you have a specific version of ceramic-one, but not planned going forward)

packages/core/src/state-management/repository.ts
```diff
@@ -226,6 +226,22 @@ export class Repository {
     const opts = { ...DEFAULT_LOAD_OPTS, ...loadOptions }
 
     const [state$, syncStatus] = await this.loadingQ.forStream(streamId).run(async () => {
+      if (process.env.CERAMIC_ENABLE_V4_MODE) {
```
dav1do, line 229: the comment about the MID indexing test relates to this code. we never take [this path](https://github.com/ceramicnetwork/js-ceramic/blob/3b54c616ea799bbedcf38ccca009ee654e19dce0/packages/core/src/state-management/repository.ts#L290) as we return early as we're always considering it already_synced. I assume it will be updated when received, and this is fine and the test will pass then, but just want to make sure.

packages/cli/src/__tests__/index-api.test.ts
```diff
@@ -71,7 +73,7 @@ afterEach(async () => {
   await tmpFolder.cleanup()
 })
 
-test('count', async () => {
+testIfV3ShouldPass('count', async () => {
```
stbrody, line 76: this file I really would have expected to pass.  Would like to understand what the issue is preventing it from passing today

stbrody: It makes me wonder if the changes I made to disable v3 functionality when the V4 env var is set are wrong somehow

packages/core/src/__tests__/ceramic-api.test.ts
```diff
@@ -116,119 +119,220 @@ describe('Ceramic API', () => {
...
-      expect(doc.content).toEqual(CONTENT0)
-      expect(addIndexSpy).toBeCalledTimes(2)
-      await doc.replace(CONTENT1, { anchor: false })
-      expect(doc.content).toEqual(CONTENT1)
-      expect(addIndexSpy).toBeCalledTimes(3)
-      addIndexSpy.mockRestore()
     })
 
-    it('will fail to create stream over size limits', async () => {
+    skipIfV4ShouldPass(
```
stbrody, line 306: again I'd expect this test to pass

packages/core/src/__tests__/ceramic-api.test.ts
```diff
@@ -36,6 +36,9 @@ const generateStringOfSize = (size): string => {
   }
   return random_data.join('')
 }
+// Should  pass on v4 if updated from TileDocument
+const describeIfV3 = process.env.CERAMIC_ENABLE_V4_MODE ? describe.skip : describe
+const skipIfV4ShouldPass = process.env.CERAMIC_ENABLE_V4_MODE ? it.skip : it
```
stbrody, line 41: why is this called `skipIfV4ShouldPass` but in other tests we use `testIfV3ShouldPass`?  Can we be consistent so it's easier to grep for these later?

packages/core/src/__tests__/ceramic-feed.test.ts
stbrody: iiiinteresting.  But even if we don't `_indexStreamIfNeeded` on the load path, I'd expect the stream to still be pinned and indexed on the write path.  If it isn't, then that's a bug.

stbrody: oh I see, but this test is creating it on a different node and then loading it on another node that has the feed.  So yeah then that makes sense since cross-node loading is disabled currently.  So yeah, this test should start passing once Recon is integrated.

packages/core/src/dispatcher.ts
stbrody: ¯\_(ツ)_/¯

If nothing's breaking I guess it's fine?  We could also update the type signature of the member variable to indicate it can be null

packages/core/src/dispatcher.ts
stbrody: I think the `IPFS_FLAVOR` env var only controls behavior of the unit tests currently.  I don't think it has any affect on the runtime behavior of the node, so doing that would be a change to the semantics of that env var.

That said changing the meaning of that env var may be totally fine.  I don't feel super strongly.  This env var is temporary anyway and something that will mostly be for us internally and maybe a handful of partners testing the alpha/beta releases of V4.

I slightly lean towards keeping them separate, but like I said I don't feel strongly, happy to combine them as well if you think that's better.

packages/stream-tests/src/__tests__/admin-api.test.ts
```diff
@@ -35,7 +35,9 @@ const MODEL_DEFINITION: ModelDefinition = {
   },
 }
 
-describe('Admin API tests', () => {
+const describeIfV3ShouldPass = process.env.CERAMIC_ENABLE_V4_MODE ? describe.skip : describe
+
+describeIfV3ShouldPass('Admin API tests', () => {
```
stbrody, line 40: why is this failing?

packages/stream-tests/src/__tests__/basic-indexing.test.ts
```diff
@@ -180,7 +180,10 @@ const envs: Array<BasicIndexingTestEnv> = [
   { dbEngine: DBEngine.postgres },
 ]
 
-describe.each(envs)('Basic end-to-end indexing query test for $dbEngine', (env) => {
+// should pass without changes other than working recon
+const describeIfV3ShouldPass = process.env.CERAMIC_ENABLE_V4_MODE ? describe.skip : describe
+
+describeIfV3ShouldPass.each(envs)('Basic end-to-end indexing query test for $dbEngine', (env) => {
```
stbrody, line 186: this is like **the test** that I wanted to make sure passes in V4 mode before merging this.  Definitely want to understand why this isn't passing.

stbrody (changes requested):
my main comments are all around why the disabled but "shouldPass" tests are failing.  I'd like to see each one have a TODO comment linking to a ticket for re-enabling them and explaining what is getting in the way of them passing now

stbrody:
Oh one other small nitpick: can we make the commit message more focused on the runtime changes to the node than the testing changes? Something like what I used on the original PR "feat: Add CERAMIC_ENABLE_V4_MODE env var that disables v3-specific features" would do

stbrody (approved):
lgtm mod changing the commit message as I mentioned in a previous comment

packages/core/src/dispatcher.ts
dav1do: haha same thought.. i'm okay leaving it as is for now /shrug

dav1do:
> Oh one other small nitpick: can we make the commit message more focused on the runtime changes to the node than the testing changes? Something like what I used on the original PR "feat: Add CERAMIC_ENABLE_V4_MODE env var that disables v3-specific features" would do

Renamed it `feat: Add CERAMIC_RECON_MODE env var that disables specific features` and will adjust the squash commit accordingly.

nathanielc (approved):

PR #3122, opened 2024-01-29

ukstv:
Replaces #3109 due to merge conflicts.

- `shouldIndex` still remains as it was previously agreed on.
- `shouldIndex` is part of opts still: this makes the most sense from DX and release management perspective.

### DIFF
--- a/packages/indexing/src/database-index-api.ts
+++ b/packages/indexing/src/database-index-api.ts
@@ -210,7 +210,7 @@ export abstract class DatabaseIndexApi<DateType = Date | number> {
   /**
    * This method inserts the stream if it is not present in the index, updates
    * the 'content' if the stream already exists in the index, or deletes the
-   * stream from the index if the 'index' arg is set to false.
+   * stream from the index if the 'shouldIndex' arg is set to false.
    * @param indexingArgs
    */
   async indexStream(
--- a/packages/indexing/src/database-index-api.ts
+++ b/packages/indexing/src/database-index-api.ts
@@ -32,6 +32,7 @@ export interface IndexStreamArgs {
   readonly tip: CID
   readonly lastAnchor: Date | null
   readonly firstAnchor: Date | null
+  readonly shouldIndex?: boolean
 }
 
 /**
@@ -207,14 +208,21 @@ export abstract class DatabaseIndexApi<DateType = Date | number> {
   }
 
   /**
-   * This method inserts the stream if it is not present in the index, or updates
-   * the 'content' if the stream already exists in the index.
+   * This method inserts the stream if it is not present in the index, updates
+   * the 'content' if the stream already exists in the index, or deletes the
+   * stream from the index if the 'index' arg is set to false.
    * @param indexingArgs
    */
   async indexStream(
     indexingArgs: IndexStreamArgs & { createdAt?: Date; updatedAt?: Date }
   ): Promise<void> {
     const tableName = asTableName(indexingArgs.model)
+    if (indexingArgs.shouldIndex === false) {
+      await this.dbConnection(tableName)
+        .where('stream_id', indexingArgs.streamID.toString())
+        .delete()
+      return
+    }
     const indexedData = this.getIndexedData(indexingArgs) as Record<string, unknown>
     const relations = this.modelRelations.get(indexingArgs.model.toString()) ?? []
     for (const relation of relations) {
--- a/packages/stream-model-instance-handler/src/__tests__/model-instance-document-handler.test.ts
+++ b/packages/stream-model-instance-handler/src/__tests__/model-instance-document-handler.test.ts
@@ -1117,7 +1117,7 @@ describe('ModelInstanceDocumentHandler', () => {
       envelope: signedCommit.jws,
     }
     await expect(handler.applyCommit(signedCommitData, context, state)).rejects.toThrow(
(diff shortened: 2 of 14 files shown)


packages/indexing/src/database-index-api.ts
```diff
@@ -207,14 +208,21 @@ export abstract class DatabaseIndexApi<DateType = Date | number> {
   }
 
   /**
-   * This method inserts the stream if it is not present in the index, or updates
-   * the 'content' if the stream already exists in the index.
+   * This method inserts the stream if it is not present in the index, updates
+   * the 'content' if the stream already exists in the index, or deletes the
+   * stream from the index if the 'index' arg is set to false.
```
PaulLeCam, line 213:
```suggestion
   * stream from the index if the 'shouldIndex' arg is set to false.
```

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
```diff
@@ -163,11 +163,18 @@ export class ModelInstanceDocumentHandler implements StreamHandler<ModelInstance
...
-          state.metadata
-        )} to ${JSON.stringify(payload.header)}\``
-      )
+      const { shouldIndex, ...others } = payload.header
+      const otherKeys = Object.keys(others)
+      if (otherKeys.length) {
+        throw new Error(
+          `Unsupported metadata changes for ModelInstanceDocument Stream ${streamId}: ${otherKeys.join(
+            ','
+          )}. Only the index argument can be changed.`
```
PaulLeCam, line 172:
```suggestion
          )}. Only the shouldIndex argument can be changed.`
```

packages/stream-model-instance-handler/src/__tests__/model-instance-document-handler.test.ts
```diff
@@ -1117,7 +1117,7 @@ describe('ModelInstanceDocumentHandler', () => {
       envelope: signedCommit.jws,
     }
     await expect(handler.applyCommit(signedCommitData, context, state)).rejects.toThrow(
-      /Updating metadata for ModelInstanceDocument Streams is not allowed/
+      `Unsupported metadata changes for ModelInstanceDocument Stream ${doc.id}: controllers. Only the index argument can be changed.`
```
PaulLeCam, line 1120:
```suggestion
      `Unsupported metadata changes for ModelInstanceDocument Stream ${doc.id}: controllers. Only the shouldIndex argument can be changed.`
```

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
```diff
@@ -163,11 +163,18 @@ export class ModelInstanceDocumentHandler implements StreamHandler<ModelInstance
...
+        )
+      }
+      if (shouldIndex !== null) {
+        state.metadata.shouldIndex = shouldIndex ?? false
+      }
```
PaulLeCam, line 177: Is it intentional to check for an explicit `null` here please?
I think the logic should be more about checking for `null` or `undefined`, no?
```suggestion
      if (shouldIndex != null) {
        state.metadata.shouldIndex = shouldIndex
      }
```

packages/stream-model-instance/src/model-instance-document.ts
```diff
@@ -264,11 +293,26 @@ export class ModelInstanceDocument<T = Record<string, any>> extends Stream {
...
     this.state$.next(updated.state)
   }
 
+  /**
+   * Set the index metadata field for the stream
+   * @param shouldIndex - Whether the stream should be indexed or not
+   * @param opts - Additional options
+   */
+  shouldIndex(shouldIndex: boolean, opts: CommonUpdateOpts = {}): Promise<void> {
+    return this.replace(this.content, { ...opts, shouldIndex: shouldIndex })
```
PaulLeCam, line 313: Wouldn't it be more efficient to use `.patch()` directly to bypass the JSON-patch logic?
```suggestion
    return this.patch([], { ...opts, shouldIndex: shouldIndex })
```

PaulLeCam (approved):

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
ukstv: First wrote about intentionality. Then realised I introduced even more complexity.

packages/stream-model-instance/src/model-instance-document.ts
ukstv: True.

packages/stream-model-instance/src/model-instance-document.ts
```diff
@@ -218,18 +239,25 @@ export class ModelInstanceDocument<T = Record<string, any>> extends Stream {
    * @param opts - Additional options
    */
   async replace(content: T | null, opts: UpdateOpts = {}): Promise<void> {
-    opts = { ...DEFAULT_UPDATE_OPTS, ...opts }
+    const { shouldIndex, ...options } = { ...DEFAULT_UPDATE_OPTS, ...opts }
```
stbrody, line 242: I'm still pretty opposed to putting this into the `UpdateOpts`.  That seems like a huge abuse of that type.  If we're going to though we should at least actually add it to the type definition in StreamOpts.ts.

PR #3124, opened 2024-01-29
feat: make the feed codecs more robust

JulissaDantes:
## Description
The current codec would misbehave for `metadata.model` field, as it is Uint8Array encoded. Also, if we do decode on the current codec, only commitId field is available. This PR introduces a StreamMetadata codec and a more complete AggregationDocument codec.

This change doesn't require a different usage to the current implementation.

### DIFF
--- a/packages/codecs/src/stream.ts
+++ b/packages/codecs/src/stream.ts
@@ -71,7 +71,7 @@ export const commitIdAsString = new Type<CommitID, string, string>(
 /**
  * codeco for StreamMetadata
  */
-export const StreamMetadata = type({
+export const StreamMetadata = sparse({
   controllers: array(string),
   model: optional(streamIdAsString),
   context: optional(streamIdAsString),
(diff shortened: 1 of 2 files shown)


packages/codecs/src/stream.ts
```diff
@@ -67,3 +67,16 @@ export const commitIdAsString = new Type<CommitID, string, string>(
   },
   (commitId) => commitId.toString()
 )
+
+/**
+ * codeco for StreamMetadata
+ */
+export const StreamMetadata = type({
```
ukstv, line 74:
```suggestion
export const StreamMetadata = sparse({
```

ukstv: `optional` is only really available inside `sparse`

ukstv (changes requested):
Pls, change `type` to `sparse` and we are good to go.

ukstv (approved):

PR #3126, opened 2024-01-30
Feature/Include pushing the different event types

JulissaDantes:
## Description
Allows to have different event codes for Firehose so that users could perform different actions according to whether an event is triggered by a stream creation, a stream anchoring or a stream update from the user.


packages/core/src/feed.ts
```diff
@@ -1,20 +1,53 @@
-import type { StreamMetadata, StreamState } from '@ceramicnetwork/common'
+import { CommitType, LogEntry, StreamMetadata, StreamState } from '@ceramicnetwork/common'
 import { Subject, type Observable } from 'rxjs'
 import { CommitID } from '@ceramicnetwork/streamid'
 import { StreamUtils } from '@ceramicnetwork/common'
 
+/**
+ * Describes all event types
+ */
+export enum EventType {
```
stbrody, line 9: I believe the standard names for these from the protocol spec now are Init Events, Data Events, and Time Events.  So presumably that means this enum should have INIT, DATA, and TIME values.

Tagging @oed to confirm.

packages/core/src/feed.ts
```diff
@@ -1,20 +1,53 @@
...
+  CREATE = 0,
+  ANCHOR = 1,
```
stbrody, line 11: toooooottaal nitpick that really doesn't matter at all, but I feel like I'll have an easier time remembering the values these represent if init/create is 0, data/update is 1, and time/anchor is 2.

packages/core/src/feed.ts
```diff
@@ -1,20 +1,53 @@
...
+ * @param log
+ * @returns EventType
+ */
+function parseCommitType(log: LogEntry[]): EventType {
+  const length = log.length
+
+  if (length == 1) return EventType.CREATE
+
+  switch (log[length - 1].type) {
+    case CommitType.ANCHOR:
```
stbrody, line 26: why not just use the CommitType enum (and make a change to the names of the values in that enum to match the current naming convention)?  Why have two enums and map between the two of them?

packages/core/src/feed.ts
```diff
@@ -1,20 +1,53 @@
...
+      return EventType.ANCHOR
+    case CommitType.GENESIS:
+      return EventType.CREATE
+    case CommitType.SIGNED:
+      return EventType.UPDATE
+    default:
```
stbrody, line 32: this worries me, can we throw an error if the log doesn't have the type set properly?  That would probably indicate a bug in our code if that ever happened.

stbrody (changes requested):

ukstv (approved):

PR #3127, opened 2024-01-30
bump major version of common, core, and http-client packages

stbrody:
Because of the recent changes to the interfaces used by the Ceramic and CeramicClient types.


dbcfd (approved):

stbrody:
@dbcfd added a new commit that also updates where these packages are referenced by other packages.  I need to do this otherwise when I run the script to generate the RC it doesn't bump these dependency versions anymore like it should.

Honestly I expected this to fail because those 4.0.0 versions haven't actually been published to NPM yet, but I surprised to see that it did in fact build for me locally.  I think because its smart enough to know to just look in the local filesystem when building locally and the version numbers match what is there locally.

PR #3129, opened 2024-01-31
chore: CommitType and eliminate dependency cycle

ukstv:
- Use CommitType and a corresponding codec
- Now /codecs depend on /common, and /common does not depend on /codecs


JulissaDantes (approved):
LGTM! Thank you

PR #3131, opened 2024-01-31
chore: No signer for AnchorService

ukstv:


dbcfd (approved):

stbrody:
Do we need any corresponding changes to the AuthenticatedEthereumAnchorService?  I believe that class is why the interface has the ceramic setter in the first place.

https://github.com/ceramicnetwork/js-ceramic/blob/eae99ad9112c5a486cbdfe1e12a30efa812b0e76/packages/core/src/anchor/ethereum/ethereum-anchor-service.ts#L145

packages/core/src/ceramic.ts
```diff
@@ -463,7 +463,6 @@ export class Ceramic implements StreamReaderWriter, StreamStateLoader {
         await this._ipfsTopology.start()
       }
 
-      this.anchorService.signer = this._signer
```
stbrody, line 466: actually, will removing this break the ability of Ceramic to authenticate to the CAS?

dbcfd: Signer is now in the auth class itself, so anchor service no longer needs it https://github.com/ceramicnetwork/js-ceramic/blob/eae99ad9112c5a486cbdfe1e12a30efa812b0e76/packages/core/src/anchor/auth/did-anchor-service-auth.ts#L24

stbrody: okay awesome.  Do we even need the AuthenticatedEthereumAnchorService class at all anymore then?

stbrody: oh wait, yes we do, I see now it overrides the sendRequest method.

But we can probably at least remove the Ceramic setter method from it though?

PR #3133, opened 2024-01-31
fix: event id can parse hex string

stephhuynh18:
We can possible receive hex strings from the recon api. We made assumptions about the even height byte length that we should not have.


dav1do (approved):
Nice,correct and easier to understand now! I still don't know why we need cbor and varint but that's a topic for a different thread.

PR #3135, opened 2024-02-01
fix: use lerna conventional-prerelease for nightlies

smrz2001:
`--canary` mode appears to have a bug where it won't update internal dependencies on publish. e.g. If package A (`v1`) and B (`v1`) are published and A refers internally to B, then canary mode will publish A => `v1.1-nightly.0` and B => `v1.1-nightly.0` but A will continue to refer to B => `v1` internally.

### DIFF
--- a/.github/workflows/reusable-release.yml
+++ b/.github/workflows/reusable-release.yml
@@ -306,7 +306,7 @@ jobs:
         name: Initialize and build code
         run: npm ci && npm run build
       -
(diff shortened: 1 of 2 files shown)


3benbox (approved):
npm plz

smrz2001:
> npm plz

You mean npm directly instead of lerna? Lol yes, one day... 😂

PR #3136, opened 2024-02-01
feat(stream-model,stream-model-handler): Add support for `relationSetFrom` view in model definition

PaulLeCam:
Enables defining a relation to a single document in a set, for example accessing a "like" on a post for a given input DID.

### DIFF
--- a/packages/stream-model-handler/src/views-utils.ts
+++ b/packages/stream-model-handler/src/views-utils.ts
@@ -7,6 +7,7 @@ const SUPPORTED_VIEW_TYPES = [
   'relationDocument',
   'relationFrom',
   'relationCountFrom',
+  'relationSetFrom',
 ]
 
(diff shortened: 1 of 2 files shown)


dbcfd:
@PaulLeCam can you describe why this is needed? Why wouldn't we grab all likes on a post? Why would we specifically grab only one?

PaulLeCam:
We can already access all the likes with the `relationFrom` view, but having to access to a single one is necessary to indicate if a SET account relation exists for a given account. For example: does the current user like the post or not?
```graphql
query ($accountID: DID!) {
  post: node(id: "...") {
    ... on Post {
      title
      # Access the unique Like document for the given account (DID) and Post, if set
      like(account: $accountID) {
        id
      }
    }
  }
}
```

ukstv (approved):

PR #3137, opened 2024-02-01
DO NOT MERGE: Prepare v3 hotfix

stbrody:
This branch sets up the version dependencies so that a hotfix can be built from the branch.  By running the hotfix workflow (https://github.com/ceramicnetwork/js-ceramic/blob/eae99ad9112c5a486cbdfe1e12a30efa812b0e76/package.json#L19), the minor versions of all packages will be bumped and published to NPM.  So `stream-model-handler` will go to v2.4.1, `cli` will go to 3.4.1, etc.  The 3 previously-bumped packages (`common`, `core`, `http-client`) will all to v3.4.1 (note there will never actually be a 3.4.0 in NPM, those will have gone from 3.3.0 to 3.4.1)


packages/common/package.json
ukstv, line 1: Oh, lerna!

ukstv: I mean, confusing at the first sight, but i get why.

ukstv (approved):

stbrody:
Wait, is this safe?  Because the code still uses the new interface types, so now we're doing a release that removes the `CeramicApi` type from `common`, but doesn't bump the major version...

ukstv:
> Wait, is this safe? Because the code still uses the new interface types, so now we're doing a release that removes the `CeramicApi` type from `common`, but doesn't bump the major version...

Oh, common, http-client, core should not be released at all.

stbrody:
taking a different approach.  Going to unpublish the last, broken release, then publish new packages that bump the major version on everything

stbrody:
nevermind, back to using this hotfix branch, but now with reverting a ton of stuff to get back to a state before the breaking type changes were made to core common and http-client

stbrody:
nevermind again, final approach made a new release based off the previous 3.3.0 release in the `release/firehost-mvp` branch.  Hotfix is now complete

PR #3138, opened 2024-02-01
feat: change all commitType name and values

JulissaDantes:
## Description
if we're going to start having users consume the CommitType enum publicly more, we should probably update the names of the types within it to match the standard names in the protocol spec. So GENESIS->INIT, SIGNED->DATA, and ANCHOR->TIME. Also rename CommitType->EventType.

### DIFF
--- a/packages/common/src/stream.ts
+++ b/packages/common/src/stream.ts
@@ -90,10 +90,10 @@ export interface StreamNext {
   metadata?: StreamMetadata
 }
 
-export enum CommitType {
-  GENESIS,
-  SIGNED,
-  ANCHOR,
+export enum EventType {
+  INIT,
+  DATA,
+  TIME,
 }
 
 /**
@@ -104,7 +104,7 @@ export interface LogEntry {
   cid: CID
 
   // Type of the commit (e.g. genesis, signed, anchor)
-  type: CommitType
+  type: EventType
 
   // Timestamp (in seconds) of when this commit was anchored (if available)
   timestamp?: number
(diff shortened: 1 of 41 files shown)


packages/common/src/stream.ts
```diff
@@ -90,10 +90,10 @@ export interface StreamNext {
   metadata?: StreamMetadata
 }
 
-export enum CommitType {
-  GENESIS,
-  SIGNED,
-  ANCHOR,
+export enum EventType {
```
stbrody, line 93: can we add comments that these used to be called genesis, signed, and anchor, and may still be referred to this way by some other code and comments?  Just to avoid confusion for folks who see types like `GenesisHeader` and aren't sure if genesis and init mean the same thing (they do).

And also maybe add some TODO comments over the other types in this file like `GenesisHeader` to rename them to be aligned with the new naming conventions.

stbrody (approved):
lgtm mod request for a few comments to clarify the name evolution to new readers of the code

packages/common/src/stream.ts
```diff
@@ -120,6 +123,7 @@ export interface LogEntry {
  * to load anything else from ipfs or the p2p Ceramic network.
  */
 export interface CommitData extends LogEntry {
+  // TODO use latest naming convention genesis -> init, signed -> data, anchor -> time
```
stbrody, line 126: I think this comment can be removed, nothing here is actually using the old naming convention

stbrody (approved):

packages/common/src/stream.ts
JulissaDantes: I added it because some properties inside that interface use the previous convention

PR #3140, opened 2024-02-02
Feature/ws2 2940 end to end tests going through the http api

JulissaDantes:
## Description
During the[ #3058](https://github.com/ceramicnetwork/js-ceramic/pull/3058) PR review the desire for this test coverage was required.


packages/cli/src/__tests__/ceramic-daemon.test.ts
```diff
@@ -493,6 +497,32 @@ describe('Ceramic interop: core <> http-client', () => {
...
+      messageEvent = decode(Codec, event.data)
+    })
+    //this will trigger a feed event
+    const MODEL_DEFINITION: ModelDefinition = {
+      name: 'myModel',
+      version: '1.0',
+      schema: { type: 'object', additionalProperties: false },
+      accountRelation: { type: 'list' },
+    }
+    // create model on different node
```
stbrody, line 516: is this actually being created on a different node?  Looks like this is the same node as we're getting the feed from

stbrody (approved):
LGTM

packages/cli/src/__tests__/ceramic-daemon.test.ts
JulissaDantes: you are right, the test was initially different and when I changed it didnt delete the comment. Thanks for catching this, just removed it.

PR #3142, opened 2024-02-03
use interval publishing metrics for anchor polling loop delay

gvelez17:
## Description

We need to track the latency of the anchor requests in the polling loop

however we need to make sure not to add excessive overhead

this pr adds reporting with interval-based publishing if prometheus metrics are turned on

Under the Hood:
     calls to the TimeableMetric.count() are added to an internal counter only 
     there is a setInterval call that runs at the requested interval that then calls the observe code which creates a scrapable observation on prometheus for the average and max over the interval
      
      This mechanism is all hidden from js-ceramic.  The reason for it underneath is that we want to avoid creating load and the actual OTLP observe has a non-negligible overhead.  So we use this lighter weight calculation for every stream, and the heavier observe only once per interval.  

## How Has This Been Tested?

- [x] unit tests
- [ ] manual verification of metrics output

## PR checklist

- [x ] I have updated the READMEs of affected packages

## References:

### DIFF
--- a/packages/core/src/anchor/anchor-processing-loop.ts
+++ b/packages/core/src/anchor/anchor-processing-loop.ts
@@ -8,7 +8,7 @@ import type { NamedTaskQueue } from '../state-management/named-task-queue.js'
 import type { StreamID } from '@ceramicnetwork/streamid'
 import { TimeableMetric, SinceField } from '@ceramicnetwork/observability'
 
-const METRICS_REPORTING_INTERVAL_MS = 10000  // 10 second reporting interval
+const METRICS_REPORTING_INTERVAL_MS = 10000 // 10 second reporting interval
 
 /**
  * Get anchor request entries from AnchorRequestStore one by one. For each entry, get CAS response,
@@ -34,7 +34,11 @@ export class AnchorProcessingLoop {
     anchorStoreQueue: NamedTaskQueue
   ) {
     this.#anchorStoreQueue = anchorStoreQueue
-    this.#anchorPollingMetrics = new TimeableMetric(SinceField.TIMESTAMP, 'anchorPollingDelay', METRICS_REPORTING_INTERVAL_MS)
+    this.#anchorPollingMetrics = new TimeableMetric(
+      SinceField.TIMESTAMP,
+      'anchorPollingDelay',
+      METRICS_REPORTING_INTERVAL_MS
+    )
     this.#loop = new ProcessingLoop(logger, store.infiniteList(batchSize), async (streamId) => {
       try {
         logger.verbose(
--- a/packages/core/src/anchor/anchor-processing-loop.ts
+++ b/packages/core/src/anchor/anchor-processing-loop.ts
@@ -36,7 +36,7 @@ export class AnchorProcessingLoop {
     this.#anchorStoreQueue = anchorStoreQueue
     this.#anchorPollingMetrics = new TimeableMetric(
       SinceField.TIMESTAMP,
-      'anchorPollingDelay',
+      'anchorRequestAge',
       METRICS_REPORTING_INTERVAL_MS
     )
     this.#loop = new ProcessingLoop(logger, store.infiniteList(batchSize), async (streamId) => {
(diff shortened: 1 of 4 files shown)


ukstv (approved):
I feel like `anchorPollingDelay` could have a better name, but have no alternatives to propose. So, let it be.

gvelez17:
@stbrody updated the description above with more detail about why/how it works

gvelez17:
> I feel like `anchorPollingDelay` could have a better name, but have no alternatives to propose. So, let it be.

what about anchorRequestAge ?
or anchorRequestLatency

ukstv:
> > I feel like `anchorPollingDelay` could have a better name, but have no alternatives to propose. So, let it be.
> 
> what about anchorRequestAge ? or anchorRequestLatency

I really like anchorRequestAge!!

packages/core/src/anchor/anchor-processing-loop.ts
```diff
@@ -46,6 +55,7 @@ export class AnchorProcessingLoop {
           `Anchor event with status ${event.status} for commit CID ${entry.cid} of Stream ${streamId} handled successfully`
         )
         if (isTerminal) {
+          this.#anchorPollingMetrics.record(entry)
```
stbrody, line 58: does this internally somehow parse the entry and understand what StreamID it applies to?  Like the age of a request in the polling loop is going to be per-stream.  You might want to take the average across all streams to be what gets exposed, but you still need to track the request lifetime per-stream.  That's the part that's hard to understand just from reading this code.  There's nothing signaling down into the metric system what field from `entry` to use to track the age of the request.  If this *isn't* somehow turning into per-streamid age tracking at the lower level, then this metric likely isn't tracking the right thing.  If you're just recording a timestamp every time this line of code gets hit, you're tracking something weird.  This line will get hit every time *any stream* reaches a terminal state.  So that means generally the loop may run many times without hitting this line, until an anchor batch runs, at which point this line will get many hundreds of times rapid fire, once per each stream anchored in the batch.

gvelez17: right, it uses entry.timestamp, the SINCE_FIELD is set on start()

stbrody: okay, I read through the code in https://github.com/ceramicnetwork/observability/pull/16, and now I understand how this is working.  It's getting the start timestamp from the entry itself.  This is super not clear to a reader of the code in this PR though, without a pretty deep understanding of what the service-metrics package is doing internally.  Can we add some comments here to explain?

The good news is now I do think this is actually doing the right thing after all, I just didn't understand it before.

packages/core/src/anchor/anchor-processing-loop.ts
```diff
@@ -30,6 +34,11 @@ export class AnchorProcessingLoop {
     anchorStoreQueue: NamedTaskQueue
   ) {
     this.#anchorStoreQueue = anchorStoreQueue
+    this.#anchorPollingMetrics = new TimeableMetric(
+      SinceField.TIMESTAMP,
```
gvelez17, line 38: @stbrody here is where it sets the "since" field ^^^^

stbrody: Yes I saw that, but seeing us set an enum that says we should be using a TIMESTAMP, did not make me immediately understand that whenever we call `record` we are comparing the current time to a timestamp taken from the document that we pass in, and that that document itself has a timestamp that was set, in a totally different part of the codebase, based on the timestamp of when the entry was persisted to the state store.  That's a lot of external context that is not available there at the call site.

packages/core/src/anchor/anchor-processing-loop.ts
stbrody, line 58:
```suggestion
          // Record the latency of how long this particular anchor request existed in the
          // AnchorRequestStore.  The "record" function will compare the current time to the
          // "timestamp" field from "entry", which was set as the current time when the request was
          // first written into the AnchorRequestStore.
          this.#anchorPollingMetrics.record(entry)
```

stbrody (approved):
Okay, lgtm with one suggested edit to add a comment explaining what is going on under the hood

packages/core/src/anchor/anchor-processing-loop.ts
gvelez17: True!

PR #3146, opened 2024-02-06

ukstv:
Again.

**BEWARE** This is a breaking change for ModelInstanceDocument API.

### DIFF
--- a/packages/stream-model-instance/src/model-instance-document.ts
+++ b/packages/stream-model-instance/src/model-instance-document.ts
@@ -232,19 +232,31 @@ export class ModelInstanceDocument<T = Record<string, any>> extends Stream {
   /**
    * Update an existing Model Instance Document by replacing its content
    * @param content - New content to replace old content
+   * @param metadata
    * @param opts - Additional options
    */
-  async replace(content: T | null, opts: UpdateOpts = {}): Promise<void> {
+  async replace(
+    content: T | null,
+    metadata: Partial<ModelInstanceDocumentMetadataArgs> | undefined | null = undefined,
+    opts: UpdateOpts = {}
+  ): Promise<void> {
     opts = { ...DEFAULT_UPDATE_OPTS, ...opts }
     validateContentLength(content)
     const signer: CeramicSigner = opts.asDID
       ? CeramicSigner.fromDID(opts.asDID)
       : opts.signer || this.api.signer
+    let header: Partial<CommitHeader> | undefined = undefined
+    if (metadata && metadata.shouldIndex != null) {
+      header = {
+        shouldIndex: metadata.shouldIndex,
+      }
+    }
     const updateCommit = await ModelInstanceDocument.makeUpdateCommit(
       signer,
       this.commitId,
       this.content,
-      content
+      content,
+      header
     )
     const updated = await this.api.applyCommit(this.id, updateCommit, opts)
     this.state$.next(updated.state)
@@ -254,6 +266,7 @@ export class ModelInstanceDocument<T = Record<string, any>> extends Stream {
    * Update the contents of an existing Model Instance Document based on a JSON-patch diff from the existing
    * contents to the desired new contents
    * @param jsonPatch - JSON patch diff of document contents
+   * @param metadata - Metadata change
    * @param opts - Additional options
    */
   async patch(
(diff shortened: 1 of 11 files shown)


PaulLeCam (approved):
LGTM but I'm wondering about versioning, as it's a breaking change for the MID API we need to bump it to a new major version, so also the HTTP client and the CLI, and maybe some other packages?

ukstv:
Yup @PaulLeCam this is indeed a breaking change.

PaulLeCam:
@ukstv @stbrody can we try to get it to today's RC please?
I think this change along with the last 2 commits (https://github.com/ceramicnetwork/js-ceramic/commit/31c70c988d476471e78ef3ab6da539459d3b3b3c and https://github.com/ceramicnetwork/js-ceramic/commit/4f86082859f2c73e59fb3ae7a9448645be3ea1bb) are all we need to unblock the new version of ComposeDB.

Regarding these 2 commits I'm not sure why they didn't get picked up in the nightly, is it because of the conventional commits format? If so, I opened a PR to bump their versions: https://github.com/ceramicnetwork/js-ceramic/pull/3147, if we can make sure they get to the RC as well?

packages/stream-model-instance/src/model-instance-document.ts
```diff
@@ -264,11 +298,26 @@ export class ModelInstanceDocument<T = Record<string, any>> extends Stream {
       prev: this.tip,
       id: this.id.cid,
     }
+    // Null check is necessary to avoid `undefined` value that can't be encoded with IPLD
+    if (metadata.shouldIndex != null) {
```
stbrody, line 302: don't we also need to check if metadata itself is null/undefined, like we do `replace` above?  Otherwise we could get an error trying to access the `shouldIndex` field of `undefined` right?

packages/stream-model-instance/src/model-instance-document.ts
```diff
@@ -264,11 +298,26 @@ export class ModelInstanceDocument<T = Record<string, any>> extends Stream {
...
     const updated = await this.api.applyCommit(this.id, commit, opts)
     this.state$.next(updated.state)
   }
 
+  /**
+   * Set the index metadata field for the stream
+   * @param shouldIndex - Whether the stream should be indexed or not
+   * @param opts - Additional options
+   */
+  shouldIndex(shouldIndex: boolean, opts: UpdateOpts = {}): Promise<void> {
```
stbrody, line 317: I'd vote for calling this method `setShouldIndex`.  A method called `shouldIndex` sounds to me like it would be a read-only accessor method that returns a boolean.

packages/stream-model-instance/src/model-instance-document.ts
```diff
@@ -291,14 +340,16 @@ export class ModelInstanceDocument<T = Record<string, any>> extends Stream {
    * @param prev - The CommitID of the current tip of the Stream that the update should be applied on top of.
    * @param oldContent - The current content of the Stream.
    * @param newContent - The new content to update the Stream with.
+   * @param header - New commit header
```
stbrody, line 343:
```suggestion
   * @param header - New commit header, used to update the stream's metadata.  Metadata fields not specified will be left alone with their current values.
```

stbrody (approved):
lgtm mod a few small comments.

When I make the RC tomorrow I'll make sure to bump the major version number of all the js-ceramic packages.

stbrody:
> Regarding these 2 commits I'm not sure why they didn't get picked up in the nightly, is it because of the conventional commits format?

I'm not sure why those didn't get into the nightly, seems like they should have.  @smrz2001 any ideas?

ukstv:
> I'd vote for calling this method setShouldIndex. A method called shouldIndex sounds to me like it would be a read-only accessor method that returns a boolean.

Oh, where have you been when we all together decided what is a proper name for the setter here!?!

ukstv:
Screenshot 2024-02-08 at 13 38 16

@stbrody

packages/stream-model-instance/src/model-instance-document.ts
PaulLeCam: Yes good catch, I fixed it

PaulLeCam:
Updated to resolve conflict and addressing some of @stbrody's comments, but as discussed with @ukstv leaving the `shouldIndex` method as-is.

stbrody:
@ukstv that conversation and screenshot you posted weren't specifically about the setter method.  That was a much broader conversation and was mostly focused on what the name of the field in the metadata should be called, and was more focused on the question of if it should be called `delete` vs `unindex` vs `shouldIndex`, rather than if the specific setter method should be called `shouldIndex` vs `setShouldIndex` etc.

That said I don't feel strongly enough to fight on this.  If you prefer it this way it's fine.

PaulLeCam:
@ukstv @stbrody are we good to merge or shall we change the method name please?
I don't mind either way about the method name, I'd just like to make sure the change makes it to the RC.

stbrody:
Going to merge this as-is so I can cut the RC.  If I manage to convince you that adding `set` into the name is preferable, we can still change it before the final release 😉

PR #3148, opened 2024-02-07
feat: allow database connection pool max size to be configurable

dbcfd:


packages/indexing/src/build-indexing.ts
```diff
@@ -84,11 +89,12 @@ export function buildIndexing(
     }
     case 'postgres':
     case 'postgresql': {
-      logger.imp('Initializing PostgreSQL connection')
+      const max = indexingConfig.maxConnectionPoolSize || 10
```
stbrody, line 92: do we want to bump the default value up at all while we're at it?

stbrody (approved):
lgtm

ukstv (approved):
🔥

packages/indexing/src/build-indexing.ts
dbcfd: I think for now I want to leave it at 10, since that's the current behavior. We can look at bumping it up if we are adjusting at lots of customers.

PR #3149, opened 2024-02-07
fix: proper serialization/deserialization of context field

stbrody:
I think there's a decent chance this fixes https://linear.app/3boxlabs/issue/WS2-2979/orbis-codec-issue as well.  FYI @JulissaDantes

### DIFF
--- a/packages/common/src/utils/stream-utils.ts
+++ b/packages/common/src/utils/stream-utils.ts
@@ -74,6 +74,10 @@ export class StreamUtils {
       cloned.header.model = uint8arrays.toString(commit.header.model, 'base64')
     }
 
+    if (commit.header?.context) {
+      cloned.header.context = uint8arrays.toString(commit.header.context, 'base64')
+    }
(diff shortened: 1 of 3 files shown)


packages/common/src/utils/stream-utils.ts
```diff
@@ -142,8 +150,8 @@ export class StreamUtils {
     if (state.metadata?.model) {
       cloned.metadata.model = state.metadata.model.toString()
     }
-    if (state.next?.metadata?.model) {
```
stbrody, line 145: this is safe to delete because only TileDocument does the thing with using `next` for unanchored changes to content and metadata.  Model and ModelInstanceDocument streams never have a `next` state, and those are the only ones that use `model` in their metadata

oed: Makes sense that we never have `model` in next.

oed (approved):
lgtm

PR #3151, opened 2024-02-08
fix: create unique package versions for nightly releases

smrz2001:
I was unnecessarily complicating the nightly releases. This PR simplifies the flow and eliminates the package versioning conflicts the workflow was running into previously.

Not having the `--preid` in the `package.json` command means that, if left unspecified when called manually, it will use the default of `beta`.

I've been able to create multiple releases using the workflow, e.g. for [stream-model](https://www.npmjs.com/package/@ceramicnetwork/stream-model?activeTab=versions), even without any changes to the package.

I was able to update the e2e tests to use the new versions and they compiled. I'm also going to run them and see if they pass after auto-updating.

### DIFF
--- a/.github/workflows/reusable-release.yml
+++ b/.github/workflows/reusable-release.yml
@@ -306,27 +306,8 @@ jobs:
         name: Initialize and build code
         run: npm ci && npm run build
       -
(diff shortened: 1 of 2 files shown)


PaulLeCam (approved):
Nice, thanks!

PR #3152, opened 2024-02-08
feat: Batch processing for an anchor loop

ukstv:
batch size = concurrency

### DIFF
--- a/packages/core/src/anchor/processing-loop.ts
+++ b/packages/core/src/anchor/processing-loop.ts
@@ -49,7 +49,7 @@ export class ProcessingLoop<T> {
   /**
    * Source of entries we process in a loop.
    */
-  private readonly source: AsyncGenerator<T>
+  private readonly source: AsyncGenerator<Array<T>>
 
   /**
    * External action that processes an entry.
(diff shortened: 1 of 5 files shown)


packages/core/src/anchor/processing-loop.ts
```diff
@@ -107,12 +107,13 @@ export class ProcessingLoop<T> {
...
-          const value = next.value
-          if (!value) {
+          const values = next.value
+          if (!values) {
             this.#logger.verbose(`No value received in ProcessingLoop, skipping this iteration`)
             continue
           }
-          await Promise.race([this.handleValue(value), rejectOnAbortSignal])
+          const handlersP = values.map((value) => this.handleValue(value))
+          await Promise.race(handlersP.concat(rejectOnAbortSignal))
```
dbcfd, line 116: Shouldn't this be 

`await Promise.race(Promise.all(handlersP), rejectOnAbortSignal)`

We want all the handlers to finish, but we want to race between handlers finishing and being aborted.

dbcfd (changes requested):

packages/core/src/anchor/processing-loop.ts
ukstv: yes!

ukstv: there now.

dbcfd (approved):

stbrody:
This works, though I have two (admittedly both on the relatively minor side) issues with the approach. 

One concern is that combining the batch size used for fetching requests from the store and the concurrency limit of processing those requests into a single knob limits us.  I think it makes sense to fetch a large number of requests from the store at once (like the 1000 value we previously had here), but we probably want to limit the concurrency of how many requests get processed at once quite a bit more than that, since we don't want to overwhelm the CAS or use too much memory on the Ceramic node holding all the CAR file responses from that many requests in memory at once.

The other possible improvement I see over this approach is that rather than doing things in batches where every single request from the batch must be fully processed before we start any requests in a new batch, we can instead specify the number of requests we want processing concurrently and then keep a steady supply of requests always in processing up to that limit.  That can keep us busier, processing things faster, and smoothing out the load.

I took a crack at implementing an alternative approach that addresses these two issues here: https://github.com/ceramicnetwork/js-ceramic/pull/3153, PTAL.

ukstv:
Okay.

PR #3153, opened 2024-02-08
feat: Add concurrency to ProcessingLoop

stbrody:
An alternative approach to https://github.com/ceramicnetwork/js-ceramic/pull/3152 that addresses the concerns raised in my comment there.

This changes the ProcessingLoop to swallow errors instead of returning an encountered error when the loop shuts down as it did before.  This is because now that multiple tasks are running concurrently, it's possible to get multiple errors at once, and it isn't clear what error should be returned by the loop.  Given that nothing outside of tests was actually doing anything with the returned error other than logging it anyway, just logging it and otherwise swallowing errors seems fine.

### DIFF
--- a/packages/core/src/anchor/anchor-processing-loop.ts
+++ b/packages/core/src/anchor/anchor-processing-loop.ts
@@ -10,6 +10,8 @@ import { TimeableMetric, SinceField } from '@ceramicnetwork/observability'
 
 const METRICS_REPORTING_INTERVAL_MS = 10000 // 10 second reporting interval
 
+const DEFAULT_CONCURRENCY = 10
+
 /**
  * Get anchor request entries from AnchorRequestStore one by one. For each entry, get CAS response,
  * and handle the response via `eventHandler.handle`.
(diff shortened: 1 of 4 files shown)


packages/core/src/anchor/anchor-processing-loop.ts
```diff
@@ -39,46 +41,54 @@ export class AnchorProcessingLoop {
...
-            const loaded = await store.load(streamId)
-            if (loaded.cid.equals(entry.cid)) {
-              await store.remove(streamId)
-            }
+
+    const concurrency =
+      Number(process.env.CERAMIC_ANCHOR_POLLING_CONCURRENCY) || DEFAULT_CONCURRENCY
+    this.#loop = new ProcessingLoop(
+      logger,
+      concurrency,
```
stbrody, line 49: this looks like a big diff but it's mostly formatting.  Literally the only change here is passing the `concurrency` arg to the `ProcessingLoop` constructor.

packages/core/src/anchor/processing-loop.ts
```diff
@@ -100,28 +110,41 @@ export class ProcessingLoop<T> {
...
+              await Promise.race([this.handleValue(value), rejectOnAbortSignal])
+            })
+            .catch((e) => {
+              this.#logger.err(`Error in ProcessingLoop: ${e}`)
+              this.#abortController.abort('ERROR')
+            })
+            .finally(() => {
+              this.#runningTasks.delete(curTaskId)
+            })
+          this.#runningTasks.set(taskId++, task)
```
dbcfd, line 138: This feels incorrect. Beyond just number overrun, this functionality should be run in one of the available 'slots' with the task pulled into a slot as others complete.

dbcfd (changes requested):

packages/core/src/anchor/processing-loop.ts
stbrody: Yeah, unfortunately no other approach I could think of allowed a task to easily remove itself from the set on completion.  For now I just modded the taskId by Number.MAX_SAFE_INTEGER to handle wrap-around

dbcfd (approved):

dbcfd:
@stbrody @ukstv both pr's are approved. There's merits in either approach, may make sense to try to load test to determine which approach to use.

PR #3154, opened 2024-02-08
feat: Disable anchoring when CERAMIC_RECON_MODE env var is set

stbrody:
This does reduce our test coverage of basic composedb features a good amount though.  I did this without modifying the body of any tests, just disabling any test that fails, but many of them fail just because the tests happens to check that the anchor status is PENDING in some place, without really relying on anchoring as a core part of the test.  I could, instead, use the env var inside of test bodies to selectively skip just the parts of the tests that are breaking, but leaving the bulk of many of the tests in tact.  It would just muddy up some of the tests a bit and to put them back later we'd have to grep for uses of the `CERAMIC_RECON_MODE` flag instead of just being able to search for `testIfV3ShouldPassWithAnchoring` and `describeIfV3ShouldPassWithAnchoring`.

I kind of think that would be a better way to go, but curious what others think.

EDIT: I went ahead and made the change.  You can see the two different commits in this PR if you want to compare the difference in the two approaches.

### DIFF
--- a/packages/core/src/state-management/repository.ts
+++ b/packages/core/src/state-management/repository.ts
@@ -585,6 +585,11 @@ export class Repository {
       return
     }
 
+    if (process.env.CERAMIC_RECON_MODE) {
+      // TODO(WS1-1471): Enable anchoring when in Recon (Prime) mode.
+      return
+    }
+
     const carFile = await this.#deps.anchorRequestCarBuilder.build(state$.id, state$.tip)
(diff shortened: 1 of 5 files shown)


dav1do (approved):
Looks good! I didn't realize how "easy" it would be to disable the anchoring logic.. I almost want this as a separate flag I can set for benchmarking runs (but at least I know how to build an image like that now). 

I'm good with the test changes, but I do question if the anchor status check is actually relevant and adding anything to the tests.

PR #3156, opened 2024-02-09
feat: enforce immutability constraints when working with interfaces

JulissaDantes:
## Description
Currently js-composedb enforces that if an interface being implemented has an immutable fields this constraint should be kept by the implementer. This PR adds the enforcement on the server side.

### DIFF
--- a/packages/stream-model-handler/src/interfaces-utils.ts
+++ b/packages/stream-model-handler/src/interfaces-utils.ts
@@ -695,15 +695,22 @@ export function isValidImmutabilityImplementation(
   expected: ModelDefinition,
   implemented: ModelDefinition
 ): boolean {
-  if (expected.version === '2.0') {
-    const expectedModelV2 = expected as ModelDefinitionV2
-    const implementedModelV2 = implemented as ModelDefinitionV2
-    // No immutable fields
-    if (!expectedModelV2.immutableFields && !implementedModelV2.immutableFields) return true
-
-    const implementedSet = new Set(implementedModelV2.immutableFields ?? [])
-    return (expectedModelV2.immutableFields ?? []).every((item) => implementedSet.has(item))
-  } else {
+  if (
+    expected.version === '1.0' ||
+    expected.immutableFields == null ||
+    expected.immutableFields.length === 0
+  ) {
+    // No immutable field expected
     return true
   }
+  if (
+    implemented.version === '1.0' ||
+    implemented.immutableFields == null ||
+    implemented.immutableFields.length === 0
+  ) {
+    // No immutable field implemented
+    return false
+  }
+  // Check all expected fields are implemented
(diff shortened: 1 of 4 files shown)


packages/stream-model-handler/src/interfaces-utils.ts
```diff
@@ -687,3 +690,11 @@ export async function validateImplementedInterfaces(
     throw new AggregateError(errors, `Interfaces validation failed for model ${model.name}`)
   }
 }
+function isValidImmutabilityImplementation(
+  expected: ModelDefinition,
+  implemented: ModelDefinition
+): boolean {
+  if (implemented.version == '1.0') return true
+  const implementedSet = new Set(implemented.immutableFields)
+  return (expected as ModelDefinitionV2).immutableFields.every((item) => implementedSet.has(item))
```
PaulLeCam, line 699: Would be good to add a type-check for the model definition here, for example we could have a v3 of the model definition in the future that changes the format of the `immutableFields`, but the type cast would hide the error here.

Also, I think `immutableFields` can be `undefined` in the definition, no? If so, we need to check it is set.

JulissaDantes: I updated this part, let me know if you think we would need something more robust.

packages/stream-model-handler/src/interfaces-utils.ts
```diff
@@ -687,3 +690,16 @@ export async function validateImplementedInterfaces(
     throw new AggregateError(errors, `Interfaces validation failed for model ${model.name}`)
   }
 }
+
+export function isValidImmutabilityImplementation(
+  expected: ModelDefinition,
+  implemented: ModelDefinition
+): boolean {
+  const expectedModelV2 = expected as ModelDefinitionV2
```
PaulLeCam, line 698: Sorry if my previous comment wasn't clear but I meant actually having a runtime check for the definition version, ex `if (expected.version === '2.0')`, not just replacing the type here.

packages/stream-model-handler/src/interfaces-utils.ts
```diff
@@ -687,3 +690,20 @@ export async function validateImplementedInterfaces(
...
+    const expectedModelV2 = expected as ModelDefinitionV2
+    const implementedModelV2 = implemented as ModelDefinitionV2
+    // No immutable fields
+    if (!expectedModelV2.immutableFields && !implementedModelV2.immutableFields) return true
+
+    const implementedSet = new Set(implementedModelV2.immutableFields ?? [])
+    return (expectedModelV2.immutableFields ?? []).every((item) => implementedSet.has(item))
+  } else {
+    return true
+  }
```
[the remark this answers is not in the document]
JulissaDantes: Ready! I copied the snippet as is.

PaulLeCam (approved):
Thanks for the changes!

PR #3157, opened 2024-02-09
feat: automatically register interest in the metamodel when using recon

stephhuynh18:
Moving the initialization of recon into the repository as it is initialized before the dispatcher, and will be used in the repository when handling the feed.


packages/core/src/recon.ts
```diff
@@ -47,10 +49,39 @@ export class ReconApi implements IReconApi {
...
+      return
+    }
+
     this.#url = await this.#config.url
+    await this.registerInterest(Model.MODEL)
+  }
+
+  async registerInterest(model: StreamID): Promise<void> {
+    if (!this.enabled) {
+      this.#logger.imp(`Recon: disabled, not registering interest in model ${model.toString()}`)
```
stbrody, line 62: maybe throw an error?

packages/core/src/recon.ts
```diff
@@ -47,10 +49,39 @@ export class ReconApi implements IReconApi {
...
+      return
+    }
+
+    try {
+      await this.#sendRequest(this.#url + `/ceramic/subscribe/model/${model.toString()}`, {
+        method: 'GET',
```
stbrody, line 68: this feels like it should be a POST.  @dav1do do you know if there's a plan to change the subscribe endpoint to use POST instead of GET?

stbrody (commented):
leaving a couple small comments on this before I head to vacation.  None of this is blocking, feel free to proceed without a final review from me

packages/core/src/recon.ts
dav1do: This route is actually deprecated in favor of interests. It's a `POST` to `/interests/{sort_key}/{sort_value}` with optional `controller` and `streamId` query parameters. 

It returns a 204 and then `/feed/events` is used to get the keys.

packages/core/src/recon.ts
```diff
@@ -47,13 +49,41 @@ export class ReconApi implements IReconApi {
...
+    if (!this.enabled) {
+      this.#logger.imp(`Recon: disabled, not putting event ${event.id}`)
+      return
+    }
+
     const body = {
-      id: event.id.toString(),
-      data: event.data.toString(),
+      eventId: event.id.toString(),
+      eventData: event.data.toString(),
```
nathanielc, line 86: I think the PR merged to change this back to `id` and `data`. @dav1do Can you confirm?

nathanielc (approved):
Looks great with one check on the current state of the Recon API.

packages/core/src/recon.ts
stephhuynh18: dangit!!!!! i gotta flip that back

dav1do: They are currently both supported.. but I was planning to delete the old version once I sort out the e2e tests: https://github.com/ceramicnetwork/rust-ceramic/pull/272

dav1do (approved):
Looks good to me. It'd be nice to change to the `id/data` body but it should work as is.

PR #3158, opened 2024-02-12
fix: Block ProcessingLoop when we've hit the concurrency limit of running tasks.

stbrody:


packages/core/src/anchor/processing-loop.ts
```diff
@@ -123,19 +123,26 @@ export class ProcessingLoop<T> {
...
-            })
-            .catch((e) => {
-              this.#logger.err(`Error in ProcessingLoop: ${e}`)
-              this.#abortController.abort('ERROR')
-            })
-            .finally(() => {
-              this.#runningTasks.delete(curTaskId)
-            })
-          this.#runningTasks.set(taskId, task)
+          await this.#semaphore.acquire().then((semRelease) => {
```
dbcfd, line 128: I'm not sure how this behavior is different than the previous behavior. `use` handles acquiring and releasing the semaphore, when given a promise, which is what we're doing.

dbcfd (changes requested):

packages/core/src/anchor/processing-loop.ts
stbrody: the issue with the old code is that the semaphore is protecting the execution of the tasks, but not the creation of them, so you can create an infinite number of tasks and add them to the #runningTasks set.  Only a fixed number will be able to execute at a time, but the job that iterates over the store spawning tasks just keeps running and running without ever blocking and we can wind up with 100s of thousands of tasks queued up 

the new version actually has the main event loop wait to acquire the semaphore, create a task to do the actual work, and store that task in the set of currently running running tasks.  Plus the semaphore doesn't release until the task completes.  So now if the processing loop hits a point where the max number of tasks are running, it will actually block before being able to make a new task

dbcfd (approved):

ukstv (commented):
Convoluted but does the trick.

ukstv (commented):
Approve.

PR #3159, opened 2024-02-13
Add test that reproduces processing the same entry multiple times at once

stbrody:
Repro for https://linear.app/3boxlabs/issue/CORE-156/anchorprocessingloop-can-process-the-same-stream-concurrently


ukstv:
FYI, I believe the test (in a slightly different _form_) passes on https://github.com/ceramicnetwork/js-ceramic/pull/3163/files

stbrody:
Superceded by https://github.com/ceramicnetwork/js-ceramic/pull/3163#pullrequestreview-1891642115

PR #3161, opened 2024-02-14
feat: create recon feed and consume events

stephhuynh18:
Manually tested 
Will be turning on tests in a future pr
Sorry about the rxjs, it was used by Zach originally, and using it felt safer than trying to create my own tools. Also it mirrors message bus so it keeps things consistent. 
I am handling events synchronously. I can parallelize them but we can discuss


nathanielc (approved):
Looks great with one small nit below

PR #3163, opened 2024-02-15
fix: AnchorProcessingLoop can process the same stream concurrently

ukstv:


packages/core/src/anchor/processing-loop.ts
```diff
@@ -93,62 +96,60 @@ export class ProcessingLoop<T> {
...
-            this.#runningTasks.set(taskId, task)
-          })
-
-          // Mod by MAX_SAFE_INTEGER just in case taskId gets really large and needs to wrap around
-          taskId = (taskId + 1) % Number.MAX_SAFE_INTEGER
+          const release = await this.#semaphore.acquire()
+          this.#logger.verbose(`ProcessingLoop: Fetching next event from source`)
+          const next = await Promise.race([this.source.next(), doneOnAbortSignal])
+          isDone = next.done
+          if (isDone) return
```
dbcfd, line 127: Need semaphore released here.

packages/core/src/anchor/processing-loop.ts
```diff
@@ -93,62 +96,60 @@ export class ProcessingLoop<T> {
...
+          const value = next.value
+          if (value === undefined) {
+            this.#logger.verbose(`No value received in ProcessingLoop, skipping this iteration`)
+            return
```
dbcfd, line 131: Release semaphore.

packages/core/src/anchor/processing-loop.ts
```diff
@@ -93,62 +96,60 @@ export class ProcessingLoop<T> {
...
-              })
-
-            // It's important that the semaphore guards adding the task to the running task set,
-            // so that we won't get more tasks created than the concurrency limit.
-            this.#runningTasks.set(taskId, task)
-          })
-
-          // Mod by MAX_SAFE_INTEGER just in case taskId gets really large and needs to wrap around
-          taskId = (taskId + 1) % Number.MAX_SAFE_INTEGER
+          const release = await this.#semaphore.acquire()
```
dbcfd, line 123: Can we move all the logic into a task, so we can release the semaphore on the finally of that task, rather than after multiple returns?

dbcfd (changes requested):

packages/core/src/anchor/processing-loop.ts
ukstv: I think I figured out how to make it. Please, have a look.

dbcfd (approved):

packages/core/src/anchor/processing-loop.ts
```diff
@@ -70,90 +78,77 @@ export class ProcessingLoop<T> {
...
+            await Promise.race([this.handleValue(value), waitForAbortSignal])
+          } catch (e) {
+            this.#logger.err(`Error in ProcessingLoop: ${e}`)
+            this.#abortController.abort('ERROR')
+          } finally {
+            taskIds.delete(uniqueTaskId)
+          }
+        })
+        if (this.#taskQueue.size >= this.#taskQueue.concurrency) {
+          await this.#taskQueue.onIdle() // Wait till we process the batch
```
stbrody, line 148: doesn't this wait until all the tasks in the queue fully completes all its running tasks?  I think we just want to wait until there's room in the queue again to take on more tasks right?

stbrody: Okay, put together a PR to keep the perf benefit of keeping the set of running tasks as full as possible up to the concurrency limit, while also keeping the cleanups and bug fixes from this PR.  Thankfully the code is much cleaner now as a result of @ukstv's work here, and the TaskQueue makes this a very easy improvement to make.  PTAL: https://github.com/ceramicnetwork/js-ceramic/pull/3167

PR #3164, opened 2024-02-15
feat: warn-cacao-near-exp

AaronGoldman:
log a warning when a cacao is verified with an expiration within 12h of the covering time event.

## Description

## How Has This Been Tested?

- [ ] Test B

## PR checklist

## References:

### DIFF
--- a/packages/common/src/utils/stream-utils.ts
+++ b/packages/common/src/utils/stream-utils.ts
@@ -1,6 +1,7 @@
 import cloneDeep from 'lodash.clonedeep'
 import * as uint8arrays from 'uint8arrays'
 import { toCID } from './cid-utils.js'
+import type { DiagnosticsLogger } from '@ceramicnetwork/common'
 import * as dagCBOR from '@ipld/dag-cbor'
 import {
   AnchorCommit,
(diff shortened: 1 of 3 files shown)


packages/common/src/utils/stream-utils.ts
```diff
@@ -366,6 +367,12 @@ export class StreamUtils {
             logEntry.expirationTime
           }. Loading the stream with 'sync: SyncOptions.ALWAYS_SYNC' will restore the stream to a usable state, by discarding the invalid commits (this means losing the data from those invalid writes!)`
         )
+      } else if (logEntry.expirationTime && logEntry.expirationTime - timestamp < 12 * 60 * 60) {
+        logger.warn(
```
stbrody, line 371: since `timestamp` can come from the anchor commit time OR from "now" if the stream isn't actually anchored yet, this log message is a bit misleading, since the message implies the stream is already anchored in all cases

stbrody, line 371: thinking about this more, I'm actually not sure how useful this log message is, and I'm afraid it might be quite noisy.  This is because there's nothing that prevents us from doing a write with a CACAO right before it's about to expire.  A standard CACAO session is valid for a week by default, but if the user logs in once, then logs in again 6 days, 23 hours, and 59 minutes later, the app will still let them re-use their old session without signing in again.  So the mere fact that the CACAO is close to expiring doesn't actually tell us that the stream took longer than expected to be anchored.

I think this log message really needs to be based on the timestamp of when the write was performed, which we only have in the ComposeDB index database.

AaronGoldman: Once we have the 24 hour grace period logging the fact that it is expired tells us basicly the same thing but it is more clear that a client is not refreshing it CACAO sessions aggressively enugh.

AaronGoldman: `anchorTime=${logEntry.timestamp}` will be null if there is no anchor but I see what you mean about the wording.

PR #3165, opened 2024-02-15
feat: turn on cross-node sync test

stephhuynh18:
I moved `testIfV3` to common test utils and used it in this test suite because one of the tests in this test suite ACTUALLY cannot work in V'. 

We cannot skip publishing in V'


packages/stream-tests/src/__tests__/cross-node-indexing.test.ts
```diff
@@ -93,19 +97,27 @@ describeIfV3ShouldPass.each(envs)(
     beforeAll(async () => {
       await pgSetup()
 
-      ipfs1 = await createIPFS()
-      ipfs2 = await createIPFS()
+      ipfs1 = await createIPFS(undefined, undefined, {
```
dav1do, line 100: just an observation: seems like having multiple constructors would be nice some day (passing undefined like this just hurts lol)

dav1do (approved):

PR #3166, opened 2024-02-16
fix: unique metadata is not properly typed as a Uint8Array

dbcfd:


dbcfd:
[This test](https://github.com/3box/ceramic-http-client-rs/blob/main/src/lib.rs#L647) will fail without this change.

ukstv (approved):

stbrody:
@dbcfd I'm confused why this PR was opened when we already had another PR with this same change in flight?  https://github.com/ceramicnetwork/js-ceramic/pull/3141

I'd still really like to understand how and why we are winding up with data that isn't in the format we are expecting.  I'm worried that this approach is papering over the symptoms rather than addressing the real source of the bug - which is whatever is generating data that doesn't have `unique` typed properly in the first place.

stbrody:
> [This test](https://github.com/3box/ceramic-http-client-rs/blob/main/src/lib.rs?rgh-link-date=2024-02-16T00%3A29%3A36Z#L647) will fail without this change.

is it possible then that the Ceramic rust client isn't constructing/serializing the `unique` field in the same way as the javascript client does?  Seems like that would be a bug in the rust client then.

PR #3168, opened 2024-02-21
fix(indexing): don't index empty deterministic streams

PaulLeCam:
Simply looking up a deterministic stream causes it to be indexed, which causes the indexer queries to return results with empty contents that don't match the expected behavior.
This PR updates the logic to ignore streams with no content when indexing.

### DIFF
--- a/packages/stream-tests/src/__tests__/deterministic-indexing.test.ts
+++ b/packages/stream-tests/src/__tests__/deterministic-indexing.test.ts
@@ -1,0 +1,110 @@
+import { jest } from '@jest/globals'
+import type { BaseQuery, IpfsApi } from '@ceramicnetwork/common'
+import { createIPFS } from '@ceramicnetwork/ipfs-daemon'
+import {
+  ModelInstanceDocument,
+  type ModelInstanceDocumentMetadataArgs,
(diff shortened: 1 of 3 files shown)


packages/stream-tests/src/__tests__/deterministic-indexing.test.ts
```diff
@@ -0,0 +1,110 @@
+import { jest } from '@jest/globals'
```
stbrody, line 1: there is already a `deterministic-model-instance-document.test.ts` file in the `stream-tests` package.  Should we add these test cases into that file instead of making a new test file?

stbrody (approved):

packages/stream-tests/src/__tests__/deterministic-indexing.test.ts
PaulLeCam: The added tests are more specific to the behavior or deterministic MIDs with indexing so I found it more convenient to have a dedicated test suite, but I can move the tests to the other suite if you prefer?

stbrody: I don't feel strongly.  If you think it makes sense to keep them separate that's fine

PR #3172, opened 2024-02-23
fix(cli,http-client,stream-model-instance): encode genesis commit in multiquery

PaulLeCam:
Encodes the genesis commit in the HTTP client if present in the multiquery request and decodes it in the HTTP handler.

As we discussed before @stbrody I also made the `ModelInstanceDocument.makeGenesis()` public as I needed it for the test, and I'll also be able to use it in the ComposeDB loader.


stbrody (approved):
lgtm

PR #3175, opened 2024-02-28
chore: consolidate schema validation implementations

stbrody:
No functional changes, just code de-duplication

Only Model and ModelInstanceDocument have their schema validation consolidated.  TileDocument is left alone because it relies on a different version of Ajv and the jsonSchema spec.


stbrody:
I no longer think this is actually a good idea to do right now.  TileDocument must continue to use a separate implementation of schema validation b/c it depends on a different version of the jsonSchema spec, and Model uses Ajv in a very different way (validating that the schema itself is a valid schema, rather than validating data *against* a schema) and likely isn't worth trying to parallelize.  So since only ModelInstanceDocument needs to move to the worker-thread based schema validation, we should keep its implementation of schema validation separate

PR #3176, opened 2024-02-29
feat: Use worker threads for jsonSchema validation of ModelInstanceDocuments

stbrody:


dav1do (approved):
LGTM. I ran a benchmark on my mac and saw create/sync in the new MIDs benchmark of 247.23/129.53 with this branch vs 220.48/144.25 with dev (only over a one minute run to make sure things didn't change substantially).

JulissaDantes:
Why havent this been merged?

stbrody:
> Why havent this been merged?

We said that we want to see that this will improve performance before merging it.  So far our testing has shown this PR having no impact on performance/throughput whatsoever, which is surprising, so we want to understand what's happening first.

Personally I'd be fine with merging and then iterating in develop, but I don't feel strongly and I believe @nathanielc prefers to wait.

PR #3177, opened 2024-03-05
feat: Update ajv-threads package to version with worker pool

stbrody:
This will make js-ceramic actually use multiple worker threads for schema validation.  When this merges is when we should be able to observe real improvements to js-ceramic throughput.

### DIFF
--- a/packages/core/src/ceramic.ts
+++ b/packages/core/src/ceramic.ts
@@ -247,7 +247,7 @@ export class Ceramic implements StreamReaderWriter, StreamStateLoader {
     const numCores = os.cpus().length
     // Use number of threads equal to half the available cores for schema validation. Leave the
     // other half for signature validation.
-    this._schemaValidator = new SchemaValidation(Math.floor(numCores / 2))
+    this._schemaValidator = new SchemaValidation(Math.max(1, Math.ceil(numCores / 2)))
     this._streamHandlers = HandlersMap.makeWithDefaultHandlers(this._logger, this._schemaValidator)
 
     // This initialization block below has to be redone.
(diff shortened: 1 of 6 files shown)


packages/core/src/ceramic.ts
```diff
@@ -244,7 +244,10 @@ export class Ceramic implements StreamReaderWriter, StreamStateLoader {
 
     this._ipfs = modules.ipfs
 
-    this._schemaValidator = new SchemaValidation()
+    const numCores = os.cpus().length
```
dav1do, line 247: I had to double check this was logical cores too.. so I think half seems reasonable since it should be cpu bound. I was thinking the full amount of cores might be fine, but I don't know if that would actually help at all or just starve things/context switch too much if both sections are going full bore.

packages/core/src/ceramic.ts
```diff
@@ -244,7 +244,10 @@ export class Ceramic implements StreamReaderWriter, StreamStateLoader {
...
+    // Use number of threads equal to half the available cores for schema validation. Leave the
+    // other half for signature validation.
+    this._schemaValidator = new SchemaValidation(Math.floor(numCores / 2))
```
dav1do, line 250: should this be at least 1? maybe `Math.max(1, Math.floor(numCores / 2)`?

dav1do: also, looks like we need to call `this._schemaValidator.init()` still

dav1do: Ha, I know that feeling 😅 Definitely did have to call `init` in a bunch of tests as well. LGTM.

dav1do (approved):

PR #3179, opened 2024-03-05
fix: remove parsing of resume token and add unit tests for recon

stephhuynh18:
feed limit is 1000, can make configurable one day

### DIFF
--- a/packages/core/src/__tests__/recon.test.ts
+++ b/packages/core/src/__tests__/recon.test.ts
@@ -3,7 +3,7 @@ import { EventID, StreamID } from '@ceramicnetwork/streamid'
 import { FetchRequest, LoggerProvider } from '@ceramicnetwork/common'
 import { jest } from '@jest/globals'
 import { type CAR } from 'cartonne'
-import { toArray, take, lastValueFrom } from 'rxjs'
+import { toArray, take, lastValueFrom, firstValueFrom, race, timer } from 'rxjs'
 
 const RECON_URL = 'http://example.com'
 const LOGGER = new LoggerProvider().getDiagnosticsLogger()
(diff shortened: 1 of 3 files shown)


packages/core/src/__tests__/recon.test.ts
```diff
@@ -0,0 +1,156 @@
...
+    })
+
+    test('should not init if already initialized', async () => {
+      // already initialized in beforeEach
+      await reconApi.init()
+
+      expect(mockSendRequest).not.toHaveBeenCalled()
+    })
+
+    test('should not start polling if feed is disabled', async () => {
```
dav1do, line 64: I don't quite understand this test.. I'd expect something like the `const received = await lastValueFrom(reconApi.pipe(take(3), toArray()))` call to check if it polled?

packages/core/src/recon.ts
```diff
@@ -96,7 +97,7 @@ export class ReconApi extends Observable<ReconEventFeedResponse> implements IRec
    * @param initialCursor
    * @returns
    */
-  async init(initialCursor = 0): Promise<void> {
+  async init(initialCursor = '0'): Promise<void> {
```
dav1do, line 100: I might want to make a note that we should parse '0' on the server side as default if we change the shape someday. I wasn't expecting the client to send anything until receiving a response but this works fine. 

Just thinking.. you can close :)

dav1do (approved):

packages/core/src/__tests__/recon.test.ts
stephhuynh18: it says should NOT start polling, so i don't expect to receive anything

But i should try to receive something and then check that I received nothing

PR #3181, opened 2024-03-12
fix: make convertCommitToSignedCommitContainer use offline mode

nathanielc:
The code paths needed for creating a new composedb model called the convertCommitToSignedCommitContainer function. However the function would always use online mode and fails in Recon mode. We should not need online mode as the commits will have just been stored locally. This changes the code to use offline mode when in recon mode.

## How Has This Been Tested?

Created a model using the composedb CLI, prior to this change it errors. With this change it no longer errors.

### DIFF
--- a/packages/common/src/utils/stream-utils.ts
+++ b/packages/common/src/utils/stream-utils.ts
@@ -280,7 +280,10 @@ export class StreamUtils {
     ipfs: IpfsApi
   ): Promise<CeramicCommit> {
     if (StreamUtils.isSignedCommit(commit)) {
-      const block = await ipfs.block.get(toCID((commit as DagJWS).link))
+      const block = await ipfs.block.get(toCID((commit as DagJWS).link), {
+        // @ts-ignore
+        offline: process.env.CERAMIC_RECON_MODE == 'true',
+      })
       return {
         jws: commit as DagJWS,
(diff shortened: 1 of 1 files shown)


smrz2001 (approved):
🥳

stephhuynh18 (approved):

packages/common/src/utils/stream-utils.ts
```diff
@@ -280,7 +280,10 @@ export class StreamUtils {
     ipfs: IpfsApi
   ): Promise<CeramicCommit> {
     if (StreamUtils.isSignedCommit(commit)) {
-      const block = await ipfs.block.get(toCID((commit as DagJWS).link))
+      const block = await ipfs.block.get(toCID((commit as DagJWS).link), {
```
stbrody, line 283: Can we change this to go through the Dispatcher that has this logic already, instead of using the raw `ipfs` handle?

nathanielc: I don't think that is easily possible. The dispatcher lives in core and this code is inside common. Currently core depends on common so this would create a cyclic dependency. Seems like a good refactor to move this logic to the dispatcher but for now its seems out of scope for this simple change.

stbrody (approved):

PR #3183, opened 2024-03-13
fix: fix log in dispatcher code

samika98:
## Description

Added more verbose logging to give more detail about what's going on.
These are the current logs : [2024-03-13T08:05:53.024Z] ERROR: Error while storing commit to IPFS: HTTPError: Not Found.

### DIFF
--- a/packages/core/src/dispatcher.ts
+++ b/packages/core/src/dispatcher.ts
@@ -367,9 +367,10 @@ export class Dispatcher {
       Metrics.count(COMMITS_STORED, 1)
       return cid
     } catch (e) {
-      this._logger.err(`Error while storing commit to IPFS: ${e}`)
+      const errorMessage = `Error in repository.ts.storeInitEvent: Error - ${e} while storing commit to ipfs: ${e.message} | recon.mode: ${this.recon.enabled}`
+      this._logger.err(errorMessage)
       Metrics.count(ERROR_STORING_COMMIT, 1)
-      throw e
+      throw new Error(errorMessage)
     }
   }
 


packages/core/src/dispatcher.ts
```diff
@@ -367,9 +367,10 @@ export class Dispatcher {
       Metrics.count(COMMITS_STORED, 1)
       return cid
     } catch (e) {
-      this._logger.err(`Error while storing commit to IPFS: ${e}`)
+      const errorMessage = `Error in repository.ts.storeInitEvent: Error - ${e} while storing commit to ipfs: ${e.message} | recon.mode: ${this.recon.enabled}`
+      this._logger.err(errorMessage)
```
smrz2001, line 371: Can we capture the backtrace here so we can see which code path the dispatcher was invoked from?

packages/core/src/dispatcher.ts
```diff
@@ -367,9 +367,10 @@ export class Dispatcher {
       Metrics.count(COMMITS_STORED, 1)
       return cid
     } catch (e) {
-      this._logger.err(`Error while storing commit to IPFS: ${e}`)
+      const errorMessage = `Error in repository.ts.storeInitEvent: Error - ${e} while storing commit to ipfs: ${e.message} | recon.mode: ${this.recon.enabled}`
```
stbrody, line 370: I think this meant to say we are in dispatcher.ts, not repository.ts.

But to be honest I don't really love putting the file name into the error message.  If the message is descriptive and unique enough, it should be easy to grep for to find what file it is in.  Putting the file name into the message feels brittle in case the code is ever refactored, and also kind like leaking information that isn't relevant to our users.  I'd be okay with a stack trace though as that at least will update if the message moves and feels like a more standard thing devs know how to use.

stbrody: Also this doesn't really feel like the right place to log the recon mode either.  I think global configuration like that should be logged once at startup, rather than every time there is an error.

packages/core/src/dispatcher.ts
stbrody: If you log an `Error` instance directly it will log the stack trace automatically I believe

PR #3184, opened 2024-03-14
feat: add log for anchor-util merkel tree depth

3benbox:
Record the merkel tree depth of an anchor.

### DIFF
--- a/packages/anchor-utils/src/merkle/merkle-tree-factory.ts
+++ b/packages/anchor-utils/src/merkle/merkle-tree-factory.ts
@@ -71,7 +71,22 @@ export class MerkleTreeFactory<TData, TLeaf extends TData, TMetadata>
     const metadata = this.metadataFn ? await this.metadataFn.generateMetadata(nodes) : null
 
     const root = await this.buildLevel(nodes, 0, metadata)
+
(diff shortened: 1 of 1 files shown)


ukstv (approved):
Approved under condition that console.log gets removed in a few days.

packages/anchor-utils/src/merkle/merkle-tree-factory.ts
```diff
@@ -71,9 +71,24 @@ export class MerkleTreeFactory<TData, TLeaf extends TData, TMetadata>
     const metadata = this.metadataFn ? await this.metadataFn.generateMetadata(nodes) : null
 
     const root = await this.buildLevel(nodes, 0, metadata)
+
+    const depth = this.getTreeDepth(root)
+    console.log(`Merkle tree generated with depth: ${depth}`)
```
stbrody, line 76: Can we pass in a DiagnosticsLogger instance and use that instead of raw console.log?

ukstv: 1. That's temporary to verify if a config change actually changes the tree depth.
2. You'd need to change CAS code back and forth to pass the logger.

Worth it?

PR #3185, opened 2024-03-15
chore: enable anchoring on recon mode

JulissaDantes:
## Description
- Unskip all skipped anchoring tests in js-ceramic
- Re-enable the anchoring code paths disabled with recon mode


nathanielc (approved):
LGTM!

stbrody:
I'm going to hold off on reviewing this until CI is green, looks like there are still some remaining issues right now

packages/core/src/anchor/__tests__/anchor-processing-loop.test.ts
```diff
@@ -152,7 +152,7 @@ test('Dont process the same entry multiple times concurrently', async () => {
   async function* overEntries(): AsyncGenerator<number> {
     do {
       // Need this sleep or else the Node runtime might never switch back to the "process" function.
-      await CommonTestUtils.delay(1)
+      await CommonTestUtils.delay(5)
```
JulissaDantes, line 155: @ukstv This is how I solved it

stbrody: what's the issue you were seeing?  This change really shouldn't be necessary.  This sleep isn't actually making the test wait for anything to happen, it's just so that the Node event loop has an interrupt point so other jobs can get scheduled, otherwise the generator can just emit events forever without the task to process those events ever running.  But increasing the duration of the sleep should have no affect on the rest of the system or the test, unless I really misunderstood something when I wrote this test.

stbrody: I'm assuming this test failure was the issue: https://app.circleci.com/pipelines/github/ceramicnetwork/js-ceramic/13094/workflows/f7e8914c-f718-40e3-bfd6-4ee07174026c/jobs/18766

Put together a PR that I believe should fix the race condition in the test: https://github.com/ceramicnetwork/js-ceramic/pull/3192

JulissaDantes:
I did not have to make changes to the code other than enabling the anchor when using recon, but I did have to adjust the test setup. A common modification was to add some delay, which is why you will see it in some tests. Ready to be reviewed by @stbrody and @nathanielc.

packages/core/src/anchor/__tests__/anchor-processing-loop.test.ts
stbrody: Can this change be reverted now that https://github.com/ceramicnetwork/js-ceramic/pull/3192 is in?

packages/stream-tests/src/__tests__/model-instance-document.test.ts
```diff
@@ -514,16 +504,32 @@ describeIfV3ShouldPass('ModelInstanceDocument API multi-node tests', () => {
   let midMetadata: ModelInstanceDocumentMetadataArgs
 
   beforeAll(async () => {
-    ipfs0 = await createIPFS()
-    ipfs1 = await createIPFS()
+    ipfs0 = await createIPFS({
+      rust: {
+        type: 'binary',
```
stbrody, line 509: Can we move this logic into `createIPFS` as the default behavior so we don't need to update the call sites like this?

packages/stream-tests/src/__tests__/model-instance-document.test.ts
```diff
@@ -514,16 +504,32 @@ describeIfV3ShouldPass('ModelInstanceDocument API multi-node tests', () => {
...
+        network: Networks.INMEMORY,
+      },
+    })
     await swarmConnect(ipfs0, ipfs1)
 
-    ceramic0 = await createCeramic(ipfs0)
-    ceramic1 = await createCeramic(ipfs1)
+    ceramic0 = await createCeramic(ipfs0, { networkName: Networks.INMEMORY })
```
stbrody, line 521: ditto, can we make this the default behavior of `createCeramic`?

packages/stream-tests/src/__tests__/model-instance-document.test.ts
```diff
@@ -549,8 +558,12 @@ describeIfV3ShouldPass('ModelInstanceDocument API multi-node tests', () => {
 
   test('load updated doc', async () => {
     const doc = await ModelInstanceDocument.create(ceramic0, CONTENT0, midMetadata)
-    await doc.replace(CONTENT1)
 
+    if (process.env.CERAMIC_RECON_MODE)
+      await TestUtils.waitForEvent(ceramic1.repository.recon, doc.tip)
+
+    await doc.replace(CONTENT1)
+    await TestUtils.delay(2000)
```
stbrody, line 566: can we use waitForEvent for this too instead of a fixed delay?

packages/stream-tests/src/__tests__/model-instance-document.test.ts
```diff
@@ -566,9 +579,12 @@ describeIfV3ShouldPass('ModelInstanceDocument API multi-node tests', () => {
   test('load updated and anchored doc', async () => {
     const doc = await ModelInstanceDocument.create(ceramic0, CONTENT0, midMetadata)
 
+    if (process.env.CERAMIC_RECON_MODE)
+      await TestUtils.waitForEvent(ceramic1.repository.recon, doc.tip)
+
     await doc.replace(CONTENT1)
     await CoreUtils.anchorUpdate(ceramic0, doc)
-
+    await TestUtils.delay(2000)
```
stbrody, line 587: ditto, can we use waitForEvent?

packages/stream-tests/src/__tests__/model.test.ts
```diff
@@ -825,12 +790,19 @@ describeIfV3ShouldPass('Model API multi-node tests', () => {
...
     const model = await Model.create(ceramic0, MODEL_DEFINITION)
+
+    if (process.env.CERAMIC_RECON_MODE)
+      await TestUtils.waitForEvent(ceramic1.repository.recon, model.tip, 70000)
+
+    await ceramic0.admin.startIndexingModelData([{ streamID: model.id }])
+    await ceramic1.admin.startIndexingModelData([{ streamID: model.id }])
     await CoreUtils.anchorUpdate(ceramic0, model)
 
+    await TestUtils.delay(2000)
```
stbrody, line 801: waitForEvent?

stbrody (commented):
looking good, a few small comments to clean things up a bit

packages/stream-tests/src/__tests__/model-instance-document.test.ts
JulissaDantes: Ready

packages/stream-tests/src/__tests__/model-instance-document.test.ts
JulissaDantes: Ready

packages/core/src/anchor/__tests__/anchor-processing-loop.test.ts
stbrody: ping

stbrody (approved):
lgtm mod undoing the change to increase the delay in anchor-processing-loop.test.ts

PR #3186, opened 2024-03-15
feat: dont log depth in anchor-utils, just return it

3benbox:
Instead of logging the depth in the library, return it so the caller can log or make a metric.


ukstv (approved):

ukstv (approved):
It will break the callers. IMO, we do not use MerkleTree class except in tests and in the factory, so not a biggie.

3benbox:
no longer needed

PR #3187, opened 2024-03-15
feat: add sepolia to allow chain for dev-unstable

3benbox:
Add Sepolia chain id to the allowed list for dev-unstable.

### DIFF
--- a/packages/anchor-utils/src/constants.ts
+++ b/packages/anchor-utils/src/constants.ts
@@ -6,15 +6,6 @@ export type SupportedNetwork =
   | 'eip155:1337' // ganache
   | 'eip155:11155111' // sepolia
 
-export const BLOCK_THRESHHOLDS: Record<SupportedNetwork, number> = {
-  'eip155:1': 1000000000,
-  'eip155:3': 1000000000,
-  'eip155:5': 1000000000,
-  'eip155:100': 1000000000,
-  'eip155:1337': 1000000,
-  'eip155:11155111': 1000000000,
-}
-
 export const ANCHOR_CONTRACT_ADDRESS = '0x231055a0852d67c7107ad0d0dfeab60278fe6adc'
 
 export const CONTRACT_TX_TYPE = 'f(bytes32)'
--- a/packages/anchor-utils/src/constants.ts
+++ b/packages/anchor-utils/src/constants.ts
@@ -4,6 +4,7 @@ export type SupportedNetwork =
   | 'eip155:5' // goerli
   | 'eip155:100' // gnosis
   | 'eip155:1337' // ganache
+  | 'eip155:11155111' // sepolia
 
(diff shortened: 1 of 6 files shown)


packages/core/src/initialization/anchoring.ts
```diff
@@ -18,7 +18,7 @@ export const DEFAULT_ANCHOR_SERVICE_URLS = {
 const SUPPORTED_CHAINS_BY_NETWORK = {
   [Networks.MAINNET]: ['eip155:1'], // Ethereum mainnet
   [Networks.TESTNET_CLAY]: ['eip155:3', 'eip155:4', 'eip155:100'], // Ethereum Ropsten, Rinkeby, Gnosis Chain
-  [Networks.DEV_UNSTABLE]: ['eip155:3', 'eip155:4', 'eip155:5'], // Ethereum Ropsten, Rinkeby, Goerli
+  [Networks.DEV_UNSTABLE]: ['eip155:3', 'eip155:4', 'eip155:5', 'eip155:11155111'], // Ethereum Ropsten, Rinkeby, Goerli, Sepolia
```
stbrody, line 21: lol what an absurd chainid

stbrody (changes requested):
I don't think this change is sufficient.  There are several other places in our codebase that reference chainids.  Just grep for "eip155:5" to see all the places we use the Goerli chainid today.  I assume most if not all of those places will need to be updated

packages/anchor-utils/src/constants.ts
```diff
@@ -4,13 +4,15 @@ export type SupportedNetwork =
   | 'eip155:5' // goerli
   | 'eip155:100' // gnosis
   | 'eip155:1337' // ganache
+  | 'eip155:11155111' // sepolia
 
 export const BLOCK_THRESHHOLDS: Record<SupportedNetwork, number> = {
```
stbrody, line 9: looks like this constant isn't actually used anywhere, let's just delete it.

stbrody: note there's a similar constant in ethereum-anchor-validator.ts that *is* used however, and still need to be updated.

stbrody: ping

3benbox: The type is used on line 9 above, no?
Can that be removed too?

stbrody: sorry, this comment is *on* line 9, so I'm not sure what you're referring to.  I'm saying the `BLOCK_THRESHHOLDS` constant in this file is defined but never used.

Also there's a `BLOCK_THRESHHOLDS` constant in `ethereum-anchor-validator.ts` that *is* used and I think still needs to be updated.

stbrody (approved):
lgtm

PR #3188, opened 2024-03-19
feat: update how we check immutability

JulissaDantes:
## Description
We used the payload data to determine if an immutable field was being altered, but this caused 2 problems:
- Not optimal developer experience when running a mutation including an immutable field but not mutating it.
- Databases built on top of js-ceramic using an `upsert` command will always throw if creating a new record for a model with an immutable field. e.g. ComposeDB

### DIFF
--- a/packages/stream-model-instance-handler/src/__tests__/model-instance-document-handler.test.ts
+++ b/packages/stream-model-instance-handler/src/__tests__/model-instance-document-handler.test.ts
@@ -850,7 +850,7 @@ describe('ModelInstanceDocumentHandler', () => {
     expect(state).toMatchSnapshot()
   })
 
-  it('MIDs with SET account relation validate signed commit fields', async () => {
+  it.only('MIDs with SET account relation validate signed commit fields', async () => {
     const genesisCommit = (await ModelInstanceDocument.makeGenesis(
       context.signer,
       null,
@@ -890,12 +890,12 @@ describe('ModelInstanceDocumentHandler', () => {
     await expect(handler.applyCommit(signedCommitDataFail, context, state)).rejects.toThrow(
       'Unique content fields value does not match metadata. If you are trying to change the value of these fields, this is causing this error: these fields values are not mutable.'
     )
-
+    console.log('voy')
     const signedCommitOK = (await ModelInstanceDocument.makeUpdateCommit(
       context.signer,
       doc.commitId,
       doc.content,
-      { one: 'foo', two: 'bar', myData: 2 }
+      { one: 'foo', two: 'bar', three: 'three', myData: 2 }
     )) as SignedCommitContainer
     await ipfs.dag.put(signedCommitOK, FAKE_CID_3)
 
@@ -917,7 +917,7 @@ describe('ModelInstanceDocumentHandler', () => {
       context.signer,
       null,
       { ...DETERMINISTIC_METADATA, model: FAKE_MODEL_SET_ID },
-      ['a', 'b', 'c']
+      ['a', 'b']
     )) as GenesisCommit
     await ipfs.dag.put(genesisCommit, FAKE_CID_1)
 
--- a/packages/stream-model-instance-handler/src/__tests__/model-instance-document-handler.test.ts
+++ b/packages/stream-model-instance-handler/src/__tests__/model-instance-document-handler.test.ts
@@ -890,7 +890,7 @@ describe('ModelInstanceDocumentHandler', () => {
     await expect(handler.applyCommit(signedCommitDataFail, context, state)).rejects.toThrow(
       'Unique content fields value does not match metadata. If you are trying to change the value of these fields, this is causing this error: these fields values are not mutable.'
     )
-    console.log('voy')
+
     const signedCommitOK = (await ModelInstanceDocument.makeUpdateCommit(
       context.signer,
       doc.commitId,
--- a/packages/stream-model-instance-handler/src/__tests__/model-instance-document-handler.test.ts
+++ b/packages/stream-model-instance-handler/src/__tests__/model-instance-document-handler.test.ts
@@ -189,6 +189,7 @@ const MODEL_DEFINITION_SET: ModelDefinition = {
     properties: {
       one: { type: 'string', minLength: 2 },
       two: { type: 'string', minLength: 2 },
+      three: { type: 'string', minLength: 2 },
       myData: {
         type: 'integer',
         maximum: 100,
@@ -197,6 +198,7 @@ const MODEL_DEFINITION_SET: ModelDefinition = {
     },
     required: ['myData'],
   },
+  immutableFields: ['three'],
 }
 
 const MODEL_DEFINITION_BLOB: ModelDefinition = {
@@ -915,7 +917,7 @@ describe('ModelInstanceDocumentHandler', () => {
       context.signer,
       null,
       { ...DETERMINISTIC_METADATA, model: FAKE_MODEL_SET_ID },
-      ['a', 'b']
+      ['a', 'b', 'c']
     )) as GenesisCommit
     await ipfs.dag.put(genesisCommit, FAKE_CID_1)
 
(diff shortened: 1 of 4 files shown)


stbrody (changes requested):
Is there a test that reproduces the issue you are trying to solve here, using the standard `ModelInstanceDocument.replace` method?  That method already constructs the json-patch to only include fields that actually changed, so I'm not sure how this could ever be a problem in practice.

JulissaDantes:
> Is there a test that reproduces the issue you are trying to solve here, using the standard `ModelInstanceDocument.replace` method? That method already constructs the json-patch to only include fields that actually changed, so I'm not sure how this could ever be a problem in practice.

Yes regarding the test, I changed the model for the set test to include an immutable field. The jsonpatch construction is not the problem, its how we validate immutability on locked fields, thats why we now will compare the previous value with the value post jsonpatch.

Also this is a problem we are currently having, you can check the linear task for a step by step reproduction, those steps will trigger the `ModelInstanceDocument.replace` method.

packages/stream-model-instance-handler/src/__tests__/model-instance-document-handler.test.ts
```diff
@@ -888,12 +890,12 @@ describe('ModelInstanceDocumentHandler', () => {
     await expect(handler.applyCommit(signedCommitDataFail, context, state)).rejects.toThrow(
       'Unique content fields value does not match metadata. If you are trying to change the value of these fields, this is causing this error: these fields values are not mutable.'
     )
-
+    console.log('voy')
```
PaulLeCam, line 893: Can you remove this log please?

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
```diff
@@ -194,9 +194,9 @@ export class ModelInstanceDocumentHandler implements StreamHandler<ModelInstance
     }
 
     const oldContent = state.content ?? {}
-    const newContent = jsonpatch.applyPatch(oldContent, payload.data).newDocument
+    const newContent = jsonpatch.applyPatch(oldContent, payload.data, undefined, false).newDocument
     const modelStream = await context.loadStream<Model>(metadata.model)
-    await this._validateContent(context, modelStream, newContent, false, payload)
+    await this._validateContent(context, modelStream, newContent, false, payload, oldContent)
```
PaulLeCam, line 199: Couldn't `state.content` be passed directly here, having the logic handle if it's `null`?

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
```diff
@@ -269,7 +270,7 @@ export class ModelInstanceDocumentHandler implements StreamHandler<ModelInstance
     // Now validate the relations
     await this._validateRelationsContent(ceramic, model, content)
     if (!genesis && payload) {
```
PaulLeCam, line 272: I think the validation could be completely bypassed when there is no existing content, rather than needing additional logic there
```suggestion
    if (!genesis && payload && oldContent != null) {
```

PaulLeCam (commented):
I think the logic works but could be simplified, seems to me the immutable fields check can simply be bypassed when the current state content is `null`, no?

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
JulissaDantes: yes it could, but the oldContent line already handles if state.content its null, what would be the benefit of doing that here?

PaulLeCam: I just think the `null` value should be provided directly, no need to provide an empty object as fallback, so the following logic can do the null check.

stbrody (commented):
I think things are correct now, but this approach introduces a lot of extra clones of the content which is not ideal.  Could we go back to the original approach of checking the jsonPatch for if it touches any of the immutable fields, but just add an explicit exception for if the stream has the SET account relation and the commit being applied is the very first data event?

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
stbrody: Rather than relying on the content being `null`, which feels like an indirect check of what we want and maybe not entirely future-proof (what if stream content could be explicitly set to null in the future?), could we instead check that the stream has a set account relation and that this is the first data commit?  That feels like the actual rule that should allow content to be set for the first time.

JulissaDantes: you mean checking the state.log and confirm this is the first data commit? or do you mean something else?

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
```diff
@@ -268,7 +280,7 @@ export class ModelInstanceDocumentHandler implements StreamHandler<ModelInstance
 
     // Now validate the relations
     await this._validateRelationsContent(ceramic, model, content)
-    if (!genesis && payload) {
+    if (!genesis && payload && !hasContent) {
```
stbrody, line 283: I'm trying to figure out how to make the behavior here more obvious to new readers of the code.  It's not immediately clear to me what `hasContent` means or why it affects the behavior of whether or not we should validate locked fields.  Can we at least update the doccomments for this function to include a description of the field and how it's used?

Maybe also consider naming the field `hasExistingContent` or something to distinguish that we're talking about whether the stream already had content before this update?

And maybe an inline comment here explaining that if this is a deterministic stream getting its content set for the first time we want to allow the immutable fields to be set to their initial value?

stbrody: I guess another possibility is renaming `hasContent` to simply `enforceImmutableFields` or something like that, then have a comment explaining the reasoning for how this gets set up in `_applySigned` at the call site?

stbrody (commented):
a few small nitpick suggestions.  Main comment is to avoid the now-unnecessary extra clone of the content in the commit handler.

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
JulissaDantes: I did a mix of the suggestions by changing the name and also updating the function documentation

packages/stream-tests/src/__tests__/model-instance-document.test.ts
```diff
@@ -224,6 +255,34 @@ describe('ModelInstanceDocument API http-client tests', () => {
...
+
+    expect(doc.content).toBeNull()
+    const newContent = { one: 'foo', two: 'bar', three: 'foobar', myData: 1 }
+    await doc.replace(newContent)
+
+    expect(doc.content).toEqual(newContent)
+    expect(doc.state.log.length).toEqual(2)
+    expect(doc.state.log[0].type).toEqual(EventType.INIT)
+    expect(doc.state.log[1].type).toEqual(EventType.DATA)
+    // single
```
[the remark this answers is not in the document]
JulissaDantes: Ready!

packages/stream-model-instance-handler/src/model-instance-document-handler.ts
```diff
@@ -268,7 +281,7 @@ export class ModelInstanceDocumentHandler implements StreamHandler<ModelInstance
 
     // Now validate the relations
     await this._validateRelationsContent(ceramic, model, content)
-    if (!genesis && payload) {
+    if (!genesis && payload && !enforceImmutableFields) {
```
stbrody, line 284: this name is backwards now.  When the value is true, you don't enforce the immutable fields.  Can we either invert how the boolean is used so it matches the name, or change the name to `skipImmutableFieldsCheck` or something like that?

JulissaDantes: I renamed it to `skipImmutableFieldsCheck`

stbrody (approved):
LGTM

PR #3189, opened 2024-03-19
chore: Use "level" instead of "level-ts"

ukstv:


packages/cli/src/s3-store.ts
```diff
@@ -1,181 +1,176 @@
...
+    this.level = level
+    this.#loadingLimit = new PQueue({
+      intervalCap: MAX_LOAD_RPS,
+      interval: 1000,
+      carryoverConcurrencyCount: true,
+    })
+  }
+
+  init(): Promise<void> {
+    throw new Error('Method not implemented. And should not be used TBH') // TODO
```
JulissaDantes, line 114: From the error message it says it shouldn not be used, what is the `TODO` for here?

packages/core/src/__tests__/ceramic-pinning.test.ts
```diff
@@ -1,8 +1,8 @@
-import { jest } from '@jest/globals'
+import { jest, describe, expect, beforeEach, afterEach } from '@jest/globals'
```
JulissaDantes, line 1: I noticed that these imports were being used, even though they weren't explicitly imported. Is there a particular reason why are they imported explicitly now?

packages/core/src/state-management/repository.ts
```diff
@@ -166,28 +166,24 @@ export class Repository {
...
+      kvFactory: factory,
     })
   }
 
   async init(): Promise<void> {
-    await this.#deps.keyValueStore.init()
-    await this.pinStore.open(this.#deps.keyValueStore)
-    await this.anchorRequestStore.open(this.#deps.keyValueStore) // Initialization hell
+    await this.pinStore.open(this.#deps.kvFactory)
+    await this.anchorRequestStore.open(this.#deps.kvFactory) // Initialization hell
```
JulissaDantes, line 181: I know this comment was here before this PR but why is this `Initialization hell`?

packages/cli/src/s3-store.ts
ukstv: Removed!

packages/core/src/__tests__/ceramic-pinning.test.ts
ukstv: For us absolute most of the times it does not matter.

We had some troubles in the past when running the test files with implicit `describe`, `expect`, etc. If I remember correctly, it was due to the fact TS compiler (in some non-trivial mode) could not actually figure out the types. If we are to move away from babel-jest, the issue will reappear.

And, you could experience an inconvenience when an editor decides it is mocha or something similar.

packages/core/src/state-management/repository.ts
ukstv: It was and is unclear in what order what store should be opened or closed hence "hell".

stbrody (approved):
nice cleanup

PR #3190, opened 2024-03-20
feat: add ability to publish metrics as ceramic model instance documents (rebased)

gvelez17:
## Description

This PR enables publishing metrics to the ceramic network as model instance documents.  It uses the same `count` and `observe` semantics as ServiceMetrics, with a configurable interval for publishing.

### Key Design notes

- avoids circular dependencies.  the observability model-metrics package accepts a ceramic api instance at run time, so never imports any js-ceramic common packages.

- singleton pattern.  because of the instance import and that its instantiated as a singleton, you can import model-metrics from a code path where the ceramic instance is not accessible, and still use it; also no need to pass it around, just import

- lightweight functions.  the `count` `observe` and `recordError` functions are very lightweight, and only update variables in memory.  At the configured interval time, a summary will be published as a model metric.  

These designs are in the [model-metrics](https://github.com/ceramicnetwork/observability/tree/master/packages/model-metrics) package in the observability repo which is a dependency of this PR.  The code is separate from js-ceramic but the npm package is published under `@ceramicnetwork/model-metrics`

## How Has This Been Tested?

- [x ] Tested manually with a local node on dev-unstable, verified that metrics are correctly produced.

To reproduce this test, add the following into daemon.config.json and run your node.
```
"metrics": {
    "metrics-publisher-enabled": true
  }
```

- [ x] Tested Manually with a harness at https://github.com/3box/runbooks-and-oncall/tree/main/MetricsDemo  (see readme for instructions)

## PR checklist

## References:


gvelez17:
Oops Nathaniel's prettier is mad at me i will fix.  Requested possibly preliminary reviews since this is a bit new stuff, would love eyes on it now and suggestions for tests/quick improvements to get this out.

gvelez17:
Note that 'totalPinnedStreams' is part of the schema, but is not currently being recorded as we do not have a way to get this out of leveldb afaik.  

'0' values are currently also 'no data' values, we could change that to '-1' or something clearer

packages/cli/src/ceramic-daemon.ts
```diff
@@ -336,6 +342,25 @@ export class CeramicDaemon {
...
+    if (opts.metrics?.metricsPublisherEnabled) {
+
+      const ipfsVersion = await ipfs.version();
+      ModelMetrics.start({ ceramic: ceramic,
+                       network: params.networkOptions.name,
+                       ceramicVersion: version,
+                       ipfsVersion: ipfsVersion.version,
+                       intervalMS: opts.metrics?.metricsPublishIntervalMS || DEFAULT_PUBLISH_INTERVAL_MS,
+                       nodeId: ipfsId.publicKey, // what makes the best ID for the node?
+                       nodeName: daemon.hostname,
```
stbrody, line 356: why have both `nodeName` and `nodeIPAddr` if they are both given the same value?

stbrody: same question for `nodeId` and `nodePeerId`

stbrody (changes requested):
Most of my comments are based off the sample metric you included in your comment:

1. Do we need both `id` and `PeerID`?  They seem to have the same value currently.
2. What does `name` represent?
3. `IPAddress` just has the value of 0.0.0.0 currently, which is likely what was given as the arg for the process to bind to, but isn't actually the node's real public IP address.  Can we get the node's actual public IP?  This may need to be changed to a list since one node can have multiple IPs if it's listening to multiple networks.
4. Can we rename `ceramicVersion` to `jsCeramicVersion`?  Since soon we'll also have the rust-ceramic/ceramic-one version we'll want to include and distinguish from the js-ceramic verison.
5. What is `recentCompletedRequests`?  Is this about anchor requests?  Write requests that came into the node via the http API?  Something else?
6. Can we group all the anchoring-related metrics into a sub-object?
7. For the two fields that say "recent" in their name, what defines "recent"?


I'd also love to see a PR for the data model(s) and interface(s) being used.

stbrody:
> IPAddress just has the value of 0.0.0.0 currently, which is likely what was given as the arg for the process to bind to, but isn't actually the node's real public IP address. Can we get the node's actual public IP? This may need to be changed to a list since one node can have multiple IPs if it's listening to multiple networks.

Alternatively, we may just want to remove this field entirely.  This actually may fall under the category of information that devs wont want to be published publicly

nathanielc:
> published

I agree collecting IP is generally considered sensitive information. We can use PeerId to uniquely identify how many nodes are on the network etc without collecting which IP they come from.

packages/cli/src/ceramic-daemon.ts
```diff
@@ -778,6 +803,8 @@ export class CeramicDaemon {
     }
 
     const indexedModels = await this.ceramic.admin.getIndexedModels()
+    console.log(`RECORDING NUMBER INDEXED MODELS: {indexedModels.length}`)
+    ModelMetrics.observe(Observable.TOTAL_INDEXED_MODELS, indexedModels.length)
```
nathanielc, line 807: How often are we collecting these metrics? I don't remember which relation type you said the model uses. Is it SINGLE or LIST? 

Also who is the controller for the stream? The did created from the node seed?

stbrody:
> '0' values are currently also 'no data' values, we could change that to '-1' or something clearer

How about `null` to mean "no data"?  I do think we will probably have semantically meaningful 0 values, so would be good to use something else to mean no data.

nathanielc (commented):
Overall I like the design. Its simple and straightforward. I have some questions below.

Also as an alternative approach if we wanted to make this more generic for any kinds of metrics instead of Ceramic specific metrics.

OpenTelemetry defines a specific structure for metric data. See their [protobuf](https://github.com/open-telemetry/opentelemetry-proto/blob/c5c8b28012583fda55b0cb16f73a820722171d49/opentelemetry/proto/metrics/v1/metrics.proto#L247) definitions. We don't need all the complexity represented in the protobuf but the general design would be to define some models like:


```
interface MetricEvent
  @createModel(description: "Required metric event content interface") {
  ts: DateTime!
  metrics: [Metric]
}

type Metric {
    attributes: [Attribute]
    name: String
    value: Float  
}

type Attribute {
   key: String
   value: String
}
```

The idea here is you can have an arbitrary set of metrics without ever having to change the schema. The downside is its not guaranteed at the schema level which metrics are present. However generally this works well in practice as conventions arise around which metrics are useful/valuable. 

To be clear, I think this PR can/should merge as is (after address all feedback) but we should consider this design when we want to generalize this workflow.

packages/cli/src/ceramic-daemon.ts
gvelez17: So in the schema, which is intended to be future-extensible, we have the generic concept of nodeId, but right now we don't have a nodeId, so we fill it in with either the PeerID or the IPAddr

gvelez17:
> > published
> 
> I agree collecting IP is generally considered sensitive information. We can use PeerId to uniquely identify how many nodes are on the network etc without collecting which IP they come from.

agreed, tho having a public endpoint for the node if there is one could be super valuable to have - this could be optional public endpoint?

packages/cli/src/ceramic-daemon.ts
stbrody: I kind of think peerId is the right thing to be the unique identifier for a node.  Either that or *maaaaybe* the node DID that we use to authenticate to the CAS.  Do we need another concept of NodeID that is distinct from either of those?

gvelez17: @stbrody so the model is here https://github.com/ceramicnetwork/observability/blob/master/packages/model-metrics/src/composites/00-simpleNodeMetrics.graphql

I am trying to make the model itself somewhat future-safe by not making a hard decision on these things - lets sync live to hammer out, am flexible to get this out

gvelez17:
> Overall I like the design. Its simple and straightforward. I have some questions below.
> 
> Also as an alternative approach if we wanted to make this more generic for any kinds of metrics instead of Ceramic specific metrics.
> 
> OpenTelemetry defines a specific structure for metric data. See their [protobuf](https://github.com/open-telemetry/opentelemetry-proto/blob/c5c8b28012583fda55b0cb16f73a820722171d49/opentelemetry/proto/metrics/v1/metrics.proto#L247) definitions. We don't need all the complexity represented in the protobuf but the general design would be to define some models like:
> 
> ```
> interface MetricEvent
>   @createModel(description: "Required metric event content interface") {
>   ts: DateTime!
>   metrics: [Metric]
> }
> 
> type Metric {
>     attributes: [Attribute]
>     name: String
>     value: Float  
> }
> 
> type Attribute {
>    key: String
>    value: String
> }
> ```
> 
> The idea here is you can have an arbitrary set of metrics without ever having to change the schema. The downside is its not guaranteed at the schema level which metrics are present. However generally this works well in practice as conventions arise around which metrics are useful/valuable.
> 
> To be clear, I think this PR can/should merge as is (after address all feedback) but we should consider this design when we want to generalize this workflow.

ok 100%, I would love to do this as version 2

I did have a generic blob one called GenericMetricEvent here https://github.com/ceramicnetwork/observability/blob/master/packages/model-metrics/src/composites/00-simpleNodeMetrics.graphql#L18 but went with an explicit schema for this round

packages/cli/src/ceramic-daemon.ts
gvelez17: yeah its the same DID used to authenticate with CAS, set here https://github.com/ceramicnetwork/js-ceramic/blob/00ceb58c9d9e785f07d9d4349d464a4f5b32199c/packages/cli/src/ceramic-daemon.ts#L325

(we dont' set this explicitly we get it from the ceramic daemon acting as an instance of CeramicAPI)

I did not set it explicitly, maybe should have, it would be LIST - there can be many for the DID.
https://github.com/ceramicnetwork/observability/blob/master/packages/model-metrics/src/composites/00-simpleNodeMetrics.graphql#L25

@nathanielc please feel free to request changes in observability package and I can make - i did not review all the PRs there as i was creating it from scratch

gvelez17:
hey @stbrody i think this and the observability change address the two things we agreed to change for this iteration; one question came up i referred to it here - https://github.com/ceramicnetwork/observability/pull/25/files - about whether we should be using the same model instance id across networks, or not.

packages/cli/src/ceramic-daemon.ts
```diff
@@ -778,6 +803,8 @@ export class CeramicDaemon {
     }
 
     const indexedModels = await this.ceramic.admin.getIndexedModels()
+    console.log(`RECORDING NUMBER INDEXED MODELS: {indexedModels.length}`)
```
stbrody, line 806: please remove the console.log

stbrody (approved):
lgtm mod removing an errant call to console.log

packages/cli/src/ceramic-daemon.ts
gvelez17: ouch.  gone.

PR #3192, opened 2024-03-22
fix: Fix race condition in anchor-processing-loop.test.ts

stbrody:


packages/core/src/anchor/__tests__/anchor-processing-loop.test.ts
```diff
@@ -153,8 +153,12 @@ test('Dont process the same entry multiple times concurrently', async () => {
     do {
       // Need this sleep or else the Node runtime might never switch back to the "process" function.
       await CommonTestUtils.delay(1)
-      for (const e of entries) yield e
-    } while (true)
+      for (const e of entries) {
+        if (!processed.has(e)) {
```
JulissaDantes, line 157: Adding this line will always avoid to process the same entry multiple times, so it sets the tests to always pass or so I think, no?

stbrody: whoops, good catch!

Guess there's no way to make the test truly race-free then (at least that I can think of), but I increased a different timeout that I think should make things more stable

JulissaDantes (approved):
LGTM!

PR #3193, opened 2024-03-25
feat: metrics for counting state of cas requests

samika98:
## Description

We need metrics for keeping track of how many streams have been anchored/errored or are in pending state. This is done in order to track performance of anchoring. 

## How Has This Been Tested?

Tested locally by creating a ceramic node with config where prometheus exporter was turned on.

- [x] Created streams and checked if the metrics were being reported accurately on url http://localhost:9464/metrics : Result : # HELP js_ceramic_cas_request_success_total description missing
# TYPE js_ceramic_cas_request_success_total counter
js_ceramic_cas_request_success_total 7513

## PR checklist

- [Yes ] I have tagged the relevant reviewers and interested parties
- [ Not needed] I have updated the READMEs of affected packages
- [Not needed ] I have made corresponding changes to the documentation

## References:

### DIFF
--- a/packages/core/src/anchor/ethereum/remote-cas.ts
+++ b/packages/core/src/anchor/ethereum/remote-cas.ts
@@ -29,7 +29,7 @@ function parseResponse(streamId: StreamID, tip: CID, json: unknown): AnchorEvent
   }
   const parsed = validation.right
   if (ErrorResponse.is(parsed)) {
-    Metrics.count('cas_request_failed', 1, )
+    Metrics.count('cas_request_failed', 1)
     return {
       status: AnchorRequestStatusName.FAILED,
       streamId: streamId,
--- a/packages/core/src/anchor/ethereum/remote-cas.ts
+++ b/packages/core/src/anchor/ethereum/remote-cas.ts
@@ -9,6 +9,7 @@ import { validate, isValid, decode } from 'codeco'
 import { deferAbortable } from '../../ancillary/defer-abortable.js'
 import { catchError, firstValueFrom, Subject, takeUntil, type Observable } from 'rxjs'
 import { DiagnosticsLogger } from '@ceramicnetwork/common'
+import { ServiceMetrics as Metrics } from '@ceramicnetwork/observability'
 
 const MAX_FAILED_REQUESTS = 3
 const MAX_MILLIS_WITHOUT_SUCCESS = 1000 * 60 // 1 minute
@@ -28,6 +29,7 @@ function parseResponse(streamId: StreamID, tip: CID, json: unknown): AnchorEvent
   }
   const parsed = validation.right
   if (ErrorResponse.is(parsed)) {
+    Metrics.count('cas_request_failed', 1, )
     return {
       status: AnchorRequestStatusName.FAILED,
       streamId: streamId,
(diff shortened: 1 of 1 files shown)


samika98:
If a stream gets polled multiple times then these counts can be off. 
Number of requests = Number of streams created
Number of streams created != cas_request_failed + cas_request_success + cas_request_pending. 

In the case where everything happens at 300 tps we should get Number of streams created = cas_request_success.

packages/core/src/anchor/ethereum/remote-cas.ts
```diff
@@ -36,6 +38,7 @@ function parseResponse(streamId: StreamID, tip: CID, json: unknown): AnchorEvent
     }
   } else {
     if (parsed.status === AnchorRequestStatusName.COMPLETED) {
+      Metrics.count('cas_request_success', 1)
```
stbrody, line 41: nitpick: can we call this `cas_request_completed`?  I want to make sure to keep things clear on the distinction between an anchor request being successfully created against the CAS vs the request being fully processed and the anchor performed successfully.

packages/core/src/anchor/ethereum/remote-cas.ts
```diff
@@ -44,6 +47,9 @@ function parseResponse(streamId: StreamID, tip: CID, json: unknown): AnchorEvent
         witnessCar: parsed.witnessCar,
       }
     }
+    if (parsed.status === AnchorRequestStatusName.PENDING) {
+      Metrics.count('cas_request_pending', 1)
```
stbrody, line 51: I'm not sure how useful this metric will be in practice.  Because we can get this PENDING response many times for the same request as the polling loop loops over the requests that haven't finished processing.  The other two metrics added in this PR don't have that problem because FAILED and COMPLETED are both terminal states that will stop the polling for that request, but PENDING can show up again and again for the same request.

So there won't really be a way to distinguish the case of having one pending request but the loop has processed that one pending request a thousand times, vs there are a thousand pending requests that have each been processed once.

stbrody: BTW this metric we already have I think will be a better proxy for the number of pending requests in flight on the node, as it tracks the size of the AnchorRequestStore: https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/core/src/store/anchor-request-store.ts#L136

samika98: Running a goose scenario on this and generating high load to observe if pending does us anything useful, I agree that it might give us something us something misleading since its not a terminal state. but it can give us somewhat of a representation of how many entries are in the pending state during a goose attack and track their growth rate. 
Will hold of merging this until I get some numbers from my Goose Scenario

packages/core/src/anchor/ethereum/remote-cas.ts
samika98: Appreciate the nitpicks :)

packages/core/src/anchor/ethereum/remote-cas.ts
samika98: The pending count was not telling us much. It was increasing fairly linearly with time and the increase, increase on the amount of requests in the anchor request store did not have any significant impact on the growth of the pending count.

stbrody (approved):
lgtm

PR #3194, opened 2024-03-27
test pinning http-request-mock to 1.8.17

gvelez17:
Just testing as a backup if this affects failing tests (may not be a good test if we are also downgrading node at same time)


gvelez17:
this was just to test a ci failure, closing

PR #3196, opened 2024-03-29
chore: Use "level" instead of "level-ts"

ukstv:
Replaces https://github.com/ceramicnetwork/js-ceramic/pull/3189

Before:

LevelDBStore::list(params, useCaseName)
LevelDBStore::close(useCaseName)
LevelDBStore encapsulated all the actual Level instances per use case. Level lifecycle is hard to reason about: when it gets opened, when it gets closed - all per Level instance, coupled with the actual calls. Too meticulous means too fragile.

After:
LevelKVFactory is an explicit container for all available LevelKVStore (=wrapper over Level) instances. If we have it through LevelKVFactory::open, it means Level instance is truly opened and available. And we have clear separation between LevelKVFactory::close - closes all Level instances, and LevelKVStore::close - closes this particular instance. These methods are idempotent.

And, of course, level-ts is gone in favour of level.


ukstv:
The only real change is https://github.com/ceramicnetwork/js-ceramic/pull/3196/commits/3cb41d03c0b1f0762aaa629715d72751c446d650 - forgotten `new Mutex()`.

stbrody:
@m0ar it's already included in the current release candidate: https://discord.com/channels/682786569857662976/759017171358580787/1225145032663634056

You could try out the RC and see if it helps?

The RC should get promoted to a full release next week.

m0ar:
@stbrody splendid, that'll do the trick! Thanks :cupcake:

PR #3197, opened 2024-03-29
test: Indexing correction

ukstv:
Uncovers actual flakiness of a test, and makes it pass. Note: actual fix is down the line.


stbrody (approved):
makes me sad, but lgtm for now to get the build green

PR #3199, opened 2024-04-01
chore: make TestUtils.waitForConditionOrTimeout throw on failure

stbrody:
Currently it returns false if it fails, which is error prone as it means the test author needs to remember to check the return value or re-check the condition after calling it.  Making it throw on error makes the behavior more obvious and harder to mis-use, at the cost of possibly getting less useful error messages in some cases.


JulissaDantes (approved):
LGTM! I think you are right, when using this function recently I was afraid that some timeout was in fact an error occurring, this will make better DX.

PR #3203, opened 2024-04-11
feat: remove event id

stephhuynh18:
## Description

We do not want to have to mirror the event-id implementation in rust-ceramic in js-ceramic. To do this we will remove all event-id implementation and references from js-ceramic and expect rust-ceramic to calculate. 

Will not pass until relevant changes in rust-ceramic are made


dav1do (commented):
Love to see all that red! It looks like the kubo tests are failing right now. Those should pass even though the rust version won't for a bit.

dav1do (approved):

PR #3206, opened 2024-04-12
fix: disable recon feed in test where it is causing flaky behavior

stephhuynh18:
## Description

The flaky behavior was caused by the node receiving the updates created by itself via recon and applying them. This caused the mock calls to include calls outside of functions we were testing for. I disabled recon in these tests as it does not test updates between nodes anyways.


ukstv (approved):
🔥

PR #3202, opened 2024-04-09
Added WebAuthN support to SignatureUtils for Passkey support

SnickerChar:
## Description

This work was requested as an improvement from the Ceramic developers (Joel Thorstensson) in conjunction with Snickerdoodle Labs. It adds WebAuthN support for DID verification, which will enable Passkey support for DIDs. It just adds the Ceramic authored @didtools/key-webauthn package and wires it into SignatureUtils.

## How Has This Been Tested?

Was not able to test on my local; the test scripts do not seem to work on a Windows machine. CI should catch any problems per devs.

## PR checklist

Before submitting this PR, please make sure:

- [x] I have tagged the relevant reviewers and interested parties
- [x] I have updated the READMEs of affected packages
- [x] I have made corresponding changes to the documentation

PR #3208, opened 2024-04-15
feat: Added WebAuthN support to SignatureUtils for Passkey support

ukstv:
Copy of #3202 to trigger test suite and merge the changes.


JulissaDantes (approved):
As a follow up PR, I would suggest adding the `@didtools/key-webauthn` to the js-did README, because I didnt know we had this.

PR #3214, opened 2024-04-23
feat: log http POST requests body in http-access log

stbrody:
Example log line from the POST request to create a new stream with the genesis commit:
```
[Tue, 23 Apr 2024 21:11:40 GMT] service=http-access ip=127.0.0.1 ts=2024-04-23T21:11:40.348Z method=POST original_url=/api/v0/streams base_url=/api/v0/streams path=/ params=-  body.type=0 body.genesis={"jws":{"payload":"AXESIDoayVr8HX8osV-ubuCJIrM9d7Xs6bJnmwTXbpO7pTJW","signatures":[{"protected":"eyJhbGciOiJFZERTQSIsImtpZCI6ImRpZDprZXk6ejZNa2oyUVQ4blpvaWF1WGNhMVA0SDZ6d0RqTjlSUDNORFdHazFBWFVkaGgyTjVwI3o2TWtqMlFUOG5ab2lhdVhjYTFQNEg2endEak45UlAzTkRXR2sxQVhVZGhoMk41cCJ9","signature":"78ikyfrUnOHtXXmSkeP6Y0nq9eEmV_Hw-tqNtWWsbIiDIMKdEgrjlZhlewlDjgrkBYTjvqtVSDqgufNA2NiYBA"}],"link":"bafyreib2dlevv7a5p4ulcx5on3qisivthv33l3hjwjtzwbgxn2j3xjjsky"},"linkedBlock":"omRkYXRhoWRzdGVwAGZoZWFkZXKiZnVuaXF1ZXA2VUM3YmxPQ2RWVW1VNUQva2NvbnRyb2xsZXJzgXg4ZGlkOmtleTp6Nk1rajJRVDhuWm9pYXVYY2ExUDRINnp3RGpOOVJQM05EV0drMUFYVWRoaDJONXA"} body.opts={"anchor":true,"publish":true,"sync":0,"syncTimeoutSeconds":0} http_version=1.1 req_header- status=200 content_length=379 content_type="application/json; charset=utf-8" ref=- user_agent="node-fetch/1.0 (+https://github.com/bitinn/node-fetch)" elapsed_ms=82.153 error_message="-" error_code=-
```

### DIFF
--- a/packages/cli/src/daemon/log-requests.ts
+++ b/packages/cli/src/daemon/log-requests.ts
@@ -3,7 +3,7 @@ import { Request, Response } from 'express'
 import morgan from 'morgan'
 
 const ACCESS_LOG_FMT =
-  'ip=:remote-addr ts=:date[iso] method=:method original_url=:original-url base_url=:base-url path=:path:params http_version=:http-version req_header:req[header] status=:status content_length=:res[content-length] content_type=":res[content-type]" ref=:referrer user_agent=":user-agent" elapsed_ms=:total-time[3] error_message=":error-message" error_code=:error-code'
+  'ip=:remote-addr ts=:date[iso] method=:method original_url=:original-url base_url=:base-url path=:path:params :body http_version=:http-version req_header:req[header] status=:status content_length=:res[content-length] content_type=":res[content-type]" ref=:referrer user_agent=":user-agent" elapsed_ms=:total-time[3] error_message=":error-message" error_code=:error-code'
 
 export function logRequests(loggerProvider: LoggerProvider): any[] {
   morgan.token<Request, Response>('error-message', (req, res: Response) => {
@@ -32,6 +32,26 @@ export function logRequests(loggerProvider: LoggerProvider): any[] {
       }
     }
     return ' params=-'
+  })
+  morgan.token<Request, Response>('body', (req) => {
+    if (req.body) {
+      const keys = Object.keys(req.body)
+      if (keys.length > 0) {
+        const body = keys.reduce((prev, curr) => {
+          const value = req.body[curr]
+          const valKeys = Object.keys(value)
+          if (valKeys.length > 0) {
+            // value is an object
+            return prev + ` body.${curr}=${JSON.stringify(value)}`
+          } else {
+            // value is a scalar
+            return prev + ` body.${curr}=${value}`
+          }
+        }, '')
+        return body
+      }
+    }
+    return ' body=-'
   })
 
   const logger = loggerProvider.makeServiceLogger('http-access')


packages/cli/src/daemon/log-requests.ts
```diff
@@ -33,6 +33,26 @@ export function logRequests(loggerProvider: LoggerProvider): any[] {
...
+  morgan.token<Request, Response>('body', (req) => {
+    if (req.body) {
+      const keys = Object.keys(req.body)
+      if (keys.length > 0) {
+        const body = keys.reduce((prev, curr) => {
+          const value = req.body[curr]
+          const valKeys = Object.keys(value)
+          if (valKeys.length > 0) {
+            // value is an object
+            return prev + ` body.${curr}=${JSON.stringify(value)}`
```
gvelez17, line 45: is this going to be extremely long?  agree we want it tho

gvelez17 (approved):
lgtm, only slightly worried about really long logs.  also assuming you tested the morgan.token code

stbrody:
> also assuming you tested the morgan.token code

Not quite sure what you mean by this?  The testing I did was running the node, sending http requests to the node, and looking at the log messages generated.

packages/cli/src/daemon/log-requests.ts
stbrody: does seem like a possible risk, yes.  This is printing the full body of the commit being sent to js-ceramic, so is users are creating large json documents in their app, this will print the full document, which could be up to 256kb large

gvelez17:
> > also assuming you tested the morgan.token code
> 
> Not quite sure what you mean by this? The testing I did was running the node, sending http requests to the node, and looking at the log messages generated.

yeah i meant if there were unit tests, its fine

Are we logging every request?  i'm still concerned we could fill up disks with this stuff - hopefully we could log this much detail only on error?

packages/cli/src/daemon/log-requests.ts
gvelez17: um if we do this for every single request on gitcoin i definitely think its a risk of crashing them by filling up the logs.  i would like to sample them

stbrody:
yeah, agreed this is too verbose right now.  We should not merge this without some controls to avoid filling the log with literally every document written to Ceramic.

gvelez17 (changes requested):
(we agreed that we need to make it a little lighter/more selective before merge)

do you want these maybe?  https://github.com/ceramicnetwork/js-ceramic/pull/3215

stbrody:
This PR needs more work to be ready to merge, and since js-ceramic is dying soon and we've made it this long without this working, going to go ahead and close this unmerged

PR #3215, opened 2024-04-24
log full body only on slow requests and only 10 per day

gvelez17:
## Description

to be combined with the PR for logging full body of requests, do so only for slow requests and limit how many

## How Has This Been Tested?

- [ ] Test B

## PR checklist

## References:

### DIFF
--- a/packages/cli/src/daemon/log-requests.ts
+++ b/packages/cli/src/daemon/log-requests.ts
@@ -2,10 +2,38 @@ import { LoggerProvider } from '@ceramicnetwork/common'
 import { Request, Response } from 'express'
 import morgan from 'morgan'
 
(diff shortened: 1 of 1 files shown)


packages/cli/src/daemon/log-requests.ts
```diff
@@ -2,10 +2,38 @@ import { LoggerProvider } from '@ceramicnetwork/common'
 import { Request, Response } from 'express'
 import morgan from 'morgan'
 
+const EXPECTED_RESPONSE_TIME_MS = 2000
+const MAX_DAILY_SLOW_REQUESTS_TO_LOG = 10
```
stbrody, line 6: What about only logging one slow query per minute instead?  I think that will still be infrequent enough to not flood the logs, while making the slow queries a lot more visible if they are happening often.

PR #3216, opened 2024-04-24
feat: FeedStream

ukstv:


packages/core/src/store/feed-aggregation-store.ts
```diff
@@ -69,9 +79,38 @@ export class FeedAggregationStore extends ObjectStore<number, StreamID> {
...
+
   async close(): Promise<void> {
     clearInterval(this.#interval)
     await this.tasks.onIdle()
     await super.close()
   }
 }
+
+class StreamIDFeedSource implements UnderlyingSource<StreamID> {
+  private gt: string | undefined
```
JulissaDantes, line 95: What does `gt` stands for?

packages/core/src/state-management/repository.ts
```diff
@@ -208,17 +205,6 @@ export class Repository {
     return this.#deps.streamUpdater
   }
 
-  /**
-   * Returns the number of streams with writes that are waiting to be anchored by the CAS.
-   */
-  get numPendingAnchors(): number {
```
JulissaDantes, line 214: Is this really no longer being used?

packages/core/src/feed.ts
```diff
@@ -29,11 +30,43 @@ class DocumentsSubject extends Subject<FeedDocument> {}
 export interface PublicFeed {
   aggregation: {
     documents: Observable<FeedDocument>
+    documentsA: (gt?: string) => ReadableStream<FeedDocument>
```
JulissaDantes, line 33: Can we have a more descriptive name for this? Maybe `readableDoc` or something like that, or document it inline?

packages/cli/src/daemon/__tests__/sse-feed.test.ts
```diff
@@ -1,86 +1,108 @@
...
+        controller.close()
+      } else {
+        controller.enqueue(next.value)
+      }
+    },
+  })
+}
 
-test('send events', () => {
+test('send events 2', async () => {
```
JulissaDantes, line 35: Why did we add the `2` in the test name?

packages/cli/src/daemon/sse-feed.ts
```diff
@@ -2,6 +2,52 @@ import type { Observable } from 'rxjs'
...
+    this.res.once('abort', () => {
+      this.abort()
+      controller.error(new ExpectedCloseError())
+    })
+  }
+
+  async write(element: T) {
+    const canWriteMore = this.res.write(`data: ${this.stringify(element)}\n\n`)
+    if (!canWriteMore) {
+      // Wait till data are drained
```
JulissaDantes, line 35: I am being nitpicky, but I would prefer `until`.

packages/core/src/store/feed-aggregation-store.ts
ukstv: I thought it is a traditional name for "greater than".
- `gt` — greater than,
- `lt` — less than,
- `gte` or `ge` — greater than or equal to,
- `lte` or `le` — less than or equal to,
- `eq` — equal to.

packages/core/src/state-management/repository.ts
ukstv: Not uncertain.

packages/cli/src/daemon/__tests__/sse-feed.test.ts
ukstv: Whoops. Fixed. Remnant of dev time.

packages/cli/src/daemon/sse-feed.ts
ukstv: Yes!

ukstv: Changed.

packages/core/src/store/feed-aggregation-store.ts
JulissaDantes: yes it is, but at first glance its not very clear what is getting saved here, Im assuming its a streamId since its used to `find` but I could be wrong, can you clarify what will be stored here?

packages/core/src/feed.ts
ukstv: Yeah, we do not need previous `documents` based on `Observable`. Deleted that. Renamed `documentsA` to `documents`.

packages/core/src/__tests__/ceramic-feed.test.ts
```diff
@@ -43,111 +37,28 @@ describe('Ceramic feed', () => {
     await ipfs2.stop()
   })
 
-  describeIfV3('tile document', () => {
-    test('add entry after creating/updating stream', async () => {
```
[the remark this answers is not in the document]
ukstv: It is for Tiles. Tiles are gone. Corollary: these test do not make any sense now.

JulissaDantes:
> seems like this still needs a lot of test coverage, especially around the resumability behavior and the http-api. Is that coming in a separate PR?

Yes @stbrody , this one: https://github.com/3box/ceramic-tests/pull/116

packages/core/src/__tests__/ceramic-feed.test.ts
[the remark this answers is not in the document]
JulissaDantes: This is going to be covered in the e2e test suite using streams with models instead of tile docs.

packages/core/src/feed.ts
```diff
@@ -21,19 +21,43 @@ export class FeedDocument {
...
+  }
+}
+
+export class StreamLoadTransformer implements Transformer<StreamID, FeedDocument> {
+  constructor(
+    private readonly fromMemoryOrStore: (streamId: StreamID) => Promise<StreamState | undefined>
+  ) {}
+
+  async transform(streamId: StreamID, controller: TransformStreamDefaultController<FeedDocument>) {
+    const found = await this.fromMemoryOrStore(streamId)
```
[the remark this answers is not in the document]
ukstv: It is `this.fromMemoryOrStore` which effectively is `Repository.streamState`, bypassing the cache and the queue: https://github.com/ceramicnetwork/js-ceramic/pull/3216/files#diff-5a05fbd648448211458a6e49c7af10e7ea19d662ff371be017c1f5997703a027R166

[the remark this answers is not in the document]
ukstv: See also: https://github.com/ceramicnetwork/js-ceramic/pull/3216/files#diff-5cf5880ceb461975489503055d311404bbdbdd365bd1cef292996a85e00edd23R46

packages/cli/src/ceramic-daemon.ts
```diff
@@ -892,14 +891,17 @@ export class CeramicDaemon {
...
-  documentsFeed(_: Request, res: Response): void {
-    const source = this.ceramic.feed.aggregation.documents
-    const feed = new SseFeed(
-      this.diagnosticsLogger,
-      source,
-      JsonAsString.pipe(AggregationDocument).encode
-    )
-    feed.send(res)
+  async documentsFeed(_: Request, res: Response): Promise<void> {
+    const readable = this.ceramic.feed.aggregation.documents()
```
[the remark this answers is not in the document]
ukstv: Really oops. Must have removed it accidentally. Thank you!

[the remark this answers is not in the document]
ukstv: Got it back.

packages/core/src/__tests__/ceramic-feed.test.ts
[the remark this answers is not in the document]
stbrody: Okay sounds good.  Yeah the test wasn't testing TileDocument, it was testing core functionality of the data feed, by using TileDocument.  So long as we have a plan to get the coverage back though that's fine.

packages/core/src/feed.ts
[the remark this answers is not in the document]
stbrody: Oh I see.  Can we rename it to `streamState` so it's consistent with the Repository function that it's coming from and less confusing to readers?

stbrody:
> > seems like this still needs a lot of test coverage, especially around the resumability behavior and the http-api. Is that coming in a separate PR?
> 
> Yes @stbrody , this one: [3box/ceramic-tests#116](https://github.com/3box/ceramic-tests/pull/116)

hmm... it feels kind of odd to me to have so little coverage in the actual js-ceramic repo, and to push all the test coverage out to the ceramic-tests repo...  I guess the biggest concern though is PR gating, and I believe we have plans to start gating merges to js-ceramic on the fast tests from ceramic-tests, so I guess once that happens that will address the biggest downside of the separation.  Although I do think it will still be a bit harder to rapidly iterate on tests and code and do proper test-driven development when the tests live in a different repo.

Certainly we *could* write all these tests in the `stream-tests` package within the js-ceramic repo.  That way they would run in CI today, and be fast and easy to run and iterate on for local development, but they'd miss out on the ability to run against our real infra.  Ideally I feel like we'd have tests in both places, but that's probably overkill right now.  So if we only write the tests once, I'm conflicted on which place is better.  I'm curious what @nathanielc and @smrz2001 think about the tradeoffs here.

packages/core/src/feed.ts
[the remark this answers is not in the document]
ukstv: Oh, okay.

packages/core/src/state-management/repository.ts
```diff
@@ -590,7 +576,9 @@ export class Repository {
             ? Model.STREAM_TYPE_ID
             : ModelInstanceDocument.STREAM_TYPE_ID
 
-        await this.handleUpdateFromNetwork(new StreamID(type, genesisCid), eventId.event, model)
+        const streamId = new StreamID(type, genesisCid)
+        await this.handleUpdateFromNetwork(streamId, eventId.event, model)
+        await this.feedAggregationStore.put(streamId, Date.now())
```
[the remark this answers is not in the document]
ukstv: Hmm, true. Removed.

packages/core/src/state-management/repository.ts
```diff
@@ -986,6 +974,21 @@ export class Repository {
         map(FeedDocument.fromStreamState)
       )
       .subscribe(this.feed.aggregation.documents)
```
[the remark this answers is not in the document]
ukstv: This one is a wasted effort. The one below, or the one left is not a wasted effort. For some reason, recon propagates records from ceramic instances in a weird way. Removing this "wasted" effort means if you try to "load" a stream, it does not get picked up.

packages/core/src/feed.ts
```diff
@@ -21,19 +21,43 @@ export class FeedDocument {
...
+
+  get aggregation() {
+    return {
+      documents: this.documents.bind(this),
+    }
+  }
+
+  documents(gt?: string): ReadableStream<FeedDocument> {
+    const transformer = new StreamLoadTransformer(this.fromMemoryOrStore)
+    return this.feedStore.streamIDs(gt).pipeThrough(new TransformStream(transformer))
```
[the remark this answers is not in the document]
ukstv: Oh, does not really matter. It is "at least once" after all. Changed to `gte`.

ukstv:
Okay, @stbrody @JulissaDantes I got back the relevant tests migrated from Tiles to MIDs. As for the time and `gte`, I am inclined to go with `gt` but with a _nanoseconds_ cursor. It is effectively impossible to have same nanoseconds for different writes.

ukstv:
BTW, image verification is due to a weird blockchain JSON RPC error, seemingly unrelated to this change.

packages/cli/src/s3-store.ts
```diff
@@ -1,11 +1,26 @@
...
+} from '@ceramicnetwork/core'
 import LevelUp from 'levelup'
 import S3LevelDOWN from 's3leveldown'
 import toArray from 'stream-to-array'
 import PQueue from 'p-queue'
 import AWSSDK from 'aws-sdk'
 import { Mutex } from 'await-semaphore'
+import type { DeepNonNullable } from 'ts-essentials'
+
+function definiteSearchParams<T extends Partial<StoreSearchParams>>(obj: T): DeepNonNullable<T> {
```
stbrody, line 17: what is this doing?  Can we add some comments here?

packages/codecs/src/feed.ts
```diff
@@ -15,6 +15,7 @@ export const JsonAsString = new Type<unknown, string, string>(
 )
 
 export const AggregationDocument = type({
+  token: string,
```
stbrody, line 18: should we call this `resumeToken` or something else to make it clearer what it's used for?

packages/core/src/store/feed-aggregation-store.ts
```diff
@@ -77,7 +81,7 @@ export class FeedAggregationStore extends ObjectStore<number, StreamID> {
     }
   }
 
-  streamIDs(gt?: string): ReadableStream<StreamID> {
+  streamIDs(gt?: string): ReadableStream<[string, StreamID]> {
```
stbrody, line 84: nitpick: I'd prefer to use a type with named fields for the result rather than a tuple where callers need to know what the values at different indices represent.  But not a big deal, if you want to leave it as-is can we at least add some comments about the structure of the return value?

stbrody (commented):
generally looking good, except I still feel pretty strongly that we should use GTE on the resume query.  Very unlikely is still possible, and I'd rather us just know for sure that we'll never miss an event rather than risk an edge case that can result in data loss.

ukstv:
> generally looking good, except I still feel pretty strongly that we should use GTE on the resume query.  Very unlikely is still possible, and I'd rather us just know for sure that we'll never miss an event rather than risk an edge case that can result in data loss.

I would like not to play telepathy here. If you have a solid proposal in mind, please, write it down.

packages/core/src/store/feed-aggregation-store.ts
```diff
@@ -69,9 +81,41 @@ export class FeedAggregationStore extends ObjectStore<number, StreamID> {
...
+  constructor(
+    private readonly find: FeedAggregationStore['find'],
+    private readonly onWrite: Observable<void>,
+    token: string = process.hrtime.bigint().toString()
+  ) {
+    this.token = token
+  }
+
+  async pull(controller: ReadableStreamController<[string, StreamID]>) {
+    const entries = await this.find({ limit: controller.desiredSize, gt: this.token })
```
stbrody, line 111: concrete proposal is to change this `gt` to `gte`.

And possibly also rename "after" to "at" or "resumeAt" (which is what rust-ceramic uses) throughout the APIs to match the updated semantics

ukstv: > concrete proposal is to change this `gt` to `gte`.
> 
> 
> 
> And possibly also rename "after" to "at" or "resumeAt" (which is what rust-ceramic uses) throughout the APIs to match the updated semantics

Consider the db containing a single entry A. Just changing GT to GTE would mean an infinite stream of As, which seems like an abuse of "at least once" semantics.

stbrody: Only if the feed connection was stopped and re-created with a resumeToken right?  If the feed is established and remains open it should just return the single document and then wait for new data right?

Yeah definitely agreed we shouldn't return the same stream over and over for a single feed session, my concern is when the feed is shut down for any reason (transient network error, js-ceramic restart, client restart, etc) and then resumed, I want to make sure there's no chance of missing updates.

ukstv: > Only if the feed connection was stopped and re-created with a resumeToken right?  If the feed is established and remains open it should just return the single document and then wait for new data right?
> 
> 
> 
> Yeah definitely agreed we shouldn't return the same stream over and over for a single feed session, my concern is when the feed is shut down for any reason (transient network error, js-ceramic restart, client restart, etc) and then resumed, I want to make sure there's no chance of missing updates.

So, without a better proposal, do you expect me to play telepathy?

packages/codecs/src/feed.ts
ukstv: Let it be resumeToken.

packages/core/src/store/feed-aggregation-store.ts
ukstv: Okay.

ukstv:
Okay @stbrody here is a fancy counter we agreed upon: https://github.com/ceramicnetwork/js-ceramic/pull/3216/files#diff-286f50cb14ccba5132a33f458e81c0a5798d5f910b518810f3d16b1f9dacc135R28 Same millisecond entries get deduplicated.

packages/core/src/__tests__/ceramic-feed.test.ts
```diff
@@ -36,134 +57,110 @@ describe('Ceramic feed', () => {
...
-      await CoreUtils.anchorUpdate(ceramic1, stream)
-
-      expect(feed.length).toEqual(2) // 2 commits = 1 genesis commit + 1 anchor commit
-      expect(feed[0].content).toEqual(feed[1].content)
-      expect(feed[0].metadata).toEqual(feed[1].metadata)
-      expect(feed[0].commitId.equals(feed[1].commitId)).toEqual(false)
-      expect(feed[0].eventType).toBe(EventType.INIT)
-      expect(feed[1].eventType).toBe(EventType.TIME)
-      s.unsubscribe()
+  testIfV3('add entry after anchoring stream', async () => {
```
stbrody, line 101: what's the reason this test can't work with rust-ceramic / in recon mode?  It's looking like most (all?) of the users of data feed api are going to be building on rust-ceramic, so seems like we'd really want all the data feed tests to work with rust-ceramic

packages/core/src/feed.ts
```diff
@@ -1,39 +1,68 @@
...
+  constructor(
+    private readonly streamState: (streamId: StreamID) => Promise<StreamState | undefined>
+  ) {}
+
+  async transform(
+    entry: AggregationStoreEntry,
+    controller: TransformStreamDefaultController<FeedDocument>
+  ) {
+    const found = await this.streamState(entry.streamID)
+    if (found) {
```
stbrody, line 63: what would it mean if we didn't find the stream state here?  Seems like that probably shouldn't ever happen right?  Should we at least log a warning or something?

packages/core/src/state-management/repository.ts
```diff
@@ -973,19 +960,6 @@ export class Repository {
    * Adds the stream's RunningState to the in-memory cache and subscribes the Repository's global feed$ to receive changes emitted by that RunningState
```
stbrody, line 960: I guess the second part of this comment isn't relevant anymore?

packages/core/src/store/feed-aggregation-store.ts
```diff
@@ -34,14 +65,20 @@ export class FeedAggregationStore extends ObjectStore<number, StreamID> {
...
 
   findKeys(params?: Partial<StoreSearchParams>): Promise<Array<string>> {
     return this.store.findKeys(params)
   }
 
+  find(params?: Partial<StoreSearchParams>): Promise<Array<IKVStoreFindResult>> {
+    return this.store.find(params)
+  }
+
   async deleteStale(lessThan: number): Promise<number> {
```
stbrody, line 80: can we make clear the units being used here?  Maybe just call the field `lessThanMs` assuming it's expecting milliseconds?

stbrody (approved):
LGTM!

Left a few small final comments, but no need to re-review after addressing.

packages/core/src/__tests__/ceramic-feed.test.ts
ukstv: No reason. It is a mistake.

JulissaDantes (approved):
LGTM! Thanks for the updating the tests.

PR #3218, opened 2024-04-26
fix: rate limit how fast anchor polling loops on small data sets

stbrody:
With this change, when the number of pending requests is small, we will poll each of them once per minute, which is aligned with how things used to work before we switched to the new polling system.  But when the number of pending requests is large, we'll poll them as fast as possible to make sure we don't fall behind.

### DIFF
--- a/packages/core/src/ceramic.ts
+++ b/packages/core/src/ceramic.ts
@@ -270,7 +270,7 @@ export class Ceramic implements StreamReaderWriter, StreamStateLoader {
     )
     const anchorRequestStore = new AnchorRequestStore(
       this._logger,
-      params.anchorLoopMinDurationMs
+      params.anchorLoopMinDurationMs >= 0
         ? params.anchorLoopMinDurationMs
         : DEFAULT_MIN_ANCHOR_LOOP_DURATION_MS
     )
@@ -435,6 +435,7 @@ export class Ceramic implements StreamReaderWriter, StreamStateLoader {
       networkOptions,
       loadOptsOverride,
       sync: config.indexing?.enableHistoricalSync,
+      anchorLoopMinDurationMs: parseInt(config.anchorLoopMinDurationMs, 10),
     }
 
     const modules: CeramicModules = {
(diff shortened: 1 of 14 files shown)


packages/core/src/store/anchor-request-store.ts
```diff
@@ -132,7 +166,17 @@ export class AnchorRequestStore extends ObjectStore<StreamID, AnchorRequestData>
...
           )
           Metrics.observe(ANCHOR_POLLING_PROCESSED, numEntries)
-          await new Promise((resolve) => setTimeout(resolve, restartDelay))
+
+          // If we iterated over all entries in the store in less time than the minLoopDuration,
+          // sleep to avoid spamming the CAS.
+          const loopEndTime = new Date()
+          // jest sometimes does weird things with time so abs is necessary
+          const loopDurationMs = Math.abs(loopEndTime.getTime() - loopStartTime.getTime())
+          if (loopDurationMs < this.#minLoopDurationMs) {
```
dbcfd, line 175: Can we change this to

```
const remainingLoopDuration = this.#minLoopDurationMs - loopDurationMs
if(remainingLoopDuration > this.#minLoopJitterMs) {
   await sleepOrAbort(remainingLoopDuration, this.#abortController.signal)
}```

That way if our remaining loop duration is something like 10ms, we don't bother awaiting.

**Review by dbcfd — CHANGES_REQUESTED** [review]

**Review by smrz2001 — APPROVED** [review]
🥳

**Review by dbcfd — APPROVED** [review]

File: packages/core/src/ceramic.ts [hunk 2]
```diff
@@ -266,11 +268,17 @@ export class Ceramic implements StreamReaderWriter, StreamStateLoader {
       this._logger,
       params.networkOptions.name
     )
+    const anchorRequestStore = new AnchorRequestStore(
+      this._logger,
+      params.anchorLoopMinDurationMs >= 0
```
stbrody, line 273: note that `null >= 0` returns true.  We might want to clean up this condition check.

PR #3223, opened 2024-05-02
feat: add anchor metrics and poll pacing

smrz2001:
This PR adds a few more anchor polling related metrics and implements a simple pacing for anchor polling with some jitter.

@stbrody, I've not added anything besides the jitter for generally preventing status requests from lining up, whether during normal operation or after a restart.

I was thinking about adding a sleep between queries in the first iteration like you suggested but wouldn't that also cause requests to line back up in the next iteration? You might have to elaborate the approach a bit more, and I can implement it.

### DIFF
--- a/packages/core/src/anchor/anchor-processing-loop.ts
+++ b/packages/core/src/anchor/anchor-processing-loop.ts
@@ -27,6 +27,10 @@ export class AnchorProcessingLoop {
    */
   readonly #anchorStoreQueue: NamedTaskQueue
   readonly #anchorPollingMetrics: TimeableMetric
+  /**
+   * Cache for the last time we polled a stream. Used to prevent polling the same stream more than once per minute.
+   */
+  readonly #pollCache: Map<string, number> = new Map()
 
   constructor(
     batchSize: number,
@@ -50,6 +54,10 @@ export class AnchorProcessingLoop {
       concurrency,
       store.infiniteList(batchSize),
       async (streamId) => {
+        // Exit early if we've already polled this stream in the last minute
+        if (!this.checkPollTime(streamId.toString())) {
+          return
+        }
         try {
           logger.verbose(
             `Loading pending anchor metadata for Stream ${streamId} from AnchorRequestStore`
@@ -78,6 +86,7 @@ export class AnchorProcessingLoop {
             await this.#anchorStoreQueue.run(streamId.toString(), async () => {
               const loaded = await store.load(streamId)
               if (loaded.cid.equals(entry.cid)) {
+                this.#pollCache.delete(streamId.toString())
                 await store.remove(streamId)
                 logger.verbose(
                   `Entry from AnchorRequestStore for Stream ${streamId} removed successfully`
@@ -110,5 +119,18 @@ export class AnchorProcessingLoop {
   async stop(): Promise<void> {
     this.#anchorPollingMetrics.stopPublishingStats()
     return this.#loop.stop()
+  }
+
+  /**
+   * Check the poll time for a stream. Updates the poll time and returns true if the stream has not been polled in the
+   * last minute.
+   */
+  checkPollTime(streamId: string): boolean {
+    if (this.#pollCache.has(streamId) && (Date.now() < this.#pollCache.get(streamId))) {
+      return false
+    }
+    // Add ±10 seconds of jitter to prevent all streams from being polled at the same time every minute after a restart
+    this.#pollCache.set(streamId, 50_000 + Math.floor(Math.random() * 20_000))
+    return true
   }
 }
--- a/packages/core/src/anchor/anchor-processing-loop.ts
+++ b/packages/core/src/anchor/anchor-processing-loop.ts
@@ -13,6 +13,9 @@ const METRICS_REPORTING_INTERVAL_MS = 10000 // 10 second reporting interval
 
 const DEFAULT_CONCURRENCY = 25
 
+const ANCHOR_POLL_PACING_MS = 60_000
+const ANCHOR_POLL_JITTER_MS = 10_000
+
 /**
  * Get anchor request entries from AnchorRequestStore one by one. For each entry, get CAS response,
  * and handle the response via `eventHandler.handle`.
@@ -28,7 +31,8 @@ export class AnchorProcessingLoop {
   readonly #anchorStoreQueue: NamedTaskQueue
   readonly #anchorPollingMetrics: TimeableMetric
   /**
-   * Cache for the last time we polled a stream. Used to prevent polling the same stream more than once per minute.
+   * Cache for the last time we polled a stream. Used to prevent polling the same stream more than once per pacing
+   * interval.
    */
   readonly #pollCache: Map<string, number> = new Map()
 
@@ -54,7 +58,7 @@ export class AnchorProcessingLoop {
       concurrency,
       store.infiniteList(batchSize),
       async (streamId) => {
-        // Exit early if we've already polled this stream in the last minute
+        // Exit early if we've already polled this stream in the last pacing interval
         if (!this.checkPollTime(streamId.toString())) {
           return
         }
@@ -123,14 +127,15 @@ export class AnchorProcessingLoop {
 
   /**
    * Check the poll time for a stream. Updates the poll time and returns true if the stream has not been polled in the
-   * last minute.
+   * last pacing interval.
    */
   checkPollTime(streamId: string): boolean {
     if (this.#pollCache.has(streamId) && (Date.now() < this.#pollCache.get(streamId))) {
       return false
     }
-    // Add ±10 seconds of jitter to prevent all streams from being polled at the same time every minute after a restart
-    this.#pollCache.set(streamId, 50_000 + Math.floor(Math.random() * 20_000))
+    // Add some jitter to prevent all streams from being polled at exactly the same time after a restart
+    const jitter = Math.floor(Math.random() * 2 * ANCHOR_POLL_JITTER_MS) - ANCHOR_POLL_JITTER_MS
+    this.#pollCache.set(streamId, ANCHOR_POLL_PACING_MS + jitter)
     return true
   }
 }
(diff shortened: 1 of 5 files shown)


packages/core/src/anchor/anchor-processing-loop.ts
```diff
@@ -50,6 +58,10 @@ export class AnchorProcessingLoop {
       concurrency,
       store.infiniteList(batchSize),
       async (streamId) => {
+        // Exit early if we've already polled this stream in the last pacing interval
+        if (!this.checkPollTime(streamId.toString())) {
+          return
```
stbrody, line 63: let's add a 'verbose' log message here that we're skipping this, with the streamid, current time, and last poll time.

packages/core/src/anchor/anchor-processing-loop.ts
```diff
@@ -50,6 +58,10 @@ export class AnchorProcessingLoop {
       concurrency,
       store.infiniteList(batchSize),
       async (streamId) => {
+        // Exit early if we've already polled this stream in the last pacing interval
+        if (!this.checkPollTime(streamId.toString())) {
```
stbrody, line 62: we shouldn't cache by streamid, we should do it by CID of the commit we are polling for.  Otherwise if the stream gets updated we risk delaying the creation of the new request against the CAS.

packages/core/src/anchor/anchor-processing-loop.ts
```diff
@@ -27,6 +30,11 @@ export class AnchorProcessingLoop {
    */
   readonly #anchorStoreQueue: NamedTaskQueue
   readonly #anchorPollingMetrics: TimeableMetric
+  /**
+   * Cache for the last time we polled a stream. Used to prevent polling the same stream more than once per pacing
+   * interval.
+   */
+  readonly #pollCache: Map<string, number> = new Map()
```
stbrody, line 37: Using a Map means the cache can grow unbounded.  Can we use an `LRUCache` instance instead?

packages/core/src/anchor/anchor-processing-loop.ts
```diff
@@ -111,4 +124,18 @@ export class AnchorProcessingLoop {
     this.#anchorPollingMetrics.stopPublishingStats()
     return this.#loop.stop()
   }
+
+  /**
+   * Check the poll time for a stream. Updates the poll time and returns true if the stream has not been polled in the
+   * last pacing interval.
+   */
+  checkPollTime(streamId: string): boolean {
+    if (this.#pollCache.has(streamId) && (Date.now() < this.#pollCache.get(streamId))) {
```
stbrody, line 133: nitpick but would be nice to only call into the cache once and just check for null vs having a separate lookup for the `has` method.  Hashtable lookups do have some performance cost due to the hashing, so it's nice to minimize how often you need to go into them.

packages/core/src/anchor/anchor-processing-loop.ts
```diff
@@ -111,4 +124,18 @@ export class AnchorProcessingLoop {
...
+      return false
+    }
+    // Add some jitter to prevent all streams from being polled at exactly the same time after a restart
```
stbrody, line 136: this seems like it only helps after the first pass over the AnchorRequestStore.  The first time through though you're going to poll everything in the store as fast as possible because nothing will be in the cache.

stbrody (commented):
the idea I had with the sleep was to add a small sleep after polling each request (possibly with some jitter), during the first pass through the store after a restart.  Then on subsequent loops through the store you won't need that anymore since everything will have gotten into the cache with different expiration times due to those sleeps.

I don't totally love the idea of having logic that runs on only the first loop and not subsequent loops though, feels kind of messy, but I haven't been able to think of anything else that will help the problem of the first loop hammering the CAS as fast as possible.

packages/core/src/anchor/anchor-processing-loop.ts
smrz2001: Yes, agreed. Only helps after the first time through the store. Adding jitter is a good idea anyway because often traffic patterns will start lining up and causing bursty processing.

packages/core/src/anchor/anchor-processing-loop.ts
smrz2001: Something like this?
```
    const nextPollTime = this.#pollCache.get(streamId)
    if (nextPollTime && (Date.now() < nextPollTime)) {
      return false
    }

```

packages/core/src/anchor/anchor-processing-loop.ts
smrz2001: I was going to use an LRU cache after you suggested it earlier but had a concern about that that I forgot to ask.

Can the LRU cache get into a pathological state where a high ingress traffic rate can cause older entries to quickly age out, and now we're just polling for many CIDs constantly because they keep getting aged out?

e.g. With an LRU cache of 10K and an ingress TPS of 100, the oldest entries will get evicted every couple of minutes, and when the loop gets back around to them, it will just poll them again right away.

A large anchor store likely means an issue with the CAS, which means as soon as it comes back up, we might start slamming it.

Maybe a large enough cache would be sufficient? 100K?

packages/core/src/anchor/anchor-processing-loop.ts
smrz2001: That's actually why I wanted to store the last polled timestamp in the DB versus keeping it in memory.

There's always the risk that at startup we slam the CAS too hard. Realistically, we might ask partners to upgrade and some of the upgrades across the network happen around the same time. Now, a bunch of Ceramic nodes will start slamming CAS with everything in their anchor stores.

Probably not a very likely scenario so if the risks of using the DB outweigh the risks of something like this happening, I'm cool with keeping the cache in-memory.

smrz2001: Although, if the node takes long enough to restart, it won't matter even if stuff is written to disk... All the timestamps would have expired anyway 😭

smrz2001:
> the idea I had with the sleep was to add a very small sleep (like 1ms or less) after polling each request (possibly with some jitter), during the first pass through the store after a restart. Then on subsequent loops through the store you won't need that anymore since everything will have gotten into the cache with different expiration times due to those sleeps.
> 
> I don't totally love the idea of having logic that runs on only the first loop and not subsequent loops though, feels kind of messy, but I haven't been able to think of anything else that will help the problem of the first loop hammering the CAS as fast as possible.

Yeah, I see what you're saying 😢

What if we don't send _any_ requests during the first iteration of the loop and just set up the next polling times? Kind of like the "trailing edge" throttling in the `lodash.throttle` function that Samika was looking at.

What do you think of this version? Requests for updated streams will also go out at the trailing edge of the pacing interval but that's ok?
```
  checkPollTime(streamId: string): boolean {
    // Add some jitter to prevent all streams from being polled at exactly the same time after a restart
    const jitter = Math.floor(Math.random() * 2 * ANCHOR_POLL_JITTER_MS) - ANCHOR_POLL_JITTER_MS
    const nextPollTime = this.#pollCache.get(streamId)
    if (nextPollTime && (Date.now() < nextPollTime)) {
      return false
    }
    this.#pollCache.set(streamId, ANCHOR_POLL_PACING_MS + jitter)
    return !!nextPollTime;
  }
```

packages/core/src/anchor/anchor-processing-loop.ts
smrz2001: Is that a big problem? We also do a fire-and-forget submission to the CAS before adding a CID to the processing loop. I don't have a strong opinion either way, just feels less noisy to create a request / get status per stream versus per CID.

packages/core/src/anchor/anchor-processing-loop.ts
christianlavoie: > Can the LRU cache get into a pathological state where a high ingress traffic rate can cause older
> entries to quickly age out, and now we're just polling for many CIDs constantly because they
> keep getting aged out?

Yes.

packages/core/src/anchor/anchor-processing-loop.ts
christianlavoie: Also, it's unlikely that partners would all restart nodes at the SAME time. Within a few hours, sure. A few minutes in the extreme. But nearing the same second? Hard to believe.

christianlavoie (commented):
Just comments.

packages/core/src/anchor/anchor-processing-loop.ts
smrz2001: Yes, that's fair.

We should still figure out a nicer way to pace status requests at startup since we could slam the CAS with 100s of TPS from a single node. This would be good to avoid even if CAS can take it.

packages/core/src/anchor/anchor-processing-loop.ts
stbrody: > e.g. With an LRU cache of 10K and an ingress TPS of 100, the oldest entries will get evicted every couple of minutes, and when the loop gets back around to them, it will just poll them again right away.

If you have that many requests, then it will probably take more than a minute to get through them all, at which point you would want to poll them again anyway.

> Maybe a large enough cache would be sufficient? 100K?

I had been thinking more like 10k, but maybe 100k would be fine.  We just need to do some back-of-the-envelope math on how much memory this will take up.

packages/core/src/anchor/anchor-processing-loop.ts
stbrody: feels more error prone to me.  You also have to make sure to remove entries from the cache if you're caching by streamid, whereas you can just let them expire naturally if you're only caching by CID.  Also CIDs are smaller than StreamIDs, so will keep the cache size down.

But no, it's not a big deal either way, if you prefer streamid I don't feel that strongly.

packages/core/src/anchor/anchor-processing-loop.ts
stbrody: yeah exactly.

packages/core/src/anchor/anchor-processing-loop.ts
stbrody: yeah, this leads to still thinking our best bet is a small sleep between each poll in just the first pass through the store after restart.

stbrody (commented):
> What if we don't send any requests during the first iteration of the loop and just set up the next polling times?

does that mean you wouldn't poll for *anything* for the first minute the node is up?  That doesn't seem ideal

ukstv (approved):
IDK. See nothing criminal here.

smrz2001:
Removed all polling updates in favor of finalizing https://github.com/ceramicnetwork/js-ceramic/pull/3218.

PR #3226, opened 2024-05-10
fix: Enforce unique field in header for set account relation MIDs

stbrody:


packages/stream-model-instance-handler/src/model-instance-document-handler.ts
```diff
@@ -351,18 +351,33 @@ export class ModelInstanceDocumentHandler implements StreamHandler<ModelInstance
...
+        }
+        break
+      case 'list':
+        if (!header.unique) {
+          throw new Error(
+            `ModelInstanceDocuments for models with LIST accountRelations must be created with a unique field`
+          )
+        }
+        break
+      // default:
```
stbrody, line 376: Uncommenting this I get the error: `TS2345: Argument of type 'string' is not assignable to parameter of type 'never'.`

I think the issue is that this is just treating it as a switch on a generic string, rather than on an enum.  The problem is that the type that I want to use here is `ModelAccountRelationV2` (https://github.com/ceramicnetwork/js-ceramic/blob/e21c53ca914a08db21eda77d5704dc4a12e2ecff/packages/stream-model/src/codecs.ts#L111-L120), but that isn't a simple union, it's something generated with codeco and I couldn't figure out how to actually use it here.

Is there something I'm missing about how to use a codeco generated type with an exhaustive switch statement like this so we can get compiler warnings about an unhandled case?  Or does this just not work with the codeco types?

@ukstv

PaulLeCam (approved):

PR #3227, opened 2024-05-13
feat: Add env var to disable Data Feed

stbrody:


JulissaDantes (commented):
Could you please provide the rationale for why we want to be able to disable the data feed?

stbrody:
> Could you please provide the rationale for why we want to be able to disable the data feed?

We are seeing an OOM crash in js-ceramic and are trying to eliminate variables.  Data Feed is the newest big thing added to js-ceramic so if it was the cause it would explain why we never observed this OOM previously.

stbrody:
Given that we seem to have discovered the source of the memory leak and it has nothing to do with DF, is this still worth adding or should we just close this PR without merging? @stephhuynh18

stephhuynh18:
> Given that we seem to have discovered the source of the memory leak and it has nothing to do with DF, is this still worth adding or should we just close this PR without merging? @stephhuynh18

i think close makes sense

PR #3228, opened 2024-05-14
fix: cache ajv schemas rather than ajv objects for schema validation

dbcfd:


stbrody:
I'd like @ukstv to take a look at this also since he added the caching here originally and has looked into how Ajv does its internal caching in the past as well

stephhuynh18 (approved):

packages/stream-model-instance-handler/src/schema-utils.ts
```diff
@@ -22,18 +19,17 @@ function buildAjv(): Ajv {
...
 
   public validateSchema(content: Record<string, any>, schema: SchemaObject, schemaId: string) {
-    let existingSchema = this.schemas.get(schemaId)
+    let existingSchema = this.ajv.getSchema(schemaId)
     if (!existingSchema) {
-      existingSchema = this.ajv.compile(schema)
-      this.schemas.set(schemaId, existingSchema)
+      this.ajv.addSchema(schema, schemaId)
+      //we've added the schema above, so ajv will have it
+      existingSchema = this.ajv.getSchema(schemaId)!
```
stbrody, line 32: is the exclamation point at the end intentional? If so what is it doing?

dbcfd: Resolving the possible null value as not null. We just inserted, so it should never be null

ukstv (approved):

stbrody (approved):
lgtm

PR #3235, opened 2024-06-05
test: Update Jest and add custom reporter to stream-tests

ukstv:
This is to help diagnose our sudden OOM failures while testing on CI. There is no clue at what point, at what test file or test case, the OOM happens. The PR adds a custom Jest reporter that prints (via `console.log`) a message to console when:
- a test file starts,
- a test file finishes,
- a test case (i.e. `test` or `it` call) starts,
- a test case finishes.

A message includes `process.memoryUsage().heapUsed` so that we could also see how the memory usage progresses.

I expect this makes it easier to diagnose a future OOM occurrence. After we fix it, the custom Jest reporter should be removed.

Now the changes:
1. Update Jest because our current version does not call reporter hooks `onTestCaseStart` and `onTestCaseResult`.
2. Add a custom Jest reporter to `stream-tests` package: the one that OOMs.
3. Change order of closing - first close anchor request loop, then anchor request store. There is a clear indication the current close order is wrong. I am not sure though if it is a reason for OOM.


ukstv:
Amazing! OOM right away!

PR #3237, opened 2024-06-05
fix: Remove history sync test

stbrody:
We suspect it may be causing the OOM we're seeing in CI.  Regardless, we don't support HDS and are about to break it entirely by removing tree metadata from the merkle trees the CAS creates.


stbrody:
Going to go ahead and merge this to see if helps stabilize CI at all, seeing so many OOM failures across PRs and CI right now

PR #3238, opened 2024-06-06
fix: sleep if loop duration is a special 0

ukstv:


stbrody:
hmm, can you explain the motivation behind this change?  The `minLoopDurationMs` isn't configurable currently and is hard-coded to 1 minute, so I don't think it could ever be 0 so this new case will never be hit.

ukstv:
> hmm, can you explain the motivation behind this change? The `minLoopDurationMs` isn't configurable currently and is hard-coded to 1 minute, so I don't think it could ever be 0 so this new case will never be hit.

https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/stream-tests/src/create-ceramic.ts#L27 Really?

ukstv:
On a serious note, pinged you on Linear.

stbrody:
ah whoops, I was only thinking about in production code, not test code.  Good catch, makes sense.

ukstv:
Yahoo!

PR #3239, opened 2024-06-13
feat: default publishing metrics on the network to true

gvelez17:
## Description

As we move towards more decentralization, publishing metrics on the ceramic network becomes more important - so defaulting it to on.  Will notify folks in an announcement and post.

## How Has This Been Tested?

- [x] run locally
- [x] unit tests

## PR checklist

## References:


docs-dev/ADVANCED_CONFIG.md
```diff
@@ -41,6 +41,7 @@ Do NOT copy the following code it is for reference only.
     "metrics": {
          "collector-host": STRING               // hostname of OTLP collector when push exporter enabled
          "metrics-exporter-enabled": BOOLEAN    // whether push exporting is enabled, default false
+         "metrics-publisher-disabled": BOOLEAN   // node metrics are published by default, set this to true to disable
```
christianlavoie, line 44: Can we avoid the double negative and just have "metrics-publisher-enabled" defaulting to true?

gvelez17: hey @christianlavoie so i initially was doing that, but in the case where there is no 'metrics' section at all, i thought it was clearer to have the disabled switch, for when it is missing entirely

so if they do not even have the switch present, it is enabled

in order to disable it, you must have a section with

metrics : {
    metrics-publisher-disabled: true
 }
 
 one thign i found when trying the first method, was that if you simply use the definition of it as 
 
  "metrics-publisher-disabled": BOOLEAN = true
  
  that does NOT in fact work, if the switch is absent, in the config.  So we'd need a different code path to have it as enabled=true for missing configs
  
  that said, I can do it that way, it will add some new code!

gvelez17: i could also just change the test in the code to 

if (! config.metrics || config.metrics.publisherEnabled)

and that might do it, for the case where there is no metrics section.  However, if the switch is missing, i think it still does not work with the code pattern we have now

gvelez17:
ok - @christianlavoie i changed it to 'enabled = true' which i agree is clearer to the end user.  The bit I didn't like is the logic making them enabled by default now is in two places, in the daemon config file and the start, for the case when there is no metrics section at all.  Added a comment, so hopefully we will notice this if we switch it later to non-default.

christianlavoie (approved):

stbrody (commented):
I believe on the call the other day we had discussed pairing this with a log message on startup informing node operators that some metrics about their node are being shared publicly, with a permalink to a docs page with more information about what information is being shared?

gvelez17:
ok @stbrody added a doc file and a log message referring to it.  My preference is to keep the documentation IN the repo with the code, so if we change the code later each version of code has a corresponding document.

The current message is as follows (if node metrics are enabled, which they are now by default)

```
[2024-06-19T22:05:42.714Z] IMPORTANT: Ceramic API running on 0.0.0.0:7007'
[2024-06-19T22:05:42.758Z] IMPORTANT: Publishing Node Metrics to the Ceramic Network.  To learn more, including how to disable publishing, please see the NODE_METRICS.md file for your branch, e.g. https://github.com/ceramicnetwork/js-ceramic/blob/develop/docs-dev/NODE_METRICS.md
```
I would like to merge this now into develop; i can also write a more extensive blog post on it after it is deployed on dev-unstable and maybe testnet.

docs-dev/NODE_METRICS.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Metric Data Published
+
+The data published is defined in the [@ceramicnetwork/node-metrics](https://www.npmjs.com/package/@ceramicnetwork/node-metrics) package.
+
+Currently, this includes the following about the Ceramic node:
+
+  - DID used to log into the Ceramic Anchor Service
+  - IP address, if known
+  - Peer ID (currently on IPFS)
```
stbrody, line 13:
```suggestion
  - Peer ID
```

docs-dev/NODE_METRICS.md
```diff
@@ -0,0 +1,41 @@
...
+  - Ceramic Version
```
stbrody, line 14:
```suggestion
  - js-ceramic version
```

docs-dev/NODE_METRICS.md
```diff
@@ -0,0 +1,41 @@
...
+  - IPFS Version
```
stbrody, line 15: ceramic
```suggestion
  - ceramic-one version, or IPFS version if still using IPFS.
```

docs-dev/NODE_METRICS.md
```diff
@@ -0,0 +1,41 @@
...
+
+As well as the following dynamic metrics published periodically (by default once per minute):
+
+  - total Pinned streams
+  - total Indexed models
+  - current pending requests to CAS
```
stbrody, line 21:
```suggestion
  - current number of pending requests to CAS
```

docs-dev/NODE_METRICS.md
```diff
@@ -0,0 +1,41 @@
...
+  - mean Anchor request age
```
stbrody, line 22:
```suggestion
  - mean pending Anchor request age
```

docs-dev/NODE_METRICS.md
```diff
@@ -0,0 +1,41 @@
...
+  - max Anchor request age
```
stbrody, line 23:
```suggestion
  - max pending Anchor request age
```

docs-dev/NODE_METRICS.md
```diff
@@ -0,0 +1,41 @@
...
+  - number of created requests in the last window
```
stbrody, line 24: is this still about anchor requests or is this some other concept of request?  Should this say "number of writes" instead?

docs-dev/NODE_METRICS.md
```diff
@@ -0,0 +1,41 @@
+# Node Metrics
+
+By default, Ceramic Nodes publish some limited metrics about their state to a model on the Ceramic Network.
```
stbrody, line 3:
```suggestion
By default, Ceramic Nodes publish some limited metrics about their state publicly to a model on the Ceramic Network.
```

stbrody (commented):
a few small suggestions/questions, mostly around adding clarity to what the different metrics represent, and future-proofing us a little bit for the switch to ceramic-one.

Also, I don't see the code that adds the log message you mentioned in your last comment?

docs-dev/NODE_METRICS.md
gvelez17: this is about anchor requests.  that is literally what it is measuring but its fine if we say "writes" - up to you

gvelez17:
ugh bad merge (I thought i'd rebased nicely!)  recreated here https://github.com/ceramicnetwork/js-ceramic/pull/3244

gvelez17:
> a few small suggestions/questions, mostly around adding clarity to what the different metrics represent, and future-proofing us a little bit for the switch to ceramic-one.
> 
> Also, I don't see the code that adds the log message you mentioned in your last comment?

Dunno what happened in the merge, its here https://github.com/ceramicnetwork/js-ceramic/pull/3239/files#diff-29a806cda83d5289fde2f9b08500d1f2335980307791376095ed3d7c600da98eR364

but this is so messy now i recreated it here https://github.com/ceramicnetwork/js-ceramic/pull/3244

in that PR the message is here

https://github.com/ceramicnetwork/js-ceramic/pull/3244/files#diff-29a806cda83d5289fde2f9b08500d1f2335980307791376095ed3d7c600da98eR364


closing this one now, asked your review on that one.  It has the latest with the committed suggestions you made @stbrody !

docs-dev/NODE_METRICS.md
stbrody: if it's measuring anchor requests i'd rather it say anchor requests 😅

PR #3244, opened 2024-06-20
feat: default node metrics publishing to on and add docs

gvelez17:
## Description

As we move towards more decentralization, publishing metrics on the ceramic network becomes more important - so defaulting it to on.  Will notify folks in an announcement and post.

## How Has This Been Tested?

- [x] run locally
- [x] unit tests

## PR checklist

## References:

### DIFF
--- a/docs-dev/NODE_METRICS.md
+++ b/docs-dev/NODE_METRICS.md
@@ -1,0 +1,41 @@
+# Node Metrics
+
+By default, Ceramic Nodes publish some limited metrics about their state publicly to a model on the Ceramic Network.
+
+## Metric Data Published
+
+The data published is defined in the [@ceramicnetwork/node-metrics](https://www.npmjs.com/package/@ceramicnetwork/node-metrics) package.
+
+Currently, this includes the following about the Ceramic node:
+
+  - DID used to log into the Ceramic Anchor Service
+  - IP address, if known
+  - Peer ID
+  - js-ceramic version
+  - ceramic-one version, or IPFS version if still using IPFS.
+
+As well as the following dynamic metrics published periodically (by default once per minute):
+
+  - total Pinned streams
+  - total Indexed models
+  - current number of pending requests to CAS
+  - mean pending Anchor request age
+  - max pending Anchor request age
+  - number of writes in the last window
+  - number of completed requests in the last window
+  - number of errors in the last window
+  - a sample of recent errors encountered
+
+These metrics are very helpful for measuring the health of the network and addressing problems early.
+
+## Disabling Node Metrics
+
+However, if you would like to *disable* metrics from your node, you can easily do so!
+
+Edit your ceramic daemon config file and set
+
+` "metrics-publisher-enabled": false `
+
+in the `metrics` section.  For more information about editing the ceramic daemon config see [ADVANCED_CONFIG](./ADVANCED_CONFIG.md)
+
+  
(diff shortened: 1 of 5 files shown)


docs-dev/NODE_METRICS.md
```diff
@@ -0,0 +1,41 @@
...
+  - ceramic-one version, or IPFS version if still using IPFS.
+
+As well as the following dynamic metrics published periodically (by default once per minute):
+
+  - total Pinned streams
+  - total Indexed models
+  - current number of pending requests to CAS
+  - mean pending Anchor request age
+  - max pending Anchor request age
+  - number of writes in the last window
```
stbrody, line 24: wait in the other PR you said this is measuring the number of anchor requests, but you changed it to say the number of writes, so now I'm confused.  Is this measuring the number of writes coming into this node, the number of http requests coming into this node, the number of anchor requests this node is making to the CAS, something else?  I'd like us to be precise about what is actually being measured here, both on this line and the line below

packages/cli/src/ceramic-daemon.ts
```diff
@@ -322,7 +322,7 @@ export class CeramicDaemon {
     if (provider) {
       await did.authenticate()
       diagnosticsLogger.imp(
-        `Node DID set to '${did.id}. This DID will be used to authenticate to the anchor service'`
+        `Node DID set to '${did.id}'. This DID will be used to authenticate to the anchor service`
```
stbrody, line 325: nice, that was bothering me too 😅

packages/cli/src/ceramic-daemon.ts
```diff
@@ -359,6 +360,9 @@ export class CeramicDaemon {
         nodePeerId: ipfsId.publicKey,
         logger: diagnosticsLogger,
       })
+      diagnosticsLogger.imp(
+        `Publishing Node Metrics to the Ceramic Network.  To learn more, including how to disable publishing, please see the NODE_METRICS.md file for your branch, e.g. https://github.com/ceramicnetwork/js-ceramic/blob/develop/docs-dev/NODE_METRICS.md`
```
stbrody, line 364:
```suggestion
        `Publishing Node Metrics publicly to the Ceramic Network.  To learn more, including how to disable publishing, please see the NODE_METRICS.md file for your branch, e.g. https://github.com/ceramicnetwork/js-ceramic/blob/develop/docs-dev/NODE_METRICS.md`
```

stbrody (commented):
last question about what exactly one of the metrics is actually tracking

docs-dev/NODE_METRICS.md
gvelez17: so we actually currently measure

```
  recentCompletedRequests: Int
  ```
  
  and this is counted here - 
  https://github.com/ceramicnetwork/js-ceramic/blob/ccfb34203e6218926e213970c9d379b7680f2cea/packages/core/src/anchor/anchor-processing-loop.ts#L72
  
  We have space in the composite reserved for 'recentCreatedRequests' but we do not actually observe those at this time. If we want to do so i think taht coudl be a separate PR?
  
  I will change the documentation to mention "recently completed anchor requests"

gvelez17: ok, removed this line, since the "completed anchor requests" line was already there.  When we add code to measure the number of created anchor requests then we can add it to the documentation as well.

stbrody (approved):
LGTM, thanks!

PR #3246, opened 2024-06-21
feat: send js-ceramic and ceramic-one versions to CAS when creating requests

stbrody:
js-ceramic half corresponding to https://github.com/ceramicnetwork/ceramic-anchor-service/pull/1232

Builds on https://github.com/ceramicnetwork/js-ceramic/pull/3245

### DIFF
--- a/packages/cli/src/__tests__/ceramic-multi-daemon.test.ts
+++ b/packages/cli/src/__tests__/ceramic-multi-daemon.test.ts
@@ -1,5 +1,5 @@
 import { jest } from '@jest/globals'
-import { Ceramic } from '@ceramicnetwork/core'
+import { Ceramic, VersionInfo } from '@ceramicnetwork/core'
 import { CeramicClient } from '@ceramicnetwork/http-client'
 import * as tmp from 'tmp-promise'
 import { CeramicDaemon } from '../ceramic-daemon.js'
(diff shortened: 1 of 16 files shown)


packages/core/src/ceramic.ts
```diff
@@ -535,11 +547,10 @@ export class Ceramic implements StreamReaderWriter, StreamStateLoader {
   }
 
   async _publishVersionMetrics() {
-    const ipfsVersion = (await this.ipfs.id()).agentVersion
     Metrics.observe(VERSION_INFO, 1, {
       jsCeramicVersion: this._versionInfo.cliPackageVersion,
       jsCeramicGitHash: this._versionInfo.gitHash,
-      ceramicOneVersion: ipfsVersion,
+      ceramicOneVersion: this._versionInfo.ceramicOneVersion,
```
stbrody, line 553: a small downside of this is that if ipfs/ceramic-one is upgraded out from under a running js-ceramic node, the js-ceramic node will continue to report the old version of ipfs/C1 until js-ceramic restarts.  Since it should be rare to upgrade C1 without also at least restarting js-ceramic, I think this is acceptable.

packages/cli/src/__tests__/ceramic-multi-daemon.test.ts
```diff
@@ -16,16 +16,20 @@ const seed = 'SEED'
...
-      db: `sqlite://${stateStoreDirectory}/ceramic.sqlite`,
-      disableComposedb: false,
+  const core = await Ceramic.create(
+    ipfs,
+    {
+      pubsubTopic: TOPIC,
+      stateStoreDirectory,
+      anchorOnRequest: false,
+      indexing: {
+        db: `sqlite://${stateStoreDirectory}/ceramic.sqlite`,
```
gvelez17, line 26: is sqlite hardcoded in?  just a side question

packages/core/src/__tests__/ceramic-pinning.test.ts
```diff
@@ -47,18 +47,22 @@ const createCeramic = async (
...
+    {
+      stateStoreDirectory,
+      anchorOnRequest,
+      indexing: {
+        db: connectionString.href,
+        allowQueriesBeforeHistoricalSync: true,
+        disableComposedb: true,
+        enableHistoricalSync: false,
+      },
+      pubsubTopic: '/ceramic/inmemory/test', // necessary so Ceramic instances can talk to each other
```
gvelez17, line 61: would like to see a constant for this

packages/cli/src/__tests__/ceramic-multi-daemon.test.ts
stbrody: This is a test file, so yes this test always uses sqlite

packages/core/src/ceramic.ts
gvelez17: yeah i think thats a really weird edge case, and we can also tell when it happens.  i think this metric is fine

packages/core/src/__tests__/ceramic-pinning.test.ts
stbrody: this is just a test file, the value used here could be anything

gvelez17 (approved):
had a small change request but not critical, approved

smrz2001 (approved):
:shipit:

PR #3248, opened 2024-06-23
fix(ceramic-node): handle case where the admin DID is not authenticated

gvelez17:
## Description

When running with simple deploy, we receive the error

```
Error: DID is not authenticated
    at get id [as id]
   ```
   
  handle this case and report the error, but do not fail to start

## How Has This Been Tested?

- [x ] Unit tests

## PR checklist

## References:

### DIFF
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -346,8 +346,9 @@ export class CeramicDaemon {
     // publishing metrics is enabled by default, even if no metrics config
     if (!opts.metrics || opts.metrics?.metricsPublisherEnabled) {
       const ipfsVersion = await ipfs.version()
+      let didId: string | null = null;
       try {
-        const didId = did.id
+        didId = did.id
       } catch (err: any) {
         diagnosticsLogger.imp(
           `Unable to publish node metrics without an authenticated DID.  If you wish your ceramic node to be able to publish metrics a privateSeedUrl is required. Error: ${err.toString()}`
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -342,11 +342,11 @@ export class CeramicDaemon {
     const daemon = new CeramicDaemon(ceramic, opts)
     await daemon.listen()
 
-    // Now that ceramic node is set up we can start publishing metrics
-    // publishing metrics is enabled by default, even if no metrics config
-    if (!opts.metrics || opts.metrics?.metricsPublisherEnabled) {
-      const ipfsVersion = await ipfs.version()
-      if (did.authenticated) {
+    if (did.authenticated) {
+      // If authenticated into the node, we can start publishing metrics
+      // publishing metrics is enabled by default, even if no metrics config
+      if (!opts.metrics || opts.metrics?.metricsPublisherEnabled) {
+        const ipfsVersion = await ipfs.version()
         NodeMetrics.start({
           ceramic: ceramic,
           network: params.networkOptions.name,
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages/cli/src/ceramic-daemon.ts [hunk 1]
```diff
@@ -344,24 +344,34 @@ export class CeramicDaemon {
...
-      })
-      diagnosticsLogger.imp(
-        `Publishing Node Metrics publicly to the Ceramic Network.  To learn more, including how to disable publishing, please see the NODE_METRICS.md file for your branch, e.g. https://github.com/ceramicnetwork/js-ceramic/blob/develop/docs-dev/NODE_METRICS.md`
-      )
+      let didId: string | null = null;
+      try {
+        didId = did.id
+      } catch (err: any) {
+        diagnosticsLogger.imp(
+          `Unable to publish node metrics without an authenticated DID.  If you wish your ceramic node to be able to publish metrics a privateSeedUrl is required. Error: ${err.toString()}`
```
stbrody, line 354: `privateSeedUrl` is required, the Ceramic daemon won't start up without it.  You need the node DID to authenticate requests to the CAS.  So there's no way the node could not have a DID.  I think there must be something else going on here.

gvelez17: I don't know, but in fact it is not present, and if i add it, then it doesn't fail

`https://github.com/ceramicstudio/simpledeploy/blob/main/k8s/base/composedb/daemon-config.json#L25`

packages/cli/src/ceramic-daemon.ts
```diff
@@ -344,24 +344,34 @@ export class CeramicDaemon {
...
-        nodeName: '', // daemon.hostname is not useful
-        nodeAuthDID: did.id,
-        nodeIPAddr: '', // daemon.hostname is not the external name
-        nodePeerId: ipfsId.publicKey,
-        logger: diagnosticsLogger,
-      })
-      diagnosticsLogger.imp(
-        `Publishing Node Metrics publicly to the Ceramic Network.  To learn more, including how to disable publishing, please see the NODE_METRICS.md file for your branch, e.g. https://github.com/ceramicnetwork/js-ceramic/blob/develop/docs-dev/NODE_METRICS.md`
-      )
+      let didId: string | null = null;
```
smrz2001, line 349: Seems like we can simplify this logic. Can we just check if `did.id` is set and then use that in the `if` check below?

gvelez17: no, because the getter for did.id is what throws an error

gvelez17: @smrz2001 this is where the error was being thrown, I believe we have to catch it: https://github.com/ceramicnetwork/js-did/blob/main/packages/dids/src/did.ts#L192

smrz2001: Can we use the `authenticated` method instead?

packages/cli/src/ceramic-daemon.ts
```diff
@@ -344,24 +344,26 @@ export class CeramicDaemon {
...
-        nodeName: '', // daemon.hostname is not useful
-        nodeAuthDID: did.id,
-        nodeIPAddr: '', // daemon.hostname is not the external name
-        nodePeerId: ipfsId.publicKey,
-        logger: diagnosticsLogger,
-      })
-      diagnosticsLogger.imp(
-        `Publishing Node Metrics publicly to the Ceramic Network.  To learn more, including how to disable publishing, please see the NODE_METRICS.md file for your branch, e.g. https://github.com/ceramicnetwork/js-ceramic/blob/develop/docs-dev/NODE_METRICS.md`
-      )
+      if (did.authenticated) {
```
stbrody, line 349: nice cleanup, I do still think logging a warning when there isn't a node DID is a good idea though.

stbrody (approved):
lgtm

smrz2001 (approved):
🥳

gvelez17:
huh i thought i saw where @stbrody you wanted a warning about not being authenticated, so i changed it to add the warning - may merge on prior approvals tho since i want to get this in

PR #3249, opened 2024-06-27
chore: unskip ceramic feed test

stephhuynh18:


JulissaDantes (approved):
Amazing!

packages/core/src/__tests__/ceramic-feed.test.ts
```diff
@@ -96,8 +96,7 @@ describe('Ceramic feed', () => {
     await doneStreaming
   })
 
-  // TODO (dav1do): Need to understand why this test is failing with callstack overflow with ceramic one.
-  test.skip('add entry after anchoring stream', async () => {
+  test('add entry after anchoring stream', async () => {
```
stbrody, line 99: do we know why it was failing in the first place or does enabling this risk the original flaky failure coming back?

stephhuynh18: @JulissaDantes mind if you answer this

JulissaDantes: No we don't know the exact reason. But there had been some enhancements and we want to check how it behaves with the latest changes in the repo

stbrody: okay that's fine, lets just plan to keep a very close eye on the tests after this merges since it seems likely the old issue will come back and cause some instability for our build that we'll want to fix ASAP.

PR #3250, opened 2024-07-01
chore: (breaking changes) use rust "ipfs" by default

dav1do:
## Description
This changes the daemon to use `IPFS_FLAVOR=rust` by default. It will start a binary if a `CERAMIC_ONE_PATH` variable is set. The docker image installs the binary and sets the path by default.

### Details

This also removes the `CERAMIC_RECON_MODE` environment variable and only uses `IPFS_FLAVOR` and changes the default from `go` to `rust`. Now, running with `CERAMIC_ONE_PATH=/Users/david/3box/rust-ceramic/target/debug/ceramic-one node ceramic.js daemon` will use a binary rust ceramic and `node ceramic.js daemon --ipfs-api http://localhost:5101` will use a remote rust ceramic endpoint. You can change the network and store directory as well: `CERAMIC_ONE_NETWORK=testnet-clay CERAMIC_ONE_STORE_DIR=~/.ceramic-one CERAMIC_ONE_PATH=/Users/david/3box/rust-ceramic/target/debug/ceramic-one node ceramic.js daemon`.

The defaults match the previous behavior to use a random directory for the store, an in-memory network, and random ports for the server. If you change the network, you must specify the store. This was to keep tests passing but ideally they're updated to pipe desired behavior in and have more repeatable defaults. 

To use kubo, you must specify `IPFS_FLAVOR=go node ceramic.js daemon`

The docker image installs a rust binary and sets default env variable values so they work and you can simply start the container with `docker run --platform=linux/x86_64 js-ceramic:dev`. `--platform` is not strictly necessary, however if you're running on an ARM cpu (e.g. apple silicon) you will get a warning and it will switch and run using emulation as we don't provide an image with the ARM version of rust ceramic currently.

We [changed](https://github.com/ceramicnetwork/rust-ceramic/pull/411) the default RPC port ceramic one uses to 5101 so this also makes that the default when not included. See the [ticket in linear](https://linear.app/3boxlabs/issue/AES-193/stretch-change-default-port-for-c1). 

### Cruft
Nothing was previously passed down to the rust ceramic IPFS info when it wasn't running in remote mode and we relied on starting with all the defaults in tests. In order to avoid changing everything here, I kept that behavior but the network is not passed through from the CLI arg at the moment. I intend to fix that in the next PR when I detect the version of a remote node rather than requiring flags.

## How Has This Been Tested?

- [x] CI ... possibly more to do still

## PR checklist

- [x] I have made corresponding changes to the documentation. Most of the documentation still refers to IPFS pinning and is specific, so I generally left it alone.

## References:


stbrody (commented):
this is a breaking change, so I think we'll want to release it alongside the other breaking changes we are planning for js-ceramic, along with a bump in the major version of the packages.

I think we should probably create a branch in the js-ceramic repo to stage all the breaking changes until we're they're all ready and then merge them all into develop at once and make the RC with the major version bump.  See https://linear.app/3boxlabs/issue/AES-188/parent-js-ceramic-breaking-changes-for-c1-ga-release for other breaking changes.

nathanielc (approved):

stephhuynh18 (approved):

packages/common/src/utils/environment-utils.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Environment related utils, that is information about the mode and system we're operating in.
+ */
+export class EnvironmentUtils {
+  /**
+   * Returns whether or not we're running using rust-ceramic
+   */
+  static useRustCeramic(): boolean {
+    return process.env.IPFS_FLAVOR !== 'go'
```
stbrody, line 9: now that we're making rust-ceramic the default for unit tests, we should make sure that just running `npm test` on a local checkout of js-ceramic still works without any extra env vars.  That probably means setting some expectation around the default location of ceramic-one, or having it in your PATH

packages/core/src/ceramic.ts
```diff
@@ -567,8 +568,10 @@ export class Ceramic implements StreamReaderWriter, StreamStateLoader {
       )
     }
 
-    if (process.env.CERAMIC_RECON_MODE) {
-      this._logger.warn(`Running Ceramic in v4 mode. This mode is still experimental.`)
+    if (EnvironmentUtils.useRustCeramic()) {
+      this._logger.warn(
+        `Running Ceramic using rust-ceramic. If you want to use Ipfs, run with IPFS_FLAVOR=go.`
```
stbrody, line 573: I think since using ceramic-one is the default now we shouldn't log a warning when doing so.  We should invert this condition and log a warning if still using kubo that you're on a deprecated configuration.

stbrody: also nitpick, but the product is `ceramic-one` not `rust-ceramic`.  I think we should always be referring to it as `ceramic-one` in all of our public facing docs and material.  CC @JustinaPetr in case there's disagreement there (wanted to tag Chris Jones also but I don't know if he's on github?)

stbrody (approved):

PR #3251, opened 2024-07-02
chore: add metrics for create stream endpoint

stephhuynh18:


packages/cli/src/ceramic-daemon.ts
```diff
@@ -57,6 +57,10 @@ const ADMIN_CODE_CACHE_CAPACITY = 50
 const STREAM_PINNED = 'stream_pinned'
 const STREAM_UNPINNED = 'stream_unpinned'
 const ERROR_QUERYING_COLLECTION = 'error_querying_collection'
+const METRIC_CREATE_STREAM_SPAN = 'http_request_create_stream_span'
```
gvelez17, line 60: i like to name them exactly the same as their string, but that's just a personal preference, i think this is fine

packages/cli/src/ceramic-daemon.ts
```diff
@@ -524,13 +528,33 @@ export class CeramicDaemon {
...
     const stream = await this.ceramic.createStreamFromGenesis(
       type,
       StreamUtils.deserializeCommit(genesis),
       opts
     )
     res.json({ streamId: stream.id.toString(), state: StreamUtils.serializeState(stream.state) })
+
+    res.removeListener('close', recordClosedAt)
+    const now = new Date()
+    const completedAt = now.getTime() - start.getTime()
```
gvelez17, line 551: semantically, i would say createDuration instead of completedAt here

gvelez17 (approved):
one semantic nit, approving the function is good, if you change that one var name maybe i would still consider approved

gvelez17:
@stephhuynh18 did you still want this?

packages/cli/src/ceramic-daemon.ts
samika98: I think it would be nice to have this metric record how long a create stream call takes. What's the best way to visualize this on Grafana @gvelez17 @stephhuynh18

PR #3253, opened 2024-07-09
chore: detect remote ipfs type and adjust accordingly

dav1do:
## Description

We now detect the remote IPFS node type and adjust the daemon config (really the environment variables) accordingly. This also now respects the --network CLI flag passed in when starting a ceramic one binary automatically.

[Linear ticket](https://linear.app/3boxlabs/issue/AES-159/js-ceramic-should-error-if-not-connected-to-c1-node).

## How Has This Been Tested?

- [x] Connected to remote ceramic one/kubo nodes with wrong IPFS_FLAVOR variable

## PR checklist

## References:

### DIFF
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -320,7 +320,7 @@ export class CeramicDaemon {
         )
       } else {
         diagnosticsLogger.warn(
-          `Igoring configuration IPFS_FLAVOR='go'. Detected remote ceramic-one running at ${ipfsId.addresses} with agent version ${ipfsId.agentVersion}. If you intended to use ipfs, specify an --ipfs-url for an ipfs node.`
+          `Ignoring configuration IPFS_FLAVOR='go'. Detected remote ceramic-one running at ${ipfsId.addresses} with agent version ${ipfsId.agentVersion}. If you intended to use ipfs, specify an --ipfs-url for an ipfs node.`
         )
       }
     }
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -289,9 +289,7 @@ export class CeramicDaemon {
     const desiredIpfsClient = EnvironmentUtils.useRustCeramic() ? 'ceramic-one' : 'kubo'
     const wrongIpfsClient =
       opts.ipfs.mode == IpfsMode.REMOTE &&
-      EnvironmentUtils.useRustCeramic() &&
-      !ipfsId.agentVersion.includes('ceramic-one')
-
+      EnvironmentUtils.useRustCeramic() !== ipfsId.agentVersion.includes('ceramic-one')
     if (wrongIpfsClient) {
       EnvironmentUtils.setIpfsFlavor(EnvironmentUtils.useRustCeramic() ? 'go' : 'rust')
     }
@@ -316,9 +314,15 @@ export class CeramicDaemon {
         .join(', ')}`
     )
     if (wrongIpfsClient) {
-      diagnosticsLogger.warn(
-        `The daemon expected the remote IPFS node to be running ${desiredIpfsClient} but it was running ${ipfsId.agentVersion}. The daemon config has been modified.`
-      )
+      if (desiredIpfsClient === 'ceramic-one') {
+        diagnosticsLogger.warn(
+          `Detected ipfs kubo running at ${ipfsId.addresses} with agent version ${ipfsId.agentVersion}. Using kubo for p2p networking is deprecated - we recommend migrating to ceramic-one instead.`
+        )
+      } else {
+        diagnosticsLogger.warn(
+          `Igoring configuration IPFS_FLAVOR='go'. Detected remote ceramic-one running at ${ipfsId.addresses} with agent version ${ipfsId.agentVersion}. If you intended to use ipfs, specify an --ipfs-url for an ipfs node.`
+        )
+      }
     }
 
     const ceramic = new Ceramic(modules, params)
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -9,6 +9,7 @@ import { DEFAULT_PUBLISH_INTERVAL_MS, NodeMetrics, Observable } from '@ceramicne
 import { IpfsConnectionFactory } from './ipfs-connection-factory.js'
 import {
   DiagnosticsLogger,
+  EnvironmentUtils,
   FieldsIndex,
   LoggerProvider,
   LogStyle,
@@ -284,6 +285,17 @@ export class CeramicDaemon {
     )
     const ipfsId = await ipfs.id()
 
+    // we need to change the values before processing the config, so we keep track to log the warning after we have logging set up
+    const desiredIpfsClient = EnvironmentUtils.useRustCeramic() ? 'ceramic-one' : 'kubo'
+    const wrongIpfsClient =
+      opts.ipfs.mode == IpfsMode.REMOTE &&
+      EnvironmentUtils.useRustCeramic() &&
+      !ipfsId.agentVersion.includes('ceramic-one')
+
+    if (wrongIpfsClient) {
+      EnvironmentUtils.setIpfsFlavor(EnvironmentUtils.useRustCeramic() ? 'go' : 'rust')
+    }
+
     const versionInfo = {
       cliPackageVersion: version,
       gitHash: commitHash,
@@ -303,6 +315,11 @@ export class CeramicDaemon {
         .map(String)
         .join(', ')}`
     )
+    if (wrongIpfsClient) {
+      diagnosticsLogger.warn(
+        `The daemon expected the remote IPFS node to be running ${desiredIpfsClient} but it was running ${ipfsId.agentVersion}. The daemon config has been modified.`
+      )
+    }
 
     const ceramic = new Ceramic(modules, params)
     let didOptions: DIDOptions = { resolver: makeResolvers(ceramic) }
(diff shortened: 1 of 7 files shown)


packages/core/src/ceramic.ts
```diff
@@ -500,10 +500,11 @@ export class Ceramic implements StreamReaderWriter, StreamStateLoader {
    */
   async _init(doPeerDiscovery: boolean): Promise<void> {
     try {
-      this._logger.imp(
-        `Connecting to ceramic network '${this._networkOptions.name}' using pubsub topic '${this._networkOptions.pubsubTopic}'`
-      )
-
+      if (!EnvironmentUtils.useRustCeramic()) {
```
dav1do, line 503: not sure if we want to keep logging this in all cases or change how we try to drive/force the bootstrapping from js-ceramic when running in remote mode

packages/core/src/ceramic.ts
```diff
@@ -569,8 +570,8 @@ export class Ceramic implements StreamReaderWriter, StreamStateLoader {
     }
 
     if (EnvironmentUtils.useRustCeramic()) {
-      this._logger.warn(
-        `Running Ceramic using rust-ceramic. If you want to use Ipfs, run with IPFS_FLAVOR=go.`
+      this._logger.imp(
+        `Running Ceramic using ceramic-one (rust-ceramic). To use IPFS (kubo), specify a --ipfs-url or run with IPFS_FLAVOR=go.`
```
smrz2001, line 574: Should this be `and` since `IPFS_FLAVOR` has to be specified to run with Kubo regardless of whether the URL is specified?
```suggestion
        `Running Ceramic using ceramic-one (rust-ceramic). To use IPFS (kubo), specify a --ipfs-url and run with IPFS_FLAVOR=go.`
```

dav1do: Right now it actually isn't required. IPFS_FLAVOR only matters for the binary with these changes, and if you're running a remote node we simply log a warning if they don't match and change the value so the rest of the process does recon feed/pubsub appropriately.

Sorry, I pretty much swapped everything else out from under you because I didn't like how required parameters were being sent is as undefined.

packages/cli/src/ceramic-daemon.ts
```diff
@@ -284,6 +285,17 @@ export class CeramicDaemon {
...
 
+    // we need to change the values before processing the config, so we keep track to log the warning after we have logging set up
+    const desiredIpfsClient = EnvironmentUtils.useRustCeramic() ? 'ceramic-one' : 'kubo'
+    const wrongIpfsClient =
+      opts.ipfs.mode == IpfsMode.REMOTE &&
+      EnvironmentUtils.useRustCeramic() &&
+      !ipfsId.agentVersion.includes('ceramic-one')
+
+    if (wrongIpfsClient) {
+      EnvironmentUtils.setIpfsFlavor(EnvironmentUtils.useRustCeramic() ? 'go' : 'rust')
```
smrz2001, line 296: We definitely want to toggle the flavor instead of erroring out? I'd be worried about an inconsistency in the config resulting in unexpected behavior.

In fact, if someone were to upgrade to this version of the code on an existing remote setup using Kubo, Ceramic will try to communicate with Kubo as if it is C1 because they might not have set `IPFS_FLAVOR=go` explicitly.

dav1do: I don't think that's true.. but I could very well be wrong 😅. As of #3250, IPFS_FLAVOR is rust by default. If we specify a remote with `--ipfs-url` in this branch, we'll detect that rust != kubo, log a warning and use kubo IPFS.

dav1do: This is targeting #3250 currently rather than develop.

packages/cli/src/ceramic-daemon.ts
```diff
@@ -284,6 +285,17 @@ export class CeramicDaemon {
     )
     const ipfsId = await ipfs.id()
 
+    // we need to change the values before processing the config, so we keep track to log the warning after we have logging set up
+    const desiredIpfsClient = EnvironmentUtils.useRustCeramic() ? 'ceramic-one' : 'kubo'
```
stbrody, line 289: does this need to happen in the ceramic-daemon or can it be pushed down into ceramic.ts?

packages/cli/src/ceramic-daemon.ts
```diff
@@ -303,6 +315,11 @@ export class CeramicDaemon {
         .map(String)
         .join(', ')}`
     )
+    if (wrongIpfsClient) {
+      diagnosticsLogger.warn(
+        `The daemon expected the remote IPFS node to be running ${desiredIpfsClient} but it was running ${ipfsId.agentVersion}. The daemon config has been modified.`
```
stbrody, line 320: I think this is kind of a confusing warning for regular users.  Especially if they didn't explicitly set any "desired" ipfs client but are just using all the defaults.

Can we change this to something more along the lines of "Detected ipfs kubo running at ${url} with agent version ${agentVersion}.  Using kubo for p2p networking is deprecated - we recommend migrating to ceramic-one instead."

packages/cli/src/ceramic-daemon.ts
stbrody: It seems like there should be a distinction between if you're using the default behavior vs if you set IPFS_FLAVOR explicitly.  If the env var is set explicitly and differs from what is detected, that feels like it should fail startup.  Whereas if it's unset and you're using the defaults, then adjusting to whatever is detected makes sense.

packages/common/src/utils/environment-utils.ts
```diff
@@ -8,4 +8,11 @@ export class EnvironmentUtils {
   static useRustCeramic(): boolean {
     return process.env.IPFS_FLAVOR !== 'go'
   }
+
+  /**
+   * Changes the server from expecting rust or go ipfs to the other
+   */
+  static setIpfsFlavor(flavor: 'go' | 'rust') {
+    process.env.IPFS_FLAVOR = flavor
```
stbrody, line 16: I don't actually know enough about Node to know if this is a good or bad idea, but instead of overriding the env var could we have a file-local static variable we override instead and that takes priority in the `useRustCeramic` call above?

packages/core/src/ceramic.ts
stbrody: I definitely think we should still log what network we are on.  So maybe if we're using ceramic-one we log a slightly different message that says the network and says that we are using Recon for data synchronization or something like that.

packages/core/src/ceramic.ts
stbrody, line 574: is this message right?  I don't think just specifying `--ipfs-url` makes you use kubo, if the url you specify points at a ceramic-one node.

TBH I would vote we just remove all reference to ipfs/kubo/go entirely in the default path.  If they're already on ceramic-one then they're where we want them, why tell them how to move backwards?

I would, however, like a warning if you're still on kubo that you're on a deprecated config and should move forward to ceramic-one.

stbrody (commented):
mostly suggestions around logging

packages/core/src/ceramic.ts
dav1do: I log a message but it could be wrong if the c1 node is remote as it's running whatever network it started on.. I could probably figure out a way to pipe it all the way down but I don't see that we know or care normally, since the ipfs node is always remote from our perspective here.

packages/cli/src/ceramic-daemon.ts
dav1do: Improved the warnings. If you specify `IPFS_FLAVOR=go` with a kubo `--ipfs-url` there is no warning, otherwise there is. We log a warning if you specified flavor=go and talked to c1 with a url, but we can just "make it work" so I'm not sure there's a need to crash. We can change the warning to an error though.

packages/cli/src/ceramic-daemon.ts
dav1do: adjusted to warnings

packages/common/src/utils/environment-utils.ts
dav1do: it worked fine but setting an env var is maybe bad form so switched to a file variable

packages/cli/src/ceramic-daemon.ts
dav1do: it needs to happen here because we have to change it before we process the config otherwise we get misconfigured pubsub/recon things and we don't tell ceramic whether we expected remote or not currently, which could probably change as it would improve the messages later, but I tried not to change more than I needed

dav1do: we could probably update the `CeramicConfig` object to include the information if we want to go that route

stbrody: this is fine

dav1do: as discussed, we actually don't care about remote at this point because it's always remote, but `_processConfig` requires the variable to be set correctly for pubsub and recon things to work, so it would have to happen in there (or use `create` from the daemon) which probably isn't worth it right now

packages/cli/src/ceramic-daemon.ts
```diff
@@ -303,6 +313,17 @@ export class CeramicDaemon {
...
         .join(', ')}`
     )
+    if (wrongIpfsClient) {
+      if (desiredIpfsClient === 'ceramic-one') {
+        diagnosticsLogger.warn(
+          `Detected ipfs kubo running at ${ipfsId.addresses} with agent version ${ipfsId.agentVersion}. Using kubo for p2p networking is deprecated - we recommend migrating to ceramic-one instead.`
+        )
+      } else {
+        diagnosticsLogger.warn(
+          `Igoring configuration IPFS_FLAVOR='go'. Detected remote ceramic-one running at ${ipfsId.addresses} with agent version ${ipfsId.agentVersion}. If you intended to use ipfs, specify an --ipfs-url for an ipfs node.`
```
stbrody, line 323:
```suggestion
          `Ignoring configuration IPFS_FLAVOR='go'. Detected remote ceramic-one running at ${ipfsId.addresses} with agent version ${ipfsId.agentVersion}. If you intended to use ipfs, specify an --ipfs-url for an ipfs node.`
```

stbrody (approved):

PR #3257, opened 2024-07-12
chore: Push metrics handling down from cli package into core package instead

stbrody:
Should be no functional changes, just code motion.

Doing this to make https://linear.app/3boxlabs/issue/AES-206/re-enable-nodemetrics-by-default easier - since we need to be able to subscribe the node to the NodeMetrics data model, which can only be done from within Ceramic core, there's no api to do that in the Daemon.

Also it just feels like this is where this code belongs anyway.


gvelez17 (approved):
This looks ok, I did not test it, but the logic looks good to me

PR #3258, opened 2024-07-16
chore: cleanup related to starting/stopping ceramic-one from daemon

dav1do:
## Description

- We now call ipfs.stop() for rust ceramic to make sure it gets shutdown when something goes wrong during setup. This required adjusting the override for the remote IPFS client as there is no shutdown endpoint on ceramic-one like kubo.
- We now default to a standard (not hidden) folder called `ceramic-one` in the local directory as the default store-dir. 
- We now force shutdown ipfs/c1 if we fail to send SIGTERM. The messaging around binary shutdown errors were adjusted as well. The acutal issue re-binding ports was due to ceramic-one shutdown and should be fixed in [rust-ceramic #440](https://github.com/ceramicnetwork/rust-ceramic/pull/440).

## How Has This Been Tested?

Local testing with [rust-ceramic #440](https://github.com/ceramicnetwork/rust-ceramic/pull/440) and CI.

## PR checklist

## References:

### DIFF
--- a/packages/cli/src/ipfs-connection-factory.ts
+++ b/packages/cli/src/ipfs-connection-factory.ts
@@ -1,6 +1,6 @@
 import mergeOpts from 'merge-options'
 import * as ipfsClient from 'ipfs-http-client'
-import { DiagnosticsLogger, EnvironmentUtils, IpfsApi, Networks } from '@ceramicnetwork/common'
+import { DiagnosticsLogger, EnvironmentUtils, IpfsApi } from '@ceramicnetwork/common'
 import { IpfsMode } from './daemon-config.js'
 import * as http from 'http'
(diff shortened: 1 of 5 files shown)


dav1do:
This isn't ready yet, there is a case where the repository can fail to start up which doesn't call shutdown on the binary so it continues running in the background. Need to review some more to figure out how to correct that.

dav1do:
> This isn't ready yet, there is a case where the repository can fail to start up which doesn't call shutdown on the binary so it continues running in the background. Need to review some more to figure out how to correct that.

Should be handled in [490e8b](https://github.com/ceramicnetwork/js-ceramic/pull/3258/commits/490e8b4c8bccbaaf9b7967dadd2691509106a2e7).

PR #3260, opened 2024-07-17
fix: Subscribe to node metrics model before publishing to it

stbrody:
Testing locally, starting the NodeMetrics publisher still fails.  It's failing because the node doesn't have the NodeMetrics data model stored locally in its block store.  We need to figure out how to make the NodeMetrics model available to nodes before this will work.  See https://linear.app/3boxlabs/issue/AES-233/make-nodemetrics-model-available-to-c1-nodes for more info.


dav1do (approved):

packages/core/src/ceramic.ts
```diff
@@ -231,6 +231,7 @@ export class Ceramic implements StreamReaderWriter, StreamStateLoader {
   private _signer: CeramicSigner
   public readonly dispatcher: Dispatcher
   public readonly loggerProvider: LoggerProvider
+  public readonly recon: IReconApi
```
gvelez17, line 234: um if we role back the node metrics changes we probably don't want to roll this stuff back?

packages/core/src/ceramic.ts
```diff
@@ -614,6 +615,12 @@ export class Ceramic implements StreamReaderWriter, StreamStateLoader {
       // If authenticated into the node, we can start publishing metrics
       // publishing metrics is enabled by default, even if no metrics config
       if (this._metricsConfig?.metricsPublisherEnabled) {
+        // First, subscribe the node to the Model used for NodeMetrics
+        const metricsModel = NodeMetrics.getModel(this._networkOptions.name)
+        await this.repository.index.indexModels([{ streamID: metricsModel }])
```
gvelez17, line 620: is there a possible failure here?  what if we fail to getModel, and its null, will this throw an error?

packages/core/src/recon.ts
```diff
@@ -57,7 +57,7 @@ export interface ReconEventFeedResponse {
  */
 export interface IReconApi extends Observable<ReconEventFeedResponse> {
   init(initialCursor?: string, initialInterests?: Array<StreamID>): Promise<void>
-  registerInterest(model: StreamID): Promise<void>
+  registerInterest(model: StreamID, controller?: string): Promise<void>
```
gvelez17, line 60: this should probably be kept separate

gvelez17 (changes requested):
i'd like to separate out the NodeMetrics changes so we don't break other stuff if we roll them back, and also make sure we are doing a try..catch around any NodeMetrics stuff since its not essential and shouldn't stop the node from running

tight try-catch around each error source is best if we can

packages/core/src/ceramic.ts
stbrody: Yes it will, would prefer to do all error handling cleanup over in https://github.com/ceramicnetwork/js-ceramic/pull/3265

packages/core/src/recon.ts
stbrody: I don't really see why tbh?  It's a very small change

packages/core/src/ceramic.ts
stbrody: why wouldn't we?  We only want this for NodeMetrics anyway.

gvelez17 (approved):

packages/core/src/recon.ts
gvelez17: yeah its an important one tho and if we wanted to roll back node metrics we wouldn't want to lose dids as a part of interests...?

stbrody: not sure what you mean by "lose dids as part of interests".  This is just a small change to the client to expose a feature of C1 that exists either way.  Right now the NodeMetrics are the only consumer of this API, nothing else is using it so it'd be fine roll back if the NodeMetrics rolls back.

PR #3261, opened 2024-07-17
chore: change ipfs default to remote mode and remove bundled ceramic-one

dav1do:
## Description

The default IPFS mode is now remote with localhost:5101 (ceramic-one). We no longer support starting the ceramic-one binary automatically in bundled mode outside of tests. It didn't provide a super elegant startup as we don't have a good cross platform install method for node so couldn't guarantee the defaults and it isn't how we want people to run things anyway. It also had some issues stopping, especially when errors occur during start up and rather than rework that handling, we're moving away from the option.

This changes the default config options to remote ceramic one, so it doesn't "just work" if you run `node ceramic.js daemon` like it would before. If you previously had bundled mode in your config, you will get an error saying it's not supported and you should pass a url.

## Note

The local CI still passes as we're starting in bundled mode for the tests, however, the daemon is no longer able to do that so the verifyImage task fails. It will require validating and merging the changes in [pipeline-tools](https://github.com/3box/pipeline-tools/pull/65) first before this is fully green.

## How Has This Been Tested?

Ran daemon locally and CI.

## PR checklist

## References:

### DIFF
--- a/packages/cli/src/ceramic-daemon.ts
+++ b/packages/cli/src/ceramic-daemon.ts
@@ -258,7 +258,7 @@ export class CeramicDaemon {
       ipfsId = await ipfs.id()
     } catch (err) {
       throw new Error(
-        `Error connecting to p2p node. Make sure ceramic-one is running or use --ipfs-api modify the expected http location (currently ${opts.ipfs?.host}): ${err}`
+        `Error connecting to p2p node. Make sure ceramic-one is running or use --ipfs-api to modify the expected http location (currently ${opts.ipfs?.host}): ${err}`
       )
     }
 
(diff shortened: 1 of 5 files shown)


packages/cli/src/ceramic-daemon.ts
```diff
@@ -253,7 +253,14 @@ export class CeramicDaemon {
       ceramicConfig.loggerProvider.getDiagnosticsLogger(),
       opts.ipfs?.host
     )
-    const ipfsId = await ipfs.id()
+    let ipfsId
+    try {
+      ipfsId = await ipfs.id()
+    } catch (err) {
+      throw new Error(
+        `Error connecting to p2p node. Make sure ceramic-one is running or use --ipfs-api modify the expected http location (currently ${opts.ipfs?.host}): ${err}`
```
stbrody, line 261:
```suggestion
        `Error connecting to p2p node. Make sure ceramic-one is running or use --ipfs-api to modify the expected http location (currently ${opts.ipfs?.host}): ${err}`
```

stbrody (approved):

PR #3263, opened 2024-07-18
chore: remove dagger image verify step from CI 

dav1do:
## Description

We're moving away from dagger and using k8s deployments and removing the verify step for now. It needs updating in order to support the remote ipfs requirement in #3261. Rather than change it, we remove it for now and replace it in a follw up.

## PR checklist

## References:


smrz2001 (approved):
🚀

PR #3265, opened 2024-07-19
feat: wait for metrics model to be available locally

stbrody:
Builds on https://github.com/ceramicnetwork/js-ceramic/pull/3260/

When starting up a clean js-ceramic node for the first time with node metrics publishing enabled, the node needs to wait for the NodeMetrics model to be synced via recon and made available locally before it tries to index it.  Example out from a clean node as it waits for this to happen below:

```
[2024-07-19T18:56:30.467Z] IMPORTANT: Connecting to IPFS node with version "ceramic-one/0.28.1" available as /ip4/127.0.0.1/udp/4101/quic-v1/p2p/12D3KooWL9zNBFnNEgEPRCtrzFXX7iyN8vnGWJGX7z6eaFkmdy5e, /ip4/10.0.0.188/udp/4101/quic-v1/p2p/12D3KooWL9zNBFnNEgEPRCtrzFXX7iyN8vnGWJGX7z6eaFkmdy5e, /ip4/127.0.0.1/tcp/4101/p2p/12D3KooWL9zNBFnNEgEPRCtrzFXX7iyN8vnGWJGX7z6eaFkmdy5e, /ip4/10.0.0.188/tcp/4101/p2p/12D3KooWL9zNBFnNEgEPRCtrzFXX7iyN8vnGWJGX7z6eaFkmdy5e
[2024-07-19T18:56:30.469Z] IMPORTANT: Initializing SQLite connection
[2024-07-19T18:56:30.509Z] IMPORTANT: Node DID set to 'did:key:z6MkhXxQCx33gofNSNSRuipu5xr8cNnW2b4aoXebyfG6FXwe'. This DID will be used to authenticate to the anchor service
[2024-07-19T18:56:30.509Z] IMPORTANT: Connecting to ceramic network 'testnet-clay' using ceramic-one with Recon for data synchronization.
[2024-07-19T18:56:30.519Z] IMPORTANT: Creating ComposeDB config table: ceramic_models
[2024-07-19T18:56:30.519Z] IMPORTANT: Creating ComposeDB config table: ceramic_config
[2024-07-19T18:56:30.519Z] IMPORTANT: Creating ComposeDB config table: ceramic_model_implements
[2024-07-19T18:56:30.536Z] IMPORTANT: Connecting to bootstrap peers for network testnet-clay
[...]

PR #3268, opened 2024-07-24
feat: Start up Node Metrics publishing after the Model becomes available locally

stbrody:
New approach instead of https://github.com/ceramicnetwork/js-ceramic/pull/3265.

Now we start a background job that waits forever for the NodeMetrics model to be synced (using exponential backoff, maxing out at checking once per minute).  If the model never becomes available, that job will keep checking for the Model once per minute, and will log an error once every 5 minutes.  If the model ever becomes available, however, it will start up the NodeMetrics publishing subsystem at that point and the background task stops.

### DIFF
--- a/packages/base-test-utils/src/index.ts
+++ b/packages/base-test-utils/src/index.ts
@@ -54,6 +54,7 @@ export class BaseTestUtils {
     throw new Error(baseErrMsg + ': ' + customMsg)
   }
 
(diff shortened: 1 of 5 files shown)


gvelez17 (approved):
Great, yes we should NOT have to registerInterest in the network wide metrics if possible, in order to publish to them!

PR #3269, opened 2024-07-29
test: Unskip a flaky feed API test

ukstv:
Ditto.

### DIFF
--- a/packages/core/src/__tests__/ceramic-feed.test.ts
+++ b/packages/core/src/__tests__/ceramic-feed.test.ts
@@ -141,7 +141,6 @@ describe('Ceramic feed', () => {
       .catch(() => {
         // Abort Signal, hence ignore
       })
-
     // create model on different node
     const model = await Model.create(ceramic2, modelDefinition())
     await doneStreaming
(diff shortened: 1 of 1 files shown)


packages/core/src/__tests__/ceramic-feed.test.ts
```diff
@@ -114,14 +113,14 @@ describe('Ceramic feed', () => {
       return emissions.length >= 4 // Recon gives you more events :(
     }, 5000)
     // model.create
-    expect(emissions[0].commitId).toEqual(model.allCommitIds[0])
+    expect(String(emissions[0].commitId)).toEqual(String(model.allCommitIds[0]))
```
stbrody, line 116: hmm, can you explain what was going wrong here and why casting to a string prevents the stack overflow?

ukstv: That is an experimental (let me be honest here) approach based on some previous experience. Loong time ago we saw something similar with Jest unable in some cases to correctly compare StreamIDs or CommitIDs into.

stbrody: Okay.  Not ideal, but worth giving it a shot and seeing how it fairs

stbrody (approved):

JulissaDantes (approved):
Thank you!

PR #3273, opened 2024-07-30
fix: correctly use controller when registering interest and correctly use node's ID when publishing metrics.

nathanielc:
## Description
In testing the node metrics publish I found two bugs:

1. Metrics interest was for all peers not just itself.
2. The node id was empty causing the published data to fail validation.

This fixes both issues.

The controller needs to be a query parameter not a POST body object.

The `ipfsId.publicKey` was empty I do not know why, however the `ipfsId.id` field was set correctly and works as expected.

## How Has This Been Tested?

Manually tested, starting a node. Previously the node saw errors everytime it tried to publish metrics. Additionally I directly inspected the interest and was able to see it was for all controllers not just itself. I have manually verified the interest is now for the specific controller.

## PR checklist

## References:

### DIFF
--- a/packages/core/src/recon.ts
+++ b/packages/core/src/recon.ts
@@ -171,11 +171,10 @@ export class ReconApi extends Observable<ReconEventFeedResponse> implements IRec
     }
     try {
       const headers = { 'Content-Type': 'application/json' }
-      const body = { ...(controller && { controller }) }
-      await this.#sendRequest(this.#url + `/ceramic/interests/model/${model.toString()}`, {
+      const query = controller ? `controller=${controller}` : ''
+      await this.#sendRequest(this.#url + `/ceramic/interests/model/${model.toString()}?${query}`, {
         method: 'POST',
         headers,
-        body,
       })
       this.#logger.debug(`Recon: added interest for model ${model.toString()}`)
     } catch (err) {
--- a/packages/core/src/recon.ts
+++ b/packages/core/src/recon.ts
@@ -171,10 +171,11 @@ export class ReconApi extends Observable<ReconEventFeedResponse> implements IRec
     }
     try {
       const headers = { 'Content-Type': 'application/json' }
-      const query = controller ? `controller=${controller}` : ''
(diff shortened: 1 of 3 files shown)


packages/core/src/recon.ts
```diff
@@ -169,11 +169,10 @@ export class ReconApi extends Observable<ReconEventFeedResponse> implements IRec
     }
     try {
       const headers = { 'Content-Type': 'application/json' }
-      const body = { ...(controller && { controller }) }
-      await this.#sendRequest(this.#url + `/ceramic/interests/model/${model.toString()}`, {
+      const query = controller ? `controller=${controller}` : ''
```
nathanielc, line 172: Do we have a canonical way of constructing query strings for URLs? This string manipulation is not ideal.

packages/core/src/recon.ts
```diff
@@ -169,11 +169,10 @@ export class ReconApi extends Observable<ReconEventFeedResponse> implements IRec
...
+      await this.#sendRequest(this.#url + `/ceramic/interests/model/${model.toString()}?${query}`, {
```
stbrody, line 173: huh, should this be considered an issue with the API that C1 is exposing?  This is a POST request, why is it using URL query parameters instead of taking its args in the POST body?  Maybe the thing to do is change it in C1 instead of here?

CC @dav1do

dav1do: ah, yeah. that's actually the old one.. we have a newer one `POST /interests` that [behaves](https://github.com/ceramicnetwork/rust-ceramic/blob/89788719e3cffa649a60fb3166e1349537842b9c/api/ceramic.yaml#L232) like that

nathanielc: Query params are not exclusive to get requests. I don't have a strong opinion either way just updated the client to match the server.

nathanielc: @dav1do Thanks for the link to the new endpoint. I changed the code to use the new endpoint and manually tested. Working well...

dav1do (approved):

PR #3274, opened 2024-07-30
fix: adjust tests for failure to start due to network mismatch

dav1do:
## Description

Related to the new network startup check for ceramic one. That error happens earlier than the cas error was, so we get a different error. On kubo, we get the old error. We also check a few fields are not null when cleaning up in case we somehow didn't manage to call init() due to an error in `ceramic._init()` but we still try to stop everything.

- [x] Ran tests locally with c1

## PR checklist

## References:


smrz2001 (approved):
:shipit:

stbrody (approved):

PR #3277, opened 2024-09-09
fix: Don't remove entries from AnchorRequestStore if applying the anchor commit fails

stbrody:
~Needs testing~ steph: Tested manually using both ipfs + c1, and added a unit test

This has the downside that if the reason applying the anchor failed is *not* transient, like if the TimeEvent is malformed or otherwise invalid, we will still leave the event in the store and keep retrying it.  I think that is probably fine for now since we don't expect to be getting any fundamentally invalid TimeEvents from the CAS.  Self-anchoring in C1 should be able to be more intelligent in understanding what types of failure modes are transient vs irrecoverable.


samika98 (approved):
LGTM

PR #3280, opened 2024-09-24
fix: Ensure loading a stream at a CommitID of an anchor can repair a stream with a CACAO error

stbrody:


packages/core/src/state-management/repository.ts
```diff
@@ -487,28 +496,40 @@ export class Repository {
     // We also skip CACAO expiration checking during this initial load as its possible
     // that the CommitID we are being asked to load may in fact be an anchor commit with
     // the timestamp information that will reveal to us that the CACAO didn't actually expire.
-    const base$ = await this.load(commitId.baseID, opts, false)
```
stbrody, line 490: this was the issue.  The "false" in the third arg here was to `skipCacaoExpiration` checks, and if the node already has the state stored locally, that worked.  But if the node didn't already have the stream in its state store, and the genesis commit has an expired CACAO, this will fail with a CACAO error when trying to build the initial state from the genesis commit.

The "right" fix would be to thread the `skipCacaoExpiration` down all the way through all the layers of commit application, so that at the point where the genesis commit is applied we can inform it to not worry about cacao expiration.  But that would require a much bigger code change.  So instead this is kind of a hack to just try our best to get the information from existing state in the state store if we have it, but if we don't have it then just go ahead and load the state directly at the CommitID given.

packages/core/src/state-management/repository.ts
```diff
@@ -487,28 +496,40 @@ export class Repository {
...
+    }
 
-  private async _atCommit(
-    commitId: CommitID,
-    existingState$: RunningState
-  ): Promise<SnapshotState> {
     return this.executionQ.forStream(commitId).run(async () => {
-      const stateAtCommit = await this.streamLoader.stateAtCommit(existingState$.state, commitId)
+      const stateAtCommit = existingState$
+        ? await this.streamLoader.resetStateToCommit(existingState$.state, commitId)
```
samika98, line 510: this was confusing to follow at first, reading the methods helps. The renaming of stateToCommit to resetStateToCommit helps

packages/core/src/state-management/repository.ts
```diff
@@ -487,28 +496,40 @@ export class Repository {
...
-      if (StreamUtils.isStateSupersetOf(stateAtCommit, existingState$.value)) {
-        existingState$.next(stateAtCommit)
-        await this._updateStateIfPinned(existingState$)
+      // Check if we need to update what state is stored in our state store based on any
+      // information learned from this CommitID.
+      if (existingState$) {
+        if (StreamUtils.isStateSupersetOf(stateAtCommit, existingState$.value)) {
+          // If the provided CommitID is ahead of what we have in the cache and state store, then we
+          // should update them to include it.
+          existingState$.next(stateAtCommit)
```
samika98, line 523: For a case were actual commits are A -> B -> C. what happens here if existing state is at 'A' and commit we learned is at 'C' 
Will this change it to - A -> C ? Or will we load ( B -> C ) and get the correct order of A -> B -> C

samika98 (approved):
Added a question in the comment

PR #3281, opened 2024-10-01
feat: disable anchoring via env var

smrz2001:


packages/core/src/state-management/repository.ts
```diff
@@ -678,6 +680,12 @@ export class Repository {
     if (state$.value.anchorStatus == AnchorStatus.ANCHORED) {
       return
     }
+    if (process.env.CERAMIC_DISABLE_ANCHORING === 'true') {
+      this.logger.warn(
```
stbrody, line 684: this will be very verbose as this code runs for every update.  I think we could log in `ceramic.ts:requestAnchor` which only runs when an anchor is explicitly requested by the user, but this code can also run for implicit automatic anchors which we probably don't want to log a warning on.

stbrody: tbh I just wouldn't worry about logging at all.  This is the right place to early return to skip anchoring, so I'd just take out the log warning and call it a day.

stbrody (approved):
lgtm mod removing warning log

PR #3282, opened 2024-10-07
feat: add metrics to track queue sizes and add operations - #AES-361

gvelez17:
## Description

From the datadog analysis, it looks like the gitcoin out of memory errors are related to queue growth specifically with the p-queue library.

This adds metrics to all instances of PQueue to track the size and add operations.

## How Has This Been Tested?

- [x] Unit tests

## PR checklist

## References:

### DIFF
--- a/packages/cli/src/s3-store.ts
+++ b/packages/cli/src/s3-store.ts
@@ -13,6 +13,7 @@ import PQueue from 'p-queue'
 import AWSSDK from 'aws-sdk'
 import { Mutex } from 'await-semaphore'
 import type { DeepNonNullable } from 'ts-essentials'
+import { ServiceMetrics as Metrics } from '@ceramicnetwork/observability'
 
 /**
  * **Remove** `undefined` fields from an S3 Level search params.
@@ -185,6 +186,8 @@ class S3KVStore implements IKVStore {
   }
 
   get(key: string): Promise<any> {
+    Metrics.count(LOAD_S3_QUEUE_ADD, 1)
+    Metrics.observe(LOAD_S3_SIZE, this.#loadingLimit.size)
     return this.#loadingLimit.add(async () => {
       const value = await this.level.get(key)
       return JSON.parse(value)
--- a/packages/cli/src/s3-store.ts
+++ b/packages/cli/src/s3-store.ts
@@ -31,6 +31,9 @@ function definiteSearchParams<T extends Partial<StoreSearchParams>>(obj: T): Dee
  */
 const MAX_LOAD_RPS = 4000
 
+const LOAD_S3_QUEUE_ADD = 'load_s3_queue_add'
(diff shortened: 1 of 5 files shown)


gvelez17:
side node, when i build on my laptop i get a unrelated error

```
src/ceramic-cli-utils.ts(530,5): error TS2322: Type 'import("/Users/gv/js-ceramic/packages/cli/node_modules/dids/dist/did").DID' is not assignable to type 'import("/Users/gv/js-ceramic/node_modules/dids/dist/did").DID'.
src/ceramic-daemon.ts(333,5): error TS2322: Type 'import("/Users/gv/js-ceramic/packages/cli/node_modules/dids/dist/did").DID' is not assignable to type 'import("/Users/gv/js-ceramic/node_modules/dids/dist/did").DID'.
  Types have separate declarations of a private property '_client'.
lerna ERR! npm run build exited 2 in '@ceramicnetwork/cli'
```

after running npm install, not sure why this is, should be unrelated.

packages/core/src/ancillary/task-queue.ts
```diff
@@ -52,6 +56,8 @@ export class TaskQueue implements TaskQueueLike {
    * Size of the queue. Counts both deferred and currently running tasks.
    */
   get size(): number {
+    Metrics.observe(TASK_QUEUE_SIZE, this.#pq.size)
+    Metrics.observe(TASK_QUEUE_SIZE_PENDING, this.#pq.pending)
```
smrz2001, line 60: These might be less useful because there will be many `TaskQueue` instances that we won't be able to differentiate since they're unnamed.

Looks like they're instantiated through `NamedTaskQueue`, so the metrics there should be sufficient.

packages/cli/src/s3-store.ts
```diff
@@ -185,6 +189,8 @@ class S3KVStore implements IKVStore {
   }
 
   get(key: string): Promise<any> {
+    Metrics.count(LOAD_S3_QUEUE_ADD, 1)
+    Metrics.observe(LOAD_S3_QUEUE_SIZE, this.#loadingLimit.size)
```
smrz2001, line 193: This is a `get` call, should we move these lines to the `put` call? I don't think the S3 store is used anymore, but adding these doesn't hurt.

gvelez17: well - the way i read this, for whatever reason the .get of the s3 storage is adding a leveldb get call into the queue.  so for monitoring the queue, i want it here.  

The put doesn't add to the queue, it just does the put synchronously it seems

packages/core/src/ancillary/task-queue.ts
gvelez17: ok, removed!

smrz2001 (approved):

PR #3283, opened 2024-10-10
fix: do not log names of all queues, cardinality is too high

gvelez17:
## Description

## How Has This Been Tested?

- [ x] unit tests

## PR checklist

## References:

### DIFF
--- a/packages/core/src/state-management/named-task-queue.ts
+++ b/packages/core/src/state-management/named-task-queue.ts
@@ -56,11 +56,11 @@ export class NamedTaskQueue {
    */
   run<A>(name: string, task: () => Promise<A>): Promise<A> {
     const queue = this.queue(name)
-    Metrics.observe(NAMED_TASK_QUEUE_SIZE, queue.size, {'method': 'run'})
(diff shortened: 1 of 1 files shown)


packages/core/src/state-management/named-task-queue.ts
```diff
@@ -4,6 +4,8 @@ import { ServiceMetrics as Metrics } from '@ceramicnetwork/observability'
 const NAMED_TASK_QUEUE_SIZE = 'named_task_queue_size'
 const NAMED_TASK_QUEUE_RUN = 'named_task_queue_run'
 const NAMED_TASK_QUEUE_ADD = 'named_task_queue_add'
+const NAMED_TASK_QUEUE_LARGE_SIZE = 'named_task_queue_large_size'
+const LARGE_QUEUE_THRESHOLD = 50
```
smrz2001, line 8: Can this be set from an env var? Doesn't need to be plumbed through the whole CLI code flow. That way, we don't have to cut a new release if the cardinality blows up again, and can just restart with a higher env var setting.
```suggestion
const LARGE_QUEUE_THRESHOLD = process.env.LARGE_QUEUE_THRESHOLD || 50
```

smrz2001 (approved):
🥳

PR #3284, opened 2024-10-10
fix: address multiquery long timeouts causing excessive queue growth …

gvelez17:
During a period of heavy traffic and possible IPFS overload or failures, many errors of the following form occurred:

```Error loading stream kjzl6kcym7
w8y6ppdchg8sfi0dnlcoixh4j5liiqhloal1dlz8yndht0nxxdafp at time undefined as part of a multiQuery request: Error: Timeout after 30000ms
```

Approximately 292K in one day.  

We also have been seeing dramatic queue size growth and p-queue memory usage during these periods of intense load.  This could be related to the large number of streams waiting for the 90 sec timeout, which still eventually fails.  It would be better to fail sooner and not increase the load further.

## How Has This Been Tested?

- [ ] Test B

## PR checklist

## References:

### DIFF
--- a/packages/core/src/ceramic.ts
+++ b/packages/core/src/ceramic.ts
@@ -67,7 +67,7 @@ import { LevelKVFactory } from './store/level-kv-factory.js'
 const METRICS_CALLER_NAME = 'js-ceramic'
 const DEFAULT_CACHE_LIMIT = 500 // number of streams stored in the cache
 const DEFAULT_QPS_LIMIT = 10 // Max number of pubsub query messages that can be published per second without rate limiting
-const DEFAULT_MULTIQUERY_TIMEOUT_MS = 30 * 1000
+const DEFAULT_MULTIQUERY_TIMEOUT_MS = 3 * 1000
 const DEFAULT_MIN_ANCHOR_LOOP_DURATION_MS = 60 * 1000 // 1 minute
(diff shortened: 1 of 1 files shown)


packages/core/src/ceramic.ts
```diff
@@ -67,7 +67,7 @@ import { LevelKVFactory } from './store/level-kv-factory.js'
 const METRICS_CALLER_NAME = 'js-ceramic'
 const DEFAULT_CACHE_LIMIT = 500 // number of streams stored in the cache
 const DEFAULT_QPS_LIMIT = 10 // Max number of pubsub query messages that can be published per second without rate limiting
-const DEFAULT_MULTIQUERY_TIMEOUT_MS = 30 * 1000
+const DEFAULT_MULTIQUERY_TIMEOUT_MS = 3 * 1000
```
stbrody, line 70: I'm not positive, but I believe when this timeout fires it will only fail the multiquery request but not clean up the work happening in the background in the execution queue.  I think probably you want to pair this with also turning down the IPFS timeouts in the dispatcher.

gvelez17: ok i had made a separate PR for those since i only saw the multiquery errors in the logs.  we could do https://github.com/ceramicnetwork/js-ceramic/pull/3285 instead.  I thought that might require more testing if its hitting more locations.

PR #3286, opened 2024-10-11
feat: Flag to fully disable pubsub and all cross-node syncing

stbrody:
Flag can be enabled by setting `ipfs.disablePeerDataSync` in the config file, or with the new `CERAMIC_DISABLE_PEER_DATA_SYNC` env var.

turning on this flag does 3 things:
1. Fully disable pubsub.  The node will not subscribe to the ceramic pubsub topic, and will not publish or receive any pubsub messages
2. Turns down the timeout when getting blocks from ipfs from 30 seconds to 1 second
3. Uses the "offline" flag whenever loading blocks form ipfs so that the request will fail immediately if the block is not found in the ipfs node's local blockstore.  IPFS will never try to load the block via bitswap.

Blocks published to ipfs with this flag enabled should still be fetchable by other ipfs nodes over bitswap.

### DIFF
--- a/packages/core/src/dispatcher.ts
+++ b/packages/core/src/dispatcher.ts
@@ -497,7 +497,7 @@ export class Dispatcher {
    * @param tip - Commit CID
    */
   publishTip(streamId: StreamID, tip: CID, model?: StreamID): Subscription {
-    if (process.env.CERAMIC_DISABLE_PUBSUB_UPDATES == 'true' || !this.enableSync) {
+    if (!this.enableSync) {
       return empty().subscribe()
     }
 
--- a/packages/core/src/dispatcher.ts
+++ b/packages/core/src/dispatcher.ts
@@ -140,7 +140,7 @@ export class Dispatcher {
     const rustCeramic = EnvironmentUtils.useRustCeramic()
     this.enableSync = rustCeramic ? false : enableSync
 
-    if (!rustCeramic) {
+    if (this.enableSync) {
       const pubsub = new Pubsub(
         _ipfs,
         topic,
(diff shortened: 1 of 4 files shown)


christianlavoie (commented):
Simpler than I would have expected.

packages/core/src/dispatcher.ts
```diff
@@ -498,7 +497,7 @@ export class Dispatcher {
    * @param tip - Commit CID
    */
   publishTip(streamId: StreamID, tip: CID, model?: StreamID): Subscription {
-    if (process.env.CERAMIC_DISABLE_PUBSUB_UPDATES == 'true' || EnvironmentUtils.useRustCeramic()) {
+    if (process.env.CERAMIC_DISABLE_PUBSUB_UPDATES == 'true' || !this.enableSync) {
       return empty().subscribe()
```
gvelez17, line 501: what is the difference between CERAMIC_DISABLE_PEER_DATA_SYNC and CERAMIC_DISABLE_PUBSUB_UPDATES ?

I take it this means we could just not publish, or we could not publish and not read or sync with other nodes?  but neither applies under rust ceramic?

If we were going to keep this i think it woudl be confusing but if we are going to leave js-ceramic entirely soon i guess it is fine.

gvelez17 (approved):
LGTM, one comment about possibly confusing option combinations, but given that we're trying to transition away anyway i think its fine

smrz2001 (approved):
:shipit:

packages/core/src/dispatcher.ts
stbrody: `CERAMIC_DISABLE_PUBSUB_UPDATES` is much more limited in scope.  It just disables publishing of update messages.  We'll still listen to pubsub and still publish query and response messages.

Honestly it's not really very useful, we put it in once a while ago to try to troubleshoot an issue gitcoin was having, I don't think anyone uses it anymore and we could frankly probably get rid of it.

